PR #1398, opened 2020-04-04
Resolve "Create integration tests for margins"

ashleyvega:
In GitLab by @pand-dev on Jan 30, 2020, 12:47

Closes #667


witgaw (approved):
TL;DR:  Everything OK, happy for it to be merged.

Reviewed margin testing cases 1-5. Increased magnitude by 10^5 to match the 5 decimal places case considered in scenario-runner with those files and to get the resulting values closer to those specified in the reference spreadsheet (after adjusting to 5 decimal places). Reference values updated to match those previously obtained and reviewed via scenario-runner..

EVODelavega (approved):

PR #1501, opened 2020-04-06
update codeowner with github names

jeremyletang:


edd (dismissed):
Thanks for catching this!

edd (approved):
Such rebase.

PR #1532, opened 2020-04-14
Overwrite original order after amending (#1511)

peterbarrow:
Instead of updating the map of orders to point to the new amended order, we overwrite the original order so everything is already pointing to the correct place.


EVODelavega (approved):
LGTM

PR #1542, opened 2020-04-16
Replace PartyID with Party

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1540


edd (commented):
GraphQL-wise this is good for me. I’ll hold back my approval as there’s enough Go here that I’d like someone else to check it, but it looks reasonable to me too.

jeremyletang (approved):
seems OK to me

edd (approved):
👍

PR #1563, opened 2020-04-21
Replace USD with VUSD in market configuration

edd:
The quoteName is different to the settlement asset. It shouldn’t be.

- Update quoteName for markets from USD to VUSD

Closes https://github.com/vegaprotocol/console/issues/3671


ashleyvega (approved):
LGTM

PR #1564, opened 2020-04-21
hex encode key

EVODelavega:


ashleyvega (approved):
LGTM, what could go wrong...

PR #1567, opened 2020-04-22
Extend API allowing more governance requests

ghost:
Closes #1541


EVODelavega (dismissed):

jeremyletang (changes requested):
All in all the number of API added here is way to big for my liking to live in the core for a single feature.
Also the usage of maps is crazy, we should probably review that, once we have more proposal types, we had a new map? so with 20 proposal types, we endup managing 20 maps just for that? I don't think that's the right way.

If the APIs are what's required for the frontend, maybe, but I think some of them should be removed as they cover the same things.

edd (commented):
The GraphQL side of this looks great 👍

jeremyletang (commented):
lgtm

jeremyletang (approved):

PR #1568, opened 2020-04-22
Remove pendingOrder from use in Prepare*Orders

peterbarrow:
Added new validation checks and update the unit tests to stress the new validation

Closes #1516


edd (approved):
The GQL changes look good.

PR #1589, opened 2020-04-27
increment token balance as well as general account in notify trader

jeremyletang:
close #1574


ponthief (approved):
already commented - accounts are topped up

EVODelavega (approved):

PR #1630, opened 2020-05-07
Fixed code that writes orders to order buffer

peterbarrow:
The ordering that the code wrote orders to the orderbuffer was incorrect and resulting in multiple order updates for a single amend. This PR fixes that problem

closes #1513


ashleyvega (approved):
LGTM

PR #1636, opened 2020-05-11
Add GraphQL complexity level to GraphQL config

ghost:
Adding complexity limit field to the node config to be used to initialise GraphQL API.

Default value is **5**


ashleyvega (approved):
LGTM, then see what breaks ... (e.g. TopGun, til that gets a better GQL query)

PR #1646, opened 2020-05-12
Increase GraphQL complexity limit to 38

ghost:
Per conversation in https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1589299950048500

Setting the limit to 38. It could be further limited down via configuration later on.


edd (approved):
Thank you!

PR #1649, opened 2020-05-13
Fix github artefact upload CI config (#1648)

ashleyvega:


edd (approved):
I'll take your word for it

PR #1654, opened 2020-05-13
Testing for version numbers

peterbarrow:
Adding a set of steps that verify the version number of an order increases after an amend and stays the same after a cancel.

Closes #788


ashleyvega (approved):
LGTM

PR #1655, opened 2020-05-13
Add comments + make race detector happy

EVODelavega:


ashleyvega (approved):
LGTM, and thanks for enabling the race detector and adding comments.

PR #1658, opened 2020-05-14
Add timestamps to votes

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1599

This will add timestamps to governance Votes. Since it is now possible to view all (more than 1) votes that were cast for a specific proposal by single party, there needs to be a way to identify when those votes were cast to confirm that behaviour is indeed true.


edd (commented):
REST & GQL changes look good to me. The rest superficially looks fine but I’ll leave that to someone who reads Go to give a quick once over.

ashleyvega (commented):
Blocking and non-blocking feedback requested.

ashleyvega (approved):
LGTM

PR #1662, opened 2020-05-15
Add generic base-event, update broker to use the base event interface…

EVODelavega:
Introduce a base event type, use context for trace ID and sequence stuff:

* added the beginnings of a generic event constructor (might not be used)
* Subscribers are typed (if desired)
* Add test cases to ensure that typed subscribers only receive events they actually need
* Subscriber ID's are unique across all types, single subscriber can subscribe to arbitrary number of events


ashleyvega (commented):
Couple of comments...

jeremyletang (approved):

PR #1676, opened 2020-05-20
Remove requirement to name every gRPC in/out type with Request/Response suffix

ghost:
Unless there is a good reason to keep the rule around, let's remove this from code style document.
If you know of a good reason why this might be beneficial please bring it up and we can close this PR.

_Note that it is still a good idea to distinguish logically unrelated things. Also it might be useful to wrap internal types as it gives a separation layer for gRPC API._

**As it is right now, the downside of having a strict rule like that - it bloats API with duplicated types defining same thing.**

E.g.
Say, there is a piece of data that defines a core structure users might be interested in - `vega.ImportantData`. To build an rpc accessor, `GetImportantData`, two wrappers are created:
```
message GetImportantDataRequest {
  // ... some arguments
}
message GetImportantDataResponse {
  vega.ImportantData data = 1;
}
```
Now you need a couple more rpc functions:
- `MakeImportantData` - allocates and registers important data, returning a copy of it
- `RemoveImportantData` - unregisters important data, returning a copy of it
- `UpdateImportantData` - updates important data, returning a copy of it
- `GetImportantDataOffline` - conditionally returns a copy of important data
- `GetBrokenImportantData` - conditionally returns a copy of important data
- `GetImportantDataInQueue` - conditionally returns a copy of important data

It is of course possible to just create 7 types with exactly same field. This will unnecessarily complicate API bringing nothing (**arguably**) in return.

Another, even worse IMO, approach would be to merge all these rpc functions into single one and perform a dispatch internally based on the supplied parameters. This is not readable and brings more complexity for no benefit in return.

A reasonable (**arguably**) approach would be to simply rename `GetImportantDataResponse` to `ImportantData` (or `ImportantDataResponse`) and return it 7 times.

### DIFF
--- a/codingstyle.md
+++ b/codingstyle.md
@@ -74,7 +74,6 @@ In addition to the Golang code review standards, we want to be consistent:
 
 * Avoid nested types as much as possible. Enums are the notable exception here.
 * Fields that are ID's should be named `ID` or `FooID` (ID is CAPS).
-* Messages used in the API use the suffix `Request` and `Response`.
 * API Request/Response types, and the service definitions belong in the `proto/api` directory (and the `api` package)
 * Message types representing a unit of data, currently used in the core (e.g. `Order`, `Market`, `Transfer`, ...) are defined in the `proto/vega.proto` file. These types are imported under the alias `types`.
 * Wherever possible, add validator tags to the proto definitions.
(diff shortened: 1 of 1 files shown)


jeremyletang (changes requested):
As discussed previously it's common practice / convention to have these type being unique for all rpc call in grpc services, so these will need to stay as they are.
Just a reference in case it's needed:
https://buf.build/docs/style-guide#services
https://buf.build/docs/lint-checkers#rpc_request_standard_name-rpc_response_standard_name-rpc_request_response_unique

PR #1679, opened 2020-05-20
Feature/1670 propagate node key with tm key

jeremyletang:
This adds a small topology package to keep track of the validators.

Very simple for now:
- uses the  validators from the genesis block
- pair each validator with a single vega node
- ensure some commands are not sent when a node is not a validator (assume a validator have the store disabled for now, we'll need to add an actual flag for that later on).


ashleyvega (changes requested):
Changes requested.

EVODelavega (approved):

ashleyvega (approved):
LGTM

PR #1681, opened 2020-05-20
Governance network parameters cleanup

ghost:
Changes here:
 - removed network parameters from governance service as those aren't used
 - pass in default network parameters into governance engine from execution engine (as this is likely where they are going to live when UpdateNetwork proposals are implemented)
 - store time related network parameters as durations (so there is no need to convert them into duration every time a proposal is processed)
 - corrected the way network parameters constraining enactment time were used
 - added sanity checks for network parameters
 - improved error handling for violation of the network parameter constrains
 - **left network parameters hardcoded** but still convenient to update via ldflags

Related to https://github.com/vegaprotocol/vega/issues/1665


ashleyvega (approved):
Two non-blocking comments.

PR #1684, opened 2020-05-21
Fixing expiring orders when amended

peterbarrow:
If we amend an order so that it's expiresAt time changes or it flips to/from the GTT time in force, we have to make sure the order is correctly handled in terms of the expiry table. This issue is to make sure this is handled correctly with tests to validate it.
Closes #1536


ashleyvega (approved):
LGTM

PR #1711, opened 2020-05-22
Remove id from new market proposal terms input in GraphQL API

ghost:
Very minor change removing market id from the proposal terms input - setting the field has no effect anyway as it is generated once proposal is read from the chain.


ashleyvega (approved):
LGTM

PR #1712, opened 2020-05-22
Allow returning proposals without ids - prepareProposal does not set it

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1680

A small fix allowing returning proposal data for newly prepared proposals which have no id or timestamp (both are set once proposal is read from the chain).


ashleyvega (approved):
LGTM

PR #1713, opened 2020-05-22
Pass in address of empty object instead of nil to work around validator

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1710

This code passes in address of empty object instead of nil to work around gRPC validator.


ashleyvega (approved):
LGTM

PR #1718, opened 2020-05-26
Prevent a user from cancelling an order they did not create

peterbarrow:
Looking to see why we can cancel an order from another user.
Closes #1611


ashleyvega (approved):
LGTM

PR #1719, opened 2020-05-26
require vega wallet at startup of processor

jeremyletang:
closes #1704 

This load thevega wallet from the nodewallet when the Processor is constructed so we would return an error straight  away instead of loading the wallet everytime we need it, and produce an error then.


ashleyvega (approved):
LGTM

PR #1726, opened 2020-05-27
MarkPrice not updated after a partial fill

peterbarrow:
During some testing it was discovered that the mark price is not updated when a partial fill takes place.
This PR is to investigate that.

Closes #1678


ashleyvega (approved):
Looks good to me, LGTM.

PR #1728, opened 2020-05-27
Remove expiry time validation from PrepareAmend call

peterbarrow:
The validation of the amended expiry time is against the current time and the createAt time. This must be done post chain.

Closes #1661


ashleyvega (approved):
LGTM

PR #1730, opened 2020-05-28
Adding ChainID value to statistics endpoint

peterbarrow:
Field Ops requires the chain_id so they can differentiate between runs of the system.
This PR is to implement that.

Closes #1612


ashleyvega (approved):
LGTM, possibly with some minor cosmetic/grammar/documentation changes.

PR #1732, opened 2020-05-28
Fix party engine issue with inability to trade on newly created markets.

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1606

Changes here:
 - untied parties from markets in the sense that parties are no longer available per market but are now stored in common pool - every party has access to every market (old or new);
- cleaned up naming and interface functions that wasn't used;
- removed half-baked concurrency synchronisation.


ashleyvega (commented):
No comments from me, but I'll leave it to others to approve.

peterbarrow (approved):

PR #1741, opened 2020-05-30
Correct timestamps in gql

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1737

Fixing the timestamps for proposals and votes and making it clearer where we use seconds and where nanoseconds


ashleyvega (approved):
LGTM

PR #1745, opened 2020-06-01
Encoding votes blob into base64 like proposal

ghost:
This will encode votes blob the same way as proposal blobs


ashleyvega (approved):
LGTM

PR #1747, opened 2020-06-01
Reusing proposal ID for market ID

ghost:
This reuses proposal id for market id on proposal enactment. Note that this means in gRPC API if user supplies a market id, it will be ignored and proposal id will always be used instead.
Also adds a check to make sure a market with no id cannot be submitted.

No impact on GraphQL API here though (GraphQL mutation for governanace API does not allow user passing market id as of today).

Related to https://github.com/vegaprotocol/vega/issues/1738


ashleyvega (dismissed):
Just the one non-blocking (grammar) comment from me. Others may have opinions on the code, so I'll leave it to them to merge or not.

ashleyvega (approved):
LGTM

edd (approved):

PR #1750, opened 2020-06-01
Add datetime to votes that returned with governance data

ghost:
GraphQL only issue.
Date time stamp is returned for votes queried via party already but governance data votes still missed the date time stamp. This fixes that minor issue.


ashleyvega (approved):
LGTM

PR #1751, opened 2020-06-02
Rewrite vote counting logic

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1603
Related tp https://github.com/vegaprotocol/vega/issues/1605

Rewriting vote counting logic. Assumptions not clear but current logic was implemented so that existing code enacts proposals right after they have passed. Also participation level was assumed to be equal to all the yes votes.

Logic in this change:
- participation is calculated as percentage of all stake that participated in voting - [spec](https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md#configuration-of-a-proposal);
 - proposal is set to be passed on (or right after) propsal closure (per its terms) - [spec](https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md#duration-of-the-proposal);
 - only proposals with participation percent greater than that set on proposal terms and yes votes stake greater than no votes stake can pass [spec](https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md#configuration-of-a-proposal);
 - proposal is enacted if it passed **and** at (or right after) proposal enactment time (per its terms) - [spec](https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md#when-changes-are-applied)

### DIFF
--- a/execution/engine.go
+++ b/execution/engine.go
@@ -495,6 +495,7 @@ func (e *Engine) enactProposal(proposal *types.Proposal) error {
 		if e.log.GetLevel() == logging.DebugLevel {
 			e.log.Debug("enacting proposal", logging.String("proposal-id", proposal.ID))
 		}
+		newMarket.Changes.Id = proposal.ID // reusing proposal ID for market ID
 		if err := e.SubmitMarket(newMarket.Changes); err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


ashleyvega (approved):
Looks fine to me, but I'll leave it to others to merge, in case they have blocking feedback.

peterbarrow (approved):

PR #1752, opened 2020-06-02
Remove incorrect validation rules used while submitting new orders

peterbarrow:
This PR removed incorrect validation rules that were applied during new orders.
Added unit tests to test the new code flow.
Updated existing unit tests to work with the new correct logic


ashleyvega (approved):
Just the one non-blocking comment from me, which can be done in a separate Issue, or rejected.

PR #1757, opened 2020-06-03
Adding validation for correct partyID when amending

peterbarrow:
This PR adds a validation step for amends that the PartyID given in the amend matches the PartyID of the original order.
Added a unit test to confirm this is correct


ghost (approved):
LGTM

PR #1763, opened 2020-06-05
Expiration settlement date for market changed to 30/10/2020-22:59:59

cdm:
Market is due to expire on 30/06, extending to October as settlement at expiry is not tested/ready yet.

```
baseName:              "GBP",
quoteName:             "VUSD",
settlementAsset:       "VUSD"
```


ashleyvega (approved):
I guess this is fine, even though 22:59:59 looks like an accidental time instead of a deliberately chosen one.

PR #1773, opened 2020-06-05
Add proto enum zero value "unspecified" to Order.Status (#737)

ashleyvega:
This PR:
- adds a zero value `STATUS_UNSPECIFIED` to the proto enum `Order.Status`
- adds a `STATUS_` prefix to all other values, as recommended by the [buf.build Style Guide](https://buf.build/docs/style-guide#enums)

Leave #737 open.

## Generic description of reasons for adding proto zero-values

The reason for adding deliberately-invalid zero values to proto enums is to avoid the case where missing or invalid values are replaced with valid zero-values which may change the meaning of an API request in a way the user did not expect.

For example, a json REST `SubmitOrder` request including `{ "Side": "Sell", ...}` will have the `Side` set to (the zero-value) `Buy` because `{ "Side": "Sell", ... }` should have been `{ "side": "Sell", ...}`.

Instead, if a missing/invalid `Side` is set to `SIDE_UNSPECIFIED`, that's clearly not a Buy or a Sell, so the API request can be rejected as obviously invalid.


ghost (approved):
LGTM

PR #1781, opened 2020-06-08
Add prefix and UNSPECIFIED to ChainStatus, AccountType, TransferType (#737)

ashleyvega:
This PR:
- adds `CHAIN_STATUS_` prefix to `ChainStatus` enum values
- adds `CHAIN_STATUS_UNSPECIFIED`
- adds `ACCOUNT_TYPE_` prefix to `AccountType` enum values
- adds `ACCOUNT_TYPE_UNSPECIFIED` as a replacement of `ACCOUNT_TYPE_ALL`
- adds a check in the accounts store to make sure an account with Type==UNSPECIFIED does not get saved
- adds `TRANSFER_TYPE_` prefix to `TransferType` enum values
- adds `TRANSFER_TYPE_UNSPECIFIED`
- adds `STATE_` prefix to `Proposal.State` enum values
- adds `STATE_UNSPECIFIED`
- adds `VALUE_` prefix to `Vote.Value` enum values
- adds `VALUE_UNSPECIFIED`

The [buf.build Style Guide](https://buf.build/docs/style-guide#enums) recommends adding enum prefixes and "unspecified" zero-values.

Leave #737 open.

## Generic description of reasons for adding proto zero-values

The reason for adding deliberately-invalid zero values to proto enums is to avoid the case where missing or invalid values are replaced with valid zero-values which may change the meaning of an API request in a way the user did not expect.

For example, a json REST `SubmitOrder` request including `{ "Side": "Sell", ...}` will have the `Side` set to (the zero-value) `Buy` because `{ "Side": "Sell", ... }` should have been `{ "side": "Sell", ...}`.

Instead, if a missing/invalid `Side` is set to `SIDE_UNSPECIFIED`, that's clearly not a Buy or a Sell, so the API request can be rejected as obviously invalid.


jeremyletang (changes requested):
The AccountType_ALL had actually the same purpose than the UNSPECIEFIED.
Can we just remove totally the ALL, and replace it's use in the code by the new UNSPECIFEID varian?

jeremyletang (approved):

PR #1784, opened 2020-06-09
Update GraphQL schema after #737 (#1783)

ashleyvega:
This PR:
- adds GraphQL `TradeType` (enum)
- adds GraphQL `Trade.Type` (type: `TradeType`)
- updates GraphQL `OrderStatus` from proto `Order.Status`, add missing:
  - `PartiallyFilled`
- updates GraphQL `RejectionReason` from proto `OrderError`, add missing:
  - `MissingGeneralAccount`
  - `InvalidSize`
  - `InvalidPersistence`
  - `InvalidType`
- adds notes in `vega.proto` and `governance.proto` to update GraphQL when adding a proto enum values
- updates GraphQL helpers and resolvers

Closes #1783.


jeremyletang (changes requested):
I may have missed the discussion about this so can you point out to it, but where was it decided to add a codegeneration tool written by our own for enum convertion? I think it was about adding check on the CI which was discussed.

I'm pretty much against running our own codegeneration tool first, and secondly even more in python.
It make the barrier to entry to contribute even higher. if we decided to add this kind of codegeneration, so it needs to be done in go so we don't have more than go needed to be installed to do it.

Written in go it will be as easy if not more that writting it in python.

edd (dismissed):
As above

jeremyletang (approved):
usually I tend to like naming function which convert from / to another type like this:
sideFromProto / sideIntoProto instead of just convertSide / unconvertSide as it do not give much of indication about what it get converted to or from, but it's not a big deal.

PR #1761, opened 2020-06-04
Required majority, required participation network parameters

ghost:
Closes https://github.com/vegaprotocol/vega/issues/1756

Per these discussions on Slack:
 - https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1591793290077900
 - https://vegaprotocol.slack.com/archives/C9E6XA8GK/p1591633108081800

Change:
 - only open proposals will be accepted to governance engine now (proposals in anything other than open will result in error and will simply be dismissed);
 - only governance engine will be changing proposal states (from open to anything);
 - proposals majority and participation will be exclusively driven by the network parameters;
 - proposal validation will only happen once - on submission;
 - proposals are no longer tracked by reference;
 - new network parameters are floating point;
 - proposal terms will no longer contain participation requirement field;
 - documented governance API in the GraphQL schema and protobuf definitions;
 - governance network parameters are now being set per proposal type (to be extended when all proposal types are handled).

**This change will break existing system tests and Console by significantly changing existing governance API.**

PR #1788, opened 2020-06-09
Check for unspecified Vote Value (#1786)

ashleyvega:
This PR adds a check during `PrepareVote` that `Vote.Value` is not `UNSPECIFIED`.

Closes #1786.


ghost (approved):
We will also need same check inside governance engine but I will be changing it a lot in https://github.com/vegaprotocol/vega/pull/1761 (per conversation in https://vegaprotocol.slack.com/archives/C9E6XA8GK/p1591633108081800) so I can do it then.

PR #1803, opened 2020-06-15
Set the updatedAt field when expiring an order

peterbarrow:
Every update to the order object should result in the `updatedAt` field being updated at the same time. When we handle an expired order we do not update that field and it can lead to a confusing situation where the order is expired but the `updateAt` field is a lower value than the `expireAt` field.

This ticket adds the code to update the `updatedAt` field when an order is expired.


ashleyvega (approved):
LGTM

PR #1825, opened 2020-06-19
adding the notary

jeremyletang:
Adds the notary, this will allow all nodes of the network to sign a message to represent the network agreement for a given action.
Will be used first in order to accept the whitelisting of an asset and later on to accept withdrawals, etc.

This is part of the split of the PR 1669


edd (changes requested):
Requesting some docs on some fields that have been missed. There are plenty of comments, this just needs a few more on:
- Protos
- A readme for notary
- One graphql field

Otherwise schema changes look good.

peterbarrow (approved):
Nonblocking Comments:
1. Is it golang standard to have variables names like `x` and `h` even if the object they represent has nothing to do with an `x` or `h`? For instance I would prefer an object of type Notary to have a variable `n` instead of `t`.
2. We use the word `Type` throughout the code like OrderType, MessageType etc. Why have you gone with `NodeSignatureKind` instead of `NodeSignatureType`?

edd (approved):
Thanks for addressing my docs feedback. I’ll change this to ‘Approved’ as I was blocking it and @peterbarrow has already approved, which is definitely a better measure!

PR #1839, opened 2020-06-22
Specify using a type switch for type mapping & explain why

EVODelavega:
Update codingstyle document to clarify why type mapping is done better through switches.


jeremyletang (approved):

ashleyvega (approved):
Thank for adding this. v.helpful.

PR #1841, opened 2020-06-22
Moved expiry time validation to post chain

peterbarrow:
We had the expiresAt validation code in the PrepareOrder function which is an optional function client side for preparing new orders. This was breaking because it can be called before the Vega node is ready. We have moved the validation code from pre-chain to post chain so the VegaTime value is correct.

Closes #1830


jeremyletang (approved):
Once build pass, good to get merged.

PR #1844, opened 2020-06-23
Start implementing governance subscribers to replace current plugin

EVODelavega:
Closes #1703 

Governance events (proposal + vote), subscribers will be used to replace the current plugin. The subscribers will be a lot simpler/lighter. For each gRPC stream, a new subscriber is created and registered with the broker. Subscriber supports custom filter callbacks, so we can use the small subscriber to stream specific events, based on things like: proposal state, reference, ID, partyID, etc...


jeremyletang (approved):
just a few comments about contexts which needs to be passed from callers.

PR #1846, opened 2020-06-23
Set order state for wash trades to rejected

peterbarrow:
A partially filled wash trade should be rejected but was wrongly being set to stopped. The code has been updated to set the order status to rejected after it has been identified as a wash trade.


jeremyletang (commented):
LGTM once build pass, should we get this document updated as well https://github.com/vegaprotocol/product/blob/master/specs/0024-order-status.md? to keep track of when this happend, this is the expected status?

EVODelavega (approved):
LGTM

PR #1857, opened 2020-06-26
Correcting type mapping section in coding style

ghost:
1. Corrected markdown formatting for readability.
2. Corrected boundary when type switch is beneficial.
3. Corrected statements regarding performance.

Reasoning:
- go compiler generates by far faster code for a type check: https://go.godbolt.org/z/jAuVA8 It is not surprising because single if-else statement is simple and any compiler will be able to optimise it away. Not so much for a types switch - a much heavier syntactic construct.
- measuring performance of a two type case, it becomes slightly less obvious but if-else variant still has less code and performs better: https://go.godbolt.org/z/-zYF9F If-else statements (even more than 1) generate code that is more friendly for CPUs with branch prediction logic in them (majority of general purpose CPUs contain that logic). It is a typical micro-optimisation in C++ to replace 1-2 case switches with if-else. Increasing number of if-else statements however does increase amount of code to read thus decreasing its readability.

### DIFF
--- a/codingstyle.md
+++ b/codingstyle.md
@@ -68,10 +68,10 @@ func NewFooEngine(fooBuf FooBuffer) *fooEngine {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


EVODelavega (changes requested):
For reasons explained in the comment on the PR as a whole, I'm against this change.

PR #1877, opened 2020-07-06
Feature/1700 positions events

EVODelavega:
Closes:
#1700
#1702 

Also prevents subscribers from starting routines that boil down to a busy loop. Subscribers now have a property used by the broker to mark them as (n)acking. If a subscriber is ack'ing, then a cleanup routine is started (waits for ctx.Done(), then closes all channels). Instead of the select loop with defaults which made the subscribers help themselves to all CPU resources available.

The positions plugin has been simplified a ton, now getting distinct events for normal settlement vs distressed settlements (as opposed to overloading the same event).

Markets buffer is also removed as part of this PR. Unlike the buffer, we don't need to batch the markets. Creating markets isn't the most common thing to have happen, and the time update event currently is a bit of a bottleneck (will be addressed later). We're just saving new markets as and when they're created


ashleyvega (commented):
Comments!

jeremyletang (approved):

PR #1880, opened 2020-07-06
Enable new asset in collateral engine

jeremyletang:
Now that we can support adding new external assets, we need to ensure that the collateral engine is aware of any new assets being added.
This PR will:
- add calls to enable new asset in collateral engine + error handling
- add api to list available asset
- new event for the broker
- startup will enable asset in the collateral as well until we have infrastructure scripts to send assert through governance.

closes #1879


edd (changes requested):
Some really nitpick changes requested in documentation

edd (approved):
Thanks for all the extra docs!

edd (commented):
Thanks for all the extra docs (changed to non-approving so that someone who will actually read the Go parts of this can do so) - except that didn't work so I give up.

PR #1890, opened 2020-07-08
proposal resolver, if a party does not exists build inplace

jeremyletang:
When a party does not exists (no deposit ever made for it before) the graphql api for proposal may fail ge the PartyByID, we then construct it in place in order to not get an error to the client.
We could also make the party optional from the Proposal type in the graphql schema, but it would make less senses as party was initially specified, it just did not exists.
 
closes #1885


EVODelavega (approved):
LGTM... I'm assuming/hoping proposals with an unknown party are marked as rejected (just checking)

peterbarrow (approved):

edd (commented):
Looks like an improvement 👍

PR #1892, opened 2020-07-08
add new proposed chain event messages

jeremyletang:
This update the chain event API (event relayed to vega from the blockchain queue, listening to foreign chains.) to be more  generic, and handle per Asset type events, instead of event handling multiple asset types.

This also update the core to use the new chain event api.
This was discussed and defined by @C0deMunk33, feel free as well to takea a look at it again, but it's exactly the API we defined together.


edd (approved):
Looks good 👍 Thanks for the extensive doc updates. I made some minor changes directly in the photo which might need regenerating - I did it directly on GitHub 😊 

This obviously covers BTC and ERC20 but not Ethereum itself. I guess that's coming in a separate PR?

EVODelavega (approved):

PR #1921, opened 2020-07-13
Extract node validation from governance

jeremyletang:
As part of the #1878 in order to implement deposit, we need to use the same mecanism than in the node validation from the 2 steps governance.

This PR extract the code doing the logic about calling checks, sending the command through the chain, checking the topology to ensure we have the minimum required checks into it's own package (part of the validators package) and use it from where it was before.


edd (commented):
One minor non-blocking comment

PR #1939, opened 2020-07-15
Finalize whitelisting

jeremyletang:
This PR cleanup some code in the processor (re processing toEnact proposal).
Then check the erc20 asset event list, validate with the notary it's all signed, and then start validation of the ethereum event
before enabling the asset in colateral and assets.

Code is almost ready but there may be an issue in the smart contract so I need this to be discussed with @C0deMunk33

### DIFF
--- a/assets/erc20/erc20.go
+++ b/assets/erc20/erc20.go
@@ -24,8 +24,9 @@ const (
 )
 
 var (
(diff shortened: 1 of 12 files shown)


edd (commented):
Banking folder needs a README. Otherwise I have some questions in here that are probably just be being naive about go

PR #1940, opened 2020-07-15
update graphql API to support asset proposals

jeremyletang:
This adds the NewAsset fields in the proposal TermsInput and the proposal Terms.

closes #1632


edd (commented):
One blocking question. Otherwise the GraphQL side looks good 👍

EVODelavega (approved):

PR #1956, opened 2020-07-17
add some validation to the prepare proposal api

jeremyletang:
Validate inputs of the governance prepare proposal api.

closes #1672


peterbarrow (approved):
2 non blocking questions:
1. Why do you compare the return of len() to <= 0? From the docs the return value can never be less than zero, but the code makes it look like it can.
2. When comparing a float/double to a value in C++ we had a standard of doing `if a<1.0`. This makes it easier to see it's a floating point comparison. Does that idea hold up in golang?

PR #1990, opened 2020-07-17
deprecate tickSize in graphql api, generate tickSize string at market submission

jeremyletang:
This mark TickSize in input parameters as deprecated.
The tickSize is mostly here for informational purpose now, and generated at market submission from the decimalPlace.
It's also a string to avoid any funny behaviour with float precision.

close #1935


edd (commented):
The change looks good to me but I will leave it for someone who is reading the code a bit closer

EVODelavega (approved):

peterbarrow (approved):

PR #2003, opened 2020-07-20
Create the fee engine

jeremyletang:
This implements the first fee engine and the logic to get the fee for the continuous trading.

close #1894


EVODelavega (approved):
LGTM

PR #2013, opened 2020-07-22
Feature/1974 market framework

EVODelavega:
Closes #1974 Adding opening auction definition to the market framework. 

Changes to market framework and some parts of the API:

Market data will have 2 additional fields, and a new market proposal gets a new field, too.

These fields currently aren't required, to make it easier on everyone to make the required changes


jeremyletang (commented):
I think a bunch of convert functions will be missing in the graphql package for the fields / message you introduced in proto.

jeremyletang (approved):

PR #2025, opened 2020-07-23
Adding auction mode

peterbarrow:
Updating the API to handle the new market type enum.
Validating the correct market type is being entered for each new order
Switching to and from auction mode and handling the parked area for orders that are parked during the switch


edd (commented):
I understand what will be needed on the API side for us to implement this. I assume the new `GoodFor` is intentionally not in `prepareOrder` in this PR, but I've highlighted it just in case. @mattrussell36  worth a skim, but it looks like it'll be a straightforward change!

jeremyletang (approved):
looks very good to me.
Just a few aesthetics changes, or thoughts there.

PR #2032, opened 2020-07-26
add a rollback function for the UpdateMarginOnOrder

jeremyletang:
In the case we create an FOK/IOC order, with a postion of 0 (only in the case the trader NEVER traded or had an order in the book before) and that the order is stopped, the margin may never be released if the trader never trade again after that, until the market is closed.

This add a check to release margin directly, only in these specifc cases.

@3jtechtest please check the integration test.

Closes #1920

### DIFF
--- a/collateral/engine.go
+++ b/collateral/engine.go
@@ -726,6 +726,62 @@ func (e *Engine) MarginUpdate(ctx context.Context, marketID string, updates []ev
 }
 
 // MarginUpdateOnOrder will run the margin updates over a set of risk events (margin updates)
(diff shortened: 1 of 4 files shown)


ponthief (approved):
seeing that the margin test has failed in system-tests, points to the fact the issue is fixed.. (test requires updating)

peterbarrow (approved):
Non blocking comment: This feels like a patch over a bigger issue. Should we be capturing the case when a party has their last open order closed/filled/stopped we trigger the margin->general account transfer? Make this a generic action once all orders are complete rather than coded against a specific order type? We will need to hold orders by party soon to support block cancels so maybe this can be looked into with that?

EVODelavega (approved):

PR #2034, opened 2020-07-27
set keep alive retry for gql

jeremyletang:


edd (approved):
Thanks. I think this is the cause of https://github.com/vegaprotocol/console/issues/3844 so 🤞

peterbarrow (approved):

PR #2040, opened 2020-07-28
Add margin checks for orders during auction

EVODelavega:
Check margins for orders submitted during auction period. Closes #2042

This PR includes execution engine/market toggling auction trading on/off after depending on the configuration of the market framework. The current flow for margin checks has been reused as much as possible. If the market is in auction, a different function on the risk engine is called which calculates the collateral required for a given order based on the order size + price.

for order amends, the old order margin is subtracted from the new order, which can either increase the margin required or release some margin to the general account. A zero transfer is possible in theory (higher price, lower volume).

### DIFF
--- a/execution/market.go
+++ b/execution/market.go
@@ -192,10 +192,14 @@ func NewMarket(
 		partyEngine:        partyEngine,
 		broker:             broker,
 		fee:                feeEngine,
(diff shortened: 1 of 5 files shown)


jeremyletang (approved):
sounds good to me, just one comment for other people to take a look at tghe calculation of the margins.

PR #2054, opened 2020-07-31
Use tendermint genesis block to initialise vega

jeremyletang:
add genesis package to callback when genesis is updated
add genesis config in governance
add genesis command to generate the new vega genesis state

closes #2051

### DIFF
--- a/.drone-github.yml
+++ b/.drone-github.yml
@@ -44,8 +44,6 @@ steps:
       - unset GITHUB_DEPLOY_SSH_PRIVATE_KEY
       # Build normal executable
       - ./script/build.sh -a build -t linux/amd64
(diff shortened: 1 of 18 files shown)


edd (commented):
Generally happy - I read this one through, one minor question that I'd like from @ashleyvega but otherwise it looks goot to me 👍

edd (approved):
Looks fine to me. I think we could improve the error message a bit to offer some more help, but I'm happy to stick that in the backlog and address it when we're looking at error messages more generally.

ponthief (approved):
looks ok, but will start with tests for fees first to make sure  that is really the case

peterbarrow (approved):

PR #2064, opened 2020-08-04
add grpc fee service api

jeremyletang:
Implements an API providing fees calculation estimates for new orders.

closes #2020


ponthief (approved):
looks Ok from the proto side of things...once system-tests added will be able to judge better

PR #2075, opened 2020-08-07
Feature/2070 faucet requests limits

jeremyletang:
add requests number limit to the faucet.
user are grey listed when they do a request to the faucet for a given asset for the duration of the cool down setted up  in config.
close #2075


edd (commented):
It looks alright to me, but needs someone better at Go. I like the rate limit and agree with a comment in Slack that you made that in a followup we need to limit the amounts that people can request.

The reason for that is https://github.com/orgs/vegaprotocol/projects/44 - someone could right now one day request 10000000000 BTC and without auction mode we're see another crash out. Which is totally possible *now* with the old notifyTrader approach, hence saying it's alright for a followup ticket.

peterbarrow (approved):

PR #2078, opened 2020-08-07
add the maxAmount to be mint by the faucet in the definition of the builtin asset

jeremyletang:
This adds a new fields in the builtins assets so we can define what's the max amount which can be requested through the faucet.


peterbarrow (approved):
Nonblocking: I've noticed many min/max functions floating around in different blocks of code. Does it make sense to put them in all one place (a maths or utils library maybe) ?

edd (approved):

PR #2093, opened 2020-08-10
Feature/2092 disable amends

jeremyletang:
This returns not implement in the PrepareAmend IPA, and not implemented also when trying to call Processor.Process with an amendment.

closes #2092


edd (commented):
The changes look good, I'm just going to file an issue for re-enabling this that we can link the commented out blocks to for future reference...

peterbarrow (approved):

PR #2107, opened 2020-08-17
Cancel Order bulk

jeremyletang:
Implements bulk cancel orders
cancel all orders
cancel all orders in markets
cancel a single order in a market.

add a lookup table in the orderbook to find order per parties

closes #2106


peterbarrow (commented):
Looks perfect to me so far

peterbarrow (approved):

PR #2111, opened 2020-08-17
fix fees from fee engine

jeremyletang:
closes #2110


ponthief (commented):
looks ok, can see where infrafee replaced by liqfee;have system-test for it

peterbarrow (approved):

PR #2114, opened 2020-08-19
Markets created on startup will be in auction mode

peterbarrow:
Start all markets in auction mode.
Handle a new market config option to specify the length of the opening auction in seconds.
Process closing the auction and moving to continuous trading.

close #2112 
close #1975 
close #1977 
close #1976


tamlyn10 (commented):
@peterbarrow I've tried to read your code to understand how the auction duration is set, but can't make it out. I assume it's being set by a network parameter, but I know we are lacking the "market lifecycle" spec at the moment! cc @barnabee

jeremyletang (commented):
Just one thing I did not know where to put it.
But there's a small comment to change at the top of the applyFees function, which is to replace with a call to if len(trade) <= 0 || mkt.IsInOpeningAuction { return nil }.

That would be needed so we dont anything threre

jeremyletang (approved):

PR #2117, opened 2020-08-20
Bugfix/2103 stop withdrawal if amount too big

jeremyletang:
Add a LockFund step to be used later which ensure we have enough funds to withdraw.
return an error if cannot or process.
closes #2103


edd (commented):
The withdraw lock account seems like a good approach. I guess it'll need a timeout or something so funds don't get stuck there forever?

peterbarrow (approved):

PR #2134, opened 2020-08-24
Feature/1524 withdraw erc20

jeremyletang:
Implements the erc20 withrawals.
Still missing are: API to build the arguments to call the eth contract, this will need to be discussed again quickly with @C0deMunk33.

Limitation: does not release funds being locked for withdrawal if it never get processed by the user for now, will be done in a further PR, it will be pretty simple, but would make this PR quite big for now.

closes #1524


edd (commented):
It's pretty giant, yes. Could you help out @vegaprotocol/qa by listing out a few scenarios that would help test this? Given it's so big it's hard to pick out errors, so I'm thinking that (after some other @vegaprotocol/core reviews) that we lean towards merging this, ramping up the testing, and then dealing with any issues in follow up tickets. Note for QA: Withdrawal is lower priority than deposits.

I'm guessing as a start (using erc20 as a generic term for some token).

- User has 1 erc20 in vega, requests withdrawal of 10 erc20 receives an error. No voting happens.
- User has 1 erc20 in vega general account, requests withdrawal of 1 erc20, receives a bundle they can submit to ethereum, after voting
- User does not exist, requests withdrawal, receives an error
- User has 1 erc20 in Vega margin account, requests withdrawal, receives an error
- User has 1 erc20 in Vega general account, requests 1 erc20 withdrawal, never executes  transaction, transaction expires?
- Submitting an expired transaction to the smart contract results in no money being withdrawn from Vega


Thanks for the proto documentation - it's nicely covered

ashleyvega (commented):
Some questions, mainly on log levels.

peterbarrow (approved):
While I didn't understand much that was going on, the code looks good.

EVODelavega (changes requested):

EVODelavega (approved):

PR #2136, opened 2020-08-25
storage market accounts, do not stop loadinng account after asset match

jeremyletang:
closes #2122


qustavo (approved):
LGTM

We have the same pattern in `GetFeeInfrastructureAccounts`, you think we need to remove the `break` statement there as well?

PR #2142, opened 2020-08-25
faucet also check if it's expired on a request

jeremyletang:
This just ensure that if a cancel is done,and the cleanup routine haven't been triggered before, the mint request will pass anyway.


peterbarrow (approved):

ponthief (approved):
seen increase in greylisting for 1 min for every single attempt in test; will check again when merged

PR #2168, opened 2020-08-28
Actually add channel to map so we only create 1 routine

EVODelavega:
Maintaining event order through an internal broker channel is indeed a valid approach. Unfortunately, things go kind of badly wrong if you then fail to actually store the channel and reuse it. Failing to do so results in new routines spawning for each event we're sending. That was the way the broker worked before, but the new routines are written in a way to never return, they assume the channel will be reused, which it isn't...

This explains that the runtime.newproc1 showed up in the pprof memory profile the way it did.


ashleyvega (approved):
LGTM

PR #2179, opened 2020-09-01
Feature/2101 event api

EVODelavega:
Start work on the event stream API's.

* Created the proto types
* Created a generic event subscriber
* Added unit tests for the generic subscriber
* Create a stream call
* Refine some interfaces (filtering on market ID and or party ID)

Optional

* Implement GQL equivalent


peterbarrow (approved):
Non blocking: Lots of new event code where I noticed the Proto() function has the same insides as the event named function directly above it. Is there any way to avoid having 2 functions doing the same thing?

edd (changes requested):
Comments required on new protos. Also a GraphQL endpoint for this subscription.

edd (approved):
Thanks for the comments on the types. The resulting documentation is better if there's a comment above every field, as well as the type, but this PR has sat in review for long enough, so I'll file that as a follow up.

PR #2181, opened 2020-09-02
uses default value in proposal when fields are not specified to handle invalid submitted proposals

jeremyletang:
This is a temporary fix (this API may need to be redone) to handle through the graphql API the proposal with missing information which got rejected after being submitted to the chain.

As the graphql API expect all fields to be valid (non-nil) we just set default/zero values for the fields which are nils so they do not get rejected / produce error when being resolved.

Ultimately we will want to have a specific api to handle rejected proposal, but this would unblock the front-end in the meantime.

closes #2056


ponthief (commented):
looks good  - will be covered by system-tests

peterbarrow (approved):

PR #2185, opened 2020-09-02
Feature/2124 implement ratelimit

qustavo:
Turns out that we need to whitelist the validators node from being rate-limited.
This PR adds that functionality


ashleyvega (approved):
LGTM

PR #2194, opened 2020-09-04
do not validate marketID and orderID by default in cancel order

jeremyletang:
This PR remove the protobuf validation from the API, so we can specify 1 or more parameters instead of all mandatory.


edd (changes requested):
One question that I don't understand, and I suspect the GraphQL schema needs an update to match the new validation (but I haven't checked)

ponthief (approved):
looks good already tested on branch

edd (approved):

EVODelavega (approved):
LGTM

PR #2206, opened 2020-09-09
remove stats debug log

qustavo:


ponthief (approved):
Ok - was making too much noise in stdout

PR #2207, opened 2020-09-09
Feature/2005 reload configuration on new blocks

jeremyletang:
Ensure the configuration are reloaded wat the end of the blocks + add some more configuration hot reloads in missing places

close #2005


peterbarrow (approved):
Nonblocking: Code looks fine, but lots of it is the same, shame it can't be refactored out into some sort of base class?

PR #2209, opened 2020-09-09
add missing enum gql in mapping function for proposal rejection reason

jeremyletang:
close #2132


edd (approved):
@mattrussell36 worth a very quick skim

PR #2222, opened 2020-09-10
load initial assets from genesis state

jeremyletang:
This configure the initial set of assets through the genesis block state.
This is still imperfect for a few reason:
- the it's handled for now in thee cmd package mostly (which may stay as this though)
- as we need to instanciate markets with assets ID, the markets aren't instanciated until the assets ar loaded.

closes #2085

### DIFF
--- a/assets/config.go
+++ b/assets/config.go
@@ -1,27 +1,17 @@
 package assets
 
 import (
-	"io/ioutil"
-	"os"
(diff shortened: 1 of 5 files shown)


ashleyvega (approved):
LGTM, possibly after fixing 1 comment

PR #2233, opened 2020-09-13
remove generate method from execution engine

jeremyletang:
closes #2156


peterbarrow (approved):
Non-blocking question: We are sending out a marketdata event for each market every time we receive a new block from tendermint. Should we be linking block time to market updates so tightly?

PR #2235, opened 2020-09-13
load chainid + bridgeAddress from genesis state instead of config

jeremyletang:
close #2225


peterbarrow (approved):

ashleyvega (changes requested):
One small change requested.

ashleyvega (dismissed):
One more change requested (config removal)

PR #2237, opened 2020-09-13
add fees payed by distressed traders in the trade between distressed trader and the network

jeremyletang:
During position resolution, in order to 0 out the position of all distressed parties, the network is placing an order in the orderbook of the market against the good trader.

This order is backed by monies from the distressed trader, and a fee is payed as this is an aggressive order.

However, while the funds comes from the distressed traders, it looks like the fee is paying paid by the network(which is actually placing the order).

The details of the fees, paid by each distressed traders now appear in the trade between the network and the distress trader when 0ing out the position.

Adding all the fees paid by the distressed traders to the network, should give an equal number to the fees paid by the network to the good traders.

close #2129

### DIFF
--- a/execution/market.go
+++ b/execution/market.go
@@ -1122,7 +1122,7 @@ func (m *Market) resolveClosedOutTraders(ctx context.Context, distressedMarginEv
 	asset, _ := m.mkt.GetAsset()
 
 	// pay the fees now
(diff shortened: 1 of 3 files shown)


ponthief (approved):
seems to be Ok...will confirm with the automated system-test

PR #2244, opened 2020-09-14
Feature/2230 remove party id from commands

jeremyletang:
close #2230


peterbarrow (approved):

edd (approved):
Well this will need some changes on the console..

PR #2245, opened 2020-09-14
positions plugins, rlock for all the scope of the functions

jeremyletang:
fix concurrent map access in position plugin.


peterbarrow (approved):
Non (b)locking question: Is it normal to use defer to schedule the unlock instead of using the explicit unlock? I'm worried that a defer will result in the lock being held for a longer time than required, even though I understand it is safer.

qustavo (approved):

PR #2247, opened 2020-09-14
better logs when signature verification fail in propagate chain event

jeremyletang:
don't log signature verification failure stragiht away.


ponthief (approved):
Ok..looks like it's removed

PR #2248, opened 2020-09-15
revert event based stats

qustavo:
After thinking about it, we'll have to revert this.
Given the nature of the event based stats, we are missing all the replay events thus, stats will appear incomplete.


nzbaen (approved):
Tested on local env and statistics were reported correctly

PR #2255, opened 2020-09-16
Fixing some GQL/gRPC event bus stream issues

EVODelavega:
> So when you send that GQL query, we're creating a new subscriber that receives all events of the specified types and buffers them. Once there's data ready, we try to push that on a blocking internal channel. If that channel isn't being read, we skip that data set-and retry. As soon as we've hit a certain number of retries, we assume the client isn't receiving any more data, and we close the stream. Although I'm not 100%, the retry limit might be a problem here. Even if it isn't, it's a variable that I'd prefer to see removed for now, so it's one less thing to worry about.

Slack ref: https://vegaprotocol.slack.com/archives/C9E6XA8GK/p1600249413031800?thread_ts=1600247729.016400&cid=C9E6XA8GK

Closes #2260 & Closes #2259

### DIFF
--- a/api/trading_data.go
+++ b/api/trading_data.go
@@ -1885,7 +1885,8 @@ func (t *tradingDataService) ObserveEventBus(in *protoapi.ObserveEventsRequest, 
 			filters = append(filters, events.GetPartyIDFilter(in.PartyID))
 		}
 	}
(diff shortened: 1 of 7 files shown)


edd (commented):
Probably needs someone more Go than me to review to see if it fixes the issues. I've linked #2259 and #2260 - does this fix any others?

qustavo (approved):

PR #2293, opened 2020-09-24
Specify early returns + unit test conventions & explain why a bit

EVODelavega:


peterbarrow (approved):

qustavo (commented):
I've seen this pattern across all the code base, I'm glad someone put it together.
Although mocks are useful in some cases, using them by default when injecting dependencies seems to be an anti-pattern. Was this a concern when deciding this?

PR #2318, opened 2020-09-29
improve documentation for governancedata proto maps (yesParty, noParty)

jeremyletang:
close #2268

### DIFF
--- a/proto/api/trading.swagger.json
+++ b/proto/api/trading.swagger.json
@@ -2846,14 +2846,14 @@
           "additionalProperties": {
             "$ref": "#/definitions/vegaVote"
           },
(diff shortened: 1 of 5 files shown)


candida-d (changes requested):
Looks good, thanks for doing this! 

Three (small) changes: 
- where you write "casted", change it to "cast". (yes I know English is weird) 
- change (publickey) to (public key) 
- when this text is all on one line, add a comma after "value" and before "(Vote)" as I've done here: 
key (string) is the party ID (publickey) value, (Vote) is the vote casted by the given party

qustavo (approved):

PR #2329, opened 2020-10-01
Unify tx related func into its own pkg

qustavo:


jeremyletang (approved):
Very nice, just one a small comment. good to merge then.

PR #2330, opened 2020-10-01
Added new auction types

peterbarrow:
Added new auction types for PRICE, OPENING, LIQUIDITY and FBA so that we can return better information to clients that request the market state.

Closes #2227


edd (approved):
Nice - thank you.

Note to self: Argh, refactor.

PR #2334, opened 2020-10-02
rewrite vegastream to use eventbus

jeremyletang:
redo vegastream so you can stream all events from a given node using grpc (filtered optionnaly with market/party)


peterbarrow (approved):
Well that's a lot more tidy.
One non blocking question, I thought we just removed the ability to subscribe to all events?

PR #2335, opened 2020-10-02
ABCI records to a path

qustavo:
When path is specified it will use it to create a file with the format `abci-record-<timestamp>`


peterbarrow (approved):

jeremyletang (changes requested):

ashleyvega (changes requested):

ashleyvega (approved):
LGTM

jeremyletang (approved):

PR #2340, opened 2020-10-03
clear opening auction only once the very first LeaveAuction call is processed fully

jeremyletang:
Right now, the call isOpeningAuction is checking if there's opening auction config in the market config to know if it's currently ongoing an opening auction.
Also as soon as we enter the LeaveAuction call, we call isOpeningAuction, and if the return value is true, we straight away set the opening auction config to nil.
By doing this, when finishing to process the LeaveAuction function, and going through the applyFee function, the market thinking it's in continuous mode, and never process the fees as in opening auction mode (no fees), but as in a continuous market.

close #2339

### DIFF
--- a/execution/market.go
+++ b/execution/market.go
@@ -536,11 +536,6 @@ func (m *Market) EnterAuction(ctx context.Context) {
 
 // LeaveAuction : Return the orderbook and market to continuous trading
 func (m *Market) LeaveAuction(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


peterbarrow (approved):
This works for now but in a few days we will have a better state system so we can reference that to check the state of the market. The line where we set m.tradeMode will need to be moved lower too when that happens.

PR #2369, opened 2020-10-09
Hotfix/2367 event bus

EVODelavega:
To cut down on the CPU utilisation in the positions stream, we're using a buffered channel.
After finding out about the use-case that had someone stream positions per party, we've made the party ID optional, and added an optional market ID (to stream all positions in a given market).

The main event stream is now also going to increase its internal channel buffer depending on the events subscribed to, so filtering received events and sending them is significantly less likely to result in events being dropped.

The event stream API will, eventually, support dynamic batching of events, but that's more of a feature and will be implemented as a separate PR. To address the underlying cause of issue #2367 takes a bit more work, but this fix takes away the reason for a lot of the more heavyweight event streams.


edd (commented):
Looks good to me - but I think someone more competent needs a look.

jeremyletang (approved):

PR #2387, opened 2020-10-13
Apply request rate-limiting and allow-list to faucet and wallet (#2384, #2388)

ashleyvega:
This PR applies request rate-limiting to the wallet, in the same way as happens for the faucet.

- Wallet creation is done by source IP address (from either in `request.RemoteAddr` or `request.Header["X-Forwarded-For"]`). The default is max 1 wallet creation per minute per source IP address
- Keypair creation is done by IP address and wallet name. The default is max 1 keypair creation per wallet per minute.
- To avoid circular imports (wallet / faucet, which both need rate-limiting), rate-limiting has been moved into a new `http` package.
- Tests added:
  - `t.Run("create wallet fail rate limit", testServiceCreateWalletFailRateLimit)`
  - `t.Run("gen keypair fail rate limit", testServiceGenKeypairFailRateLimit)`
  - `TestRateLimit`

This PR also adds an allow-list for the faucet and wallet. The default config contains standard IP ranges: `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `fe80::/10`

Closes #2384
Closes #2388


EVODelavega (changes requested):

jeremyletang (changes requested):
So it seems that this change the behaviour of the faucet.
Until now the rate limit was working in the faucet for a given party, and a given asset.
Meaning that you could request as a party, a given asset every X time. this would have work with any IP.

It seems now that we are sending to NewRequest, the party and the asset concatenated, then concatenate this internally with the IP, meaning that I could top up unlimited time if I'm sending request from any IP in the allow list.

It also means that we break the behaviour of the faucet, as any non allow listed IP, will not be able to emit a request to mint fund through the faucet.

I may have missed someting from the code, but if I understood this well, this will need to be reverted.
E.g: I most often run a local wallet when testing stuff in dev, and use call the faucet from my machine to topup funds to a pubkey that I manage locally. I believe I'm not the only person to do that, and this change would stop us to do this.

PR #2404, opened 2020-10-14
add fix for leave auction

jeremyletang:
closes #2385 

@3jtechtest could you check that the code does not panic anymore in uncrossing at Leave of auctiton?


peterbarrow (approved):

ponthief (commented):
tested this branch and it's no longer crashing on closeout traders during an auction

PR #2434, opened 2020-10-19
try fix average netorder price

jeremyletang:
closes #2180


ponthief (approved):
retested in a branch - avg. price calculated correctly

PR #2447, opened 2020-10-20
fix generation of network parameter rest GET API

jeremyletang:
fix REST API for the network parameters
POST -> GET


ashleyvega (approved):
LGTM

PR #2453, opened 2020-10-21
remove baseName from instruement

jeremyletang:
close #2435


ponthief (approved):
tested with both buildVerify and other system-tests..market creation is not negatively impacted

PR #2463, opened 2020-10-22
Fix event bus streaming (GQL/gRPC)

jeremyletang:
This bring a bunch of fixes to the event bus.

- gRPC / Graphql no works properly with no batch (batch size 0)
- batches size of 1 / small batches works to, but are not recommended as the server expecting acknowledgment when working with batches, the node won't send a second batch(of 1...) before the acknowlegement.
-  batches of any sizes works
- no connections drops after 1 batch in GQL
- now the event bus stream will retry for 2 seconds before droping a connection (almost instant before), it will do 200 retries, with a tick of 10ms each if the channel is not ready to read the data.
- The internals of the event bus did not remove properly the subscribers from the event types -> subscribers map, which means that subscribers where being broken if e.g: first someone subscribed for all events, then close the connection, then another connection open, and the same ID would be reused internally, but the list of subscribtion would have been different.
- Also full batches would have been dropped before if it was not possible to write in the stream. Now if we cannot write after 10ms, we just concatenate the next batch into the current one. meaning that if the client is too slow, and requested batches of 10, he may get batches of 20 if he was not able to consume the previous batch straight away. which also mean that he's jst accumulating delay, but that's the client problem, he'll get closed at some point.

closes #2323
closes #2444 
closes #2336

### DIFF
--- a/api/trading_data.go
+++ b/api/trading_data.go
@@ -2073,12 +2073,10 @@ func (t *tradingDataService) ObserveEventBus(
 
 	if req.BatchSize > 0 {
 		err := t.observeEventsWithAck(ctx, stream, req.BatchSize, ch, bCh)
-		fmt.Printf("ERR EVENT BUS ACK: %v\n", err)
 		return err
 
 	}
 	err = t.observeEvents(ctx, stream, ch)
-	fmt.Printf("ERR EVENT BUS ACK: %v\n", err)
 	return err
 }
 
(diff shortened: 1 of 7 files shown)


EVODelavega (approved):

edd (changes requested):
**suggestion (blocking)**: Add a PR description for what this changes, and a sensible title, and an issue link

edd (approved):
Thanks for the ticket update

PR #2479, opened 2020-10-26
Ensure the api-node runs validation on external resources as well.

jeremyletang:
By fixing the non-validators / validators behaviour in the ttopology, we also disable some checks in the ExtResChecker type.
By doing so, the APIs nodes are not validating the external resources anymore, and then not retrieving information / storing information about them.
Meaning that he resource is considered OK, but the state in the banking package is never updated.

We disable the check for validator in the ExtResChecker when starting the routing to check the resources, but keep it enabled at the time we send the node vote.

Which basically means most of the behaviour apart voting is the same than in validators.

close #2478

### DIFF
--- a/validators/ext_res_checker.go
+++ b/validators/ext_res_checker.go
@@ -172,13 +172,10 @@ func (e *ExtResChecker) StartCheck(
 
 	e.resources[id] = rs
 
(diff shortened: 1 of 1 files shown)


ashleyvega (approved):
LGTM

PR #2493, opened 2020-10-27
Add target stake calculation engine

witgaw:
Adds a target stake calculation engine that and exposes it via GetMarketData().

Closes #2421
Closes #2422


edd (commented):
Graphql looks nice :+1:

EVODelavega (approved):

PR #2502, opened 2020-10-28
fix pointer issue when amendingInPlace

jeremyletang:
In amend in place, by amending the pointer, we change the content of the original order, by the content of the amended pointer (the order pointer was originaly retrieve at the begining of the amend).
By doing so, check at the end of amend order to insert / expire an order never worked, as it was always comparing the state of the amended order with itself.

We now make a copy of the orginal order before executing the checks / amends, and then do the stuff.

closes #2500

### DIFF
--- a/matching/2500_GTT_amended_to_GTC_expires_test.go
+++ b/matching/2500_GTT_amended_to_GTC_expires_test.go
@@ -1,0 +1,42 @@
+package matching
+
+import (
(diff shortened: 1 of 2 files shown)


EVODelavega (approved):

ponthief (commented):
tested on the branch...passed

PR #2509, opened 2020-10-28
panic when genesis load produce error

jeremyletang:
We panic when the genesis state cannot be loaded by one of the package / engine which require state from genesis.
Add more details to the error when an asset cannot be loaded.


ashleyvega (approved):
LGTM

PR #2510, opened 2020-10-29
remove the external risk model

jeremyletang:
This is just a simple code cleanup.
- remove the ExternalRiskModel from proto
- remove the ExternalRiskModel from the risk pacakge

This was buil initially to work with the external risk models in python.
These models are not a thing anymore, so there's not reason for this to stay in the codebase.


qustavo (approved):
:tada:

PR #2523, opened 2020-10-31
apply some lints from staticcheck and change some log levels

jeremyletang:
These are just small changes after looking at the pprof profiles and staticcheck reports.
pprof: basically just adjusting some log levels in a few places.
staticcheck: mostly the code cleanup ones, unused code, inefficient statement.


qustavo (approved):
This is great! :tada:

PR #2528, opened 2020-11-02
Refactor price monitoring data structures 

witgaw:
Refactors DTOs used by price monitoring engine so that the triggers are a market parameter and update frequency is the network parameter. 
It also adds the network parameter containing default triggers which are to be used when no market parameter containing these is provided.

Closes #2504


edd (changes requested):
This PR needs a description

jeremyletang (approved):

edd (commented):
GraphQL still looks great

edd (approved):

PR #2530, opened 2020-11-03
Define governance asset in genesis block

jeremyletang:
This add a few things:
- asset defined in the genesis block require an ID.
 - the one defined with the genesis command already have an ID.
- the governance asset is defined now as a normal asset in the list of genesis asset
- a new network parameter is introduce to specify what asset is the governance asset (setup to VOTE token initially).
- new function to dispatch, and add a custome validation rule for each values (this was needed so we can add non static checks. e.g: for now we could check simple rules like a string is not empty, etc, now we can introduce more complex checks like ensure that this ID exists in the collateral, or in the assets list, etc).

close #2515 
close #2391

### DIFF
--- a/assets/assets.go
+++ b/assets/assets.go
@@ -8,10 +8,10 @@ import (
 
 	"code.vegaprotocol.io/vega/assets/builtin"
 	"code.vegaprotocol.io/vega/assets/erc20"
+	"code.vegaprotocol.io/vega/crypto"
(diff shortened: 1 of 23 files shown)


jeremyletang (commented):
just adding some details on the PR

EVODelavega (approved):

ashleyvega (approved):
LGTM

PR #2535, opened 2020-11-04
Switch system-tests to use develop branch

ashleyvega:
This PR updates CI so that system-tests uses the `develop` Docker image tag, which comes from the `develop` branch of the system-tests repo.


ponthief (approved):
just what we need

PR #2540, opened 2020-11-05
Adding pegged order amends

peterbarrow:
We missed out amends when adding pegged orders.
This PR adds the ability to amend the pegged order offset and reference.


edd (commented):
GraphQL side all looks great 👍

EVODelavega (approved):

PR #2555, opened 2020-11-09
Add proposal ID to the Vote type

EVODelavega:
The GQL vote type was missing the proposal ID field. This PR adds it

I've noticed we're using a `ProposalVote` type in other places which is the old `Vote` type + a proposal ID field. We should be able to unify the two types now. I can do this as part of the PR, if that works for @edd and @mattrussell36

### DIFF
--- a/gateway/graphql/generated.go
+++ b/gateway/graphql/generated.go
@@ -686,9 +686,10 @@ type ComplexityRoot struct {
 	}
 
 	Vote struct {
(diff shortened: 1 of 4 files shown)


qustavo (approved):

jeremyletang (changes requested):
Wait, I believe we will want a Proposal here, not a proposalID

jeremyletang (approved):

PR #2560, opened 2020-11-10
Added pegged order details to GraphQL 

peterbarrow:
While adding pegged orders to the system I forgot to add the new fields to GraphQL
This PR addresses that mistake


edd (approved):
The documentation on this is great, thanks

EVODelavega (approved):

PR #2576, opened 2020-11-12
Using static best bid/ask price for pegged order repricing

peterbarrow:
We were using the normal best bid/ask price for repricing pegged orders but this caused problems when there are only pegged orders in the book. We have added a new set of functions to return the static best bid/ask which will only look at persistent, non pegged orders.


edd (commented):
**question (non-blocking)**: do i as and end user care about the static best bids the ... best bid? I assume no, and in that regard the property naming in GQL is fine, but would there be a good reason for be to fetch the static one?

jeremyletang (approved):

EVODelavega (approved):

PR #2582, opened 2020-11-13
sort slice in GetHorizonYearFractions pricemonitoring

jeremyletang:
The GetHorizonYearFracions in pricemonitoring expect the slice to be sorted.
Although the slice is created from a map, which are unordered in Go, therefore the slice can't be guaranteed to be sorted if we do not sort it.


ashleyvega (approved):
LGTM

PR #2598, opened 2020-11-17
Refactor and extend price monitoring integration tests

witgaw:
This PR modifies and extends the price monitoring integration tests as discussed with @davidsiska-vega.

integration/features/price-monitoring-for-system-test.feature is meant to be used as a single test case to be implemented via system-tests. It's been chosen such that multiple code branches are explored and the trigger values / auction extensions have been chosen so that the entire run can be done in under a minute.


davidsiska-vega (approved):
Pls change "forward risk model" to "lognormal".

PR #2618, opened 2020-11-23
disable all subscribers for validators

jeremyletang:
This update the disable storage to work with the broker and storages now.
If stores-enabled=false, then no subscribers is instanciated, and set to subscribe to events in the broker

close #2619


ashleyvega (approved):
LGTM

EVODelavega (approved):

PR #2626, opened 2020-11-23
add a transaction submit type to graphql

jeremyletang:
This adds a transaction submission type to the graphql schema.
Default to old implementation -> Async submit.


edd (approved):
:+1: I was going to criticise the 'type' field's documentation, but the `SubmitTransactionType` docs are so clear there's no point. Nice.

PR #2666, opened 2020-12-01
use exponential backoff for retries in ext_res_checker

jeremyletang:
This adds a exponential backoff instead of a linear 500ms retry for ext_rest_checker.

Also adds a new counter for the request made to ethereum.

close #2663


ashleyvega (commented):
non-blocking changes requested

ashleyvega (approved):
LGTM

PR #2669, opened 2020-12-01
add some mutexes to prevent panic + metrics for evt forward

jeremyletang:
this adds a new metric to keep track of the number of evt forwarded by nodes, and a mutex to ensure a map is not accessed concurrently


peterbarrow (approved):

ashleyvega (approved):
LGTM, but wait for opinions from other devs.

PR #2707, opened 2020-12-05
Feature/vega verify

jeremyletang:
This is a spike PR for a vega verify subcommand.
The goal of this subcommand is to verify some kind of things vega use / needs / expect in order to make building against vega easier.
As an example two verification have been implemented:
- asset proposal
- genesis file

Ideally we would implement verification for all Commands (SubmitProposa, SubmitOrder, etc, etc) and the genesis file at least.

The asset proposal verification is built to verifyng payloads expected from the rest API (which is the more prone to be given wrong inputs).

The commands will then emits, warning for non mandatory things, and errors for something making the actual payload unusable.

Example usage for an asset proposal:
```
cat asset-proposal.json
{
    "reference": "some-reference",
    "partyID": "5cfa87844724df6069b94e4c8a6f03af21903d08e4251043ee9f7c",
    "terms": {
	"enactmentTimestamp": 1,
	"validationTimestamp": 1,
	"newAsset": {
	    "changes": {
		"erc20": {
		    "contractAddress": "0x323b5d4c32345ced7739f346429d"
		}
	    }
	}
    }
}
```

```
$ vega verify passet asset-proposal.json
asset-proposal.json: KO
{
  "ID": "",
  "reference": "some-reference",
  "partyID": "5cfa87844724df6069b94e4c8a6f03af21903d08e4251043ee9f7c",
  "state": "STATE_UNSPECIFIED",
  "timestamp": "0",
  "terms": {
    "closingTimestamp": "0",
    "enactmentTimestamp": "1",
    "validationTimestamp": "1",
    "newAsset": {
      "changes": {
        "erc20": {
          "contractAddress": "0x323b5d4c32345ced7739f346429d"
        }
      }
    }
  },
  "reason": "PROPOSAL_ERROR_UNSPECIFIED"
}
warn: proposal.partyID does not seems to be a valid party ID
error: prop.terms.closingTimestamp is missing or 0
warn: prop.terms.validationTimestamp may be in the past
warn: prop.terms.enactmentTimestamp may be in the past
warn: prop.terms.newAsset.changes.erc20.contractAddress may not be a valid ethereum address
error: one or more file are ill formated or invalid
```


qustavo (changes requested):
I think this is really useful :heart: 
I suspect that keeping sync between validation rules across the code and here might be hard to enforce.
How hard would it be to unify these validations into a single place (embracing a single-source-of-truth) and call them from different places.

EVODelavega (approved):

PR #2715, opened 2020-12-07
Modify price monitoring behaviour in line with spec change

witgaw:
Modify price monitoring behaviour as per https://github.com/vegaprotocol/product/issues/435: modify the engine to use the reference price (past price per trigger) rather than latest price when syncing with the risk model and to store the moves as multiplicative (rather than additive) factors internally. Also added reference price caching not to look it up more than once.

Closes #2714


davidsiska-vega (commented):
Looks good to me.

jeremyletang (approved):

PR #2730, opened 2020-12-11
Feature/update deposit proto

jeremyletang:
This is quite difficult to test as it's pretty much I/O bounded, but I tested the validators locally against an event queue using ropsten, and all was fine.

close #2728


peterbarrow (approved):
I don't understand what the code is doing but it looks clean and tidy so I'll approve.

PR #2737, opened 2020-12-14
Added code to catch amending the price of a pegged order

peterbarrow:
This PR prevents a user from amending a pegged order's price.


edd (approved):
GraphQL side and logic looks good :+1:

jeremyletang (approved):

PR #2765, opened 2020-12-18
Send Market creat event only once

jeremyletang:
change market event to be market created, sent only once
add market updated event sent on each update

close #2725


edd (approved):
Looks good to me, @EVODelavega feedback aside

EVODelavega (approved):

PR #2767, opened 2020-12-18
make the parked order price explicitly 0

jeremyletang:
close #2748


ponthief (approved):
retested on a branch and it has Passed

PR #2774, opened 2021-01-04
fix cancell all order (don't iterate + remove stuff from slice)

jeremyletang:
close #2771


ponthief (approved):
retested on the 2771 branch and it passed

PR #2778, opened 2021-01-05
Use buf for the protobuf workflow

qustavo:
Closes #1727


jeremyletang (commented):
Also it appears there' as much added line than removed line in the html generated docs? any idea? Github cannot load the diff as it's too big.

jeremyletang (approved):

PR #2795, opened 2021-01-07
grammar fix

candida-d:
- delete extra word from proposal failure description
- capitalise ID in another description


edd (approved):
**praise (non-blocking)**: Thanks!

PR #2804, opened 2021-01-09
various confirmation / eth related improvments

jeremyletang:
use networkID not chainID to check network
add defaults for the ethereumCOnfig
fix confirmations comparisons in erc20 requests...

close #2803
close #2802


ashleyvega (approved):
LGTM

PR #2821, opened 2021-01-13
Market lifecycle updates for NewMarket proposals

jeremyletang:
add new market proposal handling

new way of starting the openign auction

new way of creating market with liquidiy provision order (only when submitted via governance as of now)

fix all existing test (all test pass with all the changes actually yea

close #2787


edd (approved):
I’ve taken a run at this and it looks sane to me. @gchaincl would be best advise to review the go.

Only comments are non-blocking and trivial

PR #2827, opened 2021-01-14
Feature/2822 merge expiring orders lists

jeremyletang:
close #2822


peterbarrow (approved):
Why do the helper functions `IsExpirable` and `IsFinished` live in the proto/vega_ext.go file? 
Can we put them closer to the object they refer to?

PR #2844, opened 2021-01-19
Release v0.30.0

ashleyvega:
Closes #2841.


edd (changes requested):
**chore (blocking)**: Changelog is out of date. GitHub has been having some cache issues recently so that could be the cause, but my last edit was 9 mins ago, which is not reflected here (https://github.com/vegaprotocol/vega/issues/2841#issuecomment-762861985 - the telltale :round_pushpin: in `.spelling` gave it away)

edd (approved):
Excellent. Now round_pushpin is valid spelling.

PR #2867, opened 2021-01-27
deposits.CreditedTimestamps fixes

jeremyletang:
use creditedTimestamps in gql resolve (instead of createdTimestamp)
update creditedTimestamp when deposit is finalized in vega

fixes #2801


edd (approved):
I heartily endorse this event or product

PR #2879, opened 2021-01-28
Apply recommendations from buf to protobuf files

qustavo:
This PR:

- removes exclusions so checks are included:
  - FIELD_LOWER_SNAKE_CASE
  - ONEOF_LOWER_SNAKE_CASE
  - IMPORT_NO_PUBLIC
  - ENUM_NO_ALLOW_ALIAS
- leaves unchanged:
  - FILE_LOWER_SNAKE_CASE

## to be accounted for
- PACKAGE_VERSION_SUFFIX: As we are not in shape of providing a stable v1 release. This will/should be enabled when we have v1 candidate 

Closes #2815.


ashleyvega (approved):
~LGTM~

(changed my mind)

PR #2887, opened 2021-01-29
Normalise oracle data using oracle.Adaptors with support for Open Oracle standard

ValentinTrinque:
When an oracle relay submit new oracle data, the transaction is deserialised and normalised with its corresponding adaptor.

The adaptor to use is resolved by looking at the type of the data inside the oracle data. If the type is "Open Oracle", then the adaptor responsible to normalise Open Oracle data is used.

**Check transaction**
The data is verified. Errors are forwarded.

**Deliver transaction**
Once the data is verified, it's broadcasted to the oracle engine (yet to be implemented).

closes #2875


qustavo (approved):
This is great!

PR #2917, opened 2021-02-04
Add LP penalties and closeout handling

witgaw:
Add Liquidity Provider closeout handling and applies penalty when bond account used to cover margin requirements.


witgaw (commented):
Ready for review.

EVODelavega (changes requested):

PR #2959, opened 2021-02-10
Wire OracleEngine to Future

ValentinTrinque:
Future needs to subscribe to the oracle engine. As a result, their declaration should include oracle spec and oracle spec binding.

Close #2866


edd (commented):
The GQL looks good to me :+1:

edd (commented):
GQL side looks good to me

jeremyletang (approved):

PR #2996, opened 2021-02-18
Add OracleSpec proto for API query

ValentinTrinque:
We want to expose all active oracle spec on the network through API.

Close #2866


edd (changes requested):
Add some requests on the GraphQL API

edd (approved):

PR #3029, opened 2021-02-24
add a reference thing in liquidity provisions

jeremyletang:
close #2945


edd (commented):
GQL looks incredible to me

peterbarrow (approved):

PR #3030, opened 2021-02-24
Add debug logs on execution engine

ValentinTrinque:
Add debug logs on execution engine.

Close #2881


jeremyletang (approved):
OK to merge when tests pass.
Also I suppose we could add logs in the market.go. But knowing we want to do some major cleaning in there, I'd wait for that to be done first.

PR #3035, opened 2021-02-25
fix documentation for infra fee

jeremyletang:
close #2807


candida-d (changes requested):
One small change - missing word

candida-d (approved):

PR #3072, opened 2021-03-02
Remove staticcheck ignore directive

qustavo:
This PR removes an old `staticcheck` check (`SA4006`) that was quietly ignored in version `2020.1.4`, but causes a failure (exit code 1) in version `2020.2.2`.

See #3070 for background.


ashleyvega (approved):
LGTM

jeremyletang (approved):

PR #3075, opened 2021-03-03
fix bond account release on cancel

jeremyletang:
close #3073


qustavo (approved):
When creating LP in the test, using [WithSubmittedLiquidityProvision](https://github.com/vegaprotocol/vega/blob/develop/execution/market_test.go#L377) might help to improve readability and/or make the test more compact.

PR #3081, opened 2021-03-03
Feature/fix fee check lp orders

jeremyletang:
This should fix the fee validations for LP orders


ponthief (approved):
tested and it passed in system-tests

EVODelavega (approved):

PR #3094, opened 2021-03-05
update gql schema to use ID schalar + rename *ID -> *Id

jeremyletang:
This do a tiny bit more than what 3077 expected which was to update ID/Id to be used consistently.
It also change all IDs kinds fields types from String to the ID scalar.

Let's maybe not merge this straight away as it may break the front-end?

close #3077


ValentinTrinque (approved):

edd (approved):
Nice! Not for future me: breaking change.

PR #3105, opened 2021-03-06
add min_lp_stake to assets + change assets layout slightly

jeremyletang:
Add the min lp stake to the asset
Also move most of asset in the AssetDetail sub struct (proto). This allow us to create a proposal requiring all details of the assets as used in the network.
Also change a tiny bit the genesis state

close #2922


edd (commented):
GQL side looks good to me

EVODelavega (approved):

PR #3111, opened 2021-03-08
Adding new timestamps to market details (#2993)

peterbarrow:
Adding extra timestamp information to the APIs that get market information.


edd (approved):
Thanks, I like the encapsulation of dates in to a separate object in the GQL

jeremyletang (approved):

PR #3126, opened 2021-03-09
Add transfer for bond slashing 

EVODelavega:
Currently in margin calculation stuff, want to update this and move it out to a separate flow based on return values

Will also need to add some test cases


edd (approved):
Cowboy review

PR #3140, opened 2021-03-10
do not return an error if CancelAllOrders returns ErrTradingNotAllowed

jeremyletang:
close #3139


ponthief (approved):
tested on the fix branch - passed

PR #3177, opened 2021-03-12
Implement domain types packages

jeremyletang:
I've been trying to think how to approach the refactoring of Uint a little recently.
And mostly initially, how to introduce the domain type package, without producing massive changes due to:
- introducing new types (e.g: proto.Order vs domain.Order).
- keep compatibility with the enums
- not have to import both the domain type and the proto types for the enum mostly

What's the approach here:
- we type alias all the type required (e.g for matching Order, Trade, etc)
- twe type aliase the enums (Side, etc)
- in the same package we redefine all the enum variants with the exact same name (copy pasta of the proto ones)
- then we can safely change the import to the new domain type

This is a first step in order to introduce this new package steps by steps.
We also start to introduce types which will surely needs to be re-defined (e.g order, trade, fee, initially with a copy of all the proto fields), so we can then easily replace the price which is a uint64 now, for a uint256, without introducing all the other changes at the same time.

By doing this we should be able top introduce the changes quite incrementaly.


EVODelavega (approved):
LGTM

ValentinTrinque (approved):

PR #3223, opened 2021-03-19
move all proto events into their own package

jeremyletang:
This move all events from the proto root into another package.

Reference issue: https://github.com/vegaprotocol/vega/issues/3222


ashleyvega (approved):
LGTM

EVODelavega (approved):

PR #3284, opened 2021-03-30
Removed debug code (#3283)

peterbarrow:
Removed some debug code that was setting the log level to DEBUG no matter what the config said.


jeremyletang (approved):
Good to go once all tests are passing.

PR #3298, opened 2021-04-01
implement a simple caching strategy for orderbook

jeremyletang:
This is a possible answer to the orderbook recalculating all the time indicative price and volume.

close #3297

### DIFF
--- a/execution/market.go
+++ b/execution/market.go
@@ -165,7 +165,7 @@ type Market struct {
 	markPrice uint64
 
 	// own engines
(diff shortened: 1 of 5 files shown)


ValentinTrinque (approved):
It would be less invasive if the cache was implemented as a decorator. Do you think that's possible ? Not blocking though.

ValentinTrinque (approved):

EVODelavega (approved):

PR #3308, opened 2021-04-02
Feature/3293 add probability of trading tau scaling

jeremyletang:
Add the network parameter
Propagate it down to the suppied liquidity engine
Use it to factor the Tau from the risk model beforesending to the GetProbability of trading calculation.

close #3293


witgaw (approved):
Looks good to me, just one minor comment, feel free to ignore - can always change it later if we need to.

EVODelavega (approved):

PR #3313, opened 2021-04-06
Propagate/wrap invalid account errors

ashleyvega:
This PR slightly improves error handling by propagating the ID of invalid account.

It also uses `errors.Wrap(err, "info")` to show where some errors came from.


jeremyletang (commented):
Can we avoid using errors.Wrap, and use fmt.Errorf("... %w ...") iunstead?

jeremyletang (approved):

PR #3329, opened 2021-04-08
Ensure amending an order does not trigger price auction

EVODelavega:
There seemed to be some issues when leaving an auction. In this PR:

* Auction events have an extended field, indicated whether or not the auction was extended
* Auction events are sent when the auction is extended
* Some changes when updating price monitoring - small fixes
* Added an integration test where we leave opening auction, amend an order, and no price monitoring auction is triggered


edd (commented):
GQL side looks good 👍

jeremyletang (approved):

PR #3330, opened 2021-04-09
test: refactor step to configure network params

cityhunteur:


jeremyletang (commented):
I know it's just a draft but just a few comments:
- table changes are great (no idea why it was not like this before).
- propagating the changes of the parameter, i would suggest the following approach:

1. Instanciate a network parameter store (netparams pacakge)
2. Set the callbacks to the execution engine onto the netparams store (like here: https://github.com/vegaprotocol/vega/blob/develop/cmd/vega/node/node_pre.go#L586-L649)
3. Update the netparams using the store `netparamsStore.Update(ctx, key, value)`
4. Dispatch the params updates to the execution engine `netparamsStore.Dispatch(ctx)`

Would that make senses to you?

ValentinTrinque (changes requested):

ValentinTrinque (approved):

PR #3341, opened 2021-04-12
Added logging for party without accounts (#3287)

peterbarrow:
When a transaction is sent but the party has no accounts we log a message so we can capture the error.


jeremyletang (approved):
Just a comment, don’t matter much

PR #3358, opened 2021-04-15
Added deep copying of event messages with unit tests

peterbarrow:
Adding deep copy code to the event wrapper functions and covering in unit tests to make sure I do not break anything.


jeremyletang (commented):
I think in general I would implement a DeepCopy function for all of the types we use, so it's consistent

jeremyletang (approved):

jeremyletang (approved):

PR #3364, opened 2021-04-16
Auditing the position tracking in the execution engine

jeremyletang:
This will add a couple of things:
- ensure that we are registering / unregistering positions in the rights places
- Add panics when needed as we should always be able to register / unregister positions in the engine.

I'll try to keep the commit message sensible, please review eventually them one by one.


EVODelavega (approved):

peterbarrow (approved):
Nonblocking: Is it possible to have a log object passed in at creation time for the positions type instead of passing it into the functions that use it?

peterbarrow (approved):

PR #3394, opened 2021-04-21
use default when instanciating markete if LiquidityMonitoringParameters is nil OR LiquidityMonitoringParameters.TargetStakeParameters is nil

jeremyletang:
When instanciating the market from the governance, we use default if the liquidity monitoring parameter was not specificied, but we do not check if what's inside was. We now check for both, and will use default if any isn't.

close #3393


ashleyvega (approved):
LGTM

PR #3399, opened 2021-04-22
Small fixes to the target stake test

jeremyletang:
- register all net parameters callbacks at startup of the tests
- exit opening auction in the feature tests


cityhunteur (approved):
Thanks for fixing this. Make sense 👍

PR #3403, opened 2021-04-22
started work on fee and rewarding feature + added new required steps

jeremyletang:
close #3174


davidsiska-vega (approved):
Looks great, let's just get rid of the log-normal risk model definition there because it's not needed and it's got crazy parameters.

ValentinTrinque (changes requested):

ValentinTrinque (changes requested):

ValentinTrinque (approved):

PR #3408, opened 2021-04-23
Improve API to return vote balance and weight

ValentinTrinque:
Close #2998
Also close #3292


edd (commented):
GQL side looks good to me. cc @campbellssource

jeremyletang (approved):

PR #3436, opened 2021-04-29
Implement proposal submission command

jeremyletang:
This introduce a ProposalSubmission command instead of using the Proposal type in the command. This is a first step to separate properly our Command to submit a new proposal, and our proposal model.

Here's the difference between both data structures:
```
message ProposalSubmission {
  string reference = 1;
  ProposalTerms terms = 2 [(validator.field) = {msg_exists: true}];
}

message Proposal {
  enum State {
    // all enum state fields
  }

  string id = 1;
  string reference = 2;
  string party_id = 3 [(validator.field) = {string_not_empty : true}];
  State state = 4 [(validator.field) = {is_in_enum: true}];
  int64 timestamp = 5;
  ProposalTerms terms = 6 [(validator.field) = {msg_exists: true}];
  ProposalError reason = 7;
}
```


ashleyvega (approved):
LGTM

peterbarrow (approved):

PR #3438, opened 2021-04-30
Move all vega commands into a commands subfolder / package in the proto package

jeremyletang:
This move all commands into a commands subfolder in the proto pacakge.


ashleyvega (approved):
LGTM, plus one comment

ValentinTrinque (approved):

PR #3447, opened 2021-04-30
Clean up the order submission

ValentinTrinque:
Closes #3441 
Closes #3448


jeremyletang (commented):
A few comments to address here. 
Also after a disussion with @mattrussell36l, what you suggested in this issue: https://github.com/vegaprotocol/vega/issues/3448, is OK for the frontend, so let's add the changes in this PR so you can cleanly implement passing the submission down to the Market.

jeremyletang (approved):

PR #3457, opened 2021-05-05
use bestBidPrice and bestAskPrice to calculate probability of trading in liquidity

jeremyletang:
close #3424


ashleyvega (changes requested):
Calls to `fmt.Printf()` should be removed.

EVODelavega (approved):

PR #3484, opened 2021-05-14
Potential changes to the coding spec

peterbarrow:
Here are my thoughts on the coding spec, mainly about unit tests and using the Panic log level.


ashleyvega (approved):
LGTM, but also non-blocking questions/clarifications.

PR #3507, opened 2021-05-24
test(accounts): add test for PartyAccounts

cityhunteur:
This is to review the approach for implementing an API tests.

Tasks

- [x] Refactor `testutil` for reuse
- [x] Implement test for at leat one Account APIs


jeremyletang (commented):
Had a very quick look, just 2 things:
- Let's revert the dependencies version for all the ones that changed, just to not introduce unexpected issue with this feature.
- I'd probably have a top level folder name something like api_integrations, with all the tests for them there. At the moment we bring logic testing grpc APIs in the services folders (e.g: accounts) which apart from the tests for the API do not use / invold grpc at all.

EVODelavega (approved):

PR #3522, opened 2021-05-27
Modify drone to trigger dockerised-system-tests in Jenkins

fkondej:
Modify Drone to trigger `dockerised-system-tests` in Jenkins instead of running `system-tests` in Drone.
`system-tests` are triggered after all the other checks are ok.
Also `system-tests` are not run if there were no vega core code changes.
The `dockerised-system-tests` Jenkins job is triggered the same way as GitHub webhook does it.

You can preview how it works in this PR, by checking the results for `Split too long line` and/or `fix drone variable` commits here.

### DIFF
--- a/.drone-github.yml
+++ b/.drone-github.yml
@@ -146,10 +146,15 @@ steps:
     image: alpine
     pull: always
     environment:
+      # Jenkins trigger url with token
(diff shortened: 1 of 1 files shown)


ashleyvega (approved):
LGTM

PR #3543, opened 2021-06-04
Extend auction interaction feature test

witgaw:
Extends and modifies the auction interaction feature test + implements functionality required to make it pass.

The main issue is that the auction trigger field in market data API doesn't always get modified to reflect the trigger that has extended the auction (https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0026-auctions.md#exiting-the-auction-mode).

This PR also brings implementation in line with spec changes, please see linked issues for details.

~Blocked by: https://github.com/vegaprotocol/vega/issues/3570 https://github.com/vegaprotocol/vega/issues/3578~

Closes: https://github.com/vegaprotocol/vega/issues/3047 https://github.com/vegaprotocol/vega/issues/3570 https://github.com/vegaprotocol/vega/issues/4223 #3362 https://github.com/vegaprotocol/vega/issues/5237


witgaw (commented):
Ready for review

EVODelavega (changes requested):
Because of the amount of commits, comments, and changes over the last week or so, this PR is getting rather difficult to review overall.

Trying to find a couple of comments that I still maintain ought to be addressed before this is good to merge was a PITA, but I believe there are a couple of things that haven't been addressed fully:

* Auction integration tests: we should at least have a few of them use blocks to progress time rather than setting time to a specific value.  In a comment I outlined the reasons for this, and I think changing the auction flow without changing the tests to progress time in a block-by-block way is creating a significant blind spot in testing.
* The use of indicative price & volume, specifically during opening auctions before the auction period has expired still looks to be calling price monitoring. In doing so, we're most likely providing price and volume information to the price monitoring engine, without any trades at a given price point actually happening. That might result in price bounds being incorrect.

IMO, for sanity/clarity reasons, it might be worth while closing this PR, and reopening it so we can get rid of the noise (comments on code that has since changed, things that have been addressed elsewhere, etc...), so we can go through it all again and just focus on the issues that remain

PR #3576, opened 2021-06-09
Updating liquidity module to use uint256 data type

peterbarrow:
Updating any price variables to use the new uint256 data type.


jeremyletang (changes requested):
Calculate Supllied stake will need to return num.Uint.
LiquidityProvision and LiquidityProvisionSubmission CommitmementAmount need to be changed  to num.Uint

jeremyletang (changes requested):
CommitmentAmont in the LiquidityProvision and LiquidityProvisionSubmission needs to be changed to num.Uint.

jeremyletang (changes requested):
Seeing that there's now a Decimal alias un the num package, let's change all import from  `github.com/shopspring/decimal` to `vegaprocol.io/vega/types/num` for the Decimal type

EVODelavega (approved):

PR #3581, opened 2021-06-09
Feature/settlement 256

EVODelavega:


jeremyletang (changes requested):
FinancialAmount + Transfer needs to be implemented anew in the t ypes packages, and use uint256

peterbarrow (approved):

PR #3583, opened 2021-06-10
Feature/collateral to uint256

cityhunteur:
TODO

- [x] Fix `TestMarginUpdateOnOrder`
- [x] Fix `TestMTMLossSocialization`


jeremyletang (commented):
OK, I started reviewing mostly the engine.go file.

A couple of comments:
- when returning an account or a balance we should clone the actual bigUint
- When passing a uint as parameter, eventually cloning too if needed.
- Also let's try to avoid unecessary allocation when possible for the equivalent of a += b.

EVODelavega (changes requested):

peterbarrow (approved):

EVODelavega (approved):

PR #3594, opened 2021-06-10
Update quant library to the latest tag

witgaw:
Update quant library to the latest version which contains the normalisation for probability of trading (when constraining with min/max price from price monitoring).

Closes #3759


jeremyletang (approved):

jeremyletang (changes requested):
Do not merge straight away until we have time to cleanup the imports again, it seems you are again importing all the library I removed from the PR earlier.

jeremyletang (approved):

PR #3676, opened 2021-06-30
add tm subcommand

jeremyletang:
This PR introduce a new `tm` subcommand which will allow use to run a tendermint node via the vega subcommand. Basically all the tendermint subcommand is  exposed as a vega subcommand.

In order to implement this I had to change a bit more code than expected so the default Help feature of the command line is disable, which was preventing the tm subcommand to work properly (hence all the Help field added in all existing command to support the -h/--help flags).

@ashleyvega / @fkondej you guys may be interested with this too.

close #3633


ashleyvega (approved):
LGTM. tested and working with `dockerisedvega`.

PR #3693, opened 2021-07-05
Calculation without decimals in liquidity target

jeremyletang:
Replacing the calculation using decimals by uint256 in the liquidity target package.

We are currently spending a considerable amount of time just converting from and into decimals / uint256 in this package.


witgaw (approved):
LGTM

PR #3698, opened 2021-07-06
refactor: update FromProto helpers from func to method

cityhunteur:
This makes it consistent with other FromProto helpers and
also adds couple of missing ones.


jeremyletang (commented):
Just a couple of comment, but seems good then.

jeremyletang (approved):

PR #3729, opened 2021-07-09
Update feature issue template

gordsport:
close #3728 

Updates to make the linking between issue/work item and the specs easier


cpuja84 (approved):
Looks good

PR #3730, opened 2021-07-12
added asset insurance account and distribute the funds market insuran…

ze97286:
…ce account to global account when market is closed + bug fix in markets integration

close #3707 
close #653


jeremyletang (approved):
Just added one last comment, once this is fixed it's good to merged, approving already.

PR #3736, opened 2021-07-12
Correcting event types and adding panics to catch mistakes

peterbarrow:
After we updated the proto types to domain types a few weeks back, we have broken some of the event handlers as they are expecting a different type of object. This was causing the issue shown in the ticket where the margin values were not being updated after they were initially set.

This PR fixes the wrong types and adds panics into the event handlers to catch wrong types being sent.

Closes #3618


jeremyletang (approved):
Sounds good, these changes will need to be backported to the api node.

cc @karlem @Cityhunteur

PR #3745, opened 2021-07-13
added loss socialisation to final settlement

ze97286:
close #1908


jeremyletang (commented):
I may have not seen that, but do we have tests coverage for cases where the insurance pool is empty, and loss socialization only is used?

peterbarrow (approved):

EVODelavega (approved):

PR #3791, opened 2021-07-20
Implemented market terminated state via the oracle trigger

ze97286:
close #2790
 #3708
#2788


jeremyletang (commented):
This sounds good to me.

@cdm @ashleyvega scripts for submitting markets for governance will need to be updated to keep working as now a second oracle_spec is required in the future product.

@ValentinTrinque @EVODelavega could you take a look at this PR too.

@vega-paul You may want to synchronise with @ze97286 as this PR will break all market proposal, therefore all system tests (bvt are not passing for now), but the change is minimal ultimaltely.

peterbarrow (dismissed):

ValentinTrinque (changes requested):

ValentinTrinque (changes requested):

ValentinTrinque (changes requested):

jeremyletang (approved):

PR #3811, opened 2021-07-27
added validation to market proposal to check for matching between ora…

ze97286:
…cle spec and binding property

close #3807


jeremyletang (commented):
This looks correct to me, @ValentinTrinque any comments on this?

PR #3818, opened 2021-07-28
More editions to the Jenkinsfile

jeremyletang:
Part of #3756.


fkondej (approved):
One question and suggestion. Other than that LGTM.

PR #3832, opened 2021-08-03
Update genesis topology map

karlem:
Part of #3826


jeremyletang (commented):
Looks good to me overall, just one small comment for @ValentinTrinque,

jeremyletang (approved):

PR #3847, opened 2021-08-04
modified dummy staking account to be backed by governance token accou…

ze97286:
…nt balance

close #3827


jeremyletang (changes requested):
That's been done only  in the vegabenchmark binary which is just for test/benchmark purpose, you will want to do that in the vega binary.

jeremyletang (dismissed):

jeremyletang (approved):

PR #3872, opened 2021-08-09
Import a nodewallet now copy the file instead of saving a path to orginal file

ValentinTrinque:
Close #3853


jeremyletang (dismissed):
Should we be printing the menmonic for the vega wallet when generating it? I think we should.

jeremyletang (approved):

PR #3953, opened 2021-08-27
Run integration tests in parallel and list failing feature files (#3952)

ashleyvega:
This PR runs integration tests in parallel and displays the name of failing feature files.

Closes #3952

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ test: ## Run unit tests
 
 .PHONY: integrationtest
 integrationtest: ## run integration tests, showing ledger movements and full scenario output
-	@./script/build.sh -a integrationtest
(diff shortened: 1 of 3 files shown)


witgaw (approved):
LGTM and works fine on my machine, nice one!

EVODelavega (changes requested):
So an entire bash script replacing a single command that (because of -v and `-godog.format=pretty`) provided much more useful output.

The integration tests are hardly a bottleneck to begin with. Maintaining bash scripts on the other hand is.

The integration tests take a couple of seconds to run. They produce useful output, and work just fine as they are. IMO, we already have far too many scripts and quirks in the vega repo. We should make an effort to _reduce_ the noise, not add to it.

PR #3963, opened 2021-08-30
Load tendermint logger config

ValentinTrinque:
Close #3962


jeremyletang (commented):
Couple of same issue. Seems good apart from that. Would be good to test with the closure issue we discussed before merging though.

jeremyletang (approved):

PR #4075, opened 2021-09-20
Making LNL easier to test

EVODelavega:
After catching up with @fkondej, we agreed to prefix checkpoint files with a YmdHms time, and set default checkpoint interval to 1m to make automated testing easier


ashleyvega (approved):
LGTM :-)

PR #4106, opened 2021-09-23
Do not fail when enabling assets multiple times

EVODelavega:
Network params and assets both contain the token asset. When reloading a checkpoint, we may enable the same asset twice. This used to result in a panic (failed to restore checkpoint), but it shouldn't.

Checkpoint reloading should ignore this error. We're not checking for the specific error now, because `collateral.EnableAsset` only ever returns the `ErrAssetAlreadyEnabled` error

Closes #4089


ashleyvega (dismissed):
Working with DV. LGTM.

ze97286 (approved):

PR #4133, opened 2021-09-28
Matching engine snapshots

peterbarrow:
Adding support for snapshots in the machine engine

Closes #4144


jeremyletang (commented):
Mostly cosmetics changes, thinking about the performances impact, without getting too far (e.g copy the whole range of order into the pricelevels instead of going through it one by one), I'd try at least to address the multiple iteration / unecessary clone I mentionned.

wwestgarth (dismissed):

wwestgarth (approved):

PR #4160, opened 2021-10-04
4151 snapshot positions engine

wwestgarth:
Add snapshot hashing/serialisation for the positions engine

close #4151


ze97286 (changes requested):
positions must be sorted, otherwise the hash and serialisation will be arbitrarily different.

ze97286 (approved):

PR #4161, opened 2021-10-04
assets snapshot implemented

ze97286:


wwestgarth (dismissed):
I think this makes a good template for how we should go about doing the other engines. I've copied a lot of the structure it for the governace one I'm working on!

peterbarrow (dismissed):

wwestgarth (approved):

PR #4185, opened 2021-10-08
dedup event sent when creating the reward pool + fix owner

jeremyletang:
Reward pool owner is systemOwner.
Do not send event twice when reward pool is created


ze97286 (dismissed):

vega-paul (approved):
LGTM

PR #4190, opened 2021-10-08
Markets snapshot

karlem:
Close #4165


EVODelavega (changes requested):

EVODelavega (dismissed):

jeremyletang (changes requested):
OK, I went over most of it, a few comments, but boi, that's a lot of code.

One thing really, is that if we can move the snapshot specific code from the types out of them when possible (e.g EquityShare and ExpiringOrders).

EVODelavega (approved):

jeremyletang (approved):

PR #4210, opened 2021-10-12
added min validators net param + bug fix for overflow reward

ze97286:
close #4206 #4220


jeremyletang (dismissed):

jeremyletang (changes requested):
I might be wrong but I think there might be non-deterministic code in there. Commented in the piece of code.

wwestgarth (dismissed):

jeremyletang (approved):

PR #4219, opened 2021-10-13
enable lints

jeremyletang:


wwestgarth (dismissed):
I was going to say I quite like naked-returns, but I googled the pros/cons it turns out theres quite a lot of (justified) dislike for them out there :D so I have learnt something today.

fkondej (approved):
LGTM

PR #4212, opened 2021-10-12
Settlement at expiry scenarios updated

cpuja84:
Added tests for suspended status

Original ticket on Protocol Design Board- vegaprotocol/specs-internal#561

(WG) Blocked by ~#4304~ ~#4357~ ~#4358~ ~#4370~

Closes #3820

PR #4258, opened 2021-10-19
Settlement at expiry scenario modifications

witgaw:
Suggested changes (as part of review of the original PR: #4212 )


cpuja84 (approved):
LGTM

PR #4391, opened 2021-11-09
Refactor delegation

ze97286:
close #4387 

summary of changes:
1. removed the max delegation per validator - validator will accept as much as delegated to it
2. this allowed changing the concept of the internal data to keep the same data for current and next epoch and just promote the next epoch to current at the end of the epoch
3. in rewards the new validator function has been implemented and a new net param has been added. 
4. The feature and unit tests have been updated


witgaw (commented):
Happy with feature file changes, haven't reviewed the code changes.

peterbarrow (approved):

PR #4392, opened 2021-11-10
Update GETTING_STARTED.md to reflect the truth, and sort out the make…

wwestgarth:
closes #4140 

I've updated `GETTING_STARTED.md` to reflect my experiences of getting my environment set up, please just comment if you think anything needs adding/removing based on your own environments.

I have also removed `build.sh` because it was very out of date. I've kept the `Makefile` (because I know some devs use it and jenkins relies on it) but have simplified it alot.


EVODelavega (dismissed):

ValentinTrinque (commented):
If this is not too much, we could also get rid of the `print_check` in the Makefile in favor of the one in the golangci-lint. What about that ?

peterbarrow (approved):

PR #4398, opened 2021-11-11
set proper status on withdrawal errors

jeremyletang:
Set the proper withdrawal state when an error occur.

close #4397


wwestgarth (approved):
Ticked, but I did notice that there are no tests at all for withdrawls in `banking/engine_test.go`. Do what you like with that piece of information :p

peterbarrow (approved):

PR #4422, opened 2021-11-19
Add key rotation support

karlem:
Close  #4139


jeremyletang (commented):
Just a couple of commnts

peterbarrow (approved):

PR #4439, opened 2021-11-24
Create release_ticket.md issue template

gordsport:
Closes #4438


candida-d (dismissed):
This looks great! 

Dream scenario for documentation is being able to pull changelogs from the open repos directly into docusuarus. 

However, being able to see all the changelogs in one place, and copy & paste the markdown is next-best so that works great. I don't use the issue numbers/links, especially for the closed repos, but I assume they're useful for historical purposes.  

General questions: 
Is there any way to get me 'assigned' to the ticket, or some other way that I can get a GH notification when the whole setup ready to go? 

Is there a situation where some of the repos would get updates but /vega wouldn't - it seems very unlikely but just checking.

peterbarrow (dismissed):

peterbarrow (approved):

PR #4481, opened 2021-12-07
4450 nullchain client example

wwestgarth:
closes #4450 

This is some example client code that uses the nullblockchain time-forwarding to be able to play out a scenario of creating a market/placing orders. Makes a start at abstracting out some of the fiddling wallet management and passing around of connections.

_It is not perfect_

The error handling is very poor, and there are no CLI args to change config values etc. etc. but I had to call time on it. It works when it works.


peterbarrow (dismissed):

peterbarrow (dismissed):

jeremyletang (commented):
I realise most of my comments are useless, and I've not been trough the whole code. 
Also I think this code should be turned into a library as most of it would be useful when building application again a nullblockchain / the scenario runner./

peterbarrow (approved):

PR #4496, opened 2021-12-10
State variable engine (floating point consensus engine)

ze97286:
close #4493


wwestgarth (dismissed):

jeremyletang (commented):
I still need to go through the test really properly, will do that locally and run them as well.

jeremyletang (approved):

PR #4514, opened 2021-12-15
Add network limits service and events

pscott31:
Closes #4513


jeremyletang (commented):
Sounds good here. we'll need to add support for this event / the api in the data-node too.

jeremyletang (approved):

PR #4535, opened 2021-12-26
wiring risk factors to state var consensus engine

ze97286:
close #4530 

This is part 1 of #4530


jeremyletang (commented):
It's nice to finally get rid of this weird structure of type.RiskResult. Can we remove it as well from the proto if they are still there as well?

jeremyletang (dismissed):

PR #4543, opened 2021-12-28
integrated risk factors with consensus engine + bug fixes to state va…

ze97286:
…r framework

close #4530 

moved the PR to https://github.com/vegaprotocol/vega/pull/4547


jeremyletang (commented):
Is it possible to replace the `sv.lock.Unlock()` in the different branching in state_variable.go with `defer sv.lock.Unlock()` instead. Doing that is quite cheap since the few last versions of the compiler, and it would eventually any bug in the future where we forgot to unlock in some new branching...
Also I release that some code may call tryConsensusLock or sendEventLock so it might not be possible.

PR #4660, opened 2022-01-26
4598 add tranx id to events and ctx

ettec:
Closes #4598   Where event are triggered by a transaction the transaction hash is added to published events.


jeremyletang (commented):
Just a small thing, but looks good overall.

jeremyletang (approved):

PR #4675, opened 2022-01-28
Add integration test for P&L with decimals

EVODelavega:
Add test to check for correct P&L values when market decimals <> asset decimals

Relates to #4594

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -103,3 +103,5 @@ vegawallet
 whitepaper
 XDG
 xor
+P&L
+PnL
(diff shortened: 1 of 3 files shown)


jeremyletang (dismissed):

ze97286 (approved):

Jiajia-Cui (commented):
This test has been created based on "https://github.com/vegaprotocol/vega/blob/develop/integration/features/release-margins-issues.feature".

In this case, an asset has 5 decimals, but the market decimals are 2, the offset of 1 == 1,000 of the smallest asset unit. The balances have all been multiplied by 1k, so the values all match the original test.

PR #4723, opened 2022-02-04
move timeupdate event send to time service to ensure it is first even…

ettec:
closes  #4719 

TimeUpdate now sent by time service to ensure it is the first event.


jeremyletang (commented):
Looks good, just you gonna want to add `TimeUpdate` in quote in the changelog

pscott31 (approved):

PR #4728, opened 2022-02-07
fix execution snapshot by handling generated providers properly

wwestgarth:
closes #4694 

Mostly just handling the generated-providers from the execution engine properly.


pscott31 (commented):
I'd like to give you a green tick - I can't see anything wrong with the code itself but I don't really understand what's going on here so I don't really feel qualified to say if it's sensible or not :)

peterbarrow (approved):

PR #4838, opened 2022-02-22
Support voting using equity-like shares

ValentinTrinque:
Close #4831


wwestgarth (dismissed):

jeremyletang (commented):
Code looks fine. I think we will need to update the spam protection policy though, because it's not required for a voter to haver governance token anymore. @davidsiska-vega This might need some specs updates?

jeremyletang (approved):

jeremyletang (approved):

PR #4890, opened 2022-03-02
4880 snapshot epoch propagation

wwestgarth:
closes #4880 

When an epoch is restored from a snapshot it now notifies all subscribers so that they themselves can set the restored epoch values in their engines.


jeremyletang (commented):
I think banking might need some changes as well then.

ze97286 (commented):
just a thought, maybe instead of adding a new callback for notifyInEpochRestore, we could just notify the regular callback but with a different EpochAction and then just extend the handling in each service to do whatever they want with the restore action.

ze97286 (approved):

PR #4899, opened 2022-03-03
feat: Add ranking score information to checkpoint and snapshot and em…

ze97286:
…it an event when loaded

Closes #4898


wwestgarth (dismissed):
I did a quick snapshot-restore on this branch, and it didn't fall over.

wwestgarth (approved):

PR #4905, opened 2022-03-03
Move test from qa-scenario to Verified folder in VEGA

Jiajia-Cui:
#4902


cpuja84 (approved):
Approving the changes for tests, would be merging post CI jobs being successful

PR #5056, opened 2022-03-23
fix: add pow to commander if non-validator command

wwestgarth:
closes #5038


jeremyletang (commented):
Would it be easier / cleaner to add a function CommandWithPOW() instead? like it's a very unique case to use the commander with the POW as pretty much all other commands do not need it.

jeremyletang (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

PR #5060, opened 2022-03-23
Lognormal model

Jiajia-Cui:
added a test case using log-normal risk model, and added a test case using log-normal risk model, and deleted duplicated test

<https://github.com/vegaprotocol/vega/issues/5061>


jeremyletang (approved):

cpuja84 (approved):
LGTM

PR #5105, opened 2022-03-29
fix: base64 encoding for tm pubkey in logs

jeremyletang:
just some logging prettying.


wwestgarth (dismissed):
Nice! Thats been nothering me for a while.

ValentinTrinque (approved):

PR #5106, opened 2022-03-29
fix: ensure a market cannot be started without an auction

### LINKED ISSUE #5141: Remove ability for a market to be started without opening auction
As of now market can be started without opening auction. This should not be allowed. AuctionState also support this and should be fixed.

jeremyletang:


jeremyletang:
closing for the time being.

PR #5116, opened 2022-03-29
feature test for Insurance pool tests

cpuja84:
closes #4922


edd (approved):
I like the meta-acceptance criteria of `0015-INSR-003` here - the Acceptance Criteria that the test passes. Very circular.

PR #5134, opened 2022-03-30
add test for price during auction; bug report on traded price in feat…

Jiajia-Cui:
…ure test (line 54), should be 95 instead of 90

#5133


witgaw (changes requested):
Please clarify

witgaw (changes requested):
I don't understand what's going on here (I might be missing something, please let me know if that's the case). 

I can see value in leaving this PR open so that engineering can easily reproduce the bug when they pickup #5133, but why would you want to merge it? 

Apart from the fact that we cannot do it because the test is failing (which is good as far as engineering being able to reproduce the bug goes) and there are artefacts of an unsuccessful merge in the changelog so CI pipline won't let this PR get merged, I don't think we'd ever want to include something like "Traded price in feature test is incorrectly reported" in the change log - it's not exactly a feature or improvement we'd want to brag about :)

PR #5139, opened 2022-03-31
feat: add feature test for 0032-PRIM-price_monitoring.md

Jiajia-Cui:
#5138

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -109,3 +109,7 @@ PnL
 Vim
 vim
 VSCode
+md
+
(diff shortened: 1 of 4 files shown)


davidsiska-vega (changes requested):
I think the test would make more sense with different dps. E.g. 
asset: 6
market: 2
position: 1

davidsiska-vega (dismissed):

davidsiska-vega (approved):

PR #5228, opened 2022-04-13
use LP version field

EVODelavega:
Closes #5001 

Changed LP version field to uint64, set initial version to 1, increment version on each amend


jeremyletang (commented):
Can we add / update a test to show it increase over admendments?

jeremyletang (approved):

Issue #5237, opened 2022-04-13
Trigger state variable calculation first time indicative uncrossing price is available


witgaw:
We're currently only triggering the calculation once the opening auction time elapses.

gordsport:
Hey @witgaw, Is this issue still present?

gordsport:
@witgaw @davidsiska-vega - please can you retest this on the latest version 0.55 and let us know what needs to be done before trading enabled?

witgaw:
#5237 should've closed this, but will give it a quick look and close if it's indeed working like this now.

ze97286:
https://github.com/vegaprotocol/vega/pull/5246 has closed it.

PR #5246, opened 2022-04-19
Extend auction interaction feat test

witgaw:
Extends and modifies the auction interaction feature test + implements functionality required to make it pass.

The main issue is that the auction trigger field in market data API doesn't always get modified to reflect the trigger that has extended the auction (https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0026-auctions.md#exiting-the-auction-mode).

This PR also brings implementation in line with spec changes, please see linked issues for details.

Blocked by: ~https://github.com/vegaprotocol/vega/issues/3570 https://github.com/vegaprotocol/vega/issues/3578~ ~https://github.com/vegaprotocol/vega/pull/5298~

Closes: https://github.com/vegaprotocol/vega/issues/3047 https://github.com/vegaprotocol/vega/issues/3570 https://github.com/vegaprotocol/vega/issues/4223 https://github.com/vegaprotocol/vega/issues/3362 https://github.com/vegaprotocol/vega/issues/5237


EVODelavega (changes requested):
As pointed out in individual comments, and summarised on slack, I think we require some changes to the code still. The most important things I've noticed are:

* With the current implementation, I believe it's possible for someone to spam FOK/IOC orders at stupid prices to force a price auction to get triggered
* `ExtractOrders` can now return pointers to orders as they sit on the book, making it very easy to introduce bugs like: having an order at price level 1000, but reprice it to 2000 without updating the matching engine
* Possible changes to auction behaviours (comment on integration tests about that)
* Auction extensions effectively being treated as leaving 1 type of auction and starting a second auction immediately after (which as per spec should never happen)
* Snapshots may be broken
* If we have a horizon of, say 10s, and we get an indicative price/volume for the first time in the last seconds of the opening auction phase, the price bounds may be off, and we might trigger a panic when the opening auction ends. I think we should add an integration test that replicates this scenario to ensure this isn't the case

witgaw (commented):
Addressed most of the comments

EVODelavega (dismissed):

EVODelavega (approved):

PR #5266, opened 2022-04-21
Do not check for commitment in update market proposal

ValentinTrinque:
Close #5267


peterbarrow (dismissed):

wwestgarth (changes requested):
In the pre-merge DoD we have: `Carry out a basic happy path end-to-end check of the new code`

Given that there are no changes to test code in this PR, and also this is the second time this feature has pinged back from QA with a bug in the happy-code-path, can you put some logs or something in the PR description as evidence that this now fully works?

wwestgarth (approved):
I still think we need to change here: https://github.com/vegaprotocol/vega/blob/develop/execution/engine.go#L337

to handle that the market might not be there. If the market is settled before the updateMarket proposal is enacted the market will be removed from the map and we'll panic. But I'm happy for this to be done in a follow up ticket/PR.

PR #5274, opened 2022-04-23
feat: reward/transfer refactoring

ze97286:
Closes #5273


ze97286 (commented):
@jeremyletang we may still need to implement the logic in recurring transfers that if the market doesn't exist it doesn't invoke (but probably proposal does exist) it doesn't call the collateral engine but it does not cancel the recurring transaction.

jeremyletang (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

PR #5326, opened 2022-05-03
Add integration test verifying price bounds & decimal places

EVODelavega:
Closes #5206 

Added 2 integration tests replicating the issue reported from the system tests. As far as the core goes, all values are correct, and price monitoring auction is triggered as expected


EVODelavega:
build/vega snuck its way in to this PR somehow, PITA to remove it

PR #5374, opened 2022-05-18
fix: liquidity parameters set in the market take precedence over netw…

wwestgarth:
closes #5203 also related to #5232

The issue was that no matter what values you set in your market proposal for `TargetStakeParameters` we would *always* create the market setting those value from the equivalent global network parameters.

Why did this also cause a problem with snapshots, I hear you ask? The same code that was overwriting the market's `TargetStakeParameters` value with the netparams is also called when restoring a market. So in the case where a market-update would move the time-window away from the network parameter value, on restore we would repopulate that value correctly but then reset it to the network parameter in `propagateInitialNetParams()` and then we'd fall out of consensus.

There is a related system-test issue that will hopefully result in us getting some good coverage on this: https://github.com/vegaprotocol/system-tests/issues/783 but its actually quite hard to test because through the API it looks like we've used the values in the marketConfig, but they actually aren't what have been propagated into the engines.


pscott31 (approved):
I don't really understand what's going on here but it seems sane to me!

peterbarrow (approved):

PR #5459, opened 2022-06-09
Bug when making market order

Jiajia-Cui:
When market maker assigns a market order to the market, they get "Invalid Persistence" error message, and order is rejected despite the fact that:

1. there is enough supplied stake
2. there is enough liquidity left after the trade  

[issue](https://github.com/vegaprotocol/vega/issues/5460)


witgaw (commented):
Ready for review

PR #5474, opened 2022-06-13
Allow wash trades in opening auction 

witgaw:
While wash trades were allowed in opening  auction, the usage of GetIndicativeTrades() to check if auction will uncross resulting in trades prevented it from uncrossing at that function wouldn't return trades if these were wash trades.

Closes #5468


witgaw (commented):
Ready for review

davidsiska-vega (approved):
I've reviewed the two feature tests and they test the desired behavior.

EVODelavega (approved):

PR #5518, opened 2022-06-29
chore: heartbeat less when epochs are short to help system-tests

wwestgarth:
As @ze97286  suspected, the heartbeat rate for low epoch lengths is causing heartbeat issues, and consequently causing the new pipeline to be unstable. So I've changed it so that the heartbeat interval is atleast 50seconds and not epoch-length.

### DIFF
--- a/validators/topology.go
+++ b/validators/topology.go
@@ -251,8 +251,7 @@ func (t *Topology) OnEpochLengthUpdate(ctx context.Context, l time.Duration) err
 	// set time between hearbeats to 1% of the epoch duration in seconds as blocks
 	// e.g. if epoch is 1 day = 86400 seconds (blocks) then time between hb becomes 864
 	// if epoch is 300 seconds then blocks becomes 50 (lower bound applied).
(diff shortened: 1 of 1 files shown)


validators/topology.go
```diff
@@ -251,8 +251,7 @@ func (t *Topology) OnEpochLengthUpdate(ctx context.Context, l time.Duration) err
 	// set time between hearbeats to 1% of the epoch duration in seconds as blocks
 	// e.g. if epoch is 1 day = 86400 seconds (blocks) then time between hb becomes 864
 	// if epoch is 300 seconds then blocks becomes 50 (lower bound applied).
-	// if epoch is 5 seconds then blocks becomes 5 (lower bound applied).
-	blocks := int64(math.Max(l.Seconds()*0.01, math.Min(50.0, l.Seconds())))
+	blocks := int64(math.Max(l.Seconds()*0.01, 50.0))
```
ze97286, line 254: that's fine by me, but you can also make it a multiple of l.Seconds() if you want more flexibility

ze97286 (approved):

PR #5522, opened 2022-06-30
fix: Do not fail silently when wallet fails to start

### LINKED ISSUE #5520: wallet service fail silently when unable to start
We need to improve the feedback on the `wallet service run` command when the service is unable to start.

Today, it it makes you think the service started as a background job whereas it just failed.

ValentinTrinque:
Close #5520 

# Test

1. Run the following command:
```sh
vega wallet service run -n fairground --automatic-consent --with-console --with-token-dapp
```

2. And  next to it, retry to run:
```sh
vega wallet service run -n fairground
```

3. Acknoledge the error message

4. Cancel the command run in point 1, and acknowledge the text is not messed up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
 # Vega specific entries
 build/
+vegawallet
 bin/
 cmd/*/*-dbg
 cmd/vega/vega*
+cmd/vegawallet/vegawallet*
 cmd/vegabenchmark/vegabenchmark*
 cmd/examples/nullchain/nullchain*
 data/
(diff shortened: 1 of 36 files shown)


jeremyletang (dismissed):

jeremyletang (dismissed):

peterbarrow (commented):
I ran the tests as described in the top of the ticket and it worked correctly. The second instance correctly reported that the port was already in use and exited cleanly.

jeremyletang (approved):

PR #5534, opened 2022-07-05
fix: do not accept tx with unexpected pubkeys.

### LINKED ISSUE #5431: Reject oracle data that doesn't match any pub keys from `CheckTx()`
Add a check for every submission in in check tx. If no one is expecting the pubkey / address then the transaction itself is meant to be rejected and never be stored on chain. We should provide an error/rejection message.

This is to unblock QA from being able to test the data sourcing work

AnExsomnis:
Closes: #5431

### DIFF
--- a/execution/engine.go
+++ b/execution/engine.go
@@ -51,6 +51,7 @@ type TimeService interface {
 // OracleEngine ...
 //go:generate go run github.com/golang/mock/mockgen -destination mocks/oracle_engine_mock.go -package mocks code.vegaprotocol.io/vega/execution OracleEngine
 type OracleEngine interface {
+	ListensToPubKeys(oracles.OracleData) bool
 	Subscribe(context.Context, oracles.OracleSpec, oracles.OnMatchedOracleData) oracles.SubscriptionID
 	Unsubscribe(context.Context, oracles.SubscriptionID)
 }
--- a/oracles/engine.go
+++ b/oracles/engine.go
@@ -23,13 +23,13 @@ import (
 	"code.vegaprotocol.io/vega/logging"
 )
 
-// Broker no need to mock (use broker package mock).
+// Broker interface. Do not need to mock (use package broker/mock).
 type Broker interface {
 	Send(event events.Event)
 	SendBatch(events []events.Event)
 }
 
-// TimeService ...
+// TimeService interface.
 //go:generate go run github.com/golang/mock/mockgen -destination mocks/time_service_mock.go -package mocks code.vegaprotocol.io/vega/oracles TimeService
 type TimeService interface {
 	GetTimeNow() time.Time
@@ -62,6 +62,14 @@ func NewEngine(
 	}
 
 	return e
+}
+
+// ListensToPubKeys checks if the public keys from provided OracleData are among the keys
+// current OracleSpecs listen to.
+func (e *Engine) ListensToPubKeys(data OracleData) bool {
+	return e.subscriptions.hasAnySubscribers(func(spec OracleSpec) bool {
+		return spec.MatchPubKeys(data)
+	})
 }
 
 // BroadcastData broadcasts data to products and risk models that are interested
--- a/oracles/oracle_spec.go
+++ b/oracles/oracle_spec.go
@@ -65,7 +65,7 @@ type filter struct {
 
 type condition func(string) (bool, error)
 
-// NewOracleSpec build an OracleSpec from a types.OracleSpec in a form that
+// NewOracleSpec builds an OracleSpec from a types.OracleSpec in a form that
 // suits the processing of the filters.
 func NewOracleSpec(originalSpec types.OracleSpec) (*OracleSpec, error) {
 	if len(originalSpec.PubKeys) == 0 {
@@ -144,6 +144,12 @@ func isInternalOracleData(data OracleData) bool {
 	return true
 }
 
+// MatchPubKeys tries to match the public keys from the provided OracleData object with the ones
+// present in the Spec.
+func (s *OracleSpec) MatchPubKeys(data OracleData) bool {
+	return containsRequiredPubKeys(data.PubKeys, s.pubKeys)
+}
+
 // MatchData indicates if a given OracleData matches the spec or not.
 func (s *OracleSpec) MatchData(data OracleData) (bool, error) {
 	// if the data contains the internal oracle timestamp key, and only that key,
@@ -169,8 +175,8 @@ func (s *OracleSpec) MatchData(data OracleData) (bool, error) {
 	return true, nil
 }
 
-// containsRequiredPubKeys verifies if all the public keys is the OracleData is
-// matches the keys authorized by the OracleSpec.
+// containsRequiredPubKeys verifies if all the public keys in the OracleData
+// are within the list of currently authorized by the OracleSpec.
 func containsRequiredPubKeys(dataPKs []string, authPks map[string]struct{}) bool {
 	for _, pk := range dataPKs {
 		if _, ok := authPks[pk]; !ok {
--- a/processor/abci.go
+++ b/processor/abci.go
@@ -48,6 +48,7 @@ import (
 
 var (
 	ErrPublicKeyCannotSubmitTransactionWithNoBalance  = errors.New("public key cannot submit transaction without balance")
+	ErrUnexpectedTxPubKey                             = errors.New("no one listens to the public keys that signed this oracle data")
 	ErrTradingDisabled                                = errors.New("trading disabled")
 	ErrNoTransactionAllowedDuringBootstrap            = errors.New("no transaction allowed during the bootstraping period")
 	ErrMarketProposalDisabled                         = errors.New("market proposal disabled")
@@ -55,6 +56,9 @@ var (
 	ErrNonValidatorTransactionDisabledDuringBootstrap = errors.New("non validator transaction disabled during bootstrap")
 	ErrCheckpointRestoreDisabledDuringBootstrap       = errors.New("checkpoint restore disabled during bootstrap")
 	ErrAwaitingCheckpointRestore                      = errors.New("transactions not allowed while waiting for checkpoint restore")
+	ErrOracleDataNormalization                        = func(err error) error {
+		return fmt.Errorf("error normalizing incoming oracle data: %w", err)
+	}
 )
 
 //go:generate go run github.com/golang/mock/mockgen -destination mocks/checkpoint_mock.go -package mocks code.vegaprotocol.io/vega/processor Checkpoint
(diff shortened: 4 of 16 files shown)


processor/abci.go
```diff
@@ -746,6 +747,12 @@ func (app *App) OnCheckTx(ctx context.Context, _ tmtypes.RequestCheckTx, tx abci
 		app.log.Debug("entering checkTx", logging.String("tid", tx.GetPoWTID()), logging.String("command", tx.Command().String()))
 	}
 
+	if !app.exec.MarketsExpectKey(tx.PubKeyHex()) {
```
wwestgarth, line 750: I think we only want to do this check if the incoming transaction is a `SubmitOracleDataCommand`?

execution/engine.go
```diff
@@ -273,6 +274,19 @@ func (e *Engine) IsEligibleForProposerBonus(marketID string, value *num.Uint) bo
 	return value.ToDecimal().GreaterThan(quantum.Mul(e.npv.minLpStakeQuantumMultiple))
 }
 
+// MarketsExpectKey checks if the currently existing markets use the provided key in their subscriptions.
+func (e *Engine) MarketsExpectKey(key string) bool {
+	for _, mkt := range e.markets {
+		for k := range mkt.mkt.TradableInstrument.Instrument.GetFuture().OraclesPubKeys() {
+			if strings.Compare(k, key) {
```
wwestgarth, line 281: I'm probably missing something but whats wrong with `if k == key`?

processor/abci.go
AnExsomnis: I have corrected this here https://github.com/vegaprotocol/vega/pull/5534/commits/8ed6a685a32b5cdaf527f4bd083f72c17ff2a55a#diff-43cb75f5e3d4bfc8b8fc4540853a32c32346ec1fb151bbaed720d7bf07a4a139R822-R825
Thanks!

execution/engine.go
AnExsomnis: I have added the logic here https://github.com/vegaprotocol/vega/pull/5534/commits/8ed6a685a32b5cdaf527f4bd083f72c17ff2a55a#diff-175c28f9d4bf1098cfc45b2c0071f08ea70dc56c117b273fee2ef110a876487eR15 it to aim for a more efficient comparison:  fail faster or/and in the first not matching rune, instead of traversing the whole strings.

ValentinTrinque (changes requested):

execution/engine.go
```diff
@@ -273,6 +274,19 @@ func (e *Engine) IsEligibleForProposerBonus(marketID string, value *num.Uint) bo
 	return value.ToDecimal().GreaterThan(quantum.Mul(e.npv.minLpStakeQuantumMultiple))
 }
 
+// MarketsExpectKey checks if the currently existing markets use the provided key in their subscriptions.
+func (e *Engine) MarketsExpectKey(key string) bool {
```
ValentinTrinque, line 278: Based on my previous comments, the algorithm is a bit more complex than this. 

We consider the OracleData to match an OracleSpec when all the public keys that signed the oracle data are defined in the oracle spec.

So getting a **list** of the _watched public keys_ won't be enough. You will need a more complex data structure.

For example:
* Market A listens to 0x11111
* Market B listens to 0x22222
* An oracle data signed by 0x11111 will match
* But an oracle data signed by 0x11111 and 0x22222 won't match, because no one is listening at **both keys at the same time**

ValentinTrinque (changes requested):

processor/abci.go
```diff
@@ -48,6 +48,7 @@ import (
 
 var (
 	ErrPublicKeyCannotSubmitTransactionWithNoBalance  = errors.New("public key cannot submit transaction without balance")
+	ErrUnexpectedTxPubKey                             = errors.New("received unexpected public key in transaction")
```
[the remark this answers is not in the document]
AnExsomnis: Yes, makes sense; Thanks!

AnExsomnis:
@ValentinTrinque ,
I have reworked the logic in a new commit (for now) - https://github.com/vegaprotocol/vega/pull/5534/commits/a5d8d5bcd4180e37b763161986d92e913680b01b
The check for existing active keys is transferred to the OracleSpecs and is used by the OracleEngine (as entities that are aware of all keys and should handle this).
Also, I have added this method call inside processor/abci.go - CheckSubmitOracleData as this check could simply be done inside the handler instead of having same call for normalization somewhere else.
Please take a look when you have time and let's discuss.

processor/abci.go
```diff
@@ -818,6 +819,10 @@ func (app *App) canSubmitTx(tx abci.Tx) (err error) {
 				return ErrAssetProposalDisabled
 			}
 		}
+	case txn.SubmitOracleDataCommand:
+		if !app.exec.MarketsExpectKey(tx.PubKeyHex()) {
```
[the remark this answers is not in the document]
AnExsomnis: Thank you, this has been addressed here https://github.com/vegaprotocol/vega/pull/5534#issuecomment-1177353456

types/market.go
```diff
@@ -409,6 +409,21 @@ func (f Future) String() string {
...
+			if _, ok := uKeys[k]; ok {
+				continue
+			}
+			uKeys[k] = struct{}{}
+		}
+	}
+
+	return uKeys
+}
+
```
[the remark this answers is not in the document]
AnExsomnis: Yes, this has been addressed here https://github.com/vegaprotocol/vega/pull/5534#issuecomment-1177353456
Thanks!

execution/engine.go
AnExsomnis: Yes, this has been addressed here https://github.com/vegaprotocol/vega/pull/5534#issuecomment-1177353456
Thanks!

processor/abci.go
```diff
@@ -1244,9 +1244,12 @@ func (app *App) CheckSubmitOracleData(_ context.Context, tx abci.Tx) error {
 	}
 
 	pubKey := crypto.NewPublicKey(tx.PubKeyHex(), tx.PubKey())
-	_, err := app.oracles.Adaptors.Normalise(pubKey, *data)
+	oracleData, err := app.oracles.Adaptors.Normalise(pubKey, *data)
+	if err != nil {
+		return err
```
ValentinTrinque, line 1249: Except if the underlaying errror is very clear for what's happening, could you wrap this error with `fmt.Errof()` ?

Something that can help understand *where* things went wrong :)

oracles/oracle_spec.go
```diff
@@ -144,6 +144,12 @@ func isInternalOracleData(data OracleData) bool {
 	return true
 }
 
+// matchPubKeys tries to match the public keys from the provided OracleData object with the ones
+// present in the Spec.
+func (s *OracleSpec) matchPubKeys(data OracleData) bool {
```
ValentinTrinque, line 149: Since it's a method called by an external component, I think we should make this method pubic as it's now part of the API of the oracle spec. It states the intention better.

oracles/engine.go
```diff
@@ -64,6 +64,19 @@ func NewEngine(
...
+	_, err := e.subscriptions.filterSubscribers(func(spec OracleSpec) (bool, error) {
+		if !spec.matchPubKeys(data) {
+			return false, fmt.Errorf("provided data public keys do not match currently active in specs")
+		}
+		return true, nil
+	})
+
+	return err
+}
+
```
ValentinTrinque, line 79: Overall, this is very good. I think you can optimize this by implementing a `hasAnySubscribers() bool` method on the subscriptions structure, so it doesn't iterate over all the specs. We just needs to know if _at least one_ matches.

oracles/engine.go
```diff
@@ -64,6 +64,19 @@ func NewEngine(
 	return e
 }
 
+// ListensToPubKeys checks if the public keys from provided OracleData are among the keys
+// current OracleSpecs listen to.
+func (e *Engine) ListensToPubKeys(data OracleData) error {
```
ValentinTrinque, line 69: And to finish, could you add some tests on this ?

oracles/oracle_spec.go
AnExsomnis: All right, we logically separate them, I understand.
Thanks!

AnExsomnis: This has been corrected here https://github.com/vegaprotocol/vega/pull/5534/commits/587bd850dcca9027fd5422ef1cf618ff087a449d#diff-c31fe3f1776ec18d9a8482d6f1b00732f84023362e6183a17e18e379b0b28ebcR147-R152

processor/abci.go
AnExsomnis: Yes,
I did some optimizations on this here https://github.com/vegaprotocol/vega/pull/5534/commits/587bd850dcca9027fd5422ef1cf618ff087a449d#diff-43cb75f5e3d4bfc8b8fc4540853a32c32346ec1fb151bbaed720d7bf07a4a139R59

processor/abci.go
```diff
@@ -48,13 +48,17 @@ import (
...
+	ErrUnexpectedTxPubKey                             = errors.New("no one listens to the public keys that signed this oracle data")
 	ErrTradingDisabled                                = errors.New("trading disabled")
 	ErrNoTransactionAllowedDuringBootstrap            = errors.New("no transaction allowed during the bootstraping period")
 	ErrMarketProposalDisabled                         = errors.New("market proposal disabled")
 	ErrAssetProposalDisabled                          = errors.New("asset proposal disabled")
 	ErrNonValidatorTransactionDisabledDuringBootstrap = errors.New("non validator transaction disabled during bootstrap")
 	ErrCheckpointRestoreDisabledDuringBootstrap       = errors.New("checkpoint restore disabled during bootstrap")
 	ErrAwaitingCheckpointRestore                      = errors.New("transactions not allowed while waiting for checkpoint restore")
+	ErrOracleDataNormalization                        = func(err error) error {
+		return fmt.Errorf("error normalizing incoming oracle data: %s", err.Error())
```
ValentinTrinque, line 60: You can use `%w`, it wraps the underlaying error and works better with error API `errors.Is()`

```suggestion
		return fmt.Errorf("error normalizing incoming oracle data: %w", err)
```

ValentinTrinque (changes requested):

ValentinTrinque (dismissed):

ValentinTrinque (approved):

oracles/engine.go
```diff
@@ -64,6 +64,14 @@ func NewEngine(
 	return e
 }
 
+// ListensToPubKeys checks if the public keys from provided OracleData are among the keys
+// current OracleSpecs listen to.
+func (e *Engine) ListensToPubKeys(data OracleData) bool {
+	return e.subscriptions.hasAnySubscribers(func(spec OracleSpec) bool {
```
wwestgarth, line 70: I'm a bit late now its merged (sorry), but given `CheckTx` happens concurrently with the block processing I think we're going to need a lock on the subscribers. We could be adding a new subscriber (due to a new market creation) while at the same time an oracle-command is being checked.

PR #5535, opened 2022-07-05
feat: core side of protocol upgrade

### LINKED ISSUE #5538: Core side of protocol upgrade
# Feature Overview  (part of #5004)

- In Admin RPC
   - Add API for indicating interest to upgrade the protocol with a release tag and a upgrade block height: `ProposeNewUpgrade(upgradeBlockHeight uint64, releaseTag string) error`. In response the core submits a transaction to vega chain so that all nodes see the upgrade request. The `upgradeBlockHeight` is the last block to be processed in the chain before the chain will go down in case it is approved. 
   - Add API to query if the protocol is ready to be shutdown for upgrade: `IsReadyForUpgrade() bool`
   
- An engine is added to keep track of protocol upgrade proposals
- If a proposal doesn't get a majority of +2/3 of the *tendermint* validators until (including) the block height given in `upgradeBlockHeight` of the proposal - the proposal is rejected and dropped. 
- If it gets a majority of +2/3 of the *tendermint* validators - at the beginning of the following block - we take a snapshot at the end of which `IsReadyForUpgrade` RPC call will start returning true.

ze97286:
Closes #5538

### DIFF
--- a/admin/server.go
+++ b/admin/server.go
@@ -22,10 +22,16 @@ import (
 	"code.vegaprotocol.io/shared/paths"
 	"code.vegaprotocol.io/vega/logging"
 	"code.vegaprotocol.io/vega/nodewallets"
+	"code.vegaprotocol.io/vega/types"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/rpc"
 	"github.com/gorilla/rpc/json"
 )
+
+type ProtocolUpgradeService interface {
+	// is vega core ready to be stopped and upgraded
+	GetUpgradeStatus() types.UpgradeStatus
+}
 
 // Server implement a socket server allowing to run simple RPC commands.
 type Server struct {
@@ -33,16 +39,36 @@ type Server struct {
 	cfg Config
 	srv *http.Server
 
-	nodeWallet *NodeWallet
+	nodeWallet             *NodeWallet
+	protocolUpgradeService ProtocolUpgradeAdminService
+}
+
+func NewNonValidatorServer(log *logging.Logger,
+	config Config,
+	vegaPaths paths.Paths,
+	protocolUpgradeService ProtocolUpgradeService,
+) (*Server, error) {
+	// setup logger
+	log = log.Named(namedLogger)
+	log.SetLevel(config.Level.Get())
+
+	return &Server{
+		log:                    log,
+		cfg:                    config,
+		nodeWallet:             nil,
+		srv:                    nil,
+		protocolUpgradeService: *NewProtocolUpgradeService(protocolUpgradeService),
+	}, nil
 }
 
 // NewServer returns a new instance of the RPC socket server.
-func NewServer(
+func NewValidatorServer(
 	log *logging.Logger,
 	config Config,
 	vegaPaths paths.Paths,
 	nodeWalletPassphrase string,
 	nodeWallets *nodewallets.NodeWallets,
+	protocolUpgradeService ProtocolUpgradeService,
 ) (*Server, error) {
 	// setup logger
 	log = log.Named(namedLogger)
@@ -54,10 +80,11 @@ func NewServer(
 	}
 
 	return &Server{
-		log:        log,
-		cfg:        config,
-		nodeWallet: nodeWallet,
-		srv:        nil,
+		log:                    log,
+		cfg:                    config,
+		nodeWallet:             nodeWallet,
+		srv:                    nil,
+		protocolUpgradeService: *NewProtocolUpgradeService(protocolUpgradeService),
 	}, nil
 }
 
@@ -89,7 +116,11 @@ func (s *Server) Start() {
 	rs.RegisterCodec(json.NewCodec(), "application/json")
 	rs.RegisterCodec(json.NewCodec(), "application/json;charset=UTF-8")
 
-	rs.RegisterService(s.nodeWallet, "")
+	if s.nodeWallet != nil {
+		rs.RegisterService(s.nodeWallet, "")
+	}
+	rs.RegisterService(s.protocolUpgradeService, "protocolupgrade")
+
 	r := mux.NewRouter()
 	r.Handle(s.cfg.Server.HttpPath, rs)
 
--- a/cmd/vega/node/node.go
+++ b/cmd/vega/node/node.go
@@ -273,8 +273,14 @@ func (n *NodeCommand) startAPIs() error {
 
 	n.proxyServer = rest.NewProxyServer(n.Log, n.conf.API)
 
-	if bool(n.conf.Admin.Server.Enabled) && n.conf.IsValidator() {
-		adminServer, err := admin.NewServer(n.Log, n.conf.Admin, n.vegaPaths, n.nodeWalletPassphrase, n.nodeWallets)
+	if n.conf.IsValidator() {
+		adminServer, err := admin.NewValidatorServer(n.Log, n.conf.Admin, n.vegaPaths, n.nodeWalletPassphrase, n.nodeWallets, n.protocol.GetProtocolUpgradeService())
+		if err != nil {
+			return err
+		}
+		n.adminServer = adminServer
+	} else {
+		adminServer, err := admin.NewNonValidatorServer(n.Log, n.conf.Admin, n.vegaPaths, n.protocol.GetProtocolUpgradeService())
 		if err != nil {
 			return err
 		}
--- a/cmd/vega/propose_upgrade.go
+++ b/cmd/vega/propose_upgrade.go
@@ -1,0 +1,125 @@
+// Copyright (c) 2022 Gobalsky Labs Limited
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at https://www.mariadb.com/bsl11.
+//
+// Change Date: 18 months from the later of the date of the first publicly
+// available Distribution of this version of the repository, and 25 June 2022.
+//
+// On the date above, in accordance with the Business Source License, use
+// of this software will be governed by version 3 or later of the GNU General
+// Public License.
+
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	commandspb "code.vegaprotocol.io/protos/vega/commands/v1"
+	vgcrypto "code.vegaprotocol.io/shared/libs/crypto"
+	"code.vegaprotocol.io/shared/paths"
+	"code.vegaprotocol.io/vega/config"
+	"code.vegaprotocol.io/vega/logging"
+	"code.vegaprotocol.io/vega/txn"
+
+	"github.com/blang/semver"
+	"github.com/jessevdk/go-flags"
+)
+
+type ProposeUpgradeCmd struct {
+	config.VegaHomeFlag
+	config.OutputFlag
+	config.Passphrase `long:"passphrase-file"`
+
+	VegaReleaseTag     string `short:"v" long:"vega-release-tag" required:"true" description:"A valid vega core release tag for the upgrade proposal"`
+	DatanodeReleaseTag string `short:"d" long:"datanode-release-tag" required:"true" description:"A valid data-node release tag for upgrade proposal"`
+	UpgradeBlockHeight uint64 `short:"h" long:"height" required:"true" description:"The block height at which the upgrade should be made"`
+}
+
+var proposeUpgradeCmd ProposeUpgradeCmd
+
+func (opts *ProposeUpgradeCmd) Execute(args []string) error {
+	log := logging.NewLoggerFromConfig(logging.NewDefaultConfig())
+	defer log.AtExit()
+
+	registryPass, err := opts.Get("node wallet", false)
+	if err != nil {
+		return err
+	}
+
+	vegaPaths := paths.New(opts.VegaHome)
+
+	_, conf, err := config.EnsureNodeConfig(vegaPaths)
+	if err != nil {
+		return err
+	}
+
+	if !conf.IsValidator() {
+		return errors.New("node is not a validator")
+	}
+
+	cmd := commandspb.ProtocolUpgradeProposal{
+		VegaReleaseTag:     opts.VegaReleaseTag,
+		DataNodeReleaseTag: opts.DatanodeReleaseTag,
+		UpgradeBlockHeight: opts.UpgradeBlockHeight,
+	}
+
+	commander, blockData, cfunc, err := getNodeWalletCommander(log, registryPass, vegaPaths)
+	if err != nil {
+		return fmt.Errorf("failed to get commander: %w", err)
+	}
+
+	if opts.UpgradeBlockHeight <= blockData.Height {
+		return fmt.Errorf("upgrade block earlier than current block height")
+	}
+	_, err = semver.Parse(opts.VegaReleaseTag)
+	if err != nil {
+		return fmt.Errorf("invalid protocol version for upgrade received: version (%s), %w", opts.VegaReleaseTag, err)
+	}
+
+	_, err = semver.Parse(opts.DatanodeReleaseTag)
+	if err != nil {
+		return fmt.Errorf("invalid protocol version for upgrade received: version (%s), %w", opts.DatanodeReleaseTag, err)
+	}
+
+	defer cfunc()
+
+	tid := vgcrypto.RandomHash()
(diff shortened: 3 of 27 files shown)


protocolupgrade/engine.go
```diff
@@ -0,0 +1,243 @@
...
+	"code.vegaprotocol.io/vega/logging"
+	"code.vegaprotocol.io/vega/txn"
+	"code.vegaprotocol.io/vega/types"
+	"code.vegaprotocol.io/vega/types/num"
+	"github.com/blang/semver"
+	"github.com/cenkalti/backoff"
+	"github.com/golang/protobuf/proto"
+)
+
+var requiredQuorum = num.NewDecimalFromFloat(2).Div(num.NewDecimalFromFloat(3))
```
jeremyletang, line 35: Shouldn't we use the network parameters instead for this?

processor/abci.go
```diff
@@ -601,6 +615,17 @@ func (app *App) OnBeginBlock(req tmtypes.RequestBeginBlock) (ctx context.Context
 	app.log.Debug("entering begin block", logging.Time("at", time.Now()), logging.Uint64("height", uint64(req.Header.Height)))
 	defer func() { app.log.Debug("leaving begin block", logging.Time("at", time.Now())) }()
 
+	// if there is an approved protocol upgrade proposal and the current block height is later than the proposal's block height then take a snapshot and wait to be killed by the process manager
+	if app.protocolUpgradeService.TimeForUpgrade() {
+		app.snapshot.SnapshotNow(ctx)
```
wwestgarth, line 620: I don't think we can take the snapshot here. Assuming I follow properly we are at block `n`, the first block that will be in the new version, and effectively snapshoting block`n-1`, the last block of the old version.

When we pull down the old core process and bring up the new one with the snapshot data from block `n-1` tendermint will want the commit hash of block `n-1` during the abci `Info` call, which we won't have.  So I think you'll have to move the forcing of the snapshot to `OnCommit` of block `n-1`.

wwestgarth:
Not for this PR, but I think we are going to have to be careful more careful with which snapshots are sent/accepted when state-syncing. For example if someone if running an old version of core and state-syncs against a network with a newer version it'll just get handed a snapshot and it will just try to use it. We'll want to put something in to reject any snapshots from different versions.

It might also make sense to force a snapshot on the first block after an upgrade otherwise any one wanting to state-sync in will have to wait until the first snapshot is taken before getting anything usable.

processor/abci.go
```diff
@@ -878,6 +904,14 @@ func (app *App) RequireValidatorMasterPubKey(ctx context.Context, tx abci.Tx) er
 	return nil
 }
 
+func (app *App) DeliverProtocolUpgradeCommand(ctx context.Context, tx abci.Tx) error {
+	pu := &commandspb.ProtocolUpgradeProposal{}
+	if err := tx.Unmarshal(pu); err != nil {
+		return err
+	}
+	return app.protocolUpgradeService.UpgradeProposal(ctx, tx.PubKeyHex(), pu.Proposal.UpgradeBlockHeight, pu.Proposal.ReleaseTag)
```
wwestgarth, line 912: I'm glad to see the update height is not longer compiled in now, but who is now in charge of choosing it?

jeremyletang: Validators should discuss it together and fix a block height together.

wwestgarth:
> > Not for this PR, but I think we are going to have to be careful more careful with which snapshots are sent/accepted when state-syncing. For example if someone if running an old version of core and state-syncs against a network with a newer version it'll just get handed a snapshot and it will just try to use it. We'll want to put something in to reject any snapshots from different versions.
> > It might also make sense to force a snapshot on the first block after an upgrade otherwise any one wanting to state-sync in will have to wait until the first snapshot is taken before getting anything usable.
> 
> I think this should be covered with this? https://github.com/vegaprotocol/vega/blob/develop/processor/abci.go#L391

Maybe, but the info call happens after we would have already loaded the snapshot. 

That is a good point though, we probably need to set the new protocol-version in the consensus-params in the EndBlock response on the block before we upgrade.

protocolupgrade/engine.go
ze97286: also, I think each governance proposal type has its own so we can add a new one

admin/server.go
```diff
@@ -90,6 +100,8 @@ func (s *Server) Start() {
 	rs.RegisterCodec(json.NewCodec(), "application/json;charset=UTF-8")
 
 	rs.RegisterService(s.nodeWallet, "")
+	rs.RegisterService(s.protocolUpgradeService, "upgrade")
```
karlem, line 103: I don't think this is enough. I think you will need to implement a proper service like the one we use for `nodeWallet`.  With http request, args and reply.

wwestgarth: Is there a reason we aren't just going with a CLI thing like `vega propose_upgrade --release-tag=blah --block=blah` to shoot the command in like when a new node annouces itself?

protocolupgrade/engine.go
```diff
@@ -0,0 +1,243 @@
...
+		broker:                broker,
+		topology:              topology,
+		cmd:                   cmd,
+		hashKeys:              []string{(&types.PayloadProtocolUpgradeProposals{}).Key()},
+	}
+}
+
+// UpgradeProposal records the intention of a validator to upgrade the protocol to a release tag at block height.
+func (e *Engine) UpgradeProposal(ctx context.Context, pk string, upgradeBlockHeight uint64, releaseTag string) error {
+	if !e.topology.IsTendermintValidator(pk) {
```
karlem, line 105: Not sure what that actually means but would that not be an issue for non-validators (maybe those who runs data-node)?

admin/server.go
ze97286: we may, none of this is set in stone. whatever ends up simpler for the validators / process manager

protocolupgrade/engine.go
```diff
@@ -0,0 +1,243 @@
...
+				e.log.Info("protocol upgrade rejected", logging.String("release-tag", pup.tag), logging.Uint64("upgrade-block-height", pup.blockHeight))
+				e.broker.Send(events.NewProtocolUpgradeProposalEvent(ctx, pup.blockHeight, pup.tag, pup.approvers(), eventspb.ProtocolUpgradeProposalStatus_PROTOCOL_UPGRADE_PROPOSAL_STATUS_REJECTED))
+			}
+		}
+	}
+	if accepted != nil {
+		ID := protocolUpgradeProposalID(accepted.blockHeight, accepted.tag)
+		delete(e.activeProposals, ID)
+		delete(e.events, ID)
+		e.broker.Send(events.NewProtocolUpgradeProposalEvent(ctx, accepted.blockHeight, accepted.tag, accepted.approvers(), eventspb.ProtocolUpgradeProposalStatus_PROTOCOL_UPGRADE_PROPOSAL_STATUS_APPROVED))
```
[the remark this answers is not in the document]
ze97286: non-validators don't decide if the network is upgraded so they can't "vote" or propose an upgrade. They will be running a process manager so should react the same way as validators. The problem is that if we're not automating the binaries preparation then they need to know that they need to have built the binaries by this point and then the process manager will restart correctly.

protocolupgrade/engine.go
ze97286: ok, I'll add a network parameter for it.

protocolupgrade/snapshot.go
```diff
@@ -0,0 +1,105 @@
...
+func (e *Engine) Keys() []string {
+	return e.hashKeys
+}
+
+func (e *Engine) Stopped() bool {
+	return false
+}
+
+// get the serialised form and hash of the given key.
+func (e *Engine) serialise(k string) ([]byte, error) {
```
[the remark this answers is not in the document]
ze97286: I'm not sure what's the best strategy is. Maybe for simplicity - once a proposal gets approved - we're locking the state, i.e. rejecting everything else, so the state of the engine is clear, and auto-rejecting any incoming proposal until the upgrade takes place. what do you think? @wwestgarth @jeremyletang @karlem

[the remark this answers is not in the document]
wwestgarth: A fresh start after an upgrade seems reasonable to me.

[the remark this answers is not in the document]
ze97286: done

admin/server.go
ze97286: done

processor/abci.go
```diff
@@ -690,6 +714,12 @@ func (app *App) OnCommit() (resp tmtypes.ResponseCommit) {
 	app.updateStats()
 	app.setBatchStats()
 
+	// if there is an approved protocol upgrade proposal and the current block height is later than the proposal's block height then take a snapshot and wait to be killed by the process manager
+	if app.protocolUpgradeService.TimeForUpgrade() {
+		app.snapshot.SnapshotNow(app.blockCtx)
```
wwestgarth, line 719: This snapshot hash needs to be set as `resp.Data`. Also if it happens to already be a snapshot block, we'll be double-snapshotting and will have two snapshots for the same block height in the snapshotdb.

ze97286: re double snapshots - I'm not sure how the snapshotdb will handle that, but the reason I force a snapshot here regardless if one was taken or not is because of the asynchronous save of the snapshot which I can't rely on here.

Regarding to setting the snapshot hash in resp.Data, why do we need this? I don't want to force using it for consensus, I just want to make sure it's saved.

ze97286: ok

admin/server.go
```diff
@@ -89,7 +116,11 @@ func (s *Server) Start() {
 	rs.RegisterCodec(json.NewCodec(), "application/json")
 	rs.RegisterCodec(json.NewCodec(), "application/json;charset=UTF-8")
 
-	rs.RegisterService(s.nodeWallet, "")
+	if s.nodeWallet != nil {
```
jeremyletang, line 119: I think we have a IsValidator call we could use instead? not a big deal, but the nodewallet would be instatiated evern for a non-validators isn't it?

cmd/vega/node/node.go
```diff
@@ -284,8 +284,14 @@ func (n *NodeCommand) startAPIs() error {
...
-	if bool(n.conf.Admin.Server.Enabled) && n.conf.IsValidator() {
-		adminServer, err := admin.NewServer(n.Log, n.conf.Admin, n.vegaPaths, n.nodeWalletPassphrase, n.nodeWallets)
+	if n.conf.IsValidator() {
+		adminServer, err := admin.NewValidatorServer(n.Log, n.conf.Admin, n.vegaPaths, n.nodeWalletPassphrase, n.nodeWallets, n.protocol.GetProtocolUpgradeService())
+		if err != nil {
+			return err
+		}
+		n.adminServer = adminServer
+	} else {
+		adminServer, err := admin.NewNonValidatorServer(n.Log, n.conf.Admin, n.vegaPaths, n.protocol.GetProtocolUpgradeService())
```
jeremyletang, line 294: Alright you do have 2 constructors to instantiate it in a validator or non-validators mode indeed.

cmd/vega/propose_upgrade.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	if !conf.IsValidator() {
+		return errors.New("node is not a validator")
+	}
+
+	cmd := commandspb.ProtocolUpgradeProposal{
+		VegaReleaseTag:     opts.VegaReleaseTag,
+		DataNodeReleaseTag: opts.DatanodeReleaseTag,
+		UpgradeBlockHeight: opts.UpgradeBlockHeight,
```
jeremyletang, line 65: Should we add validation of the semver here / check if the block is non-zero but also is a future block? Just so we can add nice error handling here.

cmd/vega/propose_upgrade.go
```diff
@@ -0,0 +1,114 @@
...
+			if err != nil {
+				ch <- fmt.Errorf("failed to send protocol upgrade proposal command: %v", err)
+			}
+			close(ch)
+		}, nil, pow)
+
+	err = <-ch
+	if err != nil {
+		return err
+	}
```
jeremyletang, line 100: as I suppose we would have the error here to be fair.

netparams/defaults.go
```diff
@@ -29,6 +29,7 @@ const (
 
 func defaultNetParams() map[string]value {
 	return map[string]value{
+		ProtocolUpgradeProposalRequiredMajority: NewDecimal(DecimalGTE(num.MustDecimalFromString("0.5")), DecimalLTE(num.MustDecimalFromString("1"))).Mutable(true).MustUpdate("0.66"),
```
jeremyletang, line 32: Why not just reuse the `ValidatorsVoteRequired` ?

processor/abci.go
```diff
@@ -650,7 +674,19 @@ func (app *App) OnCommit() (resp tmtypes.ResponseCommit) {
...
-	snapHash, err := app.snapshot.Snapshot(app.blockCtx)
+
+	var snapHash []byte
+	var err error
+	// if there is an approved protocol upgrade proposal and the current block height is later than the proposal's block height then take a snapshot and wait to be killed by the process manager
+	if app.protocolUpgradeService.TimeForUpgrade() {
+		snapHash, err = app.snapshot.SnapshotNow(app.blockCtx)
+		if err == nil {
+			app.protocolUpgradeService.SetReadyForUpgrade()
+		}
```
jeremyletang, line 685: What if err != nil ? Is that possible? Should we panic if that was to happen?

jeremyletang: ah it is checked just later...

protocolupgrade/engine.go
```diff
@@ -0,0 +1,252 @@
...
+		return errors.New("upgrade block earlier than current block height")
+	}
+
+	if e.upgradeStatus.AcceptedReleaseInfo != nil {
+		return errors.New("protocol upgrade already scheduled")
+	}
+
+	if !e.topology.IsSelfTendermintValidator() {
+		return errors.New("validator is not a tendermint validator - cannot decide on upgrade")
+	}
```
jeremyletang, line 122: Not sure we should stop here if we are not a validator ourselves right? Even non-validators needs to process the rest of the state after that?

protocolupgrade/engine.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+
+	active := e.activeProposals[ID]
+	active.accepted[pk] = struct{}{}
+	evt := events.NewProtocolUpgradeProposalEvent(ctx, active.blockHeight, active.vegaReleaseTag, active.dataNodeReleaseTag, active.approvers(), eventspb.ProtocolUpgradeProposalStatus_PROTOCOL_UPGRADE_PROPOSAL_STATUS_PENDING)
+	evtProto := evt.Proto()
+	e.events[ID] = &evtProto
+	e.broker.Send(evt)
+	return nil
```
jeremyletang, line 154: If a validator registered intent for another release before, should we remove them from the approvers of the release if they register for a new one?

protocolupgrade/snapshot.go
```diff
@@ -0,0 +1,126 @@
...
+
+	sort.SliceStable(events, func(i, j int) bool {
+		if events[i].VegaReleaseTag == events[j].VegaReleaseTag {
+			if events[i].DataNodeReleaseTag == events[j].DataNodeReleaseTag {
+				return events[i].UpgradeBlockHeight < events[j].UpgradeBlockHeight
+			}
+			return events[i].DataNodeReleaseTag < events[j].DataNodeReleaseTag
+		}
+		return events[i].VegaReleaseTag < events[j].VegaReleaseTag
+	})
```
jeremyletang, line 53: why not just sort over the protocol IDs ids instead? I means your way is probably better as requires only 1 iteration though

cmd/vega/propose_upgrade.go
ze97286: added

netparams/defaults.go
ze97286: we probably can, but not sure we have to bind it to protocol implementation network param. Anyways, can change to us it and spare one network parameter

protocolupgrade/engine.go
ze97286: removed

protocolupgrade/engine.go
ze97286: we can yeah

jeremyletang (approved):

PR #5554, opened 2022-07-11
Swap and slot reduction

### LINKED ISSUE #5552: `numberOfErsatzValidators` not honoured if slot-reduction + swap happens in same epoch
# Problem encountered
If the network parameter that controls the number of ersatz validators is reduced in the same epoch that an ersatz validator is promoted, we can be left with a nodeset where `actual-number-or-ersatz > t.numberOfErsatzValidators`

I have added a unittest on this branch that flexes it: [swap-and-slot-reduction](https://github.com/vegaprotocol/vega/compare/swap-and-slot-reduction?expand=1)

and have seen it as an unstable test in this system-test run: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/4225/pipeline/277

# Observed behaviour
Sometimes when the network parameter for the number of ersatz validators is reduced, the number of actual ersatz validators never changes.

# Expected behaviour
The value of the network parameter is always honoured.

# Definition of Done

**Before Merging**

**After Merging**

wwestgarth:
closes #5552


jeremyletang (approved):

validators/validator_set.go
```diff
@@ -298,6 +298,11 @@ func (t *Topology) applyPromotion(performanceScore, rankingScore map[string]num.
 		}
 	}
 
+	// demoted tendermint validators are also ersatz so we need to add them
+	for _, id := range demotedFromTM {
+		ersatzValidators = append(ersatzValidators, t.validators[id])
+	}
+
 	// demote from ersatz to pending due to more ersatz than slots allowed
 	ersatzValidators, waitingListValidators, _ = demoteDueToLackOfSlots(ersatzValidators, waitingListValidators, ValidatorStatusErsatz, ValidatorStatusPending, t.numberOfErsatzValidators, int64(t.currentBlockHeight+1))
```
wwestgarth, line 307: The one thing I am unsure of is whether we should be then removing the `demotedFromTM` validators from `ersatzValidators` before we call promote before to stop a TM -> pending. Its unclear to me from the spec whether in: 
"If a node that submitted the transaction to join and satisfies all other conditions and has a higher score than the lowest scoring ersatz validator (scaled up by the incumbent factor)"

`lowest scoring ersatz validator` means the lowest scoring before the tendermint <-> ersatz promotions are applied, or after the tendermint <-> ersatz promotions.

validators/validator_set.go
```diff
@@ -298,6 +298,11 @@ func (t *Topology) applyPromotion(performanceScore, rankingScore map[string]num.
 		}
 	}
 
+	// demoted tendermint validators are also ersatz so we need to add them
+	for _, id := range demotedFromTM {
```
ze97286, line 302: maybe I'm misremembering this, but won't they already be there by construction? IIRC promote would already return them with the correct status in remainingValidators and then the iteration in line 293 would already put them in ersatz. what am I missing?

wwestgarth: No, in `promote` its only when the number of slots reduces that we move validators from seriesA into seriesB (and `removedFromSeriesA` is left empty). When a true "demotion" happens the demoted validator is returned in `removedFromSeriesA` but not returned in `seriesB`.

Which feels intentional?

ze97286: it feels messy mostly, but that's on me

ze97286 (approved):

PR #5557, opened 2022-07-11
fix: call unsubscribe at mkt settlement

AnExsomnis:
close #5427


wwestgarth:
I think we also want to unsubscribe the oracles when a market is rejected too, because a rejected market gets completely removed from the execution engine too.

oracles/engine.go
```diff
@@ -117,13 +117,15 @@ func (e *Engine) BroadcastData(ctx context.Context, data OracleData) error {
...
 // If cb is nil, the method panics.
-func (e *Engine) Subscribe(ctx context.Context, spec OracleSpec, cb OnMatchedOracleData) SubscriptionID {
+func (e *Engine) Subscribe(ctx context.Context, spec OracleSpec, cb OnMatchedOracleData) (SubscriptionID, Unsubscriber) {
 	if cb == nil {
 		panic(fmt.Sprintf("a callback is required for spec %v", spec))
 	}
 	updatedSubscription := e.subscriptions.addSubscriber(spec, cb, e.timeService.GetTimeNow())
 	e.sendNewOracleSpecSubscription(ctx, updatedSubscription)
-	return updatedSubscription.subscriptionID
+	return updatedSubscription.subscriptionID, func() {
```
wwestgarth, line 126: The context contains the block-height which is sent along with all events. So I think instead of having it part of the closure, you'll need to pass it in when you call in otherwise the oracle-event will have an old block-height associated with it.

AnExsomnis:
> I think we also want to unsubscribe the oracles when a market is rejected too, because a rejected market also gets completely removed from the execution engine and so its oracles should be deactivated.

Added this also here https://github.com/vegaprotocol/vega/pull/5557/commits/5a6e9af45a91d27271e05409cafc9e50c643e297#diff-66d5f93bc68001de17615b08e43dc983611f01570c0e7b96dc04f5da46ee6db3R3219
Thanks!

oracles/engine.go
AnExsomnis: Yes, I made the last changes to use the same context: https://github.com/vegaprotocol/vega/pull/5557/commits/5a6e9af45a91d27271e05409cafc9e50c643e297#diff-9030c1f6e5bde7618ca69f3b4f5ffc9c9b3cc31cac3ef5a9fb599074bb7808bfR126-R128
Thanks!

wwestgarth (dismissed):

wwestgarth (approved):

PR #5568, opened 2022-07-13
fix: return better status codes when transaction is blocked by spam

### LINKED ISSUE #5439: Wallet returns 500 when party has insufficient tokens to submit proposal request in an epoch
Test proposes an asset with a party staked less than spam.protection.proposal.min.tokens

Trx is sent through wallet
Wallet signs and send proposal submission through API /api/v1/command

Wallet returns response 500 'Internal Server Error'

Expected 4xx status code to align with error b'{"error":"InvalidArgument - party has insufficient tokens to submit proposal request in this epoch"}'

wwestgarth:
closes #5439

We now get reponses from the wallet that look like this:
```
AssertionError: http://127.0.0.1:1790/api/v1/command/sync returned HTTP 429 {"error":"party has insufficient tokens to submit proposal request in this epoch"}
```

Most of the refactor is because we were doing this silly thing in the blockchain client where we return an `err` _and_ the response if `resp.Code` was non-zero and where `err` actually was just `resp.Data`:
```
if err != nil {
    return resp, err
}
```

But in the generated protos code (both coming through tendermint, and also out from core to the wallet again) it always drops the response if there is an error:
```
if err != nil {
    nil, err
}
```

so all of our nice specific codes/details when things went wrong that we thought we were passing back were just getting lost. Which is a bit mad.

The bulk of the changes here and in the wallet is to treat `err != nil` as an _error in the rpc_ and to treat `err == nil && !resp.Success` as a _successfully failing transaction_ and by that I mean one that got sent correctly, but was for good reason rejected by core, so nothing really went "wrong". This means the wallet can now look at `resp.Code` if `resp.Success == false` to know what went on and to set a http-status that makes sense for the probably user-error.

Note: This is a breaking change because any callers (like the wallet) will now have to check `resp.Success` as well as just `err != nil` which they may not have been doing before.

### DIFF
--- a/api/core.go
+++ b/api/core.go
@@ -149,7 +149,7 @@ func (s *coreService) SubmitTransaction(ctx context.Context, req *protoapi.Submi
 	return &protoapi.SubmitTransactionResponse{
 		Success: txResult.Code == 0,
 		Code:    txResult.Code,
-		Data:    txResult.Data.String(),
+		Data:    string(txResult.Data.Bytes()),
 		Log:     txResult.Log,
 		Height:  0,
 		TxHash:  txResult.Hash.String(),
--- a/api/core.go
+++ b/api/core.go
@@ -143,33 +143,11 @@ func (s *coreService) SubmitTransaction(ctx context.Context, req *protoapi.Submi
 
 	txResult, err := s.blockchain.SubmitTransactionSync(ctx, req.Tx)
 	if err != nil {
-		// This is Tendermint's specific error signature
-		if _, ok := err.(interface {
-			Code() uint32
-			Details() string
-			Error() string
-		}); ok {
-			s.log.Debug("unable to submit transaction", logging.Error(err))
-			fullError := apiError(codes.InvalidArgument, err)
-			if txResult != nil {
-				return &protoapi.SubmitTransactionResponse{
-					Success: false,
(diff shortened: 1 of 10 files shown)


api/core.go
```diff
@@ -143,33 +143,11 @@ func (s *coreService) SubmitTransaction(ctx context.Context, req *protoapi.Submi
...
-		s.log.Debug("unable to submit transaction", logging.Error(err))
-
 		return nil, apiError(codes.Internal, err)
 	}
 
 	return &protoapi.SubmitTransactionResponse{
-		Success: true,
+		Success: txResult.Code == 0,
 		Code:    txResult.Code,
 		Data:    txResult.Data.String(),
```
jeremyletang, line 152: This always return the string of the error, e.g: I just tested it, and it returns the hex encoded string: 70617274792068617320696E73756666696369656E7420746F6B656E7320746F207375626D69742064656C65676174696F6E207265717565737420696E20746869732065706F6368
Which is: party has insufficient tokens to submit delegation request in this epoch.

Should we hex decode the txResult.Data field everytime instead?

jeremyletang (dismissed):
Just one small comment here, but it's very much better already.

jeremyletang (approved):

PR #5580, opened 2022-07-17
fix: add locks on subscribers add and remove

### LINKED ISSUE #5567: Oracle subscribers need to be locked given async transaction check
# Problem encountered
A follow up ticket for a post-merge comment left on a previous PR: https://github.com/vegaprotocol/vega/pull/5534#discussion_r916569594

# Observed behaviour
There is chance that a conconcurrent map read/write can occur during the checking of oracle transactions

# Expected behaviour
This check is made thread-safe.

# Definition of Done

**Before Merging**

**After Merging**

AnExsomnis:
close #5567

### DIFF
--- a/oracles/engine_subscribers.go
+++ b/oracles/engine_subscribers.go
@@ -15,6 +15,7 @@ package oracles
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"code.vegaprotocol.io/vega/types"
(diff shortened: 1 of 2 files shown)


oracles/engine_subscribers.go
```diff
@@ -68,6 +69,8 @@ func (r filterResult) hasMatched() bool {
 // specSubscriptions wraps the subscribers (in form of OnMatchedOracleData) to
 // the OracleSpec.
 type specSubscriptions struct {
+	mu *sync.RWMutex
```
jeremyletang, line 72: Mutexes don't need to be point, their 0 value when non pointer is a perfectly fine working mutex.

jeremyletang (dismissed):

oracles/engine_subscribers.go
AnExsomnis: Yes, but in that case the receiver `specSubscription` was a value, not a pointer itself, so we needed a memory location in it, not another value to exchange between threads.
This made me see that some of the methods have pointer receivers and some values - I made them all now to have pointer receivers so the mutex could be used as value.
Edited in a force push here: https://github.com/vegaprotocol/vega/pull/5580/commits/fe9606ae5a9b2b07306944c22f435bdd37724d5c

wwestgarth (approved):

PR #5604, opened 2022-07-22
refactor: ignore stale trigger when in auction

### LINKED ISSUE #5608: Ignore stale price monitoring trigger
As per [0032-PRIM-price_monitoring](https://github.com/vegaprotocol/specs/blob/master/protocol/0032-PRIM-price_monitoring.md#0032-PRIM-009) spec change:

- If the cumulative extentions period of various chained auctions is more than the "time horizon" in a given triplet then there is no relevant reference price and this triplet is ignored.

witgaw:
Closes #5608

### DIFF
--- a/execution/market.go
+++ b/execution/market.go
@@ -371,7 +371,7 @@ func NewMarket(
 
 	tsCalc := liquiditytarget.NewSnapshotEngine(*mkt.LiquidityMonitoringParameters.TargetStakeParameters, positionEngine, mkt.ID, positionFactor)
 
-	pMonitor, err := price.NewMonitor(asset, mkt.ID, tradableInstrument.RiskModel, mkt.PriceMonitoringSettings, stateVarEngine, log)
+	pMonitor, err := price.NewMonitor(asset, mkt.ID, tradableInstrument.RiskModel, as, mkt.PriceMonitoringSettings, stateVarEngine, log)
 	if err != nil {
 		return nil, fmt.Errorf("unable to instantiate price monitoring engine: %w", err)
 	}
--- a/monitor/price/pricemonitoring.go
+++ b/monitor/price/pricemonitoring.go
@@ -449,6 +449,13 @@ func (e *Engine) getCurrentPriceRanges(force bool) map[*bound]priceRange {
 		if !b.Active {
 			continue
 		}
+		if e.anyTriggerBreached() && len(e.pricesPast) > 0 {
+			triggerLookback := e.now.Add(time.Duration(-b.Trigger.Horizon) * time.Second)
+			// check if trigger not stale: newest reference price older than horizon lookback time
+			if (*e.mostRecentPriceTime()).Before(triggerLookback) {
+				continue
+			}
+		}
 		ref := e.getRefPrice(b.Trigger.Horizon, force)
 		var min, max num.Decimal
 
@@ -473,6 +480,26 @@ func (e *Engine) getCurrentPriceRanges(force bool) map[*bound]priceRange {
 	e.priceRangeCacheTime = e.now
 	e.stateChanged = true
 	return e.priceRangesCache
+}
+
+func (e *Engine) mostRecentPriceTime() *time.Time {
+	if len(e.pricesNow) > 0 {
+		return &e.now
+	}
+	n := len(e.pricesPast)
+	if n > 0 {
+		return &e.pricesPast[n-1].Time
+	}
+	return nil
+}
+
+func (e *Engine) anyTriggerBreached() bool {
+	for _, b := range e.bounds {
+		if !b.Active {
+			return true
+		}
+	}
+	return false
 }
 
 // clearStalePrices updates the pricesPast slice to hold only as many prices as implied by the horizon.
--- a/monitor/price/pricemonitoring.go
+++ b/monitor/price/pricemonitoring.go
@@ -449,10 +449,18 @@ func (e *Engine) getCurrentPriceRanges(force bool) map[*bound]priceRange {
 		if !b.Active {
 			continue
 		}
-		if e.anyTriggerBreached() && len(e.pricesPast) > 0 {
+		if e.anyTriggerBreached() && len(e.pricesPast)+len(e.pricesNow) > 0 {
 			triggerLookback := e.now.Add(time.Duration(-b.Trigger.Horizon) * time.Second)
-			// check if trigger not stale: newest reference price older than horizon lookback time
-			if (*e.mostRecentPriceTime()).Before(triggerLookback) {
+			// check if trigger's not stale (newest reference price older than horizon lookback time)
+			var mostRecentObservation time.Time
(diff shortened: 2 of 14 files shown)


monitor/price/pricemonitoring.go
```diff
@@ -258,9 +258,10 @@ func (e *Engine) CheckPrice(ctx context.Context, as AuctionState, trades []*type
 		e.resetPriceHistory(trades)
 		e.initialised = true
 	}
+	inAuction := as.InAuction()
 	// market is not in auction, or in batch auction
-	if fba := as.IsFBA(); !as.InAuction() || fba {
-		bounds := e.checkBounds(ctx, trades)
+	if fba := as.IsFBA(); !inAuction || fba {
+		bounds := e.checkBounds(ctx, trades, inAuction)
```
EVODelavega, line 264: This won't work once we have FBA's -> `inAuction` will then always be `true`, even if (when `CheckPrice` is called is likely to be the case), we've reached the end of a batch.

monitor/price/pricemonitoring.go
```diff
@@ -449,6 +450,21 @@ func (e *Engine) getCurrentPriceRanges(force bool) map[*bound]priceRange {
...
+			// check if trigger's not stale (newest reference price older than horizon lookback time)
+			var mostRecentObservation time.Time
+			if len(e.pricesNow) > 0 {
+				mostRecentObservation = e.now
+			} else {
+				x := e.pricesPast[len(e.pricesPast)-1]
+				mostRecentObservation = x.Time
+			}
+			if mostRecentObservation.Before(triggerLookback) {
+				b.Active = false
```
EVODelavega, line 464: IDK, but this looks like something that changes the spec a bit. AFAIK, the price bounds stay in place. We enter price auction, and can have it extended if the auction ends up uncrossing at a price point even further removed from the original bound that was breached. Having bounds that weren't breached expire during an auction (when there are no trades by definition) just looks a bit weird. Wouldn't it be possible for a market to be defined with an auction duration that basically ensures all bounds expire? Is that valid, or should we add validation on the governance level to make sure the price monitoring params are sensible?

execution/market.go
```diff
@@ -111,9 +111,9 @@ var (
 type PriceMonitor interface {
 	OnTimeUpdate(now time.Time)
 	CheckPrice(ctx context.Context, as price.AuctionState, trades []*types.Trade, persistent bool) bool
-	GetCurrentBounds() []*types.PriceMonitoringBounds
+	GetCurrentBounds(inAuction bool) []*types.PriceMonitoringBounds
```
EVODelavega, line 114: Not really a fan of this, TBH. Isn't it more sensible to handle some of the work `OnTimeUpdate`? We might end up calling `GetCurrentBounds` or `GetValidPriceRange` several times in a single block, we don't have to check the validity of the bounds each time, just on time update.

EVODelavega: If we really do need to keep track of the auction state in the price monitor, I'd argue it makes more sense to inject the auction state, rather than pass in `m.as.InAuction()` as arguments, and passing the auction state as an argument to the liquidity engine constructor. I can see why price monitoring would need to access the auction state more readily than the liquidity engine...

execution/market.go
```diff
@@ -380,7 +380,7 @@ func NewMarket(
 
 	now := timeService.GetTimeNow()
 	liqEngine := liquidity.NewSnapshotEngine(
-		liquidityConfig, log, timeService, broker, tradableInstrument.RiskModel, pMonitor, asset, mkt.ID, stateVarEngine, mkt.TickSize(), priceFactor.Clone(), positionFactor)
+		liquidityConfig, log, timeService, broker, tradableInstrument.RiskModel, as, pMonitor, asset, mkt.ID, stateVarEngine, mkt.TickSize(), priceFactor.Clone(), positionFactor)
```
EVODelavega, line 383: As mentioned above: IMO it's cleaner to give `pMonitor` access to `as`, rather than to pass in `as` here only to call `as.InAuction()` when calling the `pMonitor`

liquidity/engine.go
```diff
@@ -118,6 +118,7 @@ func NewEngine(config Config,
 	timeService TimeService,
 	broker Broker,
 	riskModel RiskModel,
+	auctionState risk.AuctionState,
```
EVODelavega, line 121: `risk.AuctionState`? Do we need `CanLeave`? I'm not a big fan of relying on interfaces from other packages.

monitor/price/pricemonitoring.go
witgaw: The spec was already changed a few months ago, this brings the implementation in line with it. You could have a minimum auction duration that is higher than each of the horizons, but that's perhaps desirable in some cases: effectively a way of disabling price auction extensions - you can still have a price auction though (continuous -> price auction, just not able to extend liquidity (or price) auction with a price auction)

execution/market.go
witgaw: Yeah I didn't love it either, but wanted to make sure it always returns actual bounds (and not bounds that would turn out to not be "bitting" when push comes to shove as they've actually expired). 

Could add a method to price monitor that get's called when market gets put into and out of auction (this is somewhat missing anyway, for example we reactive the bounds when price monitoring engine is ready to end the auction, but there are actually no guarantees that market will indeed end an auction, so it'd help with that too). What do you think?

liquidity/engine.go
witgaw: Ah no, good point, just laziness I suppose, will fix.

monitor/price/pricemonitoring.go
witgaw: hmm ok, not sure why we're anticipating it already if it's not really on the immediate horizon, but I guess since we started doing it we might as well carry on, will fix.

monitor/price/pricemonitoring.go
witgaw: Ok, your suggestion from the other comment to just give price monitoring engine access to `auctionState` is probably the cleanest, will proceed on that basis.

execution/market.go
witgaw: Ok, your suggestion to just give price monitoring engine access to auctionState is probably the cleanest, will proceed on that basis.

monitor/price/pricemonitoring.go
witgaw: Fixed now I believe.

execution/market.go
witgaw: Addressed now I believe.

execution/market.go
witgaw: Addressed now I believe.

liquidity/engine.go
witgaw: Addressed now I believe.

EVODelavega (approved):

davidsiska-vega (approved):
Feature tests look good.

PR #5611, opened 2022-07-23
fix: restore active markets from checkpoint proposals

### LINKED ISSUE #5279: Incorrect proposal handling with checkpoints
Checkpoints do not behave correctly when reloading  markets which haven't settle.
Per spec markets should be restarted if they were not closed and settle and the opening auction should start again.

https://github.com/vegaprotocol/specs-internal/blob/master/non-protocol-specs/0005-NP-LIMN-limited_network_life.md#restoring-a-checkpoint

ze97286:
Closes #5348 #5279


assets/assets.go
```diff
@@ -58,13 +58,16 @@ type Service struct {
 	isValidator bool
 }
 
+type TimeService interface {
+	GetTimeNow() time.Time
+}
+
```
jeremyletang, line 64: I don't think we need this interface?

execution/engine.go
```diff
@@ -280,8 +280,29 @@ func (e *Engine) SubmitMarketWithLiquidityProvision(ctx context.Context, marketC
...
+	return e.submitOrRestoreMarketWithLiquidityProvision(ctx, marketConfig, lp, party, lpID, deterministicId, true)
+}
+
+// RestoreMarketWithLiquidityProvision is very similar to SubmitMarketWithLiquidityProvision without updating the market tracker as it is restored separately from checkpoint.
+func (e *Engine) RestoreMarketWithLiquidityProvision(ctx context.Context, marketConfig *types.Market, lp *types.LiquidityProvisionSubmission, lpID, deterministicId string) error {
+	proposer := e.marketActivityTracker.GetProposer(marketConfig.ID)
+	if len(proposer) == 0 {
+		return ErrMarketDoesNotExist
+	}
+	return e.submitOrRestoreMarketWithLiquidityProvision(ctx, marketConfig, lp, proposer, lpID, deterministicId, false)
```
jeremyletang, line 292: Will it be fine here to use the market ID as the proposer of the market? will we not be submitting the LP for a wrong party ID?

ze97286: we're not submitting the market ID, we're submitting the proposer of the market as the proposer.

jeremyletang: oh right. I'm an idot indeed.

governance/engine.go
```diff
@@ -495,7 +498,7 @@ func (e *Engine) intoToSubmit(ctx context.Context, p *types.Proposal) (*ToSubmit
 		enactTime := time.Unix(p.Terms.EnactmentTimestamp, 0)
 		newMarket := p.Terms.GetNewMarket()
 
-		auctionDuration := enactTime.Sub(closeTime)
+		auctionDuration := time.Duration((enactTime.Unix() - closeTime.Unix()) * int64(time.Second))
```
wwestgarth, line 501: Minor: I'm sure theres a good reason you're stripping off all the precision but given that `enactTime` is already just in precision seconds, I think all you need to do is here is`closeTime := e.timeService.GetTimeNow().Truncate(time.Second)`

Also that `FIXME` comment about the auctionState, do we not also need to trunctate the time passed into `NewAuctionState`?

governance/checkpoint.go
```diff
@@ -31,31 +37,73 @@ func (e *Engine) Checkpoint() ([]byte, error) {
...
 			return err
 		}
+
+		switch prop.Terms.Change.GetTermType() {
+		case types.ProposalTermsTypeNewMarket:
+			// if the proposal is for a new market we want to restore it such that it will be in opening auction
+			if p.Terms.EnactmentTimestamp <= now.Unix() {
+				prop.Terms.EnactmentTimestamp = now.Add(duration).Unix()
+			}
+			if toSubmit, err := e.intoToSubmit(ctx, prop); err == nil {
```
wwestgarth, line 80: If `intoToSubmit` fails we're just silently ignoring it, is that intentional?

ze97286: I think it pretty much can't really fail (famous last words) but I'll add panic.

governance/engine.go
ze97286: addressed

assets/assets.go
ze97286: removed

wwestgarth (dismissed):

wwestgarth (approved):

PR #5622, opened 2022-07-25
Verify cucumber steps

### LINKED ISSUE #5636: Assure no false positives in cucumber steps
Assure that all fields in cucumber assertion steps are checked and that incorrectly named table column results in an error for all steps.

witgaw:
WG: Need to park this for a little bit while I focus on getting a paper submitted for a conference. 

I went through all cucumber steps and made sure that there are no false positives - that we never get a pass and actually don't check one or more columns in a table or proceed without fail when incorrect column name gets supplied. This PR fixes a handful of issues that I found. 

Closes https://github.com/vegaprotocol/vega/issues/5636

### DIFF
--- a/integration/steps/debug_all_events.go
+++ b/integration/steps/debug_all_events.go
@@ -1,0 +1,27 @@
+// Copyright (c) 2022 Gobalsky Labs Limited
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at https://www.mariadb.com/bsl11.
+//
+// Change Date: 18 months from the later of the date of the first publicly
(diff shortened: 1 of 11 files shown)


integration/steps/debug_all_events.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2022 Gobalsky Labs Limited
```
EVODelavega, line 1: What now?

witgaw (commented):
Ready for review

witgaw:
Moved to #5655

PR #5642, opened 2022-07-28
bug: line 118, trade did not happen as expected

Jiajia-Cui:
line 118, trade did not happen as expected while market is in continuous mode, there is enough supplied stake on the market


ze97286:
@Jiajia-Cui the trade doesn't happen because it's a wash trade. By line 118, aux3 has an outstanding order of buy 2 (with remaining 1 unit) so when they place the sell order in line 118 it is identified as wash trade and is not allowed to go through.

Jiajia-Cui:
closed with the [issue](https://github.com/vegaprotocol/vega/issues/5643)

Issue #5004, opened 2022-03-16
Implement the base protocol upgrade process


gordsport:
To implement the [base protocol upgrade](https://github.com/vegaprotocol/specs-internal/pull/874) as per the spec requirements

now:
- [ ] create a PoC of the process (using a single engine?)
- [ ] decide how to have two version of the software
- [ ] feedback the investigations and what expected impact to timings of trading enabled timeframes
- [ ] identify risks of any given approach 

Post investigations? (tbc)
- [ ] introduce new commands
- [ ] implement upgrade engine 
- [ ] use app version field in the snapshots

In the first version of this we expect to define the upgrade path (e.g. 1->2 then 2->3), in a later version we may introduce the ability to jump specified versions.

Open questions:
- system-tests to cover this
- what impact, if any, may this have on the wallet?
- how does the upgrade process impact the Tendermint weights of the nodes
- how does the upgrade process work with snapshot bugs / major or breaking changes
- how does the upgrade process work with work items like Tendermint upgrades

The expectation is that with this we will upgrade more frequently (smaller batch sizes in releases thus less likely to have huge changes in the state to have to manage)

gordsport:
Closing as completed

PR #5650, opened 2022-07-28
Add vega visor

### LINKED ISSUE #5717: Implement Visor part for protocol upgrade

karlem:
Part of #5004 

Close #5717


core/admin/server.go
```diff
@@ -40,7 +40,7 @@ type Server struct {
 	srv *http.Server
 
 	nodeWallet             *NodeWallet
-	protocolUpgradeService ProtocolUpgradeAdminService
+	protocolUpgradeService *ProtocolUpgradeAdminService
```
ze97286, line 43: why a pointer to an interface here?

ze97286: oh, I see it's not the interface it's the struct, ok makes sense.

karlem: It's not interface. It's the service stuct. And it has to be a pointer otherwise it will not get registered as service with the `RegisterService` RPC server method.

core/processor/abci.go
```diff
@@ -677,22 +677,17 @@ func (app *App) OnCommit() (resp tmtypes.ResponseCommit) {
...
-	if app.protocolUpgradeService.TimeForUpgrade() {
-		snapHash, err = app.snapshot.SnapshotNow(app.blockCtx)
-		if err == nil {
-			app.protocolUpgradeService.SetReadyForUpgrade()
-		}
-	} else {
-		snapHash, err = app.snapshot.Snapshot(app.blockCtx)
-	}
-
+	snapHash, err := app.snapshot.Snapshot(app.blockCtx)
```
ze97286, line 680: I don't think this is a good change for a couple of reasons:

1. if it's not normally a snapshot block, it will not take a snapshot - we want it to and need it to. 
2. The regular snapshot call is done and saved asynchronously - in this case we want to wait for it to complete synchronously before telling the visor that we're done and ready to die

karlem: Oh yeah, sorry I have missed that the `now` at the end. The reason why I had to change was because the previous code was leading to panic. Beacuase:
1. `TimeForUpgrade` for checking `currentBlock  == upgradeBlock`
2. But `SetReadyForUpgrade` has a check for `currentBlock - upgradeBlock == 1` otherwise panic.
I will fix it.

visor/binaries_runner.go
```diff
@@ -0,0 +1,177 @@
...
+	running     map[string]*exec.Cmd
+	binsFolder  string
+	log         *logging.Logger
+	stopTimeout time.Duration
+}
+
+func NewBinariesRunner(log *logging.Logger, binsFolder string, stopTimeout time.Duration) *BinariesRunner {
+	return &BinariesRunner{
+		binsFolder:  binsFolder,
+		running:     map[string]*exec.Cmd{},
```
ze97286, line 43: not sure if it matters, but would it make sense to keep them in order implied in the slice passed to run()? as is, the order you signal them is arbitrary.

visor/binaries_runner.go
```diff
@@ -0,0 +1,177 @@
...
+	timeout := time.After(r.stopTimeout)
+	r.mut.RUnlock()
+
+	ticker := time.NewTicker(time.Second / 10)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-timeout:
+			return fmt.Errorf("failed to gratefully shut down processes: timed out")
```
ze97286, line 164: gracefully I reckon :)

visor/binaries_runner.go
```diff
@@ -0,0 +1,177 @@
...
+func NewBinariesRunner(log *logging.Logger, binsFolder string, stopTimeout time.Duration) *BinariesRunner {
+	return &BinariesRunner{
+		binsFolder:  binsFolder,
+		running:     map[string]*exec.Cmd{},
+		log:         log,
+		stopTimeout: stopTimeout,
+	}
+}
+
+func (r *BinariesRunner) Run(ctx context.Context, binaries []config.BinaryConfig) chan error {
```
ze97286, line 49: do we need some validation that the binaries make sense? e.g. that it has at most one validator, and at least one vega (either validator or not) at most one datanode, etc.

visor/init.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if homeExists {
+		return fmt.Errorf("home folder %q already exists", homePath)
+	}
+
+	visorConf := config.DefaultVisorConfig(log, homePath)
+
+	log.Info("Initiating genesis folder")
+	if err := initDefaultFolder(visorConf.GenesisFolder(), "genesis"); err != nil {
```
ze97286, line 43: what is this supposed to be doing? it it used for anything?

visor/init.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	log.Info("Initiating upgrade folder")
+	if err := initDefaultFolder(visorConf.UpgradeFolder(), "upgrade"); err != nil {
```
ze97286, line 48: is the upgradefolder() used for anything?

visor/visor.go
```diff
@@ -0,0 +1,199 @@
...
+				r.log.Info("Preparing upgrade")
+
+				if err := binRunner.Stop(); err != nil {
+					// Force to kill if fails grateful stop fails
+					if err := binRunner.Kill(); err != nil {
+						return fmt.Errorf("failed to force kill the running processes: %w", err)
+					}
+				}
+
+				r.log.Info("Starting upgrade")
```
ze97286, line 187: but what starts the new binaries?

visor/binaries_runner.go
karlem: IMHO it should not matter. Because we `sigterm` for graceful shutdown.

visor/binaries_runner.go
karlem: we could do it. But we would need to implement some commands on the actual binaries I guess. Because the tools doesn't can't really tell otherwise..

visor/init.go
karlem: This will generate basic folder structure for the tool `home` with example configs.

visor/init.go
karlem: Yes, that is where the validator can put next binaries and configs for the upgrade.

visor/visor.go
karlem: We set the upgrade folder (the one with the new binaries) as current. So with the next loop iteration we are running a new (updated) binaries.

visor/init.go
ze97286: but it's not used functionally as far as I could see, maybe I've missed it

visor/visor.go
ze97286: ah got it, thanks

visor/init.go
karlem: Ah ok.. I see what you mean.. Yeah it is not used here in the code but it is used by the `init` command.

Capsule for example uses the init command [here](https://github.com/vegaprotocol/vegacapsule/pull/210/files#diff-d44b310e81776ebbb501940b14f82b9dec20d1d458f9b5710c36423fae69b3bbR40)

visor/binaries_runner.go
```diff
@@ -0,0 +1,220 @@
...
+	timeout := time.After(r.stopTimeout)
+	r.mut.RUnlock()
+
+	ticker := time.NewTicker(time.Second / 10)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-timeout:
+			return fmt.Errorf("failed to gratefully shut down processes: timed out")
```
jeremyletang, line 207: Gracefully I reckon too btw.

visor/binaries_runner.go
```diff
@@ -0,0 +1,220 @@
...
+		case <-ticker.C:
+			r.mut.RLock()
+			if len(r.running) == 0 {
+				return nil
+			}
+			r.mut.RUnlock()
```
jeremyletang, line 213: Is it fine to keep this locked after returning?

visor/config/visor_config.go
```diff
@@ -0,0 +1,249 @@
...
+			UpgradeFolders:           map[string]string{"vX.X.X": "vX.X.X"},
+			MaxNumberOfRestarts:      3,
+			RestartsDelaySeconds:     5,
+			StopSignalTimeoutSeconds: 15,
+			AutoInstall: AutoInstallConfig{
+				Enabled:               true,
+				GithubRepositoryOwner: "vegaprotocol",
+				GithubRepository:      "vega",
+				Assets: AssetsConfig{
+					Vega: fmt.Sprintf("vega-%s-%s", runtime.GOOS, "amd64"),
```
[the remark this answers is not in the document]
karlem: Because we don't currently produce binaries for ARM64 :/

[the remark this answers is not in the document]
jeremyletang: haha correct.

jeremyletang (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

PR #5668, opened 2022-07-30
fix: fix oracle termination before enactment

### LINKED ISSUE #5329: Panic - updating oracle termination condition and value internal oracle spec
This submission on an enacted continuous market updates the oracle term condition to be 10s in the future. When we reach the time core throws a panic.

              conditions {
                operator: OPERATOR_EQUALS
                value: "1651598980"
              }
              

```
terms {
  update_market {
    market_id: "27587c64bcb66a0f7ec17b53159b7ad5861147f2aa28cf954b2b350112be85e4"
    changes {
      instrument {
        code: "CRYPTO:VEGA CORE SYS TEST"
        future {
          quote_name: "USD"
          oracle_spec_for_settlement_price {
            pub_keys: "c49d680fc28320abe5a0d36a0ab309cbe1a40bfaad2b0bf8d22adaad005cf1bd"
            filters {
              key {
                name: "oracle.test.settlement"
                type: TYPE_INTEGER
              }
              conditions {
                operator: OPERATOR_GREATER_THAN_OR_EQUAL
                value: "0"
              }
            }
          }
          oracle_spec_for_trading_termination {
            pub_keys: "c49d680fc28320abe5a0d36a0ab309cbe1a40bfaad2b0bf8d22adaad005cf1bd"
            filters {
              key {
                name: "vegaprotocol.builtin.timestamp"
                type: TYPE_TIMESTAMP
              }
              conditions {
                operator: OPERATOR_EQUALS
                value: "1651598980"
              }
            }
          }
          oracle_spec_binding {
            settlement_price_property: "oracle.test.settlement"
            trading_termination_property: "vegaprotocol.builtin.timestamp"
          }
          settlement_price_decimals: 5
        }
      }
      metadata: "asset_class:fx/crypto"
      metadata: "product:futures"
      log_normal {
        risk_aversion_parameter: 0.001
        tau: 0.00011407711613050422
        params {
          r: 0.016
          sigma: 1.5
        }
      }
    }
  }
}
rationale {
  description: "Update market"
}
```

Enacted market
```json
{
  "id": "4725-33",
  "block": "F494941B6D7015F02EC59022B7D4A308A1975E79992ED226939728AA53BB2762",
  "type": "BUS_EVENT_TYPE_PROPOSAL",
  "proposal": {
    "id": "27587c64bcb66a0f7ec17b53159b7ad5861147f2aa28cf954b2b350112be85e4",
    "reference": "F94DE11BE1",
    "partyId": "c49d680fc28320abe5a0d36a0ab309cbe1a40bfaad2b0bf8d22adaad005cf1bd",
    "state": "STATE_ENACTED",
    "timestamp": "1651598932232997000",
    "terms": {
      "closingTimestamp": "1651598942",
      "enactmentTimestamp": "1651598944",
      "validationTimestamp": "1651598932",
      "newMarket": {
        "changes": {
          "instrument": {
            "name": "Futures System-Tests Market",
            "code": "CRYPTO:VEGA CORE SYS TEST",
            "future": {
              "settlementAsset": "b66b7c057f3b47508f8402cf8e1a7d34455e26b5f1eef98a318bd363ce5c0464",
              "quoteName": "USD",
              "oracleSpecForSettlementPrice": {
                "pubKeys": [
                  "c49d680fc28320abe5a0d36a0ab309cbe1a40bfaad2b0bf8d22adaad005cf1bd"
                ],
                "filters": [
                  {
                    "key": {
                      "name": "oracle.test.settlement",
                      "type": "TYPE_INTEGER"
                    },
                    "conditions": [
                      {
                        "operator": "OPERATOR_GREATER_THAN_OR_EQUAL",
                        "value": "0"
                      }
                    ]
                  }
                ]
              },
              "oracleSpecForTradingTermination": {
                "pubKeys": [
                  "c49d680fc28320abe5a0d36a0ab309cbe1a40bfaad2b0bf8d22adaad005cf1bd"
                ],
                "filters": [
                  {
                    "key": {
                      "name": "vegaprotocol.builtin.timestamp",
                      "type": "TYPE_TIMESTAMP"
                    },
                    "conditions": [
                      {
                        "operator": "OPERATOR_GREATER_THAN_OR_EQUAL",
                        "value": "1651599432"
                      }
                    ]
                  }
                ]
              },
              "oracleSpecBinding": {
                "settlementPriceProperty": "oracle.test.settlement",
                "tradingTerminationProperty": "vegaprotocol.builtin.timestamp"
              }
            }
          },
          "decimalPlaces": "5",
          "metadata": [
            "asset_class:fx/crypto",
            "product:futures"
          ],
          "priceMonitoringParameters": {},
          "liquidityMonitoringParameters": {
            "targetStakeParameters": {
              "timeWindow": "10",
              "scalingFactor": 5
            },
            "auctionExtension": "1"
          },
          "logNormal": {
            "riskAversionParameter": 0.001,
            "tau": 0.00011407711613050422,
            "params": {
              "r": 0.016,
              "sigma": 1.5
            }
          },
          "positionDecimalPlaces": "5"
        },
        "liquidityCommitment": {
          "commitmentAmount": "3905000000000000000000000",
          "fee": "0.3",
          "sells": [
            {
              "reference": "PEGGED_REFERENCE_BEST_ASK",
              "proportion": 13,
              "offset": "100000"
            }
          ],
          "buys": [
            {
              "reference": "PEGGED_REFERENCE_BEST_BID",
              "proportion": 2,
              "offset": "100000"
            }
          ],
          "reference": "2684B3671F"
        }
      }
    }
  },
  "version": 1,
  "chainId": "test-chain-9lCsDX"
}
```

Logs:
```logs
2022-05-03T15:48:13.498+0100    ERROR   tm.abci.socket-server   server/socket_server.go:142     Connection error [err recovered from panic: unable to start market opening auction
goroutine 118 [running]:
github.com/tendermint/tendermint/abci/server.(*SocketServer).handleRequests.func1()
        /Users/paulwebb/repos/vega/vendor/github.com/tendermint/tendermint/abci/server/socket_server.go:165 +0x85
panic({0x2296280, 0xc006b5b5b0})
        /usr/local/Cellar/go/1.18.1/libexec/src/runtime/panic.go:838 +0x207
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0002980c0, {0xc004773f00, 0x2, 0x2})
        /Users/paulwebb/repos/vega/vendor/go.uber.org/zap/zapcore/entry.go:232 +0x44c
go.uber.org/zap.(*Logger).Panic(0xc0000003c0?, {0x25c5457?, 0xc0026c2960?}, {0xc004773f00, 0x2, 0x2})
        /Users/paulwebb/repos/vega/vendor/go.uber.org/zap/logger.go:230 +0x59
code.vegaprotocol.io/vega/processor.(*App).onTick(0xc0001762c0, {0x299f498, 0xc0026c2960}, {0x29a6ea0?, 0x1?, 0x0?})
        /Users/paulwebb/repos/vega/processor/abci.go:1340 +0x369
code.vegaprotocol.io/vega/vegatime.(*Svc).notify(0xc000574000?, {0x299f498, 0xc0026c2960}, {0xc00007f201?, 0xc0000f8a60?, 0x0?})
        /Users/paulwebb/repos/vega/vegatime/service.go:84 +0x114
code.vegaprotocol.io/vega/vegatime.(*Svc).SetTimeNow(0xc0001658f0, {0x299f498, 0xc0026c2960}, {0x3510920?, 0x2295880?, 0x0?})
        /Users/paulwebb/repos/vega/vegatime/service.go:59 +0x227
code.vegaprotocol.io/vega/processor.(*App).OnBeginBlock(_, {{0xc00c0a9fc0, 0x20, 0x20}, {{0xb, 0x1}, {0xc004613038, 0x11}, 0x2211, {0x10d50178, ...}, ...}, ...})
        /Users/paulwebb/repos/vega/processor/abci.go:606 +0x5cb
code.vegaprotocol.io/vega/blockchain/abci.(*App).BeginBlock(_, {{0xc00c0a9fc0, 0x20, 0x20}, {{0xb, 0x1}, {0xc004613038, 0x11}, 0x2211, {0x10d50178, ...}, ...}, ...})
        /Users/paulwebb/repos/vega/blockchain/abci/abci.go:50 +0x75
code.vegaprotocol.io/vega/cmd/vega/node.(*appW).BeginBlock(_, {{0xc00c0a9fc0, 0x20, 0x20}, {{0xb, 0x1}, {0xc004613038, 0x11}, 0x2211, {0x10d50178, ...}, ...}, ...})
        /Users/paulwebb/repos/vega/cmd/vega/node/app_wrapper.go:60 +0x55
github.com/tendermint/tendermint/abci/server.(*SocketServer).handleRequest(0x2988ec0?, 0xc0006be4e0?, 0x299a6e0?)
        /Users/paulwebb/repos/vega/vendor/github.com/tendermint/tendermint/abci/server/socket_server.go:222 +0x975
github.com/tendermint/tendermint/abci/server.(*SocketServer).handleRequests(0xc00060cdc0, 0xc0007128a0, {0xa2d7008?, 0xc00073a008?}, 0xc0001c98b0?)
        /Users/paulwebb/repos/vega/vendor/github.com/tendermint/tendermint/abci/server/socket_server.go:189 +0x2a6
created by github.com/tendermint/tendermint/abci/server.(*SocketServer).acceptConnectionsRoutine
        /Users/paulwebb/repos/vega/vendor/github.com/tendermint/tendermint/abci/server/socket_server.go:127 +0x25f
```

AnExsomnis:
Closes #5329

### DIFF
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -774,6 +774,59 @@ func (m *Market) updateMarketValueProxy() {
 	m.stateChanged = true
 }
 
+func (m *Market) removeOrders(ctx context.Context) {
+	// remove all order from the book
+	// and send events with the stopped status
+	orders := append(m.matching.Settled(), m.peggedOrders.Settled()...)
+	orderEvents := make([]events.Event, 0, len(orders))
+	for _, v := range orders {
+		orderEvents = append(orderEvents, events.NewOrderEvent(ctx, v))
+	}
+
+	m.broker.SendBatch(orderEvents)
+}
+
+func (m *Market) cleanMarketWithState(ctx context.Context, mktState types.MarketState) error {
+	asset, _ := m.mkt.GetAsset()
+	parties := make([]string, 0, len(m.parties))
+	for k := range m.parties {
+		parties = append(parties, k)
+	}
+
+	sort.Strings(parties)
+	clearMarketTransfers, err := m.collateral.ClearMarket(ctx, m.GetID(), asset, parties)
+	if err != nil {
+		m.log.Error("Clear market error",
+			logging.MarketID(m.GetID()),
+			logging.Error(err))
+		return err
+	}
+
+	// unregister state-variables
+	m.stateVarEngine.UnregisterStateVariable(asset, m.mkt.ID)
+
+	m.broker.Send(events.NewTransferResponse(ctx, clearMarketTransfers))
+	m.mkt.State = mktState
+	m.mkt.TradingMode = types.MarketTradingModeNoTrading
+	m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+
+	return nil
+}
+
+func (m *Market) closeCancelledMarket(ctx context.Context, t time.Time) error {
+	m.tradableInstrument.Instrument.Unsubscribe(ctx)
+
+	err := m.cleanMarketWithState(ctx, types.MarketStateCancelled)
+	if err != nil {
+		return err
+	}
+
+	m.closed = true
+	m.stateChanged = true
+
+	return m.Reject(ctx)
+}
+
 func (m *Market) closeMarket(ctx context.Context, t time.Time) error {
 	// market is closed, final settlement
 	// call settlement and stuff
@@ -799,38 +852,12 @@ func (m *Market) closeMarket(ctx context.Context, t time.Time) error {
 	// this will be next block
 	m.broker.Send(events.NewTransferResponse(ctx, transfers))
 
-	asset, _ := m.mkt.GetAsset()
-	parties := make([]string, 0, len(m.parties))
-	for k := range m.parties {
-		parties = append(parties, k)
-	}
-
-	sort.Strings(parties)
-	clearMarketTransfers, err := m.collateral.ClearMarket(ctx, m.GetID(), asset, parties)
+	err = m.cleanMarketWithState(ctx, types.MarketStateSettled)
 	if err != nil {
-		m.log.Error("Clear market error",
-			logging.MarketID(m.GetID()),
-			logging.Error(err))
 		return err
 	}
 
-	// unregister state-variables
-	m.stateVarEngine.UnregisterStateVariable(asset, m.mkt.ID)
-
-	m.broker.Send(events.NewTransferResponse(ctx, clearMarketTransfers))
-	m.mkt.State = types.MarketStateSettled
-	m.mkt.TradingMode = types.MarketTradingModeNoTrading
-	m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
-
-	// remove all order from the book
-	// and send events with the stopped status
-	orders := append(m.matching.Settled(), m.peggedOrders.Settled()...)
-	orderEvents := make([]events.Event, 0, len(orders))
-	for _, v := range orders {
-		orderEvents = append(orderEvents, events.NewOrderEvent(ctx, v))
-	}
-
-	m.broker.SendBatch(orderEvents)
+	m.removeOrders(ctx)
 
 	m.stateChanged = true
 	return nil
@@ -3163,16 +3190,34 @@ func (m *Market) commandLiquidityAuction(ctx context.Context) {
 func (m *Market) tradingTerminated(ctx context.Context, tt bool) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.mkt.State = types.MarketStateTradingTerminated
-	m.mkt.TradingMode = types.MarketTradingModeNoTrading
-	m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
-	m.stateChanged = true
 
-	if m.settlementPriceInMarket == nil {
-		m.log.Debug("no settlement price", logging.MarketID(m.GetID()))
+	if m.mkt.State != types.MarketStateProposed && m.mkt.State != types.MarketStatePending {
+		m.mkt.State = types.MarketStateTradingTerminated
+		m.mkt.TradingMode = types.MarketTradingModeNoTrading
+		m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+		m.stateChanged = true
+
+		if m.settlementPriceInMarket == nil {
+			m.log.Debug("no settlement price", logging.MarketID(m.GetID()))
+			return
+		}
+		m.settlementPriceWithLock(ctx)
+	} else {
+		for party := range m.parties {
+			_, err := m.CancelAllOrders(ctx, party)
+			if err != nil {
+				m.log.Debug("could not cancel orders for party", logging.PartyID(party), logging.Error(err))
+			}
+		}
+		err := m.closeCancelledMarket(ctx, m.timeService.GetTimeNow())
+		if err != nil {
+			m.log.Debug("could not close market", logging.MarketID(m.GetID()))
+			return
+		}
+
+		m.log.Debug("market must not terminated before its enactment time", logging.MarketID(m.GetID()))
 		return
 	}
-	m.settlementPriceWithLock(ctx)
 }
 
 func (m *Market) settlementPrice(ctx context.Context, settlementPrice *num.Uint) {
@@ -3184,7 +3229,7 @@ func (m *Market) settlementPrice(ctx context.Context, settlementPrice *num.Uint)
 	m.settlementPriceWithLock(ctx)
 }
 
-// NB this musy be called with the lock already acquired.
+// NB this must be called with the lock already acquired.
 func (m *Market) settlementPriceWithLock(ctx context.Context) {
 	if m.closed {
 		return
--- a/core/governance/checkpoint.go
+++ b/core/governance/checkpoint.go
@@ -29,6 +29,11 @@ import (
 	"code.vegaprotocol.io/vega/libs/proto"
 )
 
+type enactmentTime struct {
+	current         int64
+	shouldNotVerify bool
+}
+
 func (e *Engine) Name() types.CheckpointName {
 	return types.GovernanceCheckpoint
 }
@@ -73,11 +78,14 @@ func (e *Engine) Load(ctx context.Context, data []byte) error {
 
 		switch prop.Terms.Change.GetTermType() {
 		case types.ProposalTermsTypeNewMarket:
+			enct := &enactmentTime{}
 			// if the proposal is for a new market we want to restore it such that it will be in opening auction
 			if p.Terms.EnactmentTimestamp <= now.Unix() {
 				prop.Terms.EnactmentTimestamp = now.Add(duration).Unix()
+				enct.shouldNotVerify = true
 			}
-			toSubmit, err := e.intoToSubmit(ctx, prop)
+			enct.current = prop.Terms.EnactmentTimestamp
+			toSubmit, err := e.intoToSubmit(ctx, prop, enct)
 			if err != nil {
 				e.log.Panic("Failed to convert proposal into market")
 			}
--- a/core/governance/engine.go
+++ b/core/governance/engine.go
@@ -56,6 +56,7 @@ type Broker interface {
 type Markets interface {
 	MarketExists(market string) bool
 	GetMarket(market string) (types.Market, bool)
+	GetMarketState(market string) (types.MarketState, error)
 	GetEquityLikeShareForMarketAndParty(market, party string) (num.Decimal, bool)
 	RestoreMarket(ctx context.Context, marketConfig *types.Market) error
 	RestoreMarketWithLiquidityProvision(ctx context.Context, marketConfig *types.Market, lp *types.LiquidityProvisionSubmission, lpid, deterministicID string) error
@@ -307,9 +308,9 @@ func (e *Engine) removeProposal(ctx context.Context, id string) {
 // OnTick triggers time bound state changes.
 func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteClosed) {
 	var (
-		toBeEnacted []*ToEnact
-		voteClosed  []*VoteClosed
-		toBeRemoved []string // ids
+		preparedToEnact []*ToEnact
+		voteClosed      []*VoteClosed
+		toBeRemoved     []string // ids
 	)
 
 	now := t.Unix()
@@ -332,14 +333,9 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
 					logging.Error(err))
 			} else {
 				toBeRemoved = append(toBeRemoved, proposal.ID)
-				toBeEnacted = append(toBeEnacted, enact)
+				preparedToEnact = append(preparedToEnact, enact)
 			}
 		}
-	}
-
-	// now we iterate over all proposal ids to remove them from the list
-	for _, id := range toBeRemoved {
-		e.removeProposal(ctx, id)
 	}
 
 	// then get all proposal accepted through node validation, and start their vote time.
@@ -369,10 +365,23 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
 		e.gss.changedNodeValidation = true
 	}
 
-	for _, ep := range toBeEnacted {
+	toBeEnacted := []*ToEnact{}
+	for i, ep := range preparedToEnact {
 		// this is the new market proposal, and should already be in the slice
 		prop := *ep.ProposalData()
+
 		if prop.Terms.Change.GetTermType() == types.ProposalTermsTypeNewMarket {
+			mktState, err := e.markets.GetMarketState(prop.ID)
+			if err != nil {
+				e.log.Error("could not get state of market %s", logging.String("market-id", prop.ID))
+				continue
+			}
+
+			if mktState == types.MarketStateCancelled {
+				toBeRemoved = append(toBeRemoved, prop.ID)
+				continue
+			}
+
 			// just in case the proposal wasn't added for whatever reason (shouldn't be possible)
 			found := false
 			for i, p := range e.enactedProposals {
@@ -384,11 +393,18 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
 			}
 			// no need to append
 			if found {
+				toBeEnacted = append(toBeEnacted, preparedToEnact[i])
 				continue
 			}
 		}
 		// take a copy in the state just before the proposal was enacted
 		e.enactedProposals = append(e.enactedProposals, &prop)
+		toBeEnacted = append(toBeEnacted, preparedToEnact[i])
+	}
+
+	// now we iterate over all proposal ids to remove them from the list
+	for _, id := range toBeRemoved {
+		e.removeProposal(ctx, id)
 	}
 
 	if len(toBeEnacted) != 0 {
@@ -460,7 +476,7 @@ func (e *Engine) SubmitProposal(
 		e.startProposal(p)
 	}
 
-	return e.intoToSubmit(ctx, p)
+	return e.intoToSubmit(ctx, p, &enactmentTime{current: p.Terms.EnactmentTimestamp})
 }
 
 func (e *Engine) RejectProposal(
@@ -496,7 +512,7 @@ func (e *Engine) rejectProposal(ctx context.Context, p *types.Proposal, r types.
 
 // toSubmit build the return response for the SubmitProposal
 // method.
-func (e *Engine) intoToSubmit(ctx context.Context, p *types.Proposal) (*ToSubmit, error) {
+func (e *Engine) intoToSubmit(ctx context.Context, p *types.Proposal, enct *enactmentTime) (*ToSubmit, error) {
 	tsb := &ToSubmit{p: p}
 
 	switch p.Terms.Change.GetTermType() {
@@ -510,7 +526,7 @@ func (e *Engine) intoToSubmit(ctx context.Context, p *types.Proposal) (*ToSubmit
 		enactTime := time.Unix(p.Terms.EnactmentTimestamp, 0)
 		newMarket := p.Terms.GetNewMarket()
 		auctionDuration := enactTime.Sub(closeTime)
-		if perr, err := validateNewMarketChange(newMarket, e.assets, true, e.netp, auctionDuration); err != nil {
+		if perr, err := validateNewMarketChange(newMarket, e.assets, true, e.netp, auctionDuration, enct); err != nil {
 			e.rejectProposal(ctx, p, perr, err)
 			return nil, fmt.Errorf("%w, %v", err, perr)
 		}
@@ -797,13 +813,15 @@ func (e *Engine) validateMarketUpdate(proposal *types.Proposal, params *Proposal
 
 // validates proposed change.
 func (e *Engine) validateChange(terms *types.ProposalTerms) (types.ProposalError, error) {
+	enactTime := time.Unix(terms.EnactmentTimestamp, 0)
+	enct := &enactmentTime{current: terms.EnactmentTimestamp}
+
 	switch terms.Change.GetTermType() {
 	case types.ProposalTermsTypeNewMarket:
 		closeTime := time.Unix(terms.ClosingTimestamp, 0)
-		enactTime := time.Unix(terms.EnactmentTimestamp, 0)
-		return validateNewMarketChange(terms.GetNewMarket(), e.assets, true, e.netp, enactTime.Sub(closeTime))
+		return validateNewMarketChange(terms.GetNewMarket(), e.assets, true, e.netp, enactTime.Sub(closeTime), enct)
 	case types.ProposalTermsTypeUpdateMarket:
-		return validateUpdateMarketChange(terms.GetUpdateMarket())
+		return validateUpdateMarketChange(terms.GetUpdateMarket(), enct)
 	case types.ProposalTermsTypeNewAsset:
 		return terms.GetNewAsset().Validate()
 	case types.ProposalTermsTypeUpdateAsset:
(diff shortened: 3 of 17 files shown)


core/governance/market.go
```diff
@@ -231,6 +235,27 @@ func validateFuture(future *types.FutureProduct, decimals uint64, assets Assets,
 		return types.ProposalErrorInvalidFutureProduct, ErrMissingOracleSpecForTradingTermination
 	}
 
+	for i, f := range future.OracleSpecForTradingTermination.ToOracleSpec().Filters {
+		if f.Key.Type == oraclespb.PropertyKey_TYPE_TIMESTAMP {
```
wwestgarth, line 239: I think it makes sense to do this validation when there is a timestamp condition in oracle to catch it as soon as possible, but not all oracles will have a timestamp property. And those could still terminate a market before enactment while circumventing this check.

So I imagine we'll still need some code somewhere to catch that scenario and work out what to do when terminating that orcale submission comes in when the market isn't enacted (ignore it, apply it and stop, something else)

AnExsomnis: You mean that not all oracles will have a "PropertyKey_TYPE_TIMESTAMP" set type for the `key` field, but still have a timestamp for the `value` field
AND
will contain a filter for `OracleSpecForTradingTermination`?

AnExsomnis: As far as I see now the property name "vegaprotocol.builtin.timestamp" is used to set also its type `oraclespb.PropertyKey_TYPE_TIMESTAMP` and the callback that uses both values inside the filter key for making the subscription here:
[[OracleSpecForTradingTermination](https://github.com/vegaprotocol/vega/blob/develop/core/products/future.go#L266)](https://github.com/vegaprotocol/vega/blob/develop/core/products/future.go#L266)

Maybe you mean that in `toTimestampCondiiton` - we receive only a condition

wwestgarth: The `vegaprotocol.builtin.timestamp` oracles will be fine and caught here in this new check. So thats good.

Non-time-triggered oracles will just come from a transacstion submission from whatever public-keys were in the spec. For example here are the oracle-specs on testnet for terminated/settling a market: https://explorer.fairground.wtf/oracles and the property values are all other `TypeBoolean` or `TypeInteger`.

In this case its then possible for someone to propose a market with enactment in like 5 days time, but for the oracle submissions for termination/settlement to be submitted after 4 days. Then when we hit day 5 it panics because we try to enact a market in a unexpected state. This is what (I'm assuming) happened here: https://vegaprotocol.slack.com/archives/CEE07A084/p1658934399014629

The market was settled, cleaned up and removed from the execution engine and then when enactment came we panicked because the market didn't exist anymore: 
```
processor/abci.go:1282        unable to start market opening auction        {"market-id": "89021b6e5c6bff671918ea93acd256fef79984303d6b5bcf40cf13cdbc1a8824", "error": "market does not exist"}
```

barnabee: Agree that early trading termination cannot be prevented with validation at market creation time. It’s definitely the case that some data sources will have no timestamp at all or the market may for some reason want to filter on a date/time value that’s in the past.

That means that it is absolutely possible to receive a valid (by the market and data source definition) settlement or trading terminated trigger event before the market is in a state that can receive it, and it needs to be processed safely and reasonably.

The [market lifecycle spec](https://github.com/vegaprotocol/specs/blob/master/protocol/0043-MKTL-market_lifecycle.md) details all the valid stage transitions. It should not be possible to exit the proposed or pending states except in the conditions and to the states defined in the market lifecycle. 

For example, this spec states that after the enactment date, if trading terminated is triggered, the market becomes [Cancelled](https://github.com/vegaprotocol/specs/blob/master/protocol/0043-MKTL-market_lifecycle.md#cancelled). (Note that settlement cannot occur unless the market is in trading terminated as per [here](https://github.com/vegaprotocol/specs/blob/master/protocol/0016-PFUT-product_builtin_future.md#42-final-settlement-expiry) and [here](https://github.com/vegaprotocol/specs/blob/master/protocol/0043-MKTL-market_lifecycle.md#settled).)

It’s possible we need to add some clarity to the spec on when to process and when to reject/ignore data sources in certain states. It’s also possible that we’ve missed some edge cases or complexities and should revisit the current design…

core/governance/checkpoint.go
```diff
@@ -73,11 +78,16 @@ func (e *Engine) Load(ctx context.Context, data []byte) error {
 
 		switch prop.Terms.Change.GetTermType() {
 		case types.ProposalTermsTypeNewMarket:
+			et := &enactmentTime{
```
ze97286, line 81: shouldn't this be moved after line 89? because we're potentially editing the enactment timestamp

AnExsomnis: Cleared it like this https://github.com/vegaprotocol/vega/pull/5668/commits/79f0d5890049572bdf4e3d1052d28c3d9a628d54#diff-92b369672ec5368c6bacbb63f9c3717a133c96f06f473e577ca711abefbcbe23R81

core/governance/checkpoint.go
```diff
@@ -73,11 +78,14 @@ func (e *Engine) Load(ctx context.Context, data []byte) error {
...
 		case types.ProposalTermsTypeNewMarket:
+			et := &enactmentTime{}
 			// if the proposal is for a new market we want to restore it such that it will be in opening auction
 			if p.Terms.EnactmentTimestamp <= now.Unix() {
 				prop.Terms.EnactmentTimestamp = now.Add(duration).Unix()
+				et.lastUpdate = int64(duration)
 			}
-			toSubmit, err := e.intoToSubmit(ctx, prop)
+			et.current = prop.Terms.EnactmentTimestamp
+			toSubmit, err := e.intoToSubmit(ctx, prop, et)
```
ze97286, line 88: when we're restoring from a checkpoint, I don't think we should be validating the oracle termination timestamp vs enactment time or vs anything else. If we're restoring the market, it means it was enacted already, meaning it was valid at that time. If the oracle termination time is earlier than now or than the new enactment time (i.e. now + duration) it only means that it should fire immediately to terminate the market. 

@wwestgarth @jeremyletang @EVODelavega

wwestgarth: I feel like it should go straight to terminated too, but then what value is there in even restoring it.

I'm also now wondering what happens to markets that have been terminated+settle during a checkpoint reload. Because they are removed from the execution engine but not removed as enacted proposals in governance. Do those get restored and brought back to life??

ze97286: re your first point - if it's time, it will get immediately terminated so we don't need to worry or care about it. 

re your second point, you're right. The problem is that markets are not first class citizens in checkpoints, so when we save proposals in checkpoint we have no idea what the market state was at the time of the checkpoint, so I believe when its proposal is restored the market is reinstated in opening auction. So even if only terminated, or terminated and settled will reopen from scratch. 

Slightly orthogonally I'm wondering if checkpoints and LNL would live much longer (or at all) after we have the protocol upgrade?

wwestgarth: I thought checkpoints were the safety-net for when snapshots don't work ;)

My understanding of how the code is right now is that any restored time-triggered-oracle with a timestamp in the past will pass the validation by the timestamp-shifting, and then be ignored until the market's opening auction ends. Which maybe isn't ideal, but it doesn't sound like any of the market restoring is really working properly.

ze97286: re the future of LNL I think it's something to clarify with @barnabee and @davidsiska-vega.

re termination - my point is why do this shifting at all and do the opening auction vs terminate straight away?

AnExsomnis: @ze97286 , @wwestgarth ,
Applied a change that will make the restored market skip the check as:
https://github.com/vegaprotocol/vega/pull/5668/commits/c418a28e496ffc53b057263de01569e7b7527bb1#diff-92b369672ec5368c6bacbb63f9c3717a133c96f06f473e577ca711abefbcbe23R85
So if we did not set the `ShouldNotVerify` value to true -> verify it

core/execution/market.go
```diff
@@ -3154,16 +3154,22 @@ func (m *Market) commandLiquidityAuction(ctx context.Context) {
...
+
+		if m.settlementPriceInMarket == nil {
+			m.log.Debug("no settlement price", logging.MarketID(m.GetID()))
+			return
+		}
+		m.settlementPriceWithLock(ctx)
+	} else {
+		m.log.Debug("Market is not terminated before its enactment time")
 		return
 	}
```
jeremyletang, line 3172: Seeing Barney's somment, it seems that in the cose termination is triggered before the end of the opening auction, the market should then be Cancelled. I suppose we should add the Cancelled type to the market state enum, and set it to Cancell in this else condition.

AnExsomnis: @jeremyletang ,
Added changes to `tradingTerminated` step to set market state to "Canceled" and cancel all orders for each party:
https://github.com/vegaprotocol/vega/pull/5668/commits/2a27fdbb4a9172200b26f4fe28500fe548bd52d9#diff-1782f5982916957ff0e7e82b0c7130c6d4c8ab4eb2e06f5c7b5b4a5925e15bcfR3181

As well as handing if the market state is handled in the `OnTick` step in the execution engine:
https://github.com/vegaprotocol/vega/pull/5668/commits/2a27fdbb4a9172200b26f4fe28500fe548bd52d9#diff-dedf0633c5a9f40bc6bd64eda022d287b45a334e77b08ca25c307069777582b5R375
-> removing from list, sending to the list of proposals to be rejected

core/execution/market.go
```diff
@@ -3166,16 +3207,34 @@ func (m *Market) commandLiquidityAuction(ctx context.Context) {
...
+			return
+		}
+		m.settlementPriceWithLock(ctx)
+	} else {
+		for party := range m.parties {
+			_, err := m.CancelAllOrders(ctx, party)
+			if err != nil {
+				m.log.Debug("could not cancel orders for party", logging.PartyID(party), logging.Error(err))
+			}
+			err = m.closeCancelledMarket(ctx, m.timeService.GetTimeNow())
```
ze97286, line 3228: this is definitely not right. why is this done for every party? separately why is this done at all? if the only case when a market gets into cancelled state is when it is being terminated before it is being enacted, then there can't be any orders - so there's still some cleanup to do but I don't think you need everything in closeCancelledMarket

core/execution/market.go
```diff
@@ -3166,16 +3207,34 @@ func (m *Market) commandLiquidityAuction(ctx context.Context) {
...
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.mkt.State = types.MarketStateTradingTerminated
-	m.mkt.TradingMode = types.MarketTradingModeNoTrading
-	m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
-	m.stateChanged = true
 
-	if m.settlementPriceInMarket == nil {
-		m.log.Debug("no settlement price", logging.MarketID(m.GetID()))
+	if m.mkt.State != types.MarketStateProposed && m.mkt.State != types.MarketStatePending {
```
ze97286, line 3211: shouldn't this only be != MarketStateProposed? why pending too?

core/execution/market.go
jeremyletang: There would be orders eventually. The market is in opening auction between during the PENDING state (which is meant to end when the market is enacted). This order would require margin etc.

core/execution/market.go
jeremyletang: Pending it he OpeningAuction

ze97286: I know but I'm saying that this should be legit to terminate in an opening auction, that's not the same as terminate a market pre enactment, no?

core/execution/market.go
ze97286: the way the conditions are written now, yes, but I'm asking if this is what we want - i.e. why terminating during opening auction is treated similarly to terminating before enactment

core/execution/market.go
jeremyletang: I think it is the same.

core/governance/engine.go
```diff
@@ -368,9 +369,25 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
...
+		mktState, err := e.markets.GetMarketState(prop.ID)
+		if err != nil {
+			e.log.Error("could not get state of market %s", logging.String("market-id", prop.ID))
+			continue
+		}
+		if mktState == types.MarketStateCancelled {
+			toBeRemoved = append(toBeRemoved, prop.ID)
+
+			// Remove from the list of toBeEnacted proposals that is going to be returned
+			copy(toBeEnacted[i:], toBeEnacted[i+1:])
```
ze97286, line 385: I don't think its safe to remove elements this way from toBeEnacted while iterating it.

core/governance/engine.go
```diff
@@ -368,9 +369,25 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
...
 		// this is the new market proposal, and should already be in the slice
 		prop := *ep.ProposalData()
+
+		mktState, err := e.markets.GetMarketState(prop.ID)
+		if err != nil {
+			e.log.Error("could not get state of market %s", logging.String("market-id", prop.ID))
+			continue
+		}
+		if mktState == types.MarketStateCancelled {
+			toBeRemoved = append(toBeRemoved, prop.ID)
```
ze97286, line 382: I think this has no effect because the removal already takes place earlier in line 340 in the old code

core/execution/market.go
ze97286: as confirmed in spec, this is fine - proposed and pending should be cancelled if terminated.

core/execution/market.go
```diff
@@ -3166,16 +3207,34 @@ func (m *Market) commandLiquidityAuction(ctx context.Context) {
...
+		m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+		m.stateChanged = true
+
+		if m.settlementPriceInMarket == nil {
+			m.log.Debug("no settlement price", logging.MarketID(m.GetID()))
+			return
+		}
+		m.settlementPriceWithLock(ctx)
+	} else {
+		for party := range m.parties {
```
ze97286, line 3223: why not move this loop into closeCancelledMarket?

core/execution/market.go
```diff
@@ -774,6 +774,47 @@ func (m *Market) updateMarketValueProxy() {
...
+	}
+
+	// unregister state-variables
+	m.stateVarEngine.UnregisterStateVariable(asset, m.mkt.ID)
+
+	m.broker.Send(events.NewTransferResponse(ctx, clearMarketTransfers))
+	m.mkt.TradingMode = types.MarketTradingModeNoTrading
+	m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+	m.closed = true
+	m.mkt.State = types.MarketStateCancelled
```
ze97286, line 802: should we not also have here:

`m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))`

core/execution/market.go
```diff
@@ -774,6 +774,47 @@ func (m *Market) updateMarketValueProxy() {
 	m.stateChanged = true
 }
 
+func (m *Market) closeCancelledMarket(ctx context.Context, t time.Time) error {
```
ze97286, line 777: this whole function is identical to close market from line 800 onwards. I'd suggest refactoring closeMarket to split out the settlement part rather than repeating 40 lines here.

core/execution/market.go
AnExsomnis: I dont understand, do you mean to move `CancelAllOrders` step inside the `closeCanceledMarket`?
I think these should be separate steps - the way it is handled for closing market in the normal way.

ze97286: primarily what I'm saying is that the call to closeCancelledMarket should not be done inside this loop.

core/execution/market.go
AnExsomnis: The refactoring is done here https://github.com/vegaprotocol/vega/pull/5668/commits/a11e922add05a5659da732ef070f6e0cc73cb00f#diff-1782f5982916957ff0e7e82b0c7130c6d4c8ab4eb2e06f5c7b5b4a5925e15bcfR777
Thanks!

core/execution/market.go
AnExsomnis: This has been changed here https://github.com/vegaprotocol/vega/pull/5668/commits/a11e922add05a5659da732ef070f6e0cc73cb00f#diff-1782f5982916957ff0e7e82b0c7130c6d4c8ab4eb2e06f5c7b5b4a5925e15bcfR3217
Missed that
Thanks!

core/governance/engine.go
AnExsomnis: I believe calling the removal on each element in the `toBeRemoved` list was done above, because there were no new additions to in until the end.
I moved it further now here https://github.com/vegaprotocol/vega/pull/5668/commits/a11e922add05a5659da732ef070f6e0cc73cb00f#diff-dedf0633c5a9f40bc6bd64eda022d287b45a334e77b08ca25c307069777582b5R408
Thanks!

core/governance/engine.go
AnExsomnis: This has been changed here
https://github.com/vegaprotocol/vega/pull/5668/commits/d7fbfec169e258736d071c9df4b4da92b9e6f6f6

core/execution/market.go
```diff
@@ -774,6 +774,61 @@ func (m *Market) updateMarketValueProxy() {
...
+
+func (m *Market) closeCancelledMarket(ctx context.Context, t time.Time) error {
+	m.tradableInstrument.Instrument.Unsubscribe(ctx)
+
+	err := m.clearAssetCollateral(ctx, types.MarketStateCancelled)
+	if err != nil {
+		return err
+	}
+
+	m.removeOrders(ctx)
```
ze97286, line 824: this doesn't make sense here, there won't be any settled orders, or any orders at all in the book

core/execution/market.go
```diff
@@ -774,6 +774,61 @@ func (m *Market) updateMarketValueProxy() {
...
+	orders := append(m.matching.Settled(), m.peggedOrders.Settled()...)
+	orderEvents := make([]events.Event, 0, len(orders))
+	for _, v := range orders {
+		orderEvents = append(orderEvents, events.NewOrderEvent(ctx, v))
+	}
+
+	m.broker.SendBatch(orderEvents)
+}
+
+func (m *Market) clearAssetCollateral(ctx context.Context, mktState types.MarketState) error {
```
ze97286, line 789: the name now doesn't describe what it's doing.

core/governance/engine.go
```diff
@@ -384,11 +393,18 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
 			}
 			// no need to append
 			if found {
+				toBeEnacted = append(toBeEnacted, preparedToEnact[i])
```
ze97286, line 396: I'm confused, why is this here and not in line 383?

core/execution/market.go
AnExsomnis: You mean this should be removed ?

ze97286: based on what it's currently doing, I think so, yes.

ze97286: because I think after you're calling CancelAllOrders for all parties there would be nothing to do there.

core/governance/engine.go
AnExsomnis: This loop is not working anymore with a list of toBeEnacted objects that is only looped over. We are looping over `preparedToEnact` list that has some items added to `toBeEnacted` and some not.

In line 383, the code already checked if the market state is cancelled - if it was - it was sent to the next iteration.
if not - proceed with the previously existing steps.
In case the proposal is found in the market enacted proposals -> it is added to the list of `toBeEnacted` to be returned.
And since, when it is found - the loop continues directly on the next iteration on line 397 - it is added to the new prepared list here.
In case it was not found, but still needs to be in the list to be returned - it is added in line 402.
If the addition is only in line 402 - it will miss the ones, for with the continue statement on line 397 skips to next iteration.

ze97286: I don't think this is right. I don't think this whole code block 384-399 should be doing anything ever, but I think the logic to actually append to toBeEnacted if found is wrong. I think it shouldn't do anything (as it did before) and then if not found be appended in line 402.

AnExsomnis: Line 402 did not exist in the past; this loop was looping over the enacted objects, and for those not found in the engine were simply added.
In the case of found markets with cancelled state - if it is set in the previous way - it will return a list of toBeEnacted objects that will contain porentially market state canceled

ze97286: this is not what I'm saying at all :) 

if market is cancelled, you have the continue there in line 381, that's fine and correct. 

What I'm saying is that line 396 is wrong. in fact if anything, it should be:
```
if found {
     continue
} else {
     toBeEnacted = append(toBeEnacted, preparedToEnact[i])
}
```
because at least the way I understand this block of code (384-399), it's checking if it's already in the enactedProposals of the engine, and in this case we shouldn't be adding it. 

You're right that BEFORE, it would have been there anyways so you're not breaking it more than it was broken. but I think you can fix it if you agree that it's wrong.

AnExsomnis: Okay, I understand now. Right

core/execution/market.go
AnExsomnis: Yes, right.
I removed this https://github.com/vegaprotocol/vega/pull/5668/commits/e15634c60af3eeff4dd732ebb9f6404c0422e98e#diff-1782f5982916957ff0e7e82b0c7130c6d4c8ab4eb2e06f5c7b5b4a5925e15bcfR824
Thanks!

core/governance/engine.go
AnExsomnis: Basically - since it is not a good idea to change the list that you loop over - I decided to use a new one - and for the objects that were needed to be excluded - do not add them.
But for the ones that were in the list in the previous version -add them when you meet them to the new one

AnExsomnis: Sorry if my explanations are not clear enough; please tell me if I need to explain it another way

ze97286: I don't think you're referring to my point.

What I'm saying is this: 
in develop right now: if it is already enacted (i.e. in e.enactedProposals) we leave it in the toBeEnacted - which is wrong. 
in your change: we do the same thing, which I think is wrong.

ze97286 (dismissed):

ze97286 (approved):

PR #5754, opened 2022-08-02
fix: allow minProposerEquityLikeShare to accept 0

### LINKED ISSUE #5633: The governance parameter governance.proposal.updateMarket.minProposerEquityLikeShare should be allowed to be 0
Currently when the governance parameter `governance.proposal.updateMarket.minProposerEquityLikeShare` is set to 0 in `genesis.json` the node startup fails with the error:

image

However, 0 should be an allowable value for this field, so we should relax the restriction here.

ze97286:
Closes #5633


TomMcL (dismissed):
LGTM

AnExsomnis (approved):

zale144 (approved):

PR #5758, opened 2022-08-03
Fix panic when with unregistering positions

### LINKED ISSUE #5663: Bug: Vega core crashes with "cannot unregister order" error
When running latest develop version of Vega core through a scenario in [Vega market sim](https://github.com/vegaprotocol/vega-market-sim) we hit the below error after a few time steps of trading, crashing the process.

```
2022-07-29T16:34:44.547+0100    PANIC   execution.position      positions/market_position.go:109       
 cannot unregister order with remaining > potential sell 
{"order": "ID(4126765a36b009d1a2aef53c3054ca798b5bee5e555125e40c249ff7823cd431) marketID(0fef6f8418c94f7fe6810b8f3bb598e19d83a4050924da1c97c064769632d29b) party(468547ae734d5312e58c2e6e4fdf0da98732c41225f57149822cea11a1b0cdd9) side(SIDE_SELL) price(111924800) size(1000) remaining(1000) timeInForce(TIME_IN_FORCE_GTC) type(TYPE_LIMIT) status(STATUS_ACTIVE) reference(468547ae734d5312e58c2e6e4fdf0da98732c41225f57149822cea11a1b0cdd9-259b6f1c-2fa6-4a28-9b90-b1ccc0132f9b) reason(ORDER_ERROR_UNSPECIFIED) version(2) batchID(6) liquidityProvisionID() createdAt(1638186876259066000) updatedAt(1638186938259066000) expiresAt(0) originalPrice(1119248) peggedOrder(reference(PEGGED_REFERENCE_BEST_ASK) offset(18))", 
"potential-sell": 515}
```

To replicate, run `python -m vega_sim.parameter_test.run -c TauScaling_v2` from within a fully set up Vega market sim repo and the error should be thrown within a few seconds. Currently the branch `feat/log_from_process` is best to use rather than `develop` for this.

jeremyletang:
close #5663


core/execution/pegged_orders.go
```diff
@@ -84,11 +89,25 @@ func (p *PeggedOrders) Park(o *types.Order) {
 	o.Price = num.UintZero()
 	o.OriginalPrice = num.UintZero()
 
-	p.ordersChanged = true
+	p.parked = append(p.parked, o)
+	p.isParked[o.ID] = struct{}{}
+}
+
+func (p *PeggedOrders) Unpark(o *types.Order) {
```
ze97286, line 96: I don't think this should be passed an order, ID should be enough

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -424,6 +424,16 @@ message Market {
   string settlement_price = 18;
 }
 
+message OrderPartyPair {
+  string order = 1;
```
ze97286, line 428: order_id?

core/execution/pegged_orders.go
jeremyletang: Correct. Probably just copied/past another function here to write it.

protos/sources/vega/snapshot/v1/snapshot.proto
jeremyletang: Can change that yes.

core/execution/market.go
```diff
@@ -668,19 +668,19 @@ func (m *Market) PostRestore(ctx context.Context) error {
 
 	m.liquidity.ReconcileWithOrderBook(m.matching)
 
-	if err := m.peggedOrders.ReconcileWithOrderBook(m.matching); err != nil {
-		return err
-	}
+	// if err := m.peggedOrders.ReconcileWithOrderBook(m.matching); err != nil {
```
ze97286, line 671: remove I guess

core/execution/market.go
```diff
@@ -668,19 +668,19 @@ func (m *Market) PostRestore(ctx context.Context) error {
 
 	m.liquidity.ReconcileWithOrderBook(m.matching)
 
-	if err := m.peggedOrders.ReconcileWithOrderBook(m.matching); err != nil {
```
wwestgarth, line 671: Why don't we need to do this anymore?

jeremyletang: We do not store orders all the time, only when they are parked, which means they are not on the book then

core/execution/pegged_orders.go
jeremyletang: addressed

protos/sources/vega/snapshot/v1/snapshot.proto
jeremyletang: addressed

core/execution/market.go
jeremyletang: addressed

core/execution/market.go
wwestgarth: Ok I believe you, but I'm also nervous because this bug was a bastard to hunt down a fix at the time :p

core/execution/market_for_test.go
```diff
@@ -43,13 +43,14 @@ func (m *Market) GetPeggedOrderCount() int {
...
 // GetParkedOrderCount returns hte number of parked orders in the market.
 func (m *Market) GetParkedOrderCount() int {
-	var count int
-	for _, order := range m.peggedOrders.orders {
-		if order.Status == types.OrderStatusParked {
-			count++
-		}
-	}
-	return count
+	// var count int
```
EVODelavega, line 46: shouldn't we just outright delete this code?

jeremyletang: yes we should

core/execution/pegged_orders.go
```diff
@@ -17,65 +17,70 @@ import (
...
+		p.isParked[v.ID] = struct{}{}
 	}
-	p.orders = newPeggedOrders
-	return nil
+	return p
 }
 
 func (p *PeggedOrders) Changed() bool {
-	return p.ordersChanged
+	return true
```
EVODelavega, line 61: I believe always returning true here would result in every market being re-serialised and re-hashed for each snapshot/checkpoint. Most likely (with these being set far enough apart) this will be the case anyway, but you never know... a "slow" market in auction may not have any state changed during a long auction

ze97286: orders are not saved in snapshots and now we're always serialising the whole thing for snapshot anyways. I should start removing these changed flags and functions.

core/execution/market_for_test.go
jeremyletang: done

core/execution/pegged_orders.go
jeremyletang: correct.

wwestgarth (approved):

PR #5775, opened 2022-08-05
chore: use generics for ID types

### LINKED ISSUE #5774: Refactor datanode IDs
To give different entities different types for their IDs (MarketID, PartyID etc).. but still retain the ability to call methods on a 'base' class, we currently use embedding, e.g.

``` go
// Base ID
type ID string
func (id *ID) SomeMethod() {}

// Specific ID
type MarketID struct{ ID }
func NewMarketID(id string) MarketID {
	return MarketID{ID: ID(id)}
}

// Usage
marketID := NewMarketID("foo")
```

Which is mostly fine but a bit clunky; however it adds a bit of faffery to serialisation; JSON gets its knickers in a twist a bit with embedded objects which has led to a few hacks and finally something I couldn't get around. So I propose we replace it with a generic powered solution:

```go
type ID[T any] string
func (id *ID[T]) SomeMethod() {}

type _Market struct{}  // just a dummy unique type
type MarketID = ID[_Market]

// Usage
marketID := MarketID("foo")
```

Now we still get a new type for each ID, but we get the base ID's methods through generics rather than embedding. It's probably faster, it makes debugging nicer (my variable browser now just shows the string), and you can save some typing. And it fixes a serialization pickle.

pscott31:
closes #5774 
(see issue for rationale)

I think @jeremyletang  actually had a very similar PR that got lost when we switched repos


jeremyletang:
@pscott31 here was the proposal I made: https://github.com/vegaprotocol/data-node/pull/758

Would you like to compare implementation / discuss them?

jeremyletang (approved):

PR #5798, opened 2022-08-07
fix: ensure order expiring are not amendedInPlace as it might not be supported

jeremyletang:
When amending an orde from GTC -> GTT with a time being the current time. The order is expired straight away.

However until now the code is calling amendOrderInPlace, which returns an error if the amend is not supported (not all action are supported in place). By doing so with an amend which e.g also increase the size of  the order, the amend is rejected by the book and the wrong order is returned on cancellation. This order is then sent through the bus.

We stop calling orderAmendInPlace, and just cancel the order. Also we always unregister the position of the existing order instead of the one from the book.

Consideration for future: A failed call to amend in place should panic in the bus.

close #5486


jeremyletang:
@peterbarrow, @EVODelavega  could you look at this fix as well?

@jgsbennett I ran the related test in the system tests, but it seems it expect to see the order as ACTIVE in the stream. However as the time of expiration is already in the past when sent to the network, the order is EXPIRED straight away and never get amended with an ACTIVE state, you may want to change this test.

core/execution/market.go
```diff
@@ -2585,19 +2585,20 @@ func (m *Market) amendOrder(
 
 		// Update the existing message in place before we cancel it
 		if foundOnBook {
-			m.orderAmendInPlace(existingOrder, amendedOrder)
+			// Do not amend in place, the amened could be something
```
peterbarrow, line 2588:
```suggestion
			// Do not amend in place, the amend could be something
```

peterbarrow (dismissed):

peterbarrow (approved):

PR #5824, opened 2022-08-09
Add wallet HTTP API v2 documentation

### LINKED ISSUE #5619: Document the wallet HTTP API

ValentinTrinque:
Close #5619

### DIFF
--- a/wallet/service/openapi.json
+++ b/wallet/service/openapi.json
@@ -1,0 +1,109 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Wallet API",
+    "description": "The version 2 of the wallet HTTP API leverages the wallet JSON-RPC API.\n\nSome useful links:\n- [JSON-RPC 2.0 Specification](https://www.jsonrpc.org/specification)",
+    "version": "2.0.0"
+  },
+  "externalDocs": {
+    "description": "Find out more about Vega REST APIs.",
+    "url": "https://docs.vega.xyz/docs/mainnet/api/rest/overview"
+  },
+  "paths": {
+    "/v2/requests": {
+      "post": {
+        "summary": "Submit a JSON-RPC request to the wallet API.",
+        "description": "This endpoint is the main entrypoint to submit the JSON-RPC requests. The body of the HTTP request should be formatted as a JSON-RPC v2 request. The HTTP response is formatted as the JSON-RPC v2 response.\n\nBe sure to specify the `ID` if you are interested in the response. See [the JSON-RPC notification documentation](https://www.jsonrpc.org/specification#notification) for more details.",
+        "requestBody": {
+          "description": "The JSON-RPC v2 request",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/JsonRpcRequest"
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "description": "Successful operation",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/JsonRpcResponse"
+                }
+              }
+            }
+          },
+          "204": {
+            "description": "The request is considered as a notification because it doesn't have an ID. As a result, no result shall be expected, even in case of failure."
+          },
+          "400": {
+            "description": "The submitted request does not have a valid format, or doesn't comply with the API constraints, or failed."
+          },
+          "500": {
+            "description": "Something failed internally. Nothing can be done about it."
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "schemas": {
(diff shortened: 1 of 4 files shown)


wallet/service/openapi.json
```diff
@@ -0,0 +1,95 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Wallet API",
+    "description": "The version 2 of the wallet HTTP API leverages the wallet JSON-RPC API.\n\nSome useful links:\n- [Documentation of the wallet JSON-RPC API](https://docs.vega.xyz/docs/mainnet/api/jsonrpc/wallet)\n- [JSON-RPC 2.0 Specification](https://www.jsonrpc.org/specification)",
```
edd, line 5: **question (non-blocking)**: @candida-d I think this URL works, what do you think?

candida-d: Generally yes - though with the 'mainnet' path it won't be active until this gets released to mainnet, so it might be a while. The alternative is make it testnet and then update this link once the validators release all this to mainnet, or have 2 versions in the useful links, one for whatever version is on testnet and one for whatever version is on mainnet.

ValentinTrinque: so what should I do ?

candida-d: Which of the options I suggested do you think is most feasible? Having 2 links now, one which won't work for a while, or changing it later?

wallet/service/openapi.json
```diff
@@ -0,0 +1,95 @@
...
+  },
+  "externalDocs": {
+    "description": "Find out more about Vega REST APIs.",
+    "url": "https://docs.vega.xyz/docs/mainnet/api/rest/overview"
+  },
+  "paths": {
+    "/v2/requests": {
+      "post": {
+        "summary": "Submit a JSON-RPC request to the wallet API.",
+        "description": "This endpoint is the main entrypoint to submit the JSON-RPC requests. The body of the request should be formatted as a JSON-RPC v2 request. The response is formatted as the JSON-RPC v2 response.\n\nBe sure to specify the `id` if you are interested in the response. See [the JSON-RPC notification documentation](https://www.jsonrpc.org/specification#notification) for more details.",
```
candida-d, line 16:
```suggestion
        "description": "This endpoint is the main entrypoint to submit the JSON-RPC requests. The body of the request should be formatted as a JSON-RPC v2 request. The response is formatted as the JSON-RPC v2 response.\n\nBe sure to specify the `ID` if you are interested in the response. See [the JSON-RPC notification documentation](https://www.jsonrpc.org/specification#notification) for more details.",
```

wallet/service/openapi.json
```diff
@@ -0,0 +1,95 @@
...
+        "requestBody": {
+          "description": "The JSON-RPC v2 request bidy",
```
candida-d, line 18:
```suggestion
          "description": "The JSON-RPC v2 request body",
```
Is the body the same as the description, or is there another way that this can be described to supplement the fact that it's called the "request body"? Repeating the word body doesn't help describe it much

wallet/service/openapi.json
```diff
@@ -0,0 +1,95 @@
...
+                  "$ref": "#/components/schemas/JsonRpcResponse"
+                }
+              }
+            }
+          },
+          "204": {
+            "description": "The request is considered as a notification because it doesn't have an ID. As a result, no result shall be expected, even in case of failure."
+          },
+          "400": {
+            "description": "The submitted request has not a valid format, or doesn't comply with the API constraints, or failed."
```
candida-d, line 43:
```suggestion
            "description": "The submitted request does not have a valid format, or doesn't comply with the API constraints, or failed."
```

wallet/service/openapi.json
candida-d: I'm not sure which is best btw, I think neither is an ideal solution.

ValentinTrinque: I prefer to remove it if we can't make it work now. Broken link is no better than no link.

wallet/service/openapi.json
```diff
@@ -0,0 +1,109 @@
...
+            "type": "string",
+            "enum": [
+              "2.0"
+            ]
+          },
+          "method": {
+            "type": "string"
+          },
+          "params": {
+            "type": "object"
```
edd, line 68: **nitpick (non-blocking)**: I think we can live with this, but rendered in a Swagger Editor the params field is particularly unhelpful:

Screenshot 2022-08-10 at 11 07 36

I understand why this is, and I understand that linking to the Json RPC request object docs sorta-kinda works. Let's roll with this, but I suspect we can improve it - I just have no idea how.

edd (dismissed):

candida-d (dismissed):

candida-d (approved):

PR #5840, opened 2022-08-10
Add endpoint to wallet HTTP API v2 to list available RPC methods

### LINKED ISSUE #5823: Support listing registered JSON-RPC methods on wallet HTTP API
Knowing which JSON-RPC methods are exposed to the HTTP API is not possible without exploration. This is a bad user experience. We should have a `GET /v2/methods` endpoint that list them.

ValentinTrinque:
Close #5823

### DIFF
--- a/wallet/service/openapi.json
+++ b/wallet/service/openapi.json
@@ -47,6 +47,30 @@
           }
         }
       }
+    },
+    "/v2/methods": {
+      "get": {
+        "summary": "List the available methods on the JSON-RPC API.",
+        "description": "The JSON-RPC API behind this HTTP API can have an arbitrary set of endpoints enabled.\n\nThis method allows an application to check for the existence of the JSON-RPC before calling the requests method.",
+        "responses": {
+          "200": {
+            "description": "Successful listing",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "type": "string"
+                  }
+                }
+              }
+            }
+          },
+          "500": {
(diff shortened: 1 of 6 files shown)


wallet/service/openapi.json
```diff
@@ -47,6 +47,30 @@
           }
         }
       }
+    },
+    "/v2/methods": {
+      "get": {
+        "summary": "List the available methods on the JSON-RPC API.",
+        "description": "The JSON-RPC API use behind this HTTP API can have an arbitrary setup in terms of available methods.\n\nPrior using the requests API, an application might want to check if the JSON-RPC it requires are presents. It also useful to avoid to explore the API in a die-and-repeat style.",
```
edd, line 54:
```suggestion
        "description": "The JSON-RPC API behind this HTTP API can have an arbitrary set of endpoints enabled.\n\nThis method allows an application an to check for the existence of the JSON-RPC before calling the requests method.",
```

edd (dismissed):
One suggestion in the docs, happy either way.

wallet/service/openapi.json
candida-d, line 54:
```suggestion
        "description": "The JSON-RPC API use behind this HTTP API can have an arbitrary setup in terms of available methods.\n\nPrior using the requests API, an application might want to check if the JSON-RPC it requires are present. It is also useful to avoid exploring the API in a die-and-repeat style.",
```

edd (dismissed):

jeremyletang (approved):

PR #5866, opened 2022-08-11
Add chain ID to transaction input data

### LINKED ISSUE #5802: Encode chain ID in the transaction signature
As of now a transaction for a network A cannot be differentiated from a transaction to a network B. While it's probably not the best practice to reuse a wallet over different network, a transaction being compatible on any network as soon as the pubkey exists, it could be reused from a network to the mainnet for example.

We want to add the chain ID as part of the message encoded to be signed for each transaction.
In order to support backward compatibility with older wallet, the network should:
- append the chain ID to every input data, and verify the signature
  - if its a valid signature, OK
  - if it's a invalid signature, remove the prefix ans validate again the signature
    - If it's a valid signature, trasnaction is OK
    - if not, transaction is rejected
- The wallet should prefix all input data with the chain ID, and sign this message.

ValentinTrinque:
Close #5802


commands/transaction.go
```diff
@@ -30,6 +33,34 @@ func NewInputData(height uint64) *commandspb.InputData {
...
+	if err != nil {
+		return nil, err
+	}
+
+	dataWithChainID := append([]byte(fmt.Sprintf("%s%c", chainID, ChainIDDelimiter)), data...)
+
+	return dataWithChainID, nil
+}
+
+func UnmarshalInputData(rawInputData []byte) (*commandspb.InputData, string, error) {
```
jeremyletang, line 47: Why not use bytes.HasPrefix, then slice the rest of it like so: return rawInputData[len(prefix):len(rawInputdata)]

cmd/vegawallet/commands/command_send.go
```diff
@@ -265,6 +265,14 @@ func SendCommand(w io.Writer, rf *RootFlags, req *SendCommandRequest) error {
 	ctx, cancelFn := context.WithTimeout(context.Background(), ForwarderRequestTimeout)
 	defer cancelFn()
 
+	log.Info("retrieving block height")
+	chainID, err := forwarder.GetNetworkChainID(ctx)
```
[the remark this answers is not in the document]
ValentinTrinque: alright

commands/transaction.go
ValentinTrinque: ok

commands/transaction.go
```diff
@@ -30,6 +34,34 @@ func NewInputData(height uint64) *commandspb.InputData {
...
+		return nil, err
+	}
+
+	dataWithChainID := append([]byte(fmt.Sprintf("%s%c", chainID, ChainIDDelimiter)), data...)
+
+	return dataWithChainID, nil
+}
+
+func UnmarshalInputData(rawInputData []byte, expectedChainID string) (*commandspb.InputData, error) {
+	prefix := fmt.Sprintf("%s%c", expectedChainID, ChainIDDelimiter)
```
wwestgarth, line 49: Do we not want to use the transaction version somewhere to tell us whether we should even be looking to look for a pre-fixed chainID or not?

jeremyletang (commented):
the version as now is hardcoded, wondering if we should add an proto enum for it actually? Just so it makes it explicits for people using the protos to build transactio. Also as we deprecated fully usage of old transaction we can reserve the fields.

As of now you would have:
```
enum TxVersion {
  reserved 1;
  V2 = 2;
  V3 = 3;
}
```

Thoughts?

commands/transaction.go
jeremyletang: yes it should be used in the core as well I suppose? For now we should keep supporting both the version 2 and 3 just until everyone upgrade their wallet?

ValentinTrinque:
Added the enum.

protos/sources/vega/commands/v1/transaction.proto
```diff
@@ -65,6 +65,18 @@ message InputData {
...
+enum TxVersion {
+  reserved 1;
+
+  TX_VERSION_UNSPECIFIED = 0;
+  // This version requires the Proof-of-work added to the transaction.
+  V2 = 2;
+  // This version requires the chain ID to be append in front of the input data
+  // byte, with a `\0` delimiter.
+  V3 = 3;
+}
```
jeremyletang, line 78: Linter should break here? V2 and V3 should be TX_VERSION_V2 etc

core/processor/abci.go
```diff
@@ -564,6 +564,7 @@ func (app *App) OnInitChain(req tmtypes.RequestInitChain) tmtypes.ResponseInitCh
 	app.log.Debug("ABCI service InitChain start")
 	hash := hex.EncodeToString(vgcrypto.Hash(req.AppStateBytes))
 	// let's assume genesis block is block 0
+	app.abci.SetChainID(req.ChainId)
```
wwestgarth, line 567: When we restore from a snapshot we don't call `OnInitChain` so this will also need to be set somewhere else as well.  

For `app.chainCtx` we set it in `OnBeginBlock` (next to a nice`TODO` written by me about how it isn't ideal) so you can call `app.abci.SetChainID()` there too.

jeremyletang: ah correct. we may want to have this on the snapshots then?

ze97286:
should we also as part of this PR get rid of the check in pow/engine.go that allows wallet versions < 2 to not submit pow? I think it's time.

jeremyletang:
> should we also as part of this PR get rid of the check in pow/engine.go that allows wallet versions < 2 to not submit pow? I think it's time.

yes

ValentinTrinque:
> should we also as part of this PR get rid of the check in pow/engine.go that allows wallet versions < 2 to not submit pow? I think it's time

Addressed

commands/transaction.go
```diff
@@ -38,21 +71,20 @@ func NewSignature(sig []byte, algo string, version uint32) *commandspb.Signature
...
-	if len(tx.InputData) == 0 {
-		errs.AddForProperty("tx.input_data", ErrIsRequired)
+	if tx.Version == commandspb.TxVersion_TX_VERSION_UNSPECIFIED {
+		return nil, errs.FinalAddForProperty("tx.version", ErrIsRequired)
 	}
-	if tx.Signature == nil {
-		errs.AddForProperty("tx.signature", ErrIsRequired)
-	} else {
-		errs.Merge(checkSignature(tx.Signature))
+	if tx.Version != 2 && tx.Version != 3 {
```
jeremyletang, line 84: Why not use the enum here instead of the values?

commands/transaction.go
```diff
@@ -61,52 +93,93 @@ func CheckTransaction(tx *commandspb.Transaction) (*commandspb.InputData, error)
...
+		errs.AddForProperty("tx.signature.algo", ErrUnsupportedAlgorithm)
+		errs.AddForProperty("tx.signature.version", ErrUnsupportedAlgorithm)
+	}
+
+	// We need the above check to pass, so we verify it's all good.
+	if !errs.Empty() {
+		return errs
+	}
+
+	decodedPubKey := []byte(pubKey)
```
jeremyletang, line 144: Why? this should ve just
```suggestion
	var decodedPubKey []byte{}
```

commands/transaction.go
```diff
@@ -61,52 +93,93 @@ func CheckTransaction(tx *commandspb.Transaction) (*commandspb.InputData, error)
...
+	if IsVegaPubkey(pubKey) {
```
jeremyletang, line 145: What happen if it's not a valid pubkey?

core/processor/tx.go
```diff
@@ -21,54 +22,53 @@ import (
...
+	var rawInputData []byte
+	idx := bytes.IndexByte(tx.InputData, commands.ChainIDDelimiter)
+	if idx == -1 {
+		rawInputData = tx.InputData
+	} else {
+		rawInputData = tx.InputData[idx+1:]
 	}
 
+	inputData := &commandspb.InputData{}
+	err := proto.Unmarshal(rawInputData, inputData)
```
jeremyletang, line 53: behaviour have changed here, the unmarshalling of input data would have returned an error if failing.

core/processor/tx.go
```diff
@@ -168,20 +136,14 @@ func (t Tx) Command() txn.Command {
 }
 
 func (t Tx) GetPoWNonce() uint64 {
-	if t.version > 1 && t.pow != nil {
-		return t.pow.Nonce
-	}
-	return 0
+	return t.pow.Nonce
 }
```
jeremyletang, line 140: Dont you still want the nil check here in case of a validator command?

core/processor/tx.go
```diff
@@ -168,20 +136,14 @@ func (t Tx) Command() txn.Command {
...
 
 func (t Tx) GetPoWTID() string {
-	if t.version > 1 && t.pow != nil {
-		return t.pow.Tid
-	}
-	return ""
+	return t.pow.Tid
```
jeremyletang, line 143: Same here?

core/staking/checkpoint_test.go
```diff
@@ -79,7 +79,7 @@ func TestCheckpointLoadNoDuplicates(t *testing.T) {
 
 	cp := &checkpoint.Checkpoint{}
 	if err := proto.Unmarshal(cpFile, cp); err != nil {
-		println(err)
+		t.Fatal(err)
```
jeremyletang, line 82: println!=Fatal ?

core/stats/stats.go
```diff
@@ -80,10 +80,10 @@ func (s *Stats) GetUptime() time.Time {
...
 }
 
-func (s Stats) Height() uint64 {
+func (s *Stats) Height() uint64 {
 	return s.Blockchain.Height()
 }
 
-func (s Stats) BlockHash() string {
+func (s *Stats) BlockHash() string {
 	return s.Blockchain.Hash()
```
jeremyletang, line 88: Why change them to pointers? They are getters it's unnecessary?

protos/sources/vega/commands/v1/transaction.proto
```diff
@@ -84,7 +96,7 @@ message Transaction {
 
   // A version of the transaction, to be used in the future in case changes are implemented
   // to the Transaction format.
-  uint32 version = 2000;
+  TxVersion version = 2000;
```
jeremyletang, line 99: I'm wondering how backward compatible is that? Is a transaction create with an old wallet still able to be decoded with this version of the core?

commands/transaction.go
ValentinTrinque: I just use the pubkey as it is provided. We don't have support for anything else.

core/stats/stats.go
ValentinTrinque: Since we are returning copied data, it's safe to use a point receiver. It avoids unnecessary copies of the struct. So I shift it.

core/staking/checkpoint_test.go
ValentinTrinque: Sure, but that's a test. `println` was most likely a left over. In this context it's very likely that failing the test is the expected behavior.

core/processor/tx.go
ValentinTrinque: The proof of work should be mandatory now. A transaction would be rejected without this. So there is no need to check that. However, I realize there is no verification on the commands for that. I will add it.

core/processor/tx.go
ValentinTrinque: Same as above

commands/transaction.go
ValentinTrinque: so `decodedPubKey` can have a default value. otherwise it would be empty because of the condition below.

core/processor/tx.go
ValentinTrinque: I think we misunderstood eachother last time

core/processor/tx.go
jeremyletang: Poof is not mandatory on validators commands.

commands/transaction.go
jeremyletang: But you check if it’s a valid pubkey. If it’s not one it should be handled isn’t it?

ValentinTrinque: It's checked earlier here:
https://github.com/vegaprotocol/vega/pull/5866/files#diff-ee1725107e82a8cfa620ce0297f0d97175314976b9b924c23436f44eda7f613cL60

core/processor/tx.go
ValentinTrinque: oh that change a lot of things

ValentinTrinque: It is troublesome to add the verification of the PoW right in the command verification. So I won't add it in this PR.

core/processor/tx.go
ValentinTrinque: I used the `err` to value the the `err` property of the Tx struct, because it was not valued anywhere, so it though that was a mistake to be fixed. Since it is not. I will remove that part.

jeremyletang: yea I guess, here just the proto unmarshal could have returned an error before.

core/processor/tx.go
jeremyletang: ok.

protos/sources/vega/commands/v1/transaction.proto
ValentinTrinque: It seems to work

jeremyletang (approved):

PR #5871, opened 2022-08-11
fix: clarify the error for insufficient token to submit proposal or vote

### LINKED ISSUE #5868: spam insufficient tokens log message is confusing
# Problem encountered
When the party has insufficient tokens associated for submitting a proposal they get the following error:
"party has insufficient tokens to submit proposal request in this epoch" - this is a bit confusing because it doesn't say that it's government tokens that need to be associated and the last part about the epoch is incorrect. 

# Expected behaviour
This should give the following error: "party has insufficient associated governance tokens in their staking account to submit proposal request"

ze97286:
Close #5868


jeremyletang (dismissed):

edd (dismissed):
👍

jeremyletang (approved):

PR #5908, opened 2022-08-16
fix: equity-like-shares 

guoguojin:
should be a decimal value between 0 and 1 and not an int

### DIFF
--- a/datanode/entities/marketdata.go
+++ b/datanode/entities/marketdata.go
@@ -125,14 +125,14 @@ func (bound PriceMonitoringBound) Equals(other PriceMonitoringBound) bool {
 }
 
 type LiquidityProviderFeeShare struct {
-	Party                 string `json:"party"`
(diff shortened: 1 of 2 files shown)


EVODelavega:
As mentioned on slack: around line 430, there's this:

```go
	for _, feeShare := range feeShares {
		protoFeeShare := types.LiquidityProviderFeeShare{
			Party:                 feeShare.Party,
			EquityLikeShare:       fmt.Sprintf("%d", feeShare.EquityLikeShare),
			AverageEntryValuation: fmt.Sprintf("%d", feeShare.AverageEntryValuation),
		}

		results = append(results, &protoFeeShare)
	}
```

Now the field is `num.Decimal`, this `Sprintf` should be replaced with a `feeShare.EquityLikeShare.String()` call


Now that I think of it, the `AverageEntryValuation` field is a `num.Decimal` in the core, too, perhaps this should be changed accordingly, too

jeremyletang (approved):

PR #5909, opened 2022-08-16
feat: AC for 0029-FEES-005

cdummett:
Adding additional feature test "fees/immediate-amend-fees.feat" to cover AC 0029-FEES-005 in spec 0029-FEES-fees.


jeremyletang:
Shouldn't this test be under `core/integration/features/verified`? cc @davidsiska-vega @Jiajia-Cui

Jiajia-Cui:
There are 2 scenarios in the test, are they duplicated?

cdummett:
Thanks for reviewing Jiajia, I setup the first scenario to check how fees are calculated for a placed order which resolves immediately before testing amended orders which resolve immediately. This was mostly for my own sanity, I'll remove it now as it doesn't relate to 0029-FEES-005. 

Do you have any idea why the PR is failing other checks by the way?

Jiajia-Cui:
> 

I can run it from my end, it seems fine :)

Jiajia-Cui (approved):
LGTM

PR #5911, opened 2022-08-16
fix: incorrect data types on some market data fields

guoguojin:
It has been pointed out that there is a potential for numbers that are parsed from strings can overflow as they are supposed to be num.Uint rather than uint64. This PR moves these values to num.Uint and updates the database to store the values as the aliased data type HUGEINT.


datanode/entities/marketdata.go
```diff
@@ -260,10 +270,9 @@ func priceMonitoringBoundsFromProto(bounds *types.PriceMonitoringBounds) *PriceM
 		return nil
 	}
 
-	minValidPrice, _ := strconv.ParseUint(bounds.MinValidPrice, 10, 64)
-	maxValidPrice, _ := strconv.ParseUint(bounds.MaxValidPrice, 10, 64)
-
-	referencePrice, _ := strconv.ParseUint(bounds.ReferencePrice, 10, 64)
+	minValidPrice, _ := num.UintFromString(bounds.MinValidPrice, 10)
+	maxValidPrice, _ := num.UintFromString(bounds.MaxValidPrice, 10)
+	referencePrice, _ := num.UintFromString(bounds.ReferencePrice, 10)
```
jeremyletang, line 275: The could return an error if not a valid decimal.

datanode/entities/marketdata.go
```diff
@@ -411,10 +420,10 @@ func priceMonitoringBoundsToProto(bounds []*PriceMonitoringBound) []*types.Price
 
 	for _, bound := range bounds {
 		protoBound := types.PriceMonitoringBounds{
-			MinValidPrice:  fmt.Sprintf("%d", bound.MinValidPrice),
-			MaxValidPrice:  fmt.Sprintf("%d", bound.MaxValidPrice),
+			MinValidPrice:  bound.MinValidPrice.String(),
+			MaxValidPrice:  bound.MaxValidPrice.String(),
 			Trigger:        priceMonitoringTriggerToProto(bound.Trigger),
-			ReferencePrice: fmt.Sprintf("%d", bound.ReferencePrice),
+			ReferencePrice: bound.ReferencePrice.String(),
```
jeremyletang, line 426: If these returned an error earlier you could be call .String() on a nil pointer.

jeremyletang (approved):

PR #5914, opened 2022-08-16
feat: add date ranges to some queries

### LINKED ISSUE #5681: Error when querying for historicBalances in graphQL
Error returned when querying for historicBalances in graphQL

Query:

```
query {
  historicBalances 
  (filter: {partyIds : ["f57e7a5ad0bd695f2d4567e055340c5ce5628918cc65150ffefa6057c6f2b46f"]} 
  	groupBy: [AssetId]
  	) {
    timestamp
    balance
    accountId
    partyId
    assetId
    accountType
    marketId
  }
}

```
Response:

```
{
  "errors": [
    {
      "message": "rpc error: code = Unknown desc = querying balances: scanning balances: scany: scan row into struct fields: can't scan into dest[1]: unable to assign to *string",
      "path": [
        "historicBalances"
      ]
    }
  ],
  "data": null
}
```

### LINKED ISSUE #5684: Date range filters for data sets e.g. Account Balances, Orders, Trades, Deposits, Withdrawals
We need a way to add filters e.g. `from` and `to` to specify date/time filter ranges. This is particularly useful where we want to run incentives between two dates, we might need a balance for a party at start date and then calculate their balance or change in  balance to end date e.g. profit/loss.

This would ideally be a filter that could be applied to any data store result set, and can then be sorted with asc or desc order and paginated. 

Without this feature, the very wasteful resource wise approach is to page through all results and look for start / end manually on the client, which is increasingly more 🤮 the more active / large a result set gets over time.

guoguojin:
Closes #5684 
Closes #5681


datanode/sqlstore/cursor.go
```diff
@@ -39,12 +39,14 @@ const (
 )
 
 type ColumnOrdering struct {
-	Name    string
-	Sorting Sorting
+	Name         string
+	Sorting      Sorting
+	CursorColumn bool
+	Prefix       string
```
pscott31, line 45: This is starting to get a bit muddled again, I'm not sure `CursorColumn` and `Prefix` are belong in something called `ColumnOrdering`.. And I'm not exactly sure what they're for :)

datanode/sqlstore/cursor.go
```diff
@@ -218,6 +228,13 @@ func PaginateQuery[T any, PT parserPtr[T]](
 		query = fmt.Sprintf("%s %s (%s)", query, whereOrAnd, predicate)
 	}
 
+	groupBy := ""
+	if len(grouping) > 0 {
+		groupBy = fmt.Sprintf(" GROUP BY %s", strings.Join(grouping, ","))
+	}
+
+	query = fmt.Sprintf("%s%s", query, groupBy)
+
```
pscott31, line 237: I'm not sure we want to mix up grouping and paging, they're sort of separate operations.. Presumably we want to group a thing and then we've effectively got a new shape of table and we want to have a cursor into that possibly using a different cursor as the old one might not make sense if columns have disappeared?

datanode/sqlstore/cursor.go
guoguojin: The do belong on the ColumnOrdering, I will add documentation to the struct, but basically:

CursorColumn is a flag that indicates the Column should be included as part of the Cursor ordering constructs, without it, you are automatically assuming that each column listed should be part of the cursor sorting, and that's not always the case which was why I needed to add it.

Prefix was added because in some complex queries like the one for aggregated balances where you have a more complex query involving more than one table and you need to specify which table a particular column comes from, then if prefix is defined, it will be pre-pended to the column specified. 

I did try to do it by using the full <prefix>.<column> as the column name without adding the prefix to the ColumnOrdering struct, but that causes other problems with matching column names to struct fields when you are extracting the results.

datanode/sqlstore/cursor.go
guoguojin: Unfortunately another by-product of the complex query for the aggregated balances where you need to have a group by clause in your query before the order by statement. Happy to discuss and re-implement this if there is a better way, but to me it makes sense as other queries in future may require a group by clause before the order by is constructed for the pagination.

datanode/gateway/graphql/schema.graphql
```diff
@@ -1684,6 +1686,8 @@ type Market {
 
   "Orders on a market"
   ordersConnection(
+    "DateRange to retrieve orders from/to"
```
candida-d, line 1689: Should this be in a particular format when a user puts it in?

datanode/gateway/graphql/schema.graphql
```diff
@@ -1712,7 +1716,11 @@ type Market {
     last: Int
   ): [Trade!] @deprecated(reason: "Use tradesConnection for cursor based pagination instead")
 
-  tradesConnection(pagination: Pagination): TradeConnection!
+  tradesConnection(
+    "DateRange to retrieve trades from/to"
```
candida-d, line 1720:
```suggestion
    "Date range to retrieve trades from/to"
```

datanode/gateway/graphql/schema.graphql
candida-d, line 1689:
```suggestion
    "Date range to retrieve orders from/to"
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1799,3 +1815,14 @@ message GetVegaTimeResponse {
     // Nanoseconds since the epoch, for example `1580473859111222333` corresponds to `2020-01-31T12:30:59.111222333Z`
     int64 timestamp = 1;
 }
+
+// A date range for queries that can return historical data
+// If only the start_timestamp is provided, results from the given start up to the most current timestamp should be returned.
+// If only the end_timestamp is provided, results upto the given end should be returned (exclusive).
```
candida-d, line 1821:
```suggestion
// If only the end_timestamp is provided, results up to the given end should be returned (exclusive).
```

guoguojin: Addressed with  https://github.com/vegaprotocol/vega/pull/5914/commits/cb8a00fb1ddcda3a7f2ed84e881c6d18640a4331

datanode/gateway/graphql/schema.graphql
```diff
@@ -642,6 +642,7 @@ type Query {
   historicBalances(
     filter: AccountFilter,
     groupBy: [AccountField],
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 645:
```suggestion
    "Date range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -1686,7 +1687,7 @@ type Market {
 
   "Orders on a market"
   ordersConnection(
-    "DateRange to retrieve orders from/to"
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 1690:
```suggestion
    "Date range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -1717,7 +1718,7 @@ type Market {
   ): [Trade!] @deprecated(reason: "Use tradesConnection for cursor based pagination instead")
 
   tradesConnection(
-    "DateRange to retrieve trades from/to"
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 1721:
```suggestion
    "Date range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

this one is called trades, but has switched to being described as orders - should it go back to being trades?

datanode/gateway/graphql/schema.graphql
```diff
@@ -1924,7 +1925,7 @@ type Party {
 
   "Orders relating to a party"
   ordersConnection(
-    "Date range to retrieve orders from/to"
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 1928:
```suggestion
    "Date range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -1943,8 +1944,11 @@ type Party {
   ): [Trade!] @deprecated(reason: "Use tradesConnection for cursor based pagination instead")
 
   tradesConnection(
+    "ID of the market you want to get trades for"
     marketId: ID,
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 1949:
```suggestion
    "Date range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

Should this be orders or trades?

datanode/gateway/graphql/schema.graphql
```diff
@@ -2014,6 +2018,7 @@ type Party {
 
   "The list of all withdrawals initiated by the party"
   withdrawalsConnection(
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 2021:
```suggestion
    "Date range to retrieve withdrawals from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -2023,6 +2028,7 @@ type Party {
 
   "The list of all deposits for a party by the party"
   depositsConnection(
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 2031:
```suggestion
    "Date range to retrieve deposits from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -2250,6 +2256,7 @@ type Order {
 
   "Trades relating to this order"
   tradesConnection(
+    "Date Range to retrieve orders from/to. Start and end time should be expressed as an integer value of nano-seconds passed the Unix epoch"
```
candida-d, line 2259:
```suggestion
    "Date range to retrieve trades from/to. Start and end time should be expressed as an integer value of nano-seconds past the Unix epoch"
```

datanode/gateway/graphql/schema.graphql
guoguojin: Yeah, bad copy and paste on my part, will fix it

pscott31 (dismissed):

ValentinTrinque (approved):

dexturr (approved):

PR #5929, opened 2022-08-17
feat: allow issuing of multisig add/remove signatures on demand

### LINKED ISSUE #5841: New command/TX type for new validators to be requested to build signatures
In order to request that new validators build and submit their signature for the signature bundle we need to implement a new tx type / command to require the validators to do this.

- [ ] discuss the best approach and refine this ticket
- [ ] implement the functionallity
- [ ] update the docs (node operator instruictions)

This issue requires further discussions and refinement

Also to consider (and resolve) is the impact on:

- [ ] system tests
- [ ] vegacapsule

wwestgarth:
closes #5841 

This PR adds a new command that allows a validator node to request signature bundles to remove/add a node to the multisig control contract with a given eth-address as a submitter (but only if a relevant event to do so has occurred, such as a demotion/promotion):

```
message IssueSignatures {
  // The ethereum address which will submit the signatures to the smart-contract
  string submitter = 1;
  // The kind of signatures to generate, namely for whether a signer is being added or removed
  NodeSignatureKind kind = 2;
  // The ID of the node that will be signed in or out of the smartcontract
  string validator_node_id = 3; // the node for which to emit the signatures.
} 
```

It also adds a new field to both the eth-key-rotation and announce-node command to be able designate a eth-address as a "multisig submitter" which, if supplied, means that signatures for that given submitter can be automatically issued. For example, if a submitter address is given with the key-rotation command sig-bundles for adding the new key and removing the old key will be available in the data-node without needing to submit a `IssueSignatures` command.

There is also all the fiddly snapshot related bits-and-bobs that needed to be added, as well as some subtle changes to the `GetERC20MultisigControlBundle` data-node endpoints given that more fields may need to be queried (especially the `add` one which now need filtering on submitter)

It also comes with system-test changes to account for the new way of getting the signature bundles: https://github.com/vegaprotocol/system-tests/compare/adapt-to-sig-gen-changes


core/validators/signatures.go
```diff
@@ -99,184 +127,252 @@ func (s *ERC20Signatures) EmitPromotionsSignatures(
...
+			EthAddress: signer.EthAddress,
+			Nonce:      s.lastNonce.Clone(),
+			EpochSeq:   epochSeq,
+			Added:      added,
+		}
+		s.lastNonce.AddUint64(s.lastNonce, 1)
+
+		// we're ok to override whatever is in here since we can't need to both add and remove an eth-address at the same time
+		// so we'll replace it either with the correct action, or with the same action but with a later epoch which is fine
+		// because you can still get signatures to do what you need
```
jeremyletang, line 208: Is that true? E.g in the case of an ethereum key rotation?

wwestgarth: Its true because the map is from `ethereum-address -> the sig data` (not `nodeID -> sig data`) so when you rotate your key the address would change so you'd have one element for the old-key with remove, and one for the new-key with add.

jeremyletang (approved):

core/validators/signatures.go
jeremyletang: yup I realised that down the line! thanks.

PR #5930, opened 2022-08-17
Change liquidity checks to only happen once per block

### LINKED ISSUE #5926: Liquidity auction at end-of-block - implement spec changespec
In the following specs PR there are some changes to liquidity auction:
- https://github.com/vegaprotocol/specs/pull/1180

This issue is to cover the implementation to reflect the spec changes

EVODelavega:
Closes #5926 

Only check for liquidity auctions once per block

### DIFF
--- a/core/execution/auction.go
+++ b/core/execution/auction.go
@@ -22,7 +22,6 @@ import (
 )
 
 func (m *Market) checkAuction(ctx context.Context, now time.Time) {
-	// of course, if we're not in auction, there's nothing to do here
 	if !m.as.InAuction() {
 		// new block, check liquidity, start auction if needed
 		m.checkLiquidity(ctx, nil, true)
--- a/core/execution/auction.go
+++ b/core/execution/auction.go
@@ -24,6 +24,11 @@ import (
 func (m *Market) checkAuction(ctx context.Context, now time.Time) {
 	// of course, if we're not in auction, there's nothing to do here
 	if !m.as.InAuction() {
+		// new block, check liquidity, start auction if needed
+		m.checkLiquidity(ctx, nil, true)
+		if m.as.AuctionStart() {
+			m.enterAuction(ctx)
+		}
 		return
 	}
 
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -2440,6 +2440,9 @@ func (m *Market) AmendOrder(ctx context.Context, orderAmendment *types.OrderAmen
 		allUpdatedOrders,
 		updatedOrders...,
 	)
+	if !m.as.InAuction() {
+		m.checkForReferenceMoves(ctx, allUpdatedOrders, false)
+	}
 	return conf, nil
 }
 
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -1314,8 +1314,6 @@ func (m *Market) SubmitOrder(
 	if !m.as.InAuction() {
 		m.checkForReferenceMoves(
 			ctx, allUpdatedOrders, false)
-		m.checkLiquidity(ctx, nil, true)
-		m.commandLiquidityAuction(ctx)
 	}
 
 	return conf, nil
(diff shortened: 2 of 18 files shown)


core/execution/market.go
```diff
@@ -1492,9 +1490,9 @@ func (m *Market) checkPriceAndGetTrades(ctx context.Context, order *types.Order)
 		return nil, types.OrderErrorNonPersistentOrderOutOfPriceBounds
 	}
 
-	if m.checkLiquidity(ctx, trades, persistent) {
+	/*if m.checkLiquidity(ctx, trades, persistent) {
 		return nil, types.OrderErrorInvalidPersistance
-	}
+	}*/
```
jeremyletang, line 1495: ?

core/execution/reference_price_moves.go
```diff
@@ -72,7 +75,15 @@ func (m *Market) checkForReferenceMoves(
...
 
 	// now we can start all special order repricing...
-	orderUpdates = m.repriceAllSpecialOrders(ctx, changes, orderUpdates)
+	if err == nil {
+		orderUpdates = m.repriceAllSpecialOrders(ctx, changes, orderUpdates)
+	} else {
+		cancels := m.liquidity.UndeployLPs(ctx, orderUpdates)
+		orderUpdates, err = m.updateAndCreateLPOrders(ctx, nil, cancels, nil)
+		if err != nil {
+			panic(err) // this should not be possible
```
jeremyletang, line 84: Can we use m.log.Panic("error message", logging.Error(err))
```suggestion
			m.log.Panic("could not get best static prices and unable to cancel lp orders", logging.Error(err), logging.MarketID(m.mkt.GetID())) // this should not be possible
```

core/execution/reference_price_moves.go
```diff
@@ -50,6 +50,9 @@ func (m *Market) checkForReferenceMoves(
 		return
 	}
 
+	// will be set to non-nil if a peg is missing
+	_, _, err := m.getBestStaticPricesDecimal()
```
jeremyletang, line 54: Why not get the errors on the following lines, en stop early?

core/execution/market.go
EVODelavega: can't remember exactly why this was... had an error at some point, added this back, and commented it out later after fixing it. Will remove

core/execution/reference_price_moves.go
EVODelavega: will update

core/execution/reference_price_moves.go
EVODelavega: because we would need to check 4 errors, and we still need to get the 4 relevant values to update the market fields and set `stateChanged = true`. By checking this error once, I didn't need to change any existing code, just add the error check in there. I thought about doing something with the errors returned by the existing calls, but it ended up being more verbose anyway

jeremyletang: ah right... stateChanged is unused though anyway now

core/execution/auction.go
```diff
@@ -24,6 +24,11 @@ import (
 func (m *Market) checkAuction(ctx context.Context, now time.Time) {
 	// of course, if we're not in auction, there's nothing to do here
```
peterbarrow, line 25:
```suggestion
```

core/liquidity/engine.go
```diff
@@ -530,6 +530,49 @@ func (e *Engine) CreateInitialOrders(
 	return creates, err
 }
 
+// UndeployLPs is called when a reference price is no longer available. LP orders should all be parked/set to pending
+// and should eb redeployed once possible. Pass in updated orders and update internal records first...
```
peterbarrow, line 534:
```suggestion
// and should be redeployed once possible. Pass in updated orders and update internal records first...
```

peterbarrow (dismissed):

peterbarrow (approved):

TomMcL (approved):
LGTM

PR #5931, opened 2022-08-17
5358 new equity shares

### LINKED ISSUE #5358: Update the implementation on Equity-like-share (spec 0042) given the change of the spec
change of the spec can be found here:
- https://github.com/vegaprotocol/specs-internal/pull/1088

link with the original issue [https://github.com/vegaprotocol/vega/issues/4931](url)

EVODelavega:
Closes #5358 

Still needs the integration tests to be updated, hence opening in draft.


core/execution/equity_shares.go
```diff
@@ -225,35 +275,33 @@ func (es *EquityShares) SharesFromParty(party string) num.Decimal {
...
+			es.totalPStake = es.totalPStake.Sub(v.stake)
+			es.totalVStake = es.totalVStake.Sub(v.vStake)
+			all = false
+		}
+	}
+	for id, v := range es.lps {
 		if _, ok := except[id]; ok {
 			continue
 		}
 		eq, err := es.equity(id)
```
wwestgarth, line 292: The temporary change to `es.totalPStake` and `es.totalVStake` in prep to the call to `es.equity()` feels a bit dangerous. I wonder if instead it should just be a nice and simple "normal" function and not a method on `es?` So like:
```
func equity(v *lp, totalVStake num.Decimal) (num.Decimal, error) {
	if totalVStake.IsZero() {
		return num.DecimalZero(), nil
	}
	return v.vStake.Div(totalVStake), nil
}
```
then we also don't have to worry about the error case of the party not existing, because in both calls to it we're already iterating over `e.lps`.

EVODelavega: I get what you're saying, and I do agree that this looks a bit iffy. I tried to keep the actual equity function the same throughout. The only difference here is that instead of iterating over all LP's and recalculating the totals, we're just updating them on the fly. This is more optimal when updating stake (LP submission/amendment), at the end of each time window, when getting the equity shares for all parties... the only place where it's a bit messy is here.

After the time spent, and sanity lost due to this change, I'm inclined to keep it as-is.

jeremyletang (dismissed):

Jiajia-Cui (dismissed):
LGTM

jeremyletang (approved):

Jiajia-Cui (approved):
LGTM

PR #5935, opened 2022-08-17
Ensure wallet without permissions can be deserialized

### LINKED ISSUE #5934: Crash of wallet service on permissions update
```
code.vegaprotocol.io/vega/wallet/wallet.(*HDWallet).UpdatePermissions(0xc000f5e080, {0xc000e47f18, 0x3}, {{{0x2f4021a, 0x4}, {0x0, 0x0, 0x0}}})
```

ValentinTrinque:
Close #5934


jeremyletang (dismissed):

candida-d (approved):

wallet/wallet/hd_wallet.go
```diff
@@ -326,6 +326,11 @@ func (w *HDWallet) MarshalJSON() ([]byte, error) {
 		Keys:        w.keyRing.ListKeyPairs(),
 		Permissions: w.permissions,
 	}
+
+	if jsonW.Permissions == nil {
```
ze97286, line 330: how can jsonW.Permissions be nil? it's a map not a pointer to a map isn't it?

ValentinTrinque: AFAIK, well that was my mistake. I never though map could be nil.

https://go.dev/play/p/57tl546Uth3

Seems to be possible.

ze97286: your playground is not what's happening in your code though and doesn't explain why you needed this change. This is more like it: https://go.dev/play/p/y6BHV9NYOvu

ze97286: but to be fair `tStruct.testMap == nil` evaluates to true. so you're right.

ze97286 (approved):

PR #5944, opened 2022-08-18
5678 - Add missing gql endpoints

### LINKED ISSUE #5678: Add missing gql endpoint to retrieve ethereum bundles.
GQL endpoints are missing for the following ethereum bundles and needs to be added:
- listAsset
- addSigner
- removeSigner

This will be blocking the frontend and quite soon to be run incentives.

zale144:
Closes #5678 

GraphQL endpoints for
- Erc20ListAssetBundle, 
- Erc20MultiSigSignerAddedBundles and 
- Erc20MultiSigSignerRemovedBundles
- Erc20SetAssetLimitsBundle


datanode/gateway/graphql/resolvers.go
```diff
@@ -591,6 +591,88 @@ func (r *myQueryResolver) Erc20WithdrawalApproval(ctx context.Context, wid strin
...
+		VegaAssetID: res.VegaAssetId,
+		Nonce:       res.Nonce,
+		Signatures:  res.Signatures,
+	}, nil
+}
+
+func (r *myQueryResolver) Erc20MultiSigSignerAddedBundles(ctx context.Context, nodeID string, epochSeq string, pagination *v2.Pagination) (*ERC20MultiSigSignerAddedConnection, error) {
+	res, err := r.tradingDataClientV2.GetERC20MultiSigSignerAddedBundles(
+		ctx, &v2.GetERC20MultiSigSignerAddedBundlesRequest{
+			NodeId:     nodeID,
```
wwestgarth, line 612: I think as of a couple of days ago this now has a `submitter` field too, so we should wire that in too.

datanode/gateway/graphql/schema.graphql
```diff
@@ -542,6 +542,32 @@ type Query {
...
+    pagination: Pagination
+  ): ERC20MultiSigSignerAddedConnection!
+
+  erc20MultiSigSignerRemovedBundles(
+    "The node id of the validator of which a signature bundle is required"
+    nodeId: ID!
+    "The ethereum address of the validator which will submit the bundle"
+    submitter: String!
+    "The epoch which generated the bundle i.e the epoch in which the node was demoted from a tendermint validator"
+    epochSeq: String!
```
wwestgarth, line 560: The ! means non-nullable right? I think we definitely want `epochSeq` to nullable, and maybe even `submitter`.

datanode/gateway/graphql/schema.graphql
```diff
@@ -542,6 +542,34 @@ type Query {
     withdrawalId: ID!
   ): Erc20WithdrawalApproval
 
+  erc20MultiSigSignerAddedBundles(
+    "The node id of the validator of which a signature bundle is required"
```
candida-d, line 546:
```suggestion
    "The node ID of the validator of which a signature bundle is required"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -542,6 +542,34 @@ type Query {
...
+    nodeId: ID!
+    "The ethereum address of the submitter"
+    submitter: String
+    "The epoch which generated the bundle i.e the epoch in which the node was demoted from a tendermint validator"
```
candida-d, line 550:
```suggestion
    "The epoch that generated the bundle i.e the epoch in which the node was demoted from a tendermint validator"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -542,6 +542,34 @@ type Query {
...
+    epochSeq: String
+    "Pagination"
+    pagination: Pagination
+  ): ERC20MultiSigSignerAddedConnection!
+
+  erc20MultiSigSignerRemovedBundles(
+    "The node id of the validator of which a signature bundle is required"
```
candida-d, line 557:
```suggestion
    "The node ID of the validator of which a signature bundle is required"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -542,6 +542,34 @@ type Query {
...
+    nodeId: ID!
+    "The ethereum address of the validator which will submit the bundle"
```
candida-d, line 559:
```suggestion
    "The ethereum address of the validator that will submit the bundle"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -2331,6 +2359,72 @@ type Erc20WithdrawalApproval {
...
+  pageInfo: PageInfo
+}
+
+type ERC20MultiSigSignerAddedBundleEdge {
+  node: ERC20MultiSigSignerAddedBundle!
+  cursor: String!
+}
+
+type ERC20MultiSigSignerAddedBundle {
+  "The ethereum address of the signer to be removed"
```
candida-d, line 2373: Should this be "signer to be added"?

datanode/gateway/graphql/schema.graphql
```diff
@@ -2331,6 +2359,72 @@ type Erc20WithdrawalApproval {
...
+  newSigner: String!
+  "The ethereum address of the submitter"
+  submitter: String!
+  "The nonce used in the signing operation"
+  nonce: String!
+  "unixnano timestamp for when the validator was added"
```
candida-d, line 2379:
```suggestion
  "Unix-nano timestamp for when the validator was added"
```

datanode/gateway/graphql/schema.graphql
candida-d: Should this be "the epoch in which the node was promoted to tendermint validator" if it's under `MultiSigSignerAdded`?

candida-d (commented):
Review of the schema descriptions: Questions on added/removed in some fields, otherwise just minor wording/capitalisation suggestions

datanode/gateway/graphql/schema.graphql
zale144: Honestly I don't really understand much of what these are all about, I've shamelessly copy-pasted these comments from the `.proto` file.

wwestgarth: It should be promoted (my fault, I added the protos). Its probably worth fixing the protos comments too while we're here.

pscott31 (dismissed):

wwestgarth (approved):

datanode/gateway/graphql/schema.graphql
```diff
@@ -542,6 +542,42 @@ type Query {
...
   ): Erc20WithdrawalApproval
 
+  "get the signature bundle to add a particular validator to the signer list of the multisig contract"
+  erc20MultiSigSignerAddedBundles(
+    "The node ID of the validator of which a signature bundle is required"
+    nodeId: ID!
+    "The ethereum address of the submitter"
+    submitter: String
+    "The epoch that generated the bundle i.e the epoch in which the node was promoted to tendermint validator"
+    epochSeq: String
```
dexturr, line 552: question (non-blocking): how does one get this number out of the schema from the schema as it stands?

dexturr (approved):

datanode/gateway/graphql/schema.graphql
wwestgarth: I'm not sure what you mean by "get it out of the schema" :p if you mean what-will-this-string-look-like it'll be "165" or something. If you mean where-is-it-in-the-response:
```
type ERC20MultiSigSignerAddedBundle {
  "The ethereum address of the signer to be added"
  newSigner: String!
  "The ethereum address of the submitter"
  submitter: String!
  "The nonce used in the signing operation"
  nonce: String!
  "Unix-nano timestamp for when the validator was added"
  timestamp: String!
  "The bundle of signatures from current validators to sign in the new signer"
  signatures: String!
  "The epoch in which the validator was added"
  epochSeq: String! <- it is here
}
```

dexturr: So say i want to query this, I need to pass in epochSeq, nodeId ect.  I can get nodeId from the nodes endpoint, which is cool.. Where do I know what the epochSeq from? There is nothing I see in the node type that tells me this currently or am I missing something?

wwestgarth: I get you, and thats an interesting point. I suppose you want to ask "which epoch did my node get promoted so I can ask for that bundle" and my answer would be currently you can't (although the data is in the postgres data-base so we could expose it somewhere)

In reality it will always be the bundle with the highest epoch, so you could get them all (and there won't ever be many) and take the highest, but I understand that's not a great workflow!

PR #5956, opened 2022-08-19
feat: add all grpc stream as websocket through the grpc gateway

jeremyletang:
This adds support for the GRPC steam in the rest layer (grpc-gateway). This also use this library: https://github.com/tmc/grpc-websocket-proxy, which expose these endpoints through websockets.


pscott31 (approved):
Well that's neat

PR #5962, opened 2022-08-23
Minor race in chain

wwestgarth:
Found while playing around with vegacapsule/vegatools-perftest. If `Statistics` isn't called before `LastBlockHeight` then core returns an empty string for the chainID and just reads `s.chainID` unprotected.

Also the tiniest tidy up for setting the chainCtx after snapshot restore so that we don't do it every block.

### DIFF
--- a/core/api/core.go
+++ b/core/api/core.go
@@ -102,7 +102,7 @@ func (s *coreService) LastBlockHeight(
 
 	if atomic.LoadUint32(&s.hasGenesisTimeAndChainID) == 0 {
 		if err := s.getGenesisTimeAndChainID(ctx); err != nil {
-			return nil, fmt.Errorf("failed to intialilse chainID: %w", err)
+			return nil, fmt.Errorf("failed to intialise chainID: %w", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


ValentinTrinque (dismissed):

core/api/core.go
```diff
@@ -100,9 +100,15 @@ func (s *coreService) LastBlockHeight(
 ) (*protoapi.LastBlockHeightResponse, error) {
 	defer metrics.StartAPIRequestAndTimeGRPC("LastBlockHeight")()
 
+	if atomic.LoadUint32(&s.hasGenesisTimeAndChainID) == 0 {
+		if err := s.getGenesisTimeAndChainID(ctx); err != nil {
+			return nil, fmt.Errorf("failed to intialilse chainID: %w", err)
```
peterbarrow, line 105:
```suggestion
			return nil, fmt.Errorf("failed to intialise chainID: %w", err)
```

peterbarrow (dismissed):

peterbarrow (approved):

jeremyletang (approved):

PR #5969, opened 2022-08-23
feat: equity like share total in the vote sides (gql)

### LINKED ISSUE #5682: Governance market proposal LP votes
# Feature Overview

**In order to** achieve the following designs in the Token proposal listings page:

Screenshot 2022-07-08 at 13 53 28

**We will** provide visibility in the GQL API 
**So that** frontend can display the LP vote status (based on total LP yes votes, total LP no votes) for a market proposal (if it has failed the standard participation check)

Background:

Screenshot 2022-07-08 at 13 52 00

We don’t currently have visibility of the process between when the normal participation check fails and when the proposal passes or fails (because of the LP vote).

jeremyletang:
Total equityLikeShare was missing from APIs.

close #5682

### DIFF
--- a/datanode/gateway/graphql/schema.graphql
+++ b/datanode/gateway/graphql/schema.graphql
@@ -3252,6 +3252,8 @@ type ProposalVoteSide {
   totalWeight: String!
   "Total number of governance tokens from the votes cast for this side"
   totalTokens: String!
+  "Total equity like share weight for this side (only for UpdateMarket Proposals)"
+  totalEquityLikeShareWeight: String!
 }
 
 enum VoteValue {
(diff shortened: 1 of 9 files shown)


datanode/gateway/graphql/schema.graphql
```diff
@@ -3141,6 +3143,9 @@ type Vote {
 
   "The weight of this vote based on the total of governance token"
   governanceTokenWeight: String!
+
+  "The weight of this vote based om the total equity like share"
```
candida-d, line 3147:
```suggestion
  "The weight of this vote based on the total equity like share"
```

candida-d: Is this based on the total equity-like share that the public key that's voting has?

dexturr (dismissed):

ValentinTrinque (dismissed):

ValentinTrinque (dismissed):

zale144 (approved):

ze97286 (approved):

mattrussell36 (approved):

PR #5973, opened 2022-08-23
feat: change account subscription to have only top level fields

jeremyletang:
I'm proposing to flatten all subscription types in the graphql layer. I've started only with the accounts for now. When subscribed to an account an event will raise everytime:
- mark to market happen (possibly on every trade)
- margin (low or high) move happen.

Both these events can happen multiple time per blocks.

The query used by the frontend today also request the market and the asset data for every event, which both of them are being update not very often. Even we've cach on the server side, this doesn't seem effective will the client can store these data on the frontend at startup.

part of #5974

### DIFF
--- a/datanode/gateway/graphql/generated.go
+++ b/datanode/gateway/graphql/generated.go
@@ -46,6 +46,7 @@ type Config struct {
 type ResolverRoot interface {
 	Account() AccountResolver
 	AccountEdge() AccountEdgeResolver
+	AccountUpdate() AccountUpdateResolver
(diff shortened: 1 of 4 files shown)


pscott31 (commented):
I think its a good idea, just have to get the front end on board :) [e.g. it's even worse if they then just go and send extra queries straight after to fill in the gaps]

mattrussell36 (approved):

pscott31 (approved):

PR #5976, opened 2022-08-24
(ci): #5670 add GHA for generating docs

### LINKED ISSUE #5670: Generate JSON documentation
With the following PR: https://github.com/vegaprotocol/vega/pull/5667 the protobufs definitions are being imported in the core repo.

In order to limit cluttering the repo, we stopped generating json definitions and swagger. However this is still needed by the docs team to generate the website.

## Tasks
- [ ] add a step in the pipeline to generate all json and swagger
- [ ] Create a new repository dedicated to the documentation codegen
- [ ] On merge to develop, push all update json and swagger on the repo

mkjmdski:
Closes https://github.com/vegaprotocol/vega/issues/5670

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -1,0 +1,67 @@
+---
+name: "Release Documentation"
+
+"on":
+  push:
+    tags:
+      - '**'
+
+jobs:
+  gen-docs:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Check out vega code
+        uses: actions/checkout@v3
+        with:
+          path: 'vega'
+          fetch-depth: '0'
+      -
+        name: "Setup ssh key for PR"
+        uses: webfactory/ssh-agent@v0.5.4
+        with:
+          ssh-private-key: ${{ secrets.DOCUMENTATION_KEY }}
+      -
+        name: Check out docs code
+        uses: actions/checkout@v3
+        with:
+          path: 'documentation'
+          repository: 'vegaprotocol/documentation'
+          ssh-key: ${{ secrets.DOCUMENTATION_KEY }}
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      -
+        name: Get dependencies
+        run: cd vega; bash script/gettools.sh
+      -
+        name: Generate docs
+        run: cd vega; make proto_docs; echo "generated docs:"; find protos/generated -type f
+      -
+        name: Make PR to the docs repo
+        env:
+          GH_TOKEN: ${{ secrets.VEGA_CI_BOT_GITHUB_TOKEN }}
+        run: |
+          cd vega
+          ref="$(git describe --tags --abbrev=0)"
+          short_sha="$(git rev-parse --short HEAD)"
+          cd ..
+          mkdir -p "documentation/specs/${ref}"
+          cp -r vega/protos/generated/* "documentation/specs/${ref}/"
+          cp vega/datanode/gateway/graphql/schema.graphql "documentation/specs/${ref}/datanode-schema.graphql"
+          cd documentation
+          git config --global user.email "vega-ci-bot@vega.xyz"
+          git config --global user.name "vega-ci-bot"
+          branch_name="spec-update-${ref}"
+          git checkout -b "$branch_name"
+          commit_msg="Automated specs update from vegaprotocol/vega#${short_sha}"
+          git add specs
+          git commit -m "$commit_msg"  || exit 0
+          git push -u origin "$branch_name"
+          pr_url="$(gh pr create --title "${commit_msg} to ${ref}" --body 'This is automated pull request that will be merged after pipeline passes' --reviewer vegaprotocol/docs)"
+          echo $pr_url
+          # once auto merge get's enabled on documentation repo let's do follow up
+          # sleep 5
+          # gh pr merge "${pr_url}" --auto --delete-branch --squash
--- a/script/gettools.sh
+++ b/script/gettools.sh
@@ -1,16 +1,5 @@
 #!/bin/bash -e
 
-# Note: Make sure the versions match the ones in devops-infra/docker/cipipeline/Dockerfile
-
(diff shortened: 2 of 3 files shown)


.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: cd vega; bash script/gettools.sh
+      -
+        name: Generate docs
+        run: cd vega; make proto_docs; echo "generated docs:"; find protos/generated -type f
+      -
+        name: Make PR to the docs repo
+        env:
+          GH_TOKEN: ${{ secrets.VEGA_CI_BOT_GITHUB_TOKEN }}
+        run: |
+          mkdir -p documentation/specs/$GITHUB_REF_NAME
```
edd, line 47: In https://github.com/vegaprotocol/documentation/pull/216, the folder name comes out as '#issuenumber' not 'v0.54.0'

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+          cd documentation
+          git config --global user.email "vega-ci-bot@vega.xyz"
+          git config --global user.name "vega-ci-bot"
+          branch_name="$(date +%d-%m-%Y--%H-%M)-spec-update"
+          git checkout -b "$branch_name"
+          commit_msg="Automated specs update from vegaprotocol/vega#$GITHUB_SHA"
+          git add specs
+          git commit -m "$commit_msg"  || exit 0
+          git push -u origin "$branch_name"
+          pr_url="$(gh pr create --title '$commit_msg to $GITHUB_REF_NAME' --body 'This is automated pull request that will be merged after pipeline passes')"
```
edd, line 59: This is not interpreting correctly, https://github.com/vegaprotocol/documentation/pull/216 has the title '$commit_msg to $GITHUB_REF_NAME'

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.VEGA_CI_BOT_GITHUB_TOKEN }}
+        run: |
+          mkdir -p documentation/specs/$GITHUB_REF_NAME
+          cp -r vega/protos/generated/* documentation/specs/$GITHUB_REF_NAME/
+          ls -al documentation/specs/$GITHUB_REF_NAME/
+          cd documentation
+          git config --global user.email "vega-ci-bot@vega.xyz"
+          git config --global user.name "vega-ci-bot"
+          branch_name="$(date +%d-%m-%Y--%H-%M)-spec-update"
```
edd, line 53: Could this change to 'spec-update-[tag]'? i.e. 'spec-update-v0.54.1'

edd (changes requested):

edd:
An additional requirement (which you can refuse if you like, it's expanding the scope): Could this *also* add https://github.com/vegaprotocol/vega/blob/develop/datanode/gateway/graphql/schema.graphql in the pull request?

.github/workflows/release-docs.yml
mkjmdski: GITHUB_REF_NAME is branch name of tag name depending on the event, in this PR branch name is issue number, that's why it confuses. I will just parse `git tag` for that instead

.github/workflows/release-docs.yml
mkjmdski: We can do it but I would opt for a timestamp for a reason that we might do few updates to the same tag as we want to trigger this action on push to develop

.github/workflows/release-docs.yml
mkjmdski: Yeah I will fix the quotes

mkjmdski:
todo: add assignees -> edd, candida-d

.github/workflows/release-docs.yml
edd: @mkjmdski when we chatted about this I opted for on tag instead - in which case would tag be ok here?

edd:
> todo: add assignees -> edd, candida-d

This is what we agreed earlier, but instead I made a new Github team, @vegaprotocol/docs, which means we won't need to change the workflow as the team members who should be notified of the PR change

edd (dismissed):
Thanks @mkjmdski, this looks great - and thanks for helping me find the --assign flag. Turns out I needed to scroll right.

edd (dismissed):

script/gettools.sh
```diff
@@ -2,10 +2,47 @@
...
+if ! command -v protoc; then
+	echo "protoc not detected in the path"
+	echo "Installing protoc to the version ${PROTOC_VERSION}!!"
+	(
+		cd /tmp
+		wget --quiet "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/${file_name}"
+		unzip "${file_name}"
+		mv bin/protoc /usr/local/bin/
+	)
+fi
```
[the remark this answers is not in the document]
mkjmdski: I decided to re-use this script to don't repeat dependencies install. CI Agent's do not have protoc installed so I wrote this small if to install it there.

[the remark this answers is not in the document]
jeremyletang: My point is that we should not need protoc?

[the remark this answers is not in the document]
jeremyletang: Very well, can we see what commands where failing, protoc should be unnecessary in the codebase.

edd (dismissed):

script/gettools.sh
```diff
@@ -1,16 +1,21 @@
...
+	BIN="/usr/local/bin" && \
+	BUF_VERSION="1.7.0" && \
+	curl -sSL \
+		"https://github.com/bufbuild/buf/releases/download/v${BUF_VERSION}/buf-$(uname -s)-$(uname -m)" \
+		-o "${BIN}/buf" && \
+	chmod +x "${BIN}/buf"
 fi
 
+buf --version
+which buf
```
jeremyletang, line 17: Just looking at this, not sure this is necesary as well seeing it's in the list of things we install, see line 19

jeremyletang (approved):

PR #5979, opened 2022-08-24
feat: return list of account changes in subscription

### LINKED ISSUE #5881: GraphQL accounts subscription does not return all accounts in a list
GraphQL accounts subscription does not return all accounts in a list. Instead it pushes accounts as single objects and it's inconsistent with the other subscriptions.

pscott31:
Closes #5881 

NB: Makes a slight change to GQL API that TFE will need to be aware of: The `accounts` subscription now returns a single

```[AccountUpdate!]!```

per block, instead of a flurry of individual 

```AccountUpdate!```

updates

### CODE REVIEW

File: datanode/gateway/graphql/schema.graphql [hunk 1]
```diff
@@ -69,7 +69,7 @@ type Subscription {
     assetId: ID
     "Type of the account"
     type: AccountType
-  ): AccountUpdate!
+  ): [AccountUpdate!]!
```
jeremyletang, line 72: Interesting here it's a non nullable array of non nullable AccountUpdate, if you look at orders or trades, we get a nullable list of non nullable objects

pscott31: The difference, I believe, is that with

```[AccountUpdate!]!```

you get back an empty list if there is, well an empty list.

With 

[AccountUpdate!]

With an empty list, you seem to get back 'null', which seems like more work to handle on the other end to me but I don't really mind

File: datanode/api/trading_data_v2.go [hunk 2]
```diff
@@ -138,9 +138,13 @@ func (t *tradingDataServiceV2) ObserveAccounts(req *v2.ObserveAccountsRequest,
 		t.log.Debug("Accounts subscriber - new rpc stream", logging.Uint64("ref", ref))
 	}
 
-	return observe(ctx, t.log, "Accounts", accountsChan, ref, func(account entities.AccountBalance) error {
+	return observeBatch(ctx, t.log, "Accounts", accountsChan, ref, func(accounts []entities.AccountBalance) error {
+		protoAccounts := make([]*vega.Account, len(accounts))
+		for i := 0; i < len(accounts); i++ {
+			protoAccounts[i] = accounts[i].ToProto()
+		}
```
jeremyletang, line 145: Why not just use a range expression here + append here? Not that it changes much things ultimately though.

jeremyletang (approved):

datanode/api/trading_data_v2.go
pgs31: doesn't that make an extra copy of the account object? I mean I doubt it makes much difference but I was trying to be a good little efficient go boy :)

pscott31: doesn't that make an extra copy of the account object? I mean I doubt it makes much difference but I was trying to be a good little efficient go boy :)

bglownia (approved):

datanode/api/trading_data_v2.go
jeremyletang: nope normally the generated assembly would be pretty much the same, if not the same after optimisations.

pscott31: Was curious so I had a play at https://godbolt.org/ with this little program

```
package main

type Thingy struct {
    X int
    Y string
}

func main() {
    thingies := make([]Thingy, 100)
    wotsits := make([]Thingy, len(thingies))

    // for i := 0; i<len(thingies); i++ {
    //    wotsits[i] = thingies[i]
    // }

    for _, thing := range(thingies){
        wotsits = append(wotsits, thing)
    }
    _=wotsits    
}
```

It's been a while since I had to figure out what some generated assembler is doing but I think when using the second on it is making a temporary copy of each thing as it iterates through the list - there's this bit of assembler that is in the range() version but not in the simple for loop:

```
       MOVQ    BX, main.wotsits.len+64(SP)
        MOVQ    DX, main..autotmp_17+88(SP)
        MOVQ    CX, main..autotmp_18+104(SP)
        MOVQ    R9, main..autotmp_19+80(SP)
        MOVQ    R8, main..autotmp_20+96(SP)
        MOVQ    R10, main..autotmp_21+72(SP)
```

In any case it's a bit academic :)

jeremyletang: Thingy is also not a pointer here, while in your code they are pointers, but yea Thingy is tiny here anyway.

jeremyletang: To be fair, I've approved I don't mind either

pscott31: Thanks 

FWIW the thing I am iterating over is a a `[]entities.AccountBalance`, not `[]*entities.AccountBalance`. The reason being I read somewhere that generally it causes less GC thrashing to copy around actual things rather than pointers to them as they can often live on the stack then instead of the heap.

Issue #5974, opened 2022-08-23
Implement custom streamable types for graphql


jeremyletang:
Grapqhl subscription can become complex on the server side (e.g: imagine an account event, 10 times a second, fetch everytime asset + market).

In order to reduce load, we'll flatten all types returned by subscriptions.

PR #5981, opened 2022-08-24
feat: refactor trade subscribtion to use TradeUdpate

jeremyletang:
parts of #5974


dexturr (dismissed):

datanode/gateway/graphql/schema.graphql
```diff
@@ -2310,6 +2310,54 @@ type Trade {
...
+  "The party that sold"
+  sellerId: ID!
+
+  "The aggressor indicates whether this trade was related to a BUY or SELL"
+  aggressor: Side!
+
+  "The price of the trade (probably initially the passive order price, other determination algorithms are possible though) (uint64)"
+  price: String!
+
+  "The number of contracts trades, will always be <= the remaining size of both orders immediately before the trade (uint64)"
```
candida-d, line 2339:
```suggestion
  "The number of units traded, will always be <= the remaining size of both orders immediately before the trade (uint64)"
```
I want to get confirmation on the use of "units" here but I'm sure that we shouldn't be using "contracts"

ValentinTrinque: This needs to be confirmed with @barnabee .

candida-d: Confirmed -  please go ahead with units @ValentinTrinque @jeremyletang

ze97286 (approved):

mattrussell36 (approved):

PR #5984, opened 2022-08-24
Add initial image to orders, accounts and position subscriptions

### LINKED ISSUE #5880: Send latest state to the client when opening a new subscription
When opening a connection to a GraphQL subscription the existing orders or positions should be pushed to the client, followed by any subsequent updates.

pscott31:
Closes #5880 

NB GraphQL Schema Change: As part of this work it was necessary to change the `positions` graphql schema from a single item at a time:

``` Position!```

to receiving a batch of changes:

``` [Position!]!```

Which may require some front end changes.

ALSO for the orders, accounts and position subscriptions; you will now get a bunch of extra data as the very first message in a subscription, this might have some front end impact too.

### CODE REVIEW

**Review by mattrussell36 — DISMISSED** [review]

**Comment by MM0819** [comment]
@pscott31 I just want to double-check for the orders in the initial snapshot, are you only sending open orders? 

Looking at the different order statuses, I would say we only need:

* Active
* PartiallyFilled
* Parked

There could be thousands of filled/cancelled orders for active traders, and those are certainly not relevant when subscribing to order updates as I understand they will never change in the future.

**Comment by jeremyletang** [comment]
@MM0819 you will need only Active and Parked, Partially filled is the state given to partially filled IOC orders.

**Comment by pscott31** [comment]
Yes you only get active orders. I think though that we don't consider parked orders active, I'll double check that.

**Comment by pscott31** [comment]
I was incorrect about parked orders, we do consider those live image

File: datanode/api/trading_data_v2.go [hunk 1]
```diff
@@ -149,6 +153,44 @@ func (t *tradingDataServiceV2) ObserveAccounts(req *v2.ObserveAccountsRequest,
...
+	}
+
+	accounts, pageInfo, err := t.accountService.QueryBalances(ctx, filter, entities.CursorPagination{})
+	if err != nil {
+		return errors.Wrap(err, "fetching account balance initial image")
+	}
+
+	if pageInfo.HasNextPage {
+		return fmt.Errorf("initial image spans multiple pages")
+	}
```
jeremyletang, line 180: How bit is the default page again? We may want to go against this restriction there, or eventually find a way to send all the pages one after the other. Basically put all the stuff in between line 182 and 191 in a `for pageInfo.HasNextPage {...}`

datanode/api/trading_data_v2.go
```diff
@@ -1115,13 +1159,65 @@ func (t *tradingDataServiceV2) ObservePositions(req *v2.ObservePositionsRequest,
...
+		}
+	}
+
+	// All the positions
+	if req.PartyId == nil && req.MarketId == nil {
+		positions, err = t.positionService.GetAll(ctx)
+		if err != nil {
+			return errors.Wrap(err, "getting initial positions by party")
+		}
+	}
```
jeremyletang, line 1209: Should we even allow that? Is there a use case for listening to ALL the position? isn't that a way to use loads of resources in the datanode?

datanode/api/trading_data_v2.go
pscott31: So the 'default'  page you get by making an empty `Pagination{}` is infinitely big. There's a different default you get by calling NewDefaultPaginatinon() intended to be used that as a default when people call a GRPC method without specifying, which is limited to 1000 or something

datanode/api/trading_data_v2.go
pscott31: Yeah I was in two minds about this actually but I thought given the low number of markets it's not going to be orders of magnitude more data than subscribing to a whole market.. Happy to remove it though if you like, like you say, I can't really think of a use case for it outside of testing

datanode/api/trading_data_v2.go
pscott31: @MM0819 - question for you.. Would it be OK to receive the initial image over multiple messages? @jeremyletang and I just discussing the case where the initial image might be larger than can fit in a GRPC message. We've added pagination to all the other endpoints to solve that problem.

It might be the case that because we're only ever getting live orders etc.. it's unlikely we'll be returning enough data for it to be a problem but perhaps it is prudent to design it to cope with that up front.

to be clear, i'm not suggesting changing the format of the result, adding cursors etc.. just to send initial images in batches of a few hundred or something.

MM0819: Hmmm, so there are a few things here. On other exchanges, there is typically a type field telling you how to handle the message (for example, there will be a field saying `type=snapshot` or `type=update`). The way that a snapshot messages are handled is by replacing your entire state. So, what you propose clearly wouldn't work with that design pattern. 

However, I don't know if we differentiate between the message types in the same way. In which case, I think it would work to receive the snapshot over multiple messages, as long as I don't receive any updates between snapshot messages because I might receive updates for orders where I have not yet received the snapshot.

Also, it's helpful to disable any trading strategy from making trading decisions until its state is 100% synced. This means I would need a way to identify when the final snapshot message has been received when multiple snapshot messages are being sent.

pscott31: OK so here's the solution we've come up with: The observe methods now have a response like 
```

message ObservePositionsResponse {
    oneof response {
        // An 'initial image' snapshot containing current positions
        PositionSnapshotPage snapshot = 1;
        // A list of position updates in the last block
        PositionUpdates updates = 2;
    }
}

// 'Initial image' of current positions, may be sent over multiple response messages
message PositionSnapshotPage{
    repeated vega.Position positions = 1;
    bool last_page = 2;
}

// A list of position updates in the last block
message PositionUpdates{
    repeated vega.Position positions = 1;
}
```

So you can tell if it's a snapshot or not by looking at the `oneof` in the response - and the last page of the snapshot will have `last_page=True`

AnExsomnis (dismissed):

AnExsomnis (dismissed):

ze97286 (approved):

mattrussell36 (approved):

PR #6047, opened 2022-09-02
feat: test probability of trading refactoring

### LINKED ISSUE #6004: Probability of trading refactoring
Probability of trading is currently calculated in the following way:
At certain time points  starting from the end of the opening auction and then every n minutes the calculation is triggered. Given the current best bid/ask for each side separately we calculate a maximum of 500 points per side which represents the probability of trading in offset of i * tickSize from the best level. Then when querying the probability of trading at a given level l with respect to the best bid/ask b/a we find that offset in the table and linearly interpolate (or extrapolate at the end of the table). This has significant drawback: consider the following scenario - at the end of the opening auction one bid order is left on the book after uncrossing, and this bid order is at a level equal the tick size or a small multiple of it. Then when we build the probability/offset table we only calculate 2 entries, at offset 0 and at offset 1 tick size. The former has probability 1 and the latter has probability 0. Then more orders come in the book, lets say one order at 900e18 and we have and we have an LP pegged to the best bid at offset of 1e18. When we calculate the probability of trading at 899e18 based on the pre-calculated table we will get a 0 probability which translates to 1e-8 probability which translates to a huge trading size. 

To improve this it is suggested to have the offsets as % of distance from the best rather than absolute distance. When pre-calculating the table - for bid we calculate 1000 distances at increments of 0.1% offset. For ask we calculate in offset of 0.1% until 5000% the best price. When querying for the probability of trading we calculate the offset as % of the best level and find it in the table the same way as before.

Jiajia-Cui:
closes [issue](https://github.com/vegaprotocol/vega/issues/6004)


core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
```diff
@@ -0,0 +1,147 @@
...
+      | name                                          | value |
+      | market.stake.target.timeWindow                | 24h   |
+      | market.stake.target.scalingFactor             | 1     |
+      | market.liquidity.bondPenaltyParameter         | 0.2   |
+      | market.liquidity.targetstake.triggering.ratio | 0.1   |
+      | market.liquidity.stakeToCcySiskas             | 1.0   |
+
+   And the average block duration is "1"
+
+   Scenario: 001, LP price at 0, check what's happening with LP volume
```
davidsiska-vega, line 34: Are there no AC codes that this applies to?

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
```diff
@@ -0,0 +1,147 @@
...
+    Then the order book should have the following volumes for market "ETH/MAR22":
+      | side | price | volume |
+      | sell | 990   | 1      |
+      | sell | 901   | 112    |
+      | sell | 900   | 2      |
+      | buy  | 899   | 20     |
+      | buy  | 898   | 112    |
+      | buy  | 1     | 1      | 
+
+    # lp1 which is party0 is distressed, and bond account had been slashed
```
davidsiska-vega, line 138: I don't see this happening (bond account slashed)

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
```diff
@@ -0,0 +1,147 @@
...
+    # lp1 which is party0 is distressed, and bond account had been slashed 
+    Then the market data for the market "ETH/MAR22" should be:
+      | trading mode            | supplied stake | target stake |
+      | TRADING_MODE_CONTINUOUS | 50000          | 3201         |
+
+    And the insurance pool balance should be "0" for the market "ETH/MAR22"
+
+    Then the parties should have the following account balances:
+      | party  | asset | market id | margin    | general    |
+      | party0 | USD   | ETH/MAR22 | 430243    | 4999519757 |
```
davidsiska-vega, line 147: How did the party0 get distressed and the bond account slashed if they have 4999519757 in general account?

davidsiska-vega (changes requested):
Pls address the small comments above.

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
Jiajia-Cui: its testing for a bug fixing not a specific AC, I added one AC anyway :)

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
Jiajia-Cui: I have updated the comment :) lp was distressed before I merged into the lates develop branch

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
```diff
@@ -0,0 +1,148 @@
...
+    Then the order book should have the following volumes for market "ETH/MAR22":
+      | side | price | volume |
+      | sell | 990   | 1      |
+      | sell | 901   | 112    |
+      | sell | 900   | 2      |
+      | buy  | 899   | 20     |
+      | buy  | 898   | 112    |
+      | buy  | 1     | 1      | 
+
+    # lp1 which is party0 used to be distressed before the bug fixed, and bond account had been slashed; and now lp1 is not distressed anymore
```
davidsiska-vega, line 138: If you want to talk about "the bug" maybe link to github issue? Otherwise it's a bit meaningless. Otherwise, good.

davidsiska-vega (changes requested):

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
Jiajia-Cui: issue resolved after merge develop branch

core/integration/features/verified/ProbOfTrading_Refactoring_Test.feature
Jiajia-Cui: tidied up

davidsiska-vega:
Please follow the established convention: for example
```
1919-close-out-occurs-above-maintenance.feature 
2878-close-out-orders-allowed-when-under-initial.feature
```
so `[issue-number]-[some sensible name with hyphens instead of spaces].feature`.

davidsiska-vega (approved):

PR #6065, opened 2022-09-05
Add feature test to check the change of risk model para

Jiajia-Cui:


Jiajia-Cui:
closes with [PR](https://github.com/vegaprotocol/vega/pull/6076)

Issue #6056, opened 2022-09-04
Unable to settle a market which has has a update proposal to change oracle condition and value


vega-paul:
Test test_upd_mkt_oracle

Unable to settle a market which has has a update proposal to change oracle condition and value

Have updated an active market as follows:
```
                                  oracle_settlement_condition="Equals",
                                  oracle_settlement_value='1',
                                  oracle_termination_value='True'
```

Market is then terminated - confirmed

Then try to settle a terminated market with settlement price equal 1

Wallet returns...

```
500 Internal Server Error
"there are no subscribes to the oracle data"
````

test logs

```
2022-09-04 16:47:48.156:[INFO] [MainProcess] [MainThread] [  governMod.py]:[send_message]:[1246] Sending oracle data message {'oracle.test.termination3B306': 'True'}
2022-09-04 16:47:48.180:[INFO] [MainProcess] [MainThread] [ walletsMod.py]:[wallet_sign_submit_txn_v2]:[ 249] TRX request b'{"oracleDataSubmission": {"source": "ORACLE_SOURCE_JSON", "payload": "eyJvcmFjbGUudGVzdC50ZXJtaW5hdGlvbjNCMzA2IjogIlRydWUifQ=="}, "propagate": true, "pubKey": "fe63d888177d95d5198109705fd3bf5d035ad18faf8aaf9fb16324a0fecc31d6"}'
2022-09-04 16:47:48.180:[INFO] [MainProcess] [MainThread] [ walletsMod.py]:[wallet_sign_submit_txn_v2]:[ 250] TRX submitted status 200 OK signed by trader/pubkey 4312179A5A/fe63d888177d95d5198109705fd3bf5d035ad18faf8aaf9fb16324a0fecc31d6
[...]

PR #6070, opened 2022-09-05
chore: added logging for unsubscribing of oracles and a unit test for #6056

ze97286:


ze97286:
also could not repro - see https://github.com/vegaprotocol/vega/issues/6056#issuecomment-1237119343

jeremyletang (approved):

PR #6079, opened 2022-09-06
fix: connections can fail the whole query when there are no results for the connection.

### LINKED ISSUE #6060: candlesConnection errors whole query when there aren't any candles for given parameters
# Problem encountered
Asking for candles via new `candlesConnection` errors the whole query when there aren't any candles for given `interval` and `since`.

Example query:
```gql
query CandlesConnectionPerMarket() {
  market(id: "19d3b26182313e068ddde938cbf948c37113a5d4cdafbc6b5b1364805cb442ab") {
    id
    name
    candlesConnection(interval: INTERVAL_I1D, since: "2022-08-31T11:29:24.000Z") {
      edges {
        node {
          volume
        }
      }
    }
  }
}
```

# Observed behaviour
The example query results in `null` data:
```json
{
  "errors": [
    {
      "message": "could not find candle for market 19d3b26182313e068ddde938cbf948c37113a5d4cdafbc6b5b1364805cb442ab and interval INTERVAL_I1D",
      "path": [
        "market",
        "candlesConnection"
      ]
    }
  ],
  "data": {
    "market": null
  }
}
```

# Expected behaviour
It should return market data but only `null` for `candlesConnection` if not found.

Deprecated `candles` results as expected:
```gql
query CandlesPerMarket() {
  market(id: "19d3b26182313e068ddde938cbf948c37113a5d4cdafbc6b5b1364805cb442ab") {
    id
    name
    candles(interval: INTERVAL_I1D, since: "2022-08-31T11:29:24.000Z") {
      volume
    }
  }
}
```
```json
{
  "data": {
    "market": {
      "id": "19d3b26182313e068ddde938cbf948c37113a5d4cdafbc6b5b1364805cb442ab",
      "name": "BTCUSD Monthly (30 Jun 2022)",
      "candles": null
    }
  }
}
```

# Definition of Done

**Before Merging**

**After Merging**

guoguojin:
This is due to the connection results being defined as mandatory
un-necessarily. This update fixes that to make the new connection
queries behave the same way as the queries they are replacing.

Closes #6060


pscott31:
I don't necessarily disagree with this but why would there ever not be a connection at all? I would have thought that a connection with zero edges would be the standard 'there's nothing here' response?

guoguojin:
It's when the connection is a sub-query for something else, for example in the ticket, they are querying market and want to see the candles. The query was failing the market query too because the candlesConnection was mandatory, but it should still return the data for the existing market, and null and errors for the candles when there is no data.

pscott31 (dismissed):

pscott31 (approved):

asiaznik (approved):

PR #6086, opened 2022-09-06
chore: Add core node status headers to gateway APIs

### LINKED ISSUE #5971: [data-node] [fairground] Data nodes serve stale data if their Vega node dies
# Ticket update
The original ticket suggested this was inconsistent data, which is *true*. However it's inconsistent because *n06 and n11 died*, and the data-node is serving old data from before the node died. Now the purpose of this ticket is how to represent that to the user so that they know not to trust the data they're seeing.

# Original ticket
I deposited some tDAI: https://ropsten.etherscan.io/tx/0x3aa570fa38a09e788835263b5942c4482a5708e5d94e69a0bd09da4d93f11192

My UI never successfully saw the deposit. So I went to Grafana and saw that the deposit had been seen:
image

So I went to GraphQL on n11 (the one the fairground console connects to by default) and ran:

```graphql
{
  party(id: "6eed3515524b3c6dda7a9eceaebba58e236a56faa8ef7b8129179c9d269f41f2") {
    depositsConnection {
      edges {
        node {
          asset { id name  }
          amount
          status
        }
      }
    }
  }
}
```

And got:

```json
{
  "data": {
    "party": {
      "depositsConnection": {
        "edges": null
      }
    }
  }
}
```

But how can that be? So I checked n10 instead, and got:

```json
{
  "data": {
    "party": {
      "depositsConnection": {
        "edges": [
          {
            "node": {
              "asset": {
                "id": "6d9d35f657589e40ddfb448b7ad4a7463b66efb307527fedd2aa7df1bbd5ea61",
                "name": "tDAI TEST"
              },
              "amount": "1000000000",
              "status": "STATUS_FINALIZED"
            }
          },
          {
            "node": {
              "asset": {
                "id": "5cfa87844724df6069b94e4c8a6f03af21907d7bc251593d08e4251043ee9f7c",
                "name": "tBTC TEST"
              },
              "amount": "11100000",
              "status": "STATUS_FINALIZED"
            }
          },
          {
            "node": {
              "asset": {
                "id": "5cfa87844724df6069b94e4c8a6f03af21907d7bc251593d08e4251043ee9f7c",
                "name": "tBTC TEST"
              },
              "amount": "12300000",
              "status": "STATUS_FINALIZED"
            }
          }
        ]
      }
    }
  }
}
```

A more complete survey showed that `n06-testnet` and `n11-testnet` return no TXs for my user, but `n07`, `n08`, `n09`, `n10`, `n12` return the transaction list.

# Solution
The Vega protocol design team decided the best fix for HTTP based clients is to use additional HTTP headers in the response to denote the state of the system. That way:
- We don't potentially break client implementations that rely on a `200` status code
- We don't have a bespoke REST or GraphQL endpoint that must be checked independently of the request you actually care about
- Existing implementations will be unaffected
- There is a clear indicator on HTTP responses

This doesn't address GRPC as the meeting didn't cover GRPC. If there's an easy adaptation, great.

# Task
- [ ] Every REST & GraphQL response should container the new headers:
  - [ ] `x-block-height`: '123'.
  - [ ] `x-block-timestamp: '1661338849'
- [ ] If a data node has not received blocks/events from its Node in a reasonable period of time, an additional header is set
  - [ ] `x-vega-connection: 'DISCONNECTED'` 

## Variations
- `x-block-timestamp` could also use the standard `Last modified` header if preferred. We went for block-timestamp because Last Modified could imply that the contents last changed then, which it may not have
- `x-vega-connection` could always be populated with a `CONNECTED` or `DISCONNECTED`
- `x-vega-connection` could have a clearer name

zale144:
Add all the changes necessary to add gRPC, GraphQL and REST APIs headers for communicating to the client about the state of the core node.

References #5971


jeremyletang:
Closing as another implementation has been merged now. cheers.

PR #6110, opened 2022-09-08
fix: Verify global reward is transferred to party address 0

### LINKED ISSUE #6067: A one-off transfer to an account reward type ignores the TO field
# Problem encountered
Single transfers to a reward pool should not be used, and should not be supported (#6066). However, they are currently possible. And there's a weird wrinkle:

```bash
vegawallet command send --wallet test --pubkey a35db5fcbc46962cc8283dbbccc506e82c95a548eddc913da5e35ea173c3b279 --network stagnet3 '{
    "transfer":{
        "fromAccountType": "ACCOUNT_TYPE_GENERAL",
        "toAccountType": "ACCOUNT_TYPE_GLOBAL_REWARD",
        "to":"a35db5fcbc46962cc8283dbbccc506e82c95a548eddc913da5e35ea173c3b279",
        "asset":"b4f2726571fbe8e33b442dc92ed2d7f0d810e21835b7371a7915a365f07ccd9b",
        "amount":"1",
        "oneOff":{
            "deliverOn": 0
        }
    }
}'
```

The above code snippet suggests it will transfer 0.000...1 *from* my general account, to my `ACCOUNT_TYPE_GLOBAL_REWARD` account. But a party can't have an account of that type (see: [00013-ACCT](https://github.com/vegaprotocol/specs/blob/master/protocol/0013-ACCT-accounts.md#accounts-controlled-by-parties)). What happens instead is that the `TO` field is ignored and the transfer instead targets the `ACCOUNT_TYPE_GLOBAL_REWARD` for asset `b4f2726571fbe8e33b442dc92ed2d7f0d810e21835b7371a7915a365f07ccd9b`, which it then transfers. 

# Expected behaviour
The transfer should be rejected because a party that is not 00000000000 can never have a `ACCOUNT_TYPE_GLOBAL_REWARD`.

ze97286:
Closes #6067


edd (dismissed):
👍 Thanks, this should clarify things

jeremyletang (approved):

pscott31 (approved):

PR #6112, opened 2022-09-08
fix: deadlock when removing candle subscriptions

### LINKED ISSUE #6111: Candle subscriptions not being cancelled properly
On testnet we end up with hundreds of thousands of candle subscriptions despite only having a few clients. There's a bug stopping them from unsubscribing properly.

pscott31:
Closes #6111

### DIFF
--- a/datanode/candlesv2/candle_updates.go
+++ b/datanode/candlesv2/candle_updates.go
@@ -14,118 +14,122 @@ package candlesv2
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"time"
 
 	"code.vegaprotocol.io/vega/datanode/entities"
+	"code.vegaprotocol.io/vega/datanode/utils"
 	"code.vegaprotocol.io/vega/logging"
 )
+
+var ErrCandleUpdatesStopped = errors.New("candle updates stopped")
 
 type candleSource interface {
 	GetCandleDataForTimeSpan(ctx context.Context, candleID string, from *time.Time, to *time.Time,
 		p entities.CursorPagination) ([]entities.Candle, entities.PageInfo, error)
 }
 
-type subscribeRequest struct {
-	id  string
-	out chan entities.Candle
-}
-
 type CandleUpdates struct {
-	log                *logging.Logger
-	candleSource       candleSource
-	candleID           string
-	subscribeChan      chan subscribeRequest
-	unsubscribeChan    chan string
-	nextSubscriptionID uint64
-	config             CandleUpdatesConfig
+	log          *logging.Logger
+	candleSource candleSource
+	candleID     string
+	config       CandleUpdatesConfig
+	running      atomic.Bool
+	lastCandle   atomic.Pointer[entities.Candle]
+	notifier     utils.Notifier[entities.Candle]
+	stopping     sync.Mutex
 }
 
 func NewCandleUpdates(ctx context.Context, log *logging.Logger, candleID string, candleSource candleSource,
 	config CandleUpdatesConfig) (*CandleUpdates, error,
 ) {
+	observerName := fmt.Sprintf("candles(%s)", candleID)
 	ces := &CandleUpdates{
-		log:             log,
-		candleSource:    candleSource,
-		candleID:        candleID,
-		config:          config,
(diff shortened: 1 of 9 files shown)


datanode/candlesv2/candle_updates.go
```diff
@@ -33,99 +34,83 @@ type subscribeRequest struct {
...
-	out := make(chan entities.Candle, s.config.CandleUpdatesStreamBufferSize)
+func (s *CandleUpdates) Subscribe() (<-chan []entities.Candle, uint64) {
+	ch, id := s.observer.Subscribe()
 
-	nextID := atomic.AddUint64(&s.nextSubscriptionID, 1)
-	subscriptionID := fmt.Sprintf("%s-%d", s.candleID, nextID)
-	s.subscribeChan <- subscribeRequest{
-		id:  subscriptionID,
-		out: out,
+	if lastCandle := s.lastCandle.Load(); lastCandle != nil {
```
ettec, line 101: this is I think is a race condition that could result in the same candle being sent twice, or worse, see 2 scenarios below, this is why the last candle logic was in the same state space as the per candle go routine in the current implementation: - 

Subscribe thread -> add observer
Send Thread-> store last candle 
Subscribe thread -> reads last candle
Send Thread->  Sends a shed ton of new candles, including to the new observer
Subscribe thread -> Sends now very old last candle to the new observer

Or another less drastic but similar possibility:

Subscribe thread -> add observer
Send Thread-> store last candle 
Subscribe thread -> send last candle
Send Thread -> sends same last candle to subscriber

ettec (changes requested):

datanode/candlesv2/service.go
```diff
@@ -49,54 +48,51 @@ func NewService(ctx context.Context, log *logging.Logger, config Config, candleS
...
-		return "", nil, fmt.Errorf("no candle exists for candle id:%s", candleID)
+		return 0, nil, fmt.Errorf("no candle exists for candle id:%s", candleID)
 	}
 
-	if _, ok := cs.candleIDToUpdatesStream[candleID]; !ok {
-		updates, err := NewCandleUpdates(cs.ctx, cs.log, candleID, cs, cs.Config.CandleUpdates)
+	updatesStream, ok := cs.candleIDToUpdatesStream[candleID]
+
+	// If we don't have a stream for this candle, or existing update stream has stopped running, make a new one.
+	if !ok || (ok && !updatesStream.IsRunning()) {
```
ettec, line 77: Possiblility here that a subscribing client ends us subscribed to a dead updates stream, as the updates stream dies after this call, but the subscriber thread go on to subscribe thinking all is well, but it will now be subscribed to a dead stream

datanode/candlesv2/candle_updates.go
pscott31: You're quite right. I've fixed this by adding a `SubscribeAndNotify` method to the Notifier (previously called Observer) and using that.

datanode/candlesv2/service.go
pscott31: Yes you are right again. Spent an inordinate amount of time trying to come up with the best solution to this and I think I am hopefully there now. There's now a mutex that gets taken when the UpdateStream is in the process of shutting down that prevents new subscriptions.

jeremyletang (approved):

datanode/candlesv2/candle_updates.go
```diff
@@ -14,118 +14,122 @@ package candlesv2
...
+			if s.notifier.GetSubscribersCount() > 0 {
+				candles, err := s.getCandleUpdates(ctx, s.lastCandle.Load())
 				if err != nil {
 					s.log.Errorf("failed to get candles, closing stream for candle id %s: %w", s.candleID, err)
 					return
 				}
 
 				if len(candles) > 0 {
-					lastCandle = &candles[len(candles)-1]
+					s.lastCandle.Store(&candles[len(candles)-1])
```
ettec, line 90: Possible race condition here:
Sender Thread-> Store last candle
Subscriber Thread->  Gets last candle sends it  and suscribes
Sender Thread-> Sends all candles including the last candle already sent to all subscribers

I think all you need to do is flip the notify and store last candle order in the sender thread

PR #6117, opened 2022-09-09
feat: add batch market instructions

### LINKED ISSUE #5961: Implement Batch Instruction
Implement functionallity that adds a transaction type that allows a user of the protocol to submit multiple market instructions (e.g. SubmitOrder, CancelOrder, AmendOrder) in a single transaction.

Spec:
https://github.com/vegaprotocol/specs/blob/master/protocol/0074-BTCH-batch-market-instructions.md

jeremyletang:
close #5961


core/processor/abci.go
```diff
@@ -169,6 +170,8 @@ type App struct {
 
 	nilPow  bool
 	nilSpam bool
+
+	maxBatchSize uint64
```
wwestgarth, line 174: I think we need a lock on this incase the netparam is being updated while a incoming batch transaction is being length checked.

jeremyletang: fair enough. will do.

jeremyletang: addressed

wwestgarth (dismissed):

ValentinTrinque (approved):

PR #6125, opened 2022-09-09
chore: Add core node status headers to gateway APIs

### LINKED ISSUE #5971: [data-node] [fairground] Data nodes serve stale data if their Vega node dies

zale144:
Add all the changes necessary to add gRPC, GraphQL and REST APIs headers for communicating to the client about the state of the core node.

Closes #5971


pscott31 (dismissed):
This looks good to me, though I admit I don't quite follow all the interceptor magic fully it seems sensible.

One thing that worries me a little bit is that by the looks of it this is going to add one grpc call to from datanode->core for every grpc call that happens on datanode. Some single GQL queries resolve into very many grpc calls - e.g. something `{trades{market{id}}` has a separate GRPC call to get the market for every trade.

Maybe GetState is so utterly trivial that it doesn't matter but its worth keeping an eye on, if it becomes a bother maybe we can fetch it once a second or something and cache it.

zale144:
> This looks good to me, though I admit I don't quite follow all the interceptor magic fully it seems sensible.
> 
> One thing that worries me a little bit is that by the looks of it this is going to add one grpc call to from datanode->core for every grpc call that happens on datanode. Some single GQL queries resolve into very many grpc calls - e.g. something `{trades{market{id}}` has a separate GRPC call to get the market for every trade.
> 
> Maybe GetState is so utterly trivial that it doesn't matter but its worth keeping an eye on, if it becomes a bother maybe we can fetch it once a second or something and cache it.

`GetState` doesn't need any network calls or anything. It merely checks the value on a struct field `state` in 

```
type connectivityStateManager struct {
	mu         sync.Mutex
	state      connectivity.State
	notifyChan chan struct{}
	channelzID *channelz.Identifier
}
```

which is in the `google.golang.org/grpc` library. 

However `GetLastBlock` is a bit more expensive, it comes from the `BlockService` and ends up querying the db with this query: `SELECT vega_time, height, hash FROM blocks order by vega_time desc limit 1`

Like I have mentioned in the Slack thread, I still need to find a way to only call these methods once per request, because now if we use GQL, they get called double, because gRPC interceptors make one call to each method, and GQL middleware makes the another call, since the calls to the GQL are propagated to the gRPC client.

jeremyletang:
> Like I have mentioned in the Slack thread, I still need to find a way to only call these methods once per request, because now if we use GQL, they get called double, because gRPC interceptors make one call to each method, and GQL middleware makes the another call, since the calls to the GQL are propagated to the gRPC client.

Makes sense, have you figured a way to do this? How about passing a flag on the grpc / rest call to indicate if you dont want the extra informations in the headers? This way you can just avoid all the calls made by the grpc handlers when going via graphql, and instead add it directly from the  graphql layer?

datanode/gateway/graphql/server.go
```diff
@@ -23,15 +23,18 @@ import (
 	"strconv"
 	"time"
 
+	"golang.org/x/crypto/acme/autocert"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
```
jeremyletang, line 28: It's just style, but usually we try to have imports with the following order in the codebase:
- std import
- project import (code.vegaprotocol.io/*)
- external dependencies
Each block of imports being separated with a blank line.

datanode/gateway/middleware.go
```diff
@@ -64,3 +65,55 @@ func MetricCollectionMiddleware(log *logging.Logger, next http.Handler) http.Han
...
+}
+
+type InjectableResponseWriter struct {
+	http.ResponseWriter
+	headers http.Header
+}
+
+type key string
+
+const injectableWriterKey key = "injectable-writer-key"
```
jeremyletang, line 95: We usually inject key like in the following file in the context: libs/context/context.go.

I also kinda seems to remember its the recommended way to do it.

zale144:
> > Like I have mentioned in the Slack thread, I still need to find a way to only call these methods once per request, because now if we use GQL, they get called double, because gRPC interceptors make one call to each method, and GQL middleware makes the another call, since the calls to the GQL are propagated to the gRPC client.
> 
> Makes sense, have you figured a way to do this? How about passing a flag on the grpc / rest call to indicate if you dont want the extra informations in the headers? This way you can just avoid all the calls made by the grpc handlers when going via graphql, and instead add it directly from the graphql layer?

OK, I've pushed some changes. Basically, I'll be getting the headers from gRPC server, so no need to query the values from GQL API again

pscott31 (approved):

jeremyletang (approved):

PR #6143, opened 2022-09-12
chore: some fixes / style in execution market

jeremyletang:


core/execution/engine.go
```diff
@@ -215,8 +215,13 @@ func (e *Engine) Hash() []byte {
...
+
+	// get the accounts hash + add it at end of all markets hash
+	accountsHash := e.collateral.Hash()
+	e.log.Debug("accounts state hash", logging.Hash(accountsHash))
+
 	bytes := []byte{}
-	for _, h := range hashes {
+	for _, h := range append(hashes, string(accountsHash)) {
 		bytes = append(bytes, []byte(h)...)
 	}
```
jeremyletang, line 226: Moved that here as we don't need to hash the collateral  state for every markets.

core/execution/market.go
```diff
@@ -1478,6 +1419,9 @@ func (m *Market) submitValidatedOrder(ctx context.Context, order *types.Order) (
 		if isPegged {
 			m.peggedOrders.Park(order)
 		}
+		// parking the order, needs to unregister it first
+		_ = m.position.UnregisterOrder(ctx, order)
+		m.broker.Send(events.NewOrderEvent(ctx, order))
 		return &types.OrderConfirmation{Order: order}, nil, nil
```
jeremyletang, line 1425: @EVODelavega @peterbarrow Just to confirm, when reading the code I realised that we may not have been unregistering position in this specific case of submit order. Would be good to have another pair of eyes there.

core/execution/market.go
```diff
@@ -2612,7 +2557,10 @@ func (m *Market) amendOrder(
 			if needToRemoveExpiry {
 				m.expiringOrders.RemoveOrder(expiresAt, existingOrder.ID)
 			}
-			if needToAddExpiry {
+			// need to make sure the order haven't been matched with the
+			// amend, consuming the remain volume as well or we would
+			// add an order while it's not needed to the expiring list
+			if needToAddExpiry && cnf != nil && !cnf.Order.IsFinished() {
```
jeremyletang, line 2563: I suppose we would have added order in expiry list which might have traded fully due to the amend...

core/execution/market.go
```diff
@@ -2984,13 +2929,11 @@ func (m *Market) orderCancelReplace(
 	// the ones with the fees embedded
 	conf.Trades = trades
 
-	// pegged order might have been parked
+	// pegged order might have been parked ??? NO?
 	if m.peggedOrders.IsParked(newOrder.ID) {
 		m.peggedOrders.Unpark(newOrder.ID)
 	}
```
jeremyletang, line 2935: Pretty sure that can't happen, as we handle parked orders earlier in the flow, by just submitting them, also we would panic a few lines before trying to remove it from book.

core/matching/orderbook.go
```diff
@@ -747,10 +747,7 @@ func (b *OrderBook) ReplaceOrder(rm, rpl *types.Order) (*types.OrderConfirmation
 	if _, err := b.CancelOrder(rm); err != nil {
 		return nil, err
 	}
-	// Because other collections might be pointing at the original order
-	// use it's memory when inserting the new version
-	*rm = *rpl
-	return b.SubmitOrder(rm)
```
jeremyletang, line 753: Should not be needed anymore as I suppose that was needed for the pegged orders. So would rather avoid using that

core/execution/engine.go
```diff
@@ -215,8 +215,13 @@ func (e *Engine) Hash() []byte {
...
```
wwestgarth, line 215: If you can be bothered because its unrelated: this loop above:
```
for _, m := range e.markets {
	hash := m.Hash()
	e.log.Debug("market app state hash", logging.Hash(hash), logging.String("market-id", m.GetID()))
	hashes = append(hashes, string(hash))
}
```
if over a map, so between nodes the logs of the market hashes are sometimes in different orders which always caused me endless pain when comparing logs during snapshot bugs.

Wondering if it makes sense to use `e.marketsCpy` instead.

jeremyletang: sounds good. will do.

jeremyletang: addressed

core/execution/market.go
EVODelavega: Perhaps, that being said, having the `position.UnregisterOrder` call outside of the `if isPegged` does imply that we'd be calling this for both pegged orders and LPs, when in auction. I'm not 100% on whether we need it for both (although I'd need to take a bit more time to go through all possible scenarios

jeremyletang: hum good point. we should look into that yea.

jeremyletang: Alright, just looked  and that's be OK for pegged orders at least.

jeremyletang: Also looked at the liquidity orders, and that should work well for them as well. but please look on your side. I think that was actually really a bug before.

core/matching/orderbook.go
EVODelavega: Probably mixing 3 things up in my mind, but there's something about this deref/assign... Perhaps this is part of some trickery that gives the caller (or some caller higher up the call chain) access to the same memory so we can update market/original prices for example when updating LP order prices?

Probably not, but just thought I'd mention it

EVODelavega (dismissed):

EVODelavega (approved):

PR #6165, opened 2022-09-13
feat: bug in feature test, initial order from trader3 should be rejected but got accepted in the feature test

### LINKED ISSUE #6188: Bug in feature test, initial order from trader3 should be rejected but got accepted in the feature test
please see details in the [PR](https://github.com/vegaprotocol/vega/pull/6165)

- In the feature test in line 79, the initial order from trader3 should be rejected since they do not have enough collateral to cover the initial margin level.

- I think its a bug existing in feature test.

- I did the same test in market-sim (with the same risk parameter/net work parameter and same orders on the market), and the order from trader3 does get rejected.

Jiajia-Cui:
- In line 79, the initial order from trader3 should be rejected since they do not have enough collateral to cover the initial margin level.

- I think its a bug existing in feature test.

- I did the same test in market-sim (with the same risk parameter/net work parameter and same orders on the market), and the order from trader3 does get rejected.


Jiajia-Cui:
close with [issue](https://github.com/vegaprotocol/vega/issues/6188)

PR #6177, opened 2022-09-14
Add {Begin|End}Block and ProtocolUpgradeStarted events

jeremyletang:


wwestgarth:
I don't know how the data-node plans on handling the new end-block event, but I do know that we send on check-point events in `OnCommit()` which will have the block-height of the block we claimed we just ended.

core/processor/abci.go
```diff
@@ -626,7 +627,13 @@ func (app *App) OnEndBlock(req tmtypes.RequestEndBlock) (ctx context.Context, re
 		logging.String("previous-datetime", vegatime.Format(app.previousTimestamp)),
 	)
 
-	app.epoch.OnBlockEnd(ctx)
+	app.broker.Send(
```
wwestgarth, line 630: I feel like should we not being sending as the last thing we do in`OnEndBlock()` once the engines have finished all their end-block processing and sent out their events?

jeremyletang: addressed

pscott31 (approved):

Issue #6123, opened 2022-09-09
Snapshots: manually run soak-tests and fix all bugs (again)


wwestgarth:
# Feature Overview

**In order to** be sure snapshots are still working
**We will** manually run the snapshot soak test script
**So that** we can fix any bugs

This is in leiu of https://github.com/vegaprotocol/devops-infra/issues/1444

As QA add system-tests to flex more AC's it puts core into new states and we need to make sure that all the snapshots taken at all these states are functional and can be loaded by loading from every single one.

We probably also need to decide when the "last" time we should run this script before trading is enabled will be. It needs to be late enough that we have as many tests as possible, but also giving core enough time to fix any bugs that it might throw up....

Its worth noting that the test set `full a-f` produces snapshots that cannot be loaded, so there is at least one outstanding bug somewhere.

## Specs
- [Link](xyz) to spec or section within a spec

# Tasks
A checklist of the tasks that are needed to develop the feature and meet the acceptance criteria and feature test scenarios. Ideally, tasks would reflect the issues/pull requests likely to be created when developing the feature. 
- [ ]
- [ ]

# Acceptance Criteria
A list of criteria (aim for 3!) that have to be met for this feature to be accepted as 'done' by the product owner.

- It is possible to

Acceptance Criteria and Feature Test Scenarios can, in some cases, be closely related. If acceptance criteria become fully covered by feature test scenarios they can be removed leaving on acceptance criteria that can't be directly or fully proven with tests.

# Test Scenarios
Detailed scenarios (1-3!) that can be executed as feature tests to verify that the feature has been implemented as expected. We use the follow format:

GIVEN (setup/context) 
WHEN (action) 
THEN (assertion) For example...
See [here](https://github.com/vegaprotocol/vega/tree/develop/integration/) for more format information and examples.

# Additional Details (optional)
Any additional information that provides context or gives information that will help us develop the feature. 

# Examples (optional)
Code snippets from the spec for reference

# Definition of Done
>ℹ️ Not every issue will need every item checked, however, every item on this list should be properly considered and actioned to meet the [DoD](https://github.com/vegaprotocol/vega/blob/develop/DEFINITION_OF_DONE.md).

**Before Merging**
- [ ] Create relevant for [system-test](https://github.com/vegaprotocol/system-tests/issues) tickets with feature labels
- [ ] Code refactored to meet SOLID and other code design principles
- [ ] Code is compilation error, warning, and hint free
[...]

PR #6199, opened 2022-09-15
Snapshot fixes

wwestgarth:
Today's snapshot fixes

Related to #6123


core/execution/market_snapshot.go
```diff
@@ -161,6 +161,10 @@ func NewMarketFromSnapshot(
 	market.tradableInstrument.Instrument.Product.NotifyOnSettlementPrice(market.settlementPrice)
 	liqEngine.SetGetStaticPricesFunc(market.getBestStaticPricesDecimal)
 
+	if mkt.State == types.MarketStateTradingTerminated {
+		market.tradableInstrument.Instrument.Product.UnsubscribeTradingTerminated(ctx)
+	}
```
jeremyletang, line 166: Why not d othe check at line 160, and just not register in the first time?

wwestgarth: 160 is the callback from the instrument into the market. The subscription to the oracles happens deep in the heart of `markets.NewTradableInstrument(ctx, log, mkt.TradableInstrument, oracleEngine)`: https://github.com/vegaprotocol/vega/blob/develop/core/products/future.go#L260

jeremyletang: right makes sense. just feels a bit weird that we register it just to unregister soon after.

wwestgarth: I hear you. I'll do a whole `NewTradableInstrumentFromSnapshot()` type thing.

jeremyletang: ah sounds good.

ValentinTrinque (approved):

PR #6209, opened 2022-09-16
fix: update positions to 0 when market is settled

### LINKED ISSUE #5998: [graphql] Positions should not have non-zero unrealisedPNL or openVolume
# Problem encountered
A user has a position in a market that has settled. GraphQL API is returning non-zero `openVolume` and `unrealisedPNL`

Query run on testnet node 10:
```graphql
{
 party(id:"3aec463efd76df0c584b5f90180acd2a3d5f3d3e1c3d7b0d235337d77be6cce9"){
    positionsConnection{
      edges {
        node {
          market { name state }
      	  realisedPNL
          unrealisedPNL
          openVolume
        }
      }
    }
  }
}
```
```json
{
  "data": {
    "party": {
      "positionsConnection": {
        "edges": [
          {
            "node": {
              "market": {
                "name": "OP/USD AUG-SEP22 - Incentive.",
                "state": "STATE_ACTIVE"
              },
              "realisedPNL": "0",
              "unrealisedPNL": "-380",
              "openVolume": "2"
            }
          },
          {
            "node": {
              "market": {
                "name": "OP/USD AUG-SEP22 - Incentive",
                "state": "STATE_SETTLED"
              },
              "realisedPNL": "0",
              "unrealisedPNL": "0",
              "openVolume": "0"
            }
          },
          {
            "node": {
              "market": {
                "name": "OP/USD AUG-SEP22",
                "state": "STATE_SETTLED"
              },
              "realisedPNL": "24",
              "unrealisedPNL": "52",
              "openVolume": "4"
            }
          },
          {
            "node": {
              "market": {
                "name": "BTCUSD (1 days) - 8 prices",
                "state": "STATE_SETTLED"
              },
              "realisedPNL": "-8111953085",
              "unrealisedPNL": "0",
              "openVolume": "0"
            }
          }
        ]
      }
    }
  }
}
```

**Note** Diving deeper in to the OP/USD AUG-SEP22 market, all of the user's orders are finalised (filled/stopped/cancelled/rejected) so it's likely to be the positions API having not done all the tidyup, as per @barnabee's note below.

# Expected behaviour
All orders are closed out and settled. `openVolume` is zero. `unrealisedPNL` is 0 and `realisedPNL` is set appropriately

# Steps to reproduce

## Manual
1. Propose a market
2. Vote the market in
3. Provide liquidity so the market gets out of opening auction
4. Trade a bit
5. Settle the market (JSON oracle, unsure if this is important)
6. Perform the query above

> So core probably emits a “market settled” event and then deletes the entire data structure. No need to update a position to zero if you're going to delete it.

> Data node on the other hand cares about historic information so it needs to decide the position is now zero, either by interpreting the “market settled” event in a way that updates all positions on the market to zero, or by asking core to emit more of some other event that would cause the position to be updated to zero.

guoguojin:
Closes #5998


pscott31:
I think this is probably OK, but I think it would be better if core emitted some synthetic trades that close down the position when the market is settled.

With this, now when you sum up all your trades after a market has settled they won't add up to 0. If you make a spreadsheet of your trades to work out your PnL you are going have to make special accounting for what happens at settlement; you're going to have to dig into oracle data to figure out what your settlement price was.

'Realised PnL' gets set to 0 here as well, so then you're going to have to dig around to find the transfer of your last bit of PnL into your general account.

But I guess this is better than doing nothing at all when it settles, as we currently are :)

davidsiska-vega:
> if core emitted some synthetic trades that close down the position when the market is settled. 
That would make sense - we should already have a precedent of synthetic trades when a party is closed out due to breach of margin.

guoguojin:
I'm closing this one as I messed up the merging a bit as it has been sitting around for a while. The new PR to fix this is https://github.com/vegaprotocol/vega/pull/6262

PR #6226, opened 2022-09-20
feat: close out using log-normal model

Jiajia-Cui:


core/integration/features/verified/closeout-2 partyies_lognormal.feature
```diff
@@ -0,0 +1,130 @@
...
+      | buy  | 50     | 40       | 
+      | buy  | 100    | 10       | 
+      | sell | 1000   | 10       | 
+      | sell | 1055   | 223      | 
+
+    And the parties should have the following margin levels:
+      | party     | market id  | maintenance | search | initial | release |
+      | trader3   | ETH/DEC19  | 81          | 121    | 162     | 243     |
+
+    # how come trader3's order is not rejected? they only have 90 in the collateral and the initial margin level is 162
```
witgaw, line 89: Isn't this the issue recently discussed on Slack where a spec change was proposed? I think the comment should be rephrased to explain why this is allowed.

core/integration/features/verified/closeout-2 partyies_lognormal.feature
```diff
@@ -0,0 +1,130 @@
...
+    And the insurance pool balance should be "0" for the market "ETH/DEC19"
+
+    Then the parties should have the following profit and loss:
+      | party      | volume | unrealised pnl | realised pnl |
+      | auxiliary1 | -10    | 0              | 0            |
+      | auxiliary2 | 10     | 0              | 0            |
+   # setup trader3 position and close it out
+    When the parties place the following orders:
+      | party   | market id | side | volume| price | resulting trades | type       | tif     | reference      |
+      | trader2 | ETH/DEC19 | sell | 10    | 100   | 1                | TYPE_LIMIT | TIF_GTC | sell-provider-1|
```
witgaw, line 104: Wouldn't it be cleaner to use one of the auxiliary traders to move the market here? I struggle to understand if trader2 really get's forcibly closed out or if he just trades -10 here and then his +10 order trades with networks effectively closing him out (not "forcibly" though if that makes sense). Might also be worth adding a check for the resulting trades to demonstrate that `trader3` acquires a position of +10 as a result of this order.

witgaw (changes requested):
Please see the comments below, file name should also be tidied up.

core/integration/features/verified/closeout-2 partyies_lognormal.feature
Jiajia-Cui: replaced trader2 with auxiliary2

witgaw (changes requested):
Perhaps a name `closeout-lognormal.feature` would be appropriate? Don't think we really have 2 parties being closed out anymore.

witgaw (changes requested):
Everything looks good, it's just that the binary was accidentally added so needs to be removed before merging (cmd/vega/vega)

witgaw (approved):

PR #6231, opened 2022-09-20
6202 always check margin

### LINKED ISSUE #6202: Bug in margin calc when trader reduces their order size
please see the feature file in this [PR](https://github.com/vegaprotocol/vega/pull/6147)

party1 has reduced their order size from 20 to 2 (line 69), but there is no change in margin account and margin level; while party4 has doubled their order size from 20 to 40 (line 70), and the margin account and margin levels had been all doubled.

So the bug in margin calc only existing when trader reduces their order size

EVODelavega:
Closes #6202 

Always recalculate margins on order amendments, and release excess margins if the amendment was a reduction in order size (ie if party potentially reduced their potential position).

### DIFF
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -2676,11 +2676,10 @@ func (m *Market) amendOrder(
 	}
 
 	// from here these are the normal amendment
-	var priceIncrease, priceShift, sizeIncrease, sizeDecrease, expiryChange, timeInForceChange bool
+	var priceShift, sizeIncrease, sizeDecrease, expiryChange, timeInForceChange bool
 
 	if amendedOrder.Price.NEQ(existingOrder.Price) {
 		priceShift = true
-		priceIncrease = existingOrder.Price.LT(amendedOrder.Price)
 	}
 
 	if amendedOrder.Size > existingOrder.Size {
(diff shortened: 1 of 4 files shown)


EVODelavega:
Added that `GetPositionsByParty` function on positions engine in case we need to implement something similar to this for auctions, or we need to update passive affected orders/parties in some cases. I don't think we need that right now, but made sense to have something like that just in case

core/execution/market.go
```diff
@@ -2753,6 +2751,10 @@ func (m *Market) amendOrder(
 		// amended in place. Maybe a panic would be better
 		ret, err := m.orderAmendInPlace(existingOrder, amendedOrder)
 		if err == nil {
+			if sizeDecrease {
+				// ensure we release excess if party reduced the size of their order
+				m.recheckMargin(ctx, m.position.GetPositionsByParty(amendedOrder.Party))
+			}
```
jeremyletang, line 2757: Is that necessary? Wouldn't the call at line 2718 be enough?

EVODelavega: No, it's not enough. That call only checks for insufficient margin, but does not release excess. The reason for this (IIRC) is the margin check is used when repricing LP's and adding new orders. Releasing excess - especially when repricing LP's - leads to a lot of redundant transfers, and potentially excessive bond slashing (release -> bond account -> place next order -> take from bond + slash, etc...)

jeremyletang: hum interesting.

Jiajia-Cui (dismissed):
LGTM

party1 amend the order size from 20 to 2, and margin account moved from 19218 to 1922;
party4 amend the order size from 20 to 40, and margin account moved from 85366 to 170732

jeremyletang (approved):

PR #6239, opened 2022-09-20
feat: datanode snapshots

### LINKED ISSUE #5706: Data synchronisation when a NEW NODE joins the network
### Task Overview

In order to ensure that the data-node has all the correct information on the addition of a new node
We will ensure there is a data synchronisation on a new node joining the network

**Before Merging**

**After Merging**

ettec:
closes #5706 sort of, this is the part to create deterministic delta snapshots and the part to load up a datanode from this data.  The sharing of the data is a tbd


datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -9,6 +9,34 @@ create table blocks
...
+    hash          BYTEA                    NOT NULL
+);
+
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION update_last_block()
+    RETURNS TRIGGER
+    LANGUAGE PLPGSQL AS
+$$
+BEGIN
+    Delete from last_block;
```
pscott31, line 28: maybe add a one_row_check like chain then you can just update instead of insert/delete

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -168,13 +196,15 @@ CREATE TABLE orders_history (
...
-    vega_time_to      TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT 'infinity'
+    vega_time_to      TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT 'infinity',
+    PRIMARY KEY(vega_time, seq_num)
 );
 
 SELECT create_hypertable('orders_history', 'vega_time', chunk_time_interval => INTERVAL '1 day');
 CREATE INDEX ON orders_history (market_id, vega_time DESC);
 CREATE INDEX ON orders_history (party_id, vega_time DESC);
 CREATE INDEX ON orders_history (id, vega_time_to);
+CREATE INDEX ON orders_history (vega_time, seq_num);
```
pscott31, line 207: don't need if primary key

ettec, line 207: Remove this - got PK

datanode/snapshot/config.go
```diff
@@ -0,0 +1,24 @@
...
+	DatabaseSnapshotsPath string `long:"database-snapshot-path" description:"the snapshots path relative the database working directory"`
+}
+
+// NewDefaultConfig creates an instance of the package specific configuration, given a
+// pointer to a logger instance to be used for logging within the package.
+func NewDefaultConfig() Config {
+	return Config{
+		DatabaseSnapshotsPath:    "/snapshots",
+		RemoveSnapshotsOnStartup: false,
+		BlockInterval:            1000, // Roughly every 15 mins at 1 block per second
```
wwestgarth, line 21: I'm not really sure how/where these deltas will be published, but will it could it cause issues if different data-nodes have `BlockInterval` set to different values?

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+)
+
+type Meta struct {
+	ChainID                  string
+	HeightFrom               int64
+	HeightTo                 int64
+	CurrentStateSnapshotFile string
+	HistorySnapshotFile      string
+	WriteLockFile            string
```
wwestgarth, line 27: `WriteLockFile` is only used in a test. I think the test can work it out for itself based on what it passes into `CreateSnapshot`, and then we don't have this test-only field in production code.

datanode/snapshot/current_snapshot.go
```diff
@@ -0,0 +1,145 @@
...
+			tableCopySQL := fmt.Sprintf(`copy (select * from %s order by %s) to '%s'`, tableName,
+				meta.SortOrder, snapshotFile)
+			copySQL = append(copySQL, tableCopySQL)
+		}
+	}
+
+	return copySQL
+}
+
+func FileExists(snapshotsDir string, file string) (bool, error) {
```
wwestgarth, line 53: This is essentially [this](https://github.com/vegaprotocol/vega/blob/develop/libs/fs/fs.go#L39), so we should reuse.

datanode/snapshot/current_snapshot.go
```diff
@@ -0,0 +1,145 @@
...
+			return false, nil
+		}
+		return false, err
+	}
+
+	return true, nil
+}
+
+func GetCurrentStateSnapshots(snapshotsDir string) (string, map[int64]CurrentStateSnapshot, error) {
+	files, err := ioutil.ReadDir(snapshotsDir)
```
wwestgarth, line 65: `ioutils` is deprecated? https://pkg.go.dev/io/ioutil#ReadDir

pscott31 (commented):
Maybe move test data into 'testdata' folder?

datanode/snapshot/config.go
ettec: yes, very much, ideally this will need to a network parameter or similar to make sure they are in sync with other datanodes and core.  I've put this in the TBD column for now and left in config. But we will need to address this.

jeremyletang: let's use the existing network parameter that we use for the core for now? and then we'll figure if it needs to be changed? Having core and datanode take snapshots at the same time everytime shall make things simpler at first.

cmd/data-node/commands/start/node_pre.go
```diff
@@ -61,122 +58,71 @@ func (l *NodeCommand) persistentPre([]string) (err error) {
...
 			l.vegaPaths.StatePathFor(paths.DataNodeStorageHome))
 		if err != nil {
-			return fmt.Errorf("failed to start embedded postgres: %w", err)
+			return nil, fmt.Errorf("failed to start embedded postgres: %w", err)
 		}
 	}
 
+	hasVegaSchema, err := snapshot.HasVegaSchema(l.ctx, l.conf.SQLStore.ConnectionConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if database if empty: %w", err)
```
jeremyletang, line 96:
```suggestion
		return nil, fmt.Errorf("failed to check if database is empty: %w", err)
```

datanode/snapshot/compression.go
```diff
@@ -0,0 +1,142 @@
...
+func decompressAndUntarSnapshot(snapshotsDir string, snapshot snapshot) (databaseVersion int64, err error) {
+	decompressedFilesDestination := filepath.Join(snapshotsDir, snapshot.UncompressedDataDir())
+
+	err = os.Mkdir(decompressedFilesDestination, fs.ModePerm)
+	if err != nil {
+		return 0, fmt.Errorf("failed to make target diretory for uncompressed files %s: %w", decompressedFilesDestination, err)
+	}
+
+	compressedFilePath := filepath.Join(snapshotsDir, snapshot.CompressedFileName())
+	sourceFile, err := os.Open(compressedFilePath)
```
jeremyletang, line 95: Might be mistaken but can't see where this is closed.

wwestgarth:
When I run `vega datanode get-all-available-history` on a fresh `vega datanode init -f` setup I get:
```
failed to get all available history: failed to get current state snapshots:failed to get files in snapshot directory:open /Users/wwestgarth/Library/Application Support/vega/data-node/snapshots: no such file or directory
```
should init be creating that directory?

Also if I then make that directory myself I get the expected error message, but along with some logging which is a bit unexpected to be as default from a CLI tool:
```
2022-09-21T12:46:01.981+0100	INFO	cfgwatcher	config/watcher.go:90	config watcher started successfully	{"config": "/Users/wwestgarth/Library/Application Support/vega/data-node/config.toml"}
No history found and the datanode is currently empty
```

datanode/snapshot/config.go
```diff
@@ -0,0 +1,24 @@
...
+	PanicOnError  encoding.Bool `long:"panic-on-error" description:"true if service should panic on snapshot creation error"`
+
+	DatabaseSnapshotsPath string `long:"database-snapshot-path" description:"the snapshots path relative the database working directory"`
+}
+
+// NewDefaultConfig creates an instance of the package specific configuration, given a
+// pointer to a logger instance to be used for logging within the package.
+func NewDefaultConfig() Config {
+	return Config{
+		DatabaseSnapshotsPath:    "/snapshots",
```
jeremyletang, line 19: Shouldn't this be set in the paths library? cc @ValentinTrinque

datanode/snapshot/config.go
```diff
@@ -0,0 +1,24 @@
...
+		RemoveSnapshotsOnStartup: false,
+		BlockInterval:            1000, // Roughly every 15 mins at 1 block per second
+		Publish:                  false,
```
jeremyletang, line 22: Shouldn't this default to true? By default I'd expect all datanode to be "nice player" and publish every snapshots they do.

wwestgarth:
Can't help but feel that `get-all-available-history` and `load-all-available-history` should be subcommands of something, either their own snapshot section or under `postgres`. So you'd do like `vega datanode postgres load-all-available-history`

datanode/snapshot/current_snapshot.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			currentStateSnapshots[csSnapshot.Height] = *csSnapshot
+		}
+	}
+
+	return chainID, currentStateSnapshots, nil
+}
+
+func fromCurrentStateSnapshotFileName(fileName string) *CurrentStateSnapshot {
+	re := regexp.MustCompile("(.*)-(\\d+)-" + currentStateSnapshotIdentifier + ".tar.gz")
```
jeremyletang, line 104: Must compile will panic if there's an error. It's nice when you want something like a global var I suppose, here, we may want to have some kind of error handling instead?

datanode/snapshot/database_meta_data.go
```diff
@@ -0,0 +1,167 @@
...
+	dbVersion, err := goose.GetDBVersion(db)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get goose database version:%w", err)
+	}
+	return dbVersion, nil
+}
+
+func GetAllTableNames(ctx context.Context, conn *pgxpool.Pool) ([]string, error) {
+	tableNameRows, err := conn.Query(ctx, "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' and table_type = 'BASE TABLE' and"+
+		" table_name != 'goose_db_version' order by table_name")
```
jeremyletang, line 100: why is that 2 string concatenated here?

datanode/snapshot/database_meta_data.go
```diff
@@ -0,0 +1,167 @@
...
+		return false, fmt.Errorf("failed to get all table names:%w", err)
+	}
+
+	return len(tableNames) != 0, nil
+}
+
+func getTableSortOrders(ctx context.Context, conn *pgxpool.Pool) (map[string]string, error) {
+	var primaryKeyIndexes []IndexInfo
+	err := pgxscan.Select(ctx, conn, &primaryKeyIndexes,
+		`select tablename, Indexname, Indexdef from pg_indexes where schemaname ='public' and Indexname like '%_pkey' order by tablename`)
```
jeremyletang, line 135: I suppose this is correct, just weird that some of the selected column here are Capitalised.

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return Meta{}, fmt.Errorf("failed to set timezone to UTC:%w", err)
+	}
+
+	_, err = copyDataTx.Exec(ctx, "SET TRANSACTION ISOLATION LEVEL SERIALIZABLE")
+	if err != nil {
+		return Meta{}, fmt.Errorf("failed to set transaction isolation level to serilizable:%w", err)
+	}
+
+	writeLockFileName := filepath.Join(b.snapshotsPath, WriteLockFileName(chainID, toHeight))
```
jeremyletang, line 95: I used to use that in the path to create lock files. https://github.com/gofrs/flock

I realise it's really easy to do yourself, but using that help abstract away the details / filesystem stuff in the middle of the logic here and there.

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			b.logError(fmt.Errorf("failed to get md5 hash of snapshot:%w", err))
+		}
+
+		metrics.SetLastSnapshotRowcount(float64(rowsCopied))
+		metrics.SetLastSnapshotCurrentStateBytes(float64(compressedCurrentStateByteCount))
+		metrics.SetLastSnapshotHistoryBytes(float64(compressedHistoryByteCount))
+		metrics.SetLastSnapshotSeconds(time.Since(start).Seconds())
+
+		b.log.Infof("finished creating snapshot for chain:%s, height:%d, time taken:%s, rows copied:%d, total compressed data size:%d, md5 hash:%s", chainID, toHeight,
```
jeremyletang, line 164: could we use structured logging in general? we do everywhere, makes things easier to parse usually.

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+			time.Since(start), rowsCopied, compressedCurrentStateByteCount+compressedHistoryByteCount, snapshotHash)
+	}()
+
+	return meta, nil
+}
+
+func (b *Service) logError(err error) {
+	if b.config.PanicOnError {
+		panic(err)
```
jeremyletang, line 173: Could we use b.log.Panic("message", logging.Error(err)) instead?

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	b.log.Errorf(":%w", err)
```
jeremyletang, line 176: b.log.Error("message", logging.Error(err))

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func GetHistoryMd5Hash(log *logging.Logger, snapshot Meta) (string, error) {
+	snapshotHash := md5.New()
```
jeremyletang, line 180: Why md5? I know there's no security issue here really or anything, but could we just use libs/crypto.Hash intead which is sha3_256?

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err := file.Close()
+		if err != nil {
+			log.Errorf("failed to close file %s after hashing:%w", filePath, err)
+		}
+	}()
+	_, err = io.Copy(hash, file)
```
jeremyletang, line 217: Have we tested that with massive files? e.g: 200gb?

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,257 @@
...
+
+	return nil
+}
+
+func WriteLockFileName(chainID string, height int64) string {
+	return fmt.Sprintf("%s-%d.writelock", chainID, height)
+}
+
+func (b *Service) createUncompressedDataDirectory(uncompressedDataDir string) (string, error) {
+	dir := filepath.Join(b.snapshotsPath, uncompressedDataDir)
```
jeremyletang, line 230: For all things path related, would be good to have those things covered with the paths libs I suppose. cc @ValentinTrinque

wwestgarth:
I tried running it locally with `vega datanode postgres run` in one terminal and kicking off the data-node in another reading in the intergration test's event-file with the snapshot interval set at 10 blocks, and it hangs for me (I added some logs):
```
2022-09-21T13:04:34.160+0100	INFO	snapshot	snapshot/service.go:70	creating snapshot for chain chain-ze7mNt, from height 0, to height 10
2022-09-21T13:04:34.179+0100	INFO	snapshot	snapshot/service_create_snapshot.go:50	connected
2022-09-21T13:04:34.230+0100	INFO	snapshot	snapshot/service_create_snapshot.go:57	new meta
2022-09-21T13:04:34.230+0100	INFO	snapshot	snapshot/service_create_snapshot.go:61	stuff
2022-09-21T13:04:34.237+0100	INFO	snapshot	snapshot/service_create_snapshot.go:84	executing SET TIMEZONE
```

I also couldn't exit the process and had to kill it manually.

datanode/snapshot/service_load_snapshot.go
```diff
@@ -0,0 +1,433 @@
...
+			}
+		}
+
+		b.log.Infof("snapshotting all datanode data")
+		meta, err := b.CreateSnapshot(ctx, chainID, oldestHistoryBlock.Height, lastBlock.Height)
+		if err != nil {
+			return 0, 0, fmt.Errorf("failed to snapshot all datanode data: %w", err)
+		}
+
+		b.log.Infof("snapshotted all datanode data from height %d to %d", meta.HeightFrom, meta.HeightTo)
```
jeremyletang, line 60: structured logging here too.

datanode/snapshot/service_load_snapshot.go
```diff
@@ -0,0 +1,433 @@
...
+	allSnapshotData := appendCurrentStateSnapshotOntoHistory(contiguousHistory, currentStateSnapshot)
+
+	start := time.Now()
+	b.log.Infof("restoring data node from snapshot data %+q", allSnapshotData)
+
+	b.log.Infof("creating database")
+	err = sqlstore.RecreateVegaDatabase(ctx, b.log, b.connConfig)
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to create vega database: %w", err)
+	}
```
jeremyletang, line 72: I do favour this form usually when the err is used only in one context to scope it
```suggestion
	if err = sqlstore.RecreateVegaDatabase(ctx, b.log, b.connConfig); if err != nil {
		return 0, 0, fmt.Errorf("failed to create vega database: %w", err)
	}
```

jeremyletang (commented):
Phils going pick up the Balance_Tests after merge.

There's still a bunch of commented test to fix.

datanode/snapshot/database_meta_data.go
ettec: to split across the IDE, my monitors not that wide :)  do we care?

datanode/snapshot/service_create_snapshot.go
ettec: I think the semantics are slightly different, in this case the lock file is temporary thing exists while the data is being created, once its removed it basically means the write is complete and the data is fully dumped, after that point it is removed. In the case of flock it states:
`This will release the lock and close the underlying file descriptor.
// It will not remove the file from disk, that's up to your application.`

Which sort of defeats the point of using it.

datanode/snapshot/service_create_snapshot.go
ettec: on reflection I think the panic on error flag is unnecessary, will just log error

datanode/snapshot/service_create_snapshot.go
wwestgarth: I think `flock` is what we want here because it also locks the file from edits and more importantly being deleted by any other process while locked. The fact `flock` leaves the the file in existence afterwards is irrelevant because its not the existence of the file that means its "locked".

I'm guessing this lock just exists to prevent listing an "in progress" snapshot when you if you call `get-all-available-history` when a snapshot is still in progress? I'm wondering if we should just cut the complexity and have a single file lock for "data node is still running" because we'll need something to prevent someone calling `load-all-available-history` if the data-node is running at all, regardless of is there is a snapshot in progress.

wwestgarth:
>Using postgres, is as much as possible, to be kept as an impl detail (in practice, not so hidden), the datanode shouldn't refer to this in the commands IMO. P.s. fixed the spurios config line output.
> 

Ok, well I still think it should have a subcommand though just to group the functionality and reduce the need for such long names. For example:
```
vega datanode snapshots --list
vega datanode snapshots --load
```

datanode/snapshot/service_create_snapshot.go
ettec: regarding the load all available, that ensures the datanode is dead before proceeding, it kills all connections to the database before proceeding to prevent accidental concurrent running.

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,269 @@
...
+
+	writeLockFileName := filepath.Join(b.snapshotsPath, WriteLockFileName(chainID, toHeight))
+	_, err = os.Create(writeLockFileName)
+	if err != nil {
+		return Meta{}, fmt.Errorf("failed to create write lock file:%w", err)
+	}
+
+	if async {
+		go func() {
+			err = b.snapshotData(ctx, copyDataTx, currentSnapshot, dbMetaData, historySnapshot, currentDataDir,
```
wwestgarth, line 105: I think we need to do more here than just kick off this all in a go-routine. If someone sigterms the data-node process while a snapshot is still in progress we need to wait for the snapshot to finish before stopping the process, otherwise we'll have a half-baked snapshot hanging around.

cmd/data-node/commands/start/node_pre.go
```diff
@@ -61,122 +58,71 @@ func (l *NodeCommand) persistentPre([]string) (err error) {
...
 	}
 
+	hasVegaSchema, err := snapshot.HasVegaSchema(l.ctx, l.conf.SQLStore.ConnectionConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if database is empty: %w", err)
+	}
+
+	// If it's an empty database, recreate it with correct locale settings
+	if !hasVegaSchema {
+		sqlstore.RecreateVegaDatabase(l.ctx, l.Log, l.conf.SQLStore.ConnectionConfig)
```
wwestgarth, line 101: Should probably handle the error that could be returned  by `RecreateVegaDatabase()`.

datanode/snapshot/service.go
```diff
@@ -0,0 +1,186 @@
...
+		log:           log.Named("snapshot"),
+		config:        config,
+		connConfig:    connConfig,
+		snapshotsPath: snapshotsPath,
+		blockStore:    blockStore,
+		chainService:  chainService,
+	}
+
+	if service.snapshotsEnabled() {
+		_, err := os.Stat(service.snapshotsPath)
```
wwestgarth, line 40: Shouldn't we be looking in `config.DatabaseSnapshotsPath` if its set? Same when we in the `if config.RemoveSnapshotsOnStartup {}` statement.

cmd/data-node/commands/history/show.go
```diff
@@ -0,0 +1,78 @@
...
+	snapshotsPath := vegaPaths.StatePathFor(paths.DataNodeSnapshotHome)
+
+	connSource, err := sqlstore.NewTransactionalConnectionSource(log, cmd.Config.SQLStore.ConnectionConfig)
+	if err != nil {
+		return fmt.Errorf("unable to connect to database: %w", err)
+	}
+
+	blockStore := sqlstore.NewBlocks(connSource)
+
+	err = cmd.showHistory(blockStore, snapshotsPath)
```
wwestgarth, line 44: Same here, we're not honouring the database snapshot path option.

cmd/data-node/commands/load_all_available_history.go
```diff
@@ -0,0 +1,70 @@
+package commands
```
wwestgarth, line 1: I think this whole file can be removed now that its been moved into `cmd/data-node/commands/history/load.go`?

datanode/snapshot/service_create_snapshot.go
```diff
@@ -0,0 +1,269 @@
...
+	compressedCurrentStateByteCount, err := tarAndCompressSnapshot(b.snapshotsPath, currentSnapshot, dbMetaData.DatabaseVersion)
+	if err != nil {
+		return fmt.Errorf("failed to compress snapshot:%w", err)
+	}
+	b.log.Infof("compressed current state snapshot data size: %dB", compressedCurrentStateByteCount)
+
+	b.log.Infof("compressing history data")
+	compressedHistoryByteCount, err := tarAndCompressSnapshot(b.snapshotsPath, historySnapshot, dbMetaData.DatabaseVersion)
+	if err != nil {
+		return fmt.Errorf("failed to compress history snapshot:%w", err)
```
wwestgarth, line 155: Whats the consequence if for some reason a particular database snapshot does fail? Do we just miss a slice of history and thats ok? I just no in core if something goes wrong we have to panic because its part of consensus but that might not be true here.

datanode/snapshot/config.go
wwestgarth: I think it should be in the paths library too, so when you do `vega paths list` you can see the default location for where things get dumped. It then opens up us having an absolute path here, which is a bit more flexible.

datanode/snapshot/service.go
ettec: the database snapshot path is different from the nodes snapshot path (which is created using the Paths) and will depend on where and how the database is run as it is a path relative to the DB, and hence in the config, defaulted to the most usual case.

datanode/snapshot/service_create_snapshot.go
ettec: we don't have massive files, only deltas.

datanode/snapshot/service_create_snapshot.go
ettec: we could, but its just for logging purposed to be able to easily confirm nodes are producing the same data, the hash algo doesn't matter

datanode/snapshot/service_create_snapshot.go
ettec: hmm, the problem is losing a history slice across all nodes could make the history unrecoverable if we cannot get hold of that slice then only history after that point would be loadable, this would be mitigated if we have a cmd to create a slice which I was considering adding at one point but seemed overkill.  This is probably the better thing to do than take down the whole network on some issue that causes a cross network snapshot fail, that way we could at least manually dig our way out of that hole.  I'll leave adding this for another PR if ok.

datanode/snapshot/config.go
ettec: its a work in progress, once the publisher is done, yes, true would be the default, as the moment I've just hooked the publisher in, its not doing anything yet.

datanode/snapshot/config.go
ettec: related to later comment, the db snapshot path is a path relative to where the DB is run, different to the nodes snapshot path, the nodes snapshot path is part of the paths list.  The db path will depend on where and how the DB is run, its sets to the most common value in config, but we don't mandate that.

datanode/snapshot/current_snapshot.go
ettec: this was fixed (and is still), what code do you see?

datanode/snapshot/service_create_snapshot.go
ettec: it wasn't being read any more, I've removed it.

datanode/snapshot/current_snapshot.go
wwestgarth: Ah you're right, my grep picked up the one in `func fromHistoryFileName()` which should probably be changed too.

ettec: sorted that one as well

datanode/snapshot/service_create_snapshot.go
wwestgarth: If it makes no difference which we use, then it makes sense to use `libs/crypto.Hash()` and keep the algorithm consistent across the rest of the code.

wwestgarth:
Logs from the CI run aren't looking too hot: https://jenkins.ops.vega.xyz/job/common/job/system-tests-wrapper/10756/artifact/testnet/logs/testnet-nodeset-full-2-full/data-node-2.stdout-2022-09-26T15%3A13%3A06Z.log

```
2022-09-26T15:13:24.365Z	INFO	snapshot	snapshot/service_create_snapshot.go:36	creating snapshot for chain test-chain-TkmxfG, from height 25, to height 36
2022-09-26T15:13:24.448Z	INFO	snapshot	snapshot/service_create_snapshot.go:89	copying all table data....
2022-09-26T15:13:24.451Z	ERROR	snapshot	runtime/asm_amd64.s:1594	failed to snapshot data:failed to rollback snapshot transaction:%!s(<nil>), rollback caused by:failed to execute copy copy (select * from erc20_multisig_signer_events order by id) to '/snapshots/test-chain-TkmxfG-36-currentstatesnapshot/erc20_multisig_signer_events': ERROR: could not open file "/snapshots/test-chain-TkmxfG-36-currentstatesnapshot/erc20_multisig_signer_events" for writing: No such file or directory (SQLSTATE 58P01)
```

datanode/snapshot/config.go
ettec: yeah, they have to be on the same machine unless using some network disk etc

datanode/snapshot/service_create_snapshot.go
ettec: the eventual hash of the file will be a IPFS Content ID which will be different to crypto hash

wwestgarth (approved):

PR #6240, opened 2022-09-20
Refactor TransferResponses into LedgerMovements

### LINKED ISSUE #5925: Tidy up LedgerEntry messages in TransferResponse events
Currently there is a `reference` field and a `type` field which are both strings and and in a random event I got from a system test run look  backwards to me. 

image

Looking at a few in the DB seems to show inconsistent usage:

image

In particular this lower case 'settlement' seems a bit odd. It would be good to clarify the purpose of these fields and make their usage consistent.

`type` should probably be an ENUM rather than a string.

Bonus Task:

In the datanode we currently have to decipher which accounts `(party, asset, market, type)` a ledger entry is referring to by decoding the `accountId` which is a string concatenation with special characters for no market/system owner and this is rather fragile and relies on quirks like party IDs being 64 characters long. It would be better to have explicit  `(optional) party_id`, `(optional) market_id`, `(required) asset_id`, `(required) account_type` fields

jeremyletang:
Rename TransferResponse into LedgerMovement.
Use AccountDetails type instead of account IDs or full account struct. 

Barely touched the datanode part, just enough so it can work and do not annoy too much with other PR in progress.

close #5925


pscott31 (dismissed):
Looks very sensible to me!

jeremyletang:
```
{
"system-tests": "feature/transfer-response-to-ledger-movements"
}
```

witgaw (dismissed):
Verified test change looks legit.

AnExsomnis:
IMO good to go!

AnExsomnis (dismissed):

witgaw (dismissed):

core/types/collateral.go
```diff
@@ -28,6 +80,15 @@ type Account struct {
 	Type     AccountType
 }
 
+func (a Account) ToDetails() *AD {
+	return &AD{
```
ValentinTrinque, line 84: Can we name AD as AccountDetails ?

jeremyletang: We can. I just liked better AD as it’s shorted and I thought I’d use it more in the codebase / it looks slightly like ID as well, which was it relates too

jeremyletang: addressed.

ValentinTrinque (approved):

PR #6246, opened 2022-09-21
fix: error if signature is invalid

ze97286:


karlem (dismissed):

jeremyletang (dismissed):

wwestgarth:
I was just looking at the commit that introduce this bug, and a noticed another subtle change that may or may not have been intentional.

Before we would verify the sigantures in only in `DecodeTx()` where we also validate all the other fields of the transaction, and in the tendermint client [here](https://github.com/vegaprotocol/vega/blob/develop/core/blockchain/client.go#L129) we would only check the fields and _not_ verify the signature.

Now we do it in both places? Do we know if thats intentional or are we now verifying twice as much for no reason?

ze97286:
> I was just looking at the commit that introduce this bug to understand why it wasn't caught, and a noticed another subtle change that may or may not have been intentional.
> 
> Before we would verify the sigantures only in `core/processor/DecodeTx()` where we also validate all the other fields of the transaction, and in the tendermint client [here](https://github.com/vegaprotocol/vega/blob/develop/core/blockchain/client.go#L129) we would only check the fields and _not_ verify the signature.
> 
> Now we do it in both places? Do we know if thats intentional or are we now verifying twice as much for no reason?

DecodeTx must do all verification. I don't think there's much point in calling CheckTransaction from SubmitTransactionSync because checkTx is called synchronously and do the same check so it seems redundant and does not add to security.

jeremyletang (approved):

PR #6250, opened 2022-09-21
feat: add feature test and AC for position resolution

Jiajia-Cui:


core/integration/features/verified/verified-positions-resolution-5.feature
```diff
@@ -0,0 +1,133 @@
...
+
+    And the markets:
+      | id        | quote name | asset | risk model                | margin calculator   | auction duration | fees         | price monitoring  | oracle config          |
+      | ETH/DEC19 | ETH        | USD   | lognormal-risk-model-fish | margin-calculator-1 | 1                | default-none | default-none      | default-eth-for-future |
+
+    And the following network parameters are set:
+      | name                           | value |
+      | market.auction.minimumDuration | 1     |
+
+    Scenario: using lognormal risk model, set "designatedLooser" closeout while the position of "designatedLooser" is not fully covered by orders on the order book; 0012-POSR-002, 0012-POSR-005
```
davidsiska-vega: I don't really see how we're testing 0012-POSR-005 but that can be because the AC isn't worded brilliantly. It says
```
Non-distressed traders who trade with the network because their open orders are hit during the close out trade have their positions settled correctly.
``` 
I'd say that this test checks that the positions are *updated* (not settled) correctly. 

If we want to claim *settled* then we should either
a) do a mark price update and check that we're getting correct mark-to-market settlement or 
b) send in a final settlement price and check that everyone gets the correct cashflow (and any potential shortfall comes from insurance pool). 

Whichever you end up doing here, it would be good to update the AC in the spec to be more precise.

davidsiska-vega (changes requested):

davidsiska-vega (approved):

PR #6253, opened 2022-09-21
Prepend chainID to input data only when signing the transaction

### LINKED ISSUE #6260: Prepend chainID to input data only when signing the transaction
Today, we append the chain ID to the input data and sending with the transaction.

That's adding unnecessary bytes to the chain.

A better strategy would be to append the chain ID only during to the transaction signature.

During verification the node just need to append the chain ID to the input data to verify the signature.

With this mechanism, the chain ID marks the transaction without taking anymore space on the blockchain.

jeremyletang:
Close #6260


ValentinTrinque:
```
{
   "system-tests": "fix_chain_id"
}
```

ValentinTrinque (approved):

PR #6262, opened 2022-09-21
fix: positions should be zero when market is settled

### LINKED ISSUE #5998: [graphql] Positions should not have non-zero unrealisedPNL or openVolume

guoguojin:
Closes #5998


datanode/entities/position.go
```diff
@@ -111,6 +117,17 @@ func (p *Position) UpdateWithSettleDestressed(e settleDestressed) {
 	p.TxHash = TxHash(e.TxHash())
 }
 
+func (p *Position) UpdateWithSettleMarket(e settleMarket) {
+	markPriceDec := num.DecimalFromUint(e.SettledPrice())
+	openVolumeDec := num.DecimalFromInt64(p.OpenVolume)
+
+	unrealisedPnl := openVolumeDec.Mul(markPriceDec.Sub(p.AverageEntryPrice)).Div(e.PositionFactor())
+	p.RealisedPnl = p.RealisedPnl.Add(unrealisedPnl)
+	p.UnrealisedPnl = decimal.Zero
```
EVODelavega, line 126: Not a fan of this, TBH... Of course we have `type Decimal = decimal.Decimal` in the `num` package, but if we ever were to swap out the underlying type, having the underlying type bleed in to the code base would be an issue

jeremyletang: that's on the datanode though, it initially started with using the decimal package directly.

core/execution/market.go
```diff
@@ -3232,6 +3232,7 @@ func (m *Market) settlementPriceWithLock(ctx context.Context) {
 
 			// send the market data with all updated stuff
 			m.broker.Send(events.NewMarketDataEvent(ctx, m.GetMarketData()))
+			m.broker.Send(events.NewMarketSettled(ctx, m.GetID(), m.timeService.GetTimeNow().UnixNano(), m.markPrice, m.positionFactor))
 		} else {
```
EVODelavega, line 3236: nit-pick: this `else` seems pointless. Return early, and remove the `else`

datanode/entities/position.go
guoguojin: Thanks, I've updated the other places it was used in the other positions methods too with this commit https://github.com/vegaprotocol/vega/pull/6262/commits/dd3080400fa705880944db06480782ac0a45ee6b. It'll be used a lot in other places in data-node so maybe we should raise a low priority ticket to clean that up at some point.

core/execution/market.go
guoguojin: Addressed with this commit https://github.com/vegaprotocol/vega/pull/6262/commits/dd3080400fa705880944db06480782ac0a45ee6b

EVODelavega (approved):

PR #6280, opened 2022-09-23
feat: first version of block explorer backend

### LINKED ISSUE #6163: Draft: New Tendermint APIs
# Feature Overview

**In order to** Support the current block explorer
**We will** Develop APIs to repalce the current tendermint APIs
**So that** We can deliver a block explorer and use a SQL indexer (which does not support the TM REST API)

## Specs

Current Endpoints used from the TM RPC API:

* Paginate list of blocks (/blockchain)
* Get block by id (/block?height=${block})
* Genesis data (/genesis)
* Paginated list of transactions (N/A tx_search crashed nodes so we elected not to use it)
* Unconfirmed transactions (/unconfirmed_txs)
* Transaction by hash (/tx?hash=${txHash})
* Validators (/validators) Returns different data to nodes query, could add this to existing nodes query.

The chain explorer itself currently gives us back decoded data from the following endpoints:

* Transaction by hash (/tx?hash=${txHash})
* Get block by id (/block?height=${block})

We also use the tendermint websocket API:
* Websocket (/websocket)

As a result of these changes, we may need to update the block 

# Tasks
A checklist of the tasks that are needed to develop the feature and meet the acceptance criteria and feature test scenarios. Ideally, tasks would reflect the issues/pull requests likely to be created when developing the feature. 
- [ ]
- [ ]

# Acceptance Criteria
A list of criteria (aim for 3!) that have to be met for this feature to be accepted as 'done' by the product owner.

- It is possible to

Acceptance Criteria and Feature Test Scenarios can, in some cases, be closely related. If acceptance criteria become fully covered by feature test scenarios they can be removed leaving on acceptance criteria that can't be directly or fully proven with tests.

# Test Scenarios
Detailed scenarios (1-3!) that can be executed as feature tests to verify that the feature has been implemented as expected. We use the follow format:

GIVEN (setup/context) 
WHEN (action) 
See [here](https://github.com/vegaprotocol/vega/tree/develop/integration/) for more format information and examples.

# Additional Details (optional)

# Examples (optional)
Code snippets from the spec for reference

# Definition of Done

**Before Merging**

**After Merging**

pscott31:
Makes a good start at closing #6163 

There's a bit more to do but the front end guys are chomping at the bit


blockexplorer/store/config.go
```diff
@@ -0,0 +1,36 @@
...
+
+var namedLogger = "postgres.store"
+
+type Config struct {
+	Postgres config.PostgresConnection `group:"database" namespace:"postgres"`
+}
+
+func NewDefaultConfig() Config {
+	return Config{
+		Postgres: config.PostgresConnection{
```
ze97286, line 27: we'd need a new database every time we restart from a checkpoint right?

ze97286:
looks good, just a general question, why a separate app vs inside datanode? in my mind the pro is that it can evolve into a full block explorer back end serving all of the end points used by the block explorer UI (ours and others'). On the other hand doing it inside datanode would have saved 95% of the code in this PR. Were there other considerations?

pscott31:
Well my rationale for doing it as a separate app was basically that it was a sort of orthogonal project to datanode. It doesn't need our database and stuff, or a connection to core. All it looks at is the Postgres DB that Tendermint writes to so it seemed cleaner to have it as its own thing.

- In fact it's sort of orthogonal to vega - the only vega-specific thing it does use our proto definitions to decode the transaction payload into JSON so I thought maybe it might one day be something we could relase as a generic Tendermint block explorer thing.

- If it crashes or slows down or whatever it won't take down data node and potentially the network.

- Like this it's super easy to iterate on it as we don't need new datanode/vega releases necessarily and network restarts and all that when the front end want changes to this API (as we're hoping to iterate quite quickly on this) 

- We had some troubles with networks coming down because of boxes slowing down so I thought it would be nice to be able to run this on a machine where data-node wasn't

- That said I have added it in as a 'vega' subcommand so it is baked in in that sense, in the end users won't have to download/run a separate binary once we're done developing.

- I'm not 100% convinced that tendermint can write to postgres fast enough to make this viable and at least this way it's easy to unpick if we have to come up with another solution

Also when I started I didn't envisage as much boilerplate to set up the app and whatnot, had I known then what I know now I might not have bothered but we're here now.

blockexplorer/store/config.go
pscott31: yeah, that is a good point actually I am not sure what tendermint will do actually when it starts a new chain when there is stuff already in the database from an old one. I suspect just write it alongside what is there, so I probably need to add an 'unsafe-reset-all' command or something - and/or make chainID a mandatory field when querying.

ze97286:
@pscott31 to me it all makes sense. I like the idea of this growing into a backend where developers can build block explorers on top of. 
Regarding postgres indexer, I agree with the first part of the sentence - I also believe it won't be fast enough but I think the solution would be to have this back end scan the data from tendermint and store it efficiently

ze97286 (approved):

PR #6281, opened 2022-09-23
Positions potential size bug lookout

### LINKED ISSUE #6278: Crasher: panic: trade with a potential buy position < to the trade size
Caused by running RL training on vega-market-sim for a few hours (iteration 138). 

[vega.err.txt](https://github.com/vegaprotocol/vega/files/9628955/vega.err.txt)

I also have `vega.out` but it's 14 GB so I am not uploading it but here are the last 10000 lines: 
[vega.out.tail.10000.txt](https://github.com/vegaprotocol/vega/files/9629056/vega.out.tail.10000.txt)

Data node reported no errors.

[data_node.out.txt](https://github.com/vegaprotocol/vega/files/9628957/data_node.out.txt)

jeremyletang:
close #6278

### DIFF
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -2932,6 +2932,7 @@ func (m *Market) orderAmendInPlace(
 		// panic here, no good reason for a failure at this point
 		m.log.Panic("Failure after amend order from matching engine (amend-in-place)",
 			logging.OrderWithTag(*amendOrder, "new-order"),
+			logging.OrderWithTag(*originalOrder, "old-order"),
 			logging.Error(err))
 	}
 
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -2654,12 +2654,7 @@ func (m *Market) amendOrder(
 			}
 			// we were parked, need to unpark
 			m.peggedOrders.Unpark(amendedOrder.ID)
-			orderConf, orderUpdts, err := m.submitValidatedOrder(ctx, amendedOrder)
-			if err != nil {
-				// If we cannot submit a new order then the amend has failed, return the error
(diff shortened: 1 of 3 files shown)


core/execution/market.go
```diff
@@ -2725,39 +2711,31 @@ func (m *Market) amendOrder(
...
+		return ret, nil, nil
 	}
 
 	if m.log.GetLevel() == logging.DebugLevel {
 		m.log.Debug("Order amendment not allowed", logging.Order(*existingOrder))
 	}
+
+	// for some reason, we got here without doing anything to the order,
+	// or we would have returned earlier.
+	// should we panic, as that doesn't seems right,
```
peterbarrow, line 2735: I'd panic as we have no idea why we should be here

peterbarrow (dismissed):

peterbarrow (dismissed):

core/matching/side.go
```diff
@@ -160,7 +160,7 @@ func (s *OrderBookSide) amendOrder(orderAmend *types.Order) (uint64, error) {
 		return 0, types.ErrOrderAmendFailure
 	}
 
-	if oldOrder.Size != orderAmend.Size &&
+	if oldOrder.Size < orderAmend.Size &&
 		oldOrder.Remaining < orderAmend.Size {
```
peterbarrow, line 164: Does this line need to be here? I think if line 163 is true we are in trouble because we cannot update in place.

jeremyletang: yea indeed, that's the point. If order the newOrder have an increased side, that's a bug. and should never happend.

peterbarrow (approved):

PR #6284, opened 2022-09-23
Add commit hash to the version build tags

### LINKED ISSUE #6283: Add commit hash to verison if is development version
Some developers get confused by the wallet not raising compatibility issues between different development version of the software.

To strengthen the version check, we could add the commit hash (first 8 characters) behind the `+dev` build tag. 

It would result in:

| Version | Hash | Before | After |
| -- | -- | -- | -- |
| 0.56.0 | 8sdf6df8 | 0.56.0 | 0.56.0 |
| 0.56.0+dev | 8sdf6df8 | 0.56.0+dev |  0.56.0+dev.8sdf6df8 |
| 0.56.0+dev | 8sdf6df8-modified |  0.56.0+dev | 0.56.0+dev.8sdf6df8.modified |

This will help the version verification to be extremely pedantic on the version to use against development environment.

ValentinTrinque:
Close #6283

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -14,13 +14,12 @@ package version
 
 import (
 	"runtime/debug"
-
-	"github.com/blang/semver"
+	"strings"
 )
 
 var (
-	cLIVersionHash = ""
-	cLIVersion     = "v0.56.0+dev"
+	cliVersionHash = ""
+	cliVersion     = "v0.56.0+dev"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


version/version.go
```diff
@@ -29,25 +28,45 @@ func init() {
...
+		cliVersionHash += "-modified"
 	}
 }
 
-func SemverGet() semver.Version {
-	return semver.MustParse(cLIVersion[1:])
-}
-
+// Get returns the version of the software. If the version happens to be a
+// development version, and only if, the first 8 characters of the git hash, is
```
guoguojin, line 43: I think the sentence:

```
... If the version happens to be a 
development version, and only if, ....
```

reads better if you changed it to:

```
When the version is a development version
```

PR #6285, opened 2022-09-23
fix: node signature connection errors

guoguojin:
Closes #6230 

This fix includes:
 - nodeSignatureConnection should return an empty list instead of null
 - nodeSignatureConnection should return an appropriate message when an invalid resource ID in given.

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -2218,7 +2218,7 @@ func (t *tradingDataServiceV2) ListNodeSignatures(ctx context.Context, req *v2.L
 
 	sigs, pageInfo, err := t.notaryService.GetByResourceID(ctx, req.Id, pagination)
 	if err != nil {
-		return nil, apiError(codes.NotFound, err)
+		return nil, fmt.Errorf("could not retrieve resource: %w", err)
 	}
 
 	edges, err := makeEdges[*v2.NodeSignatureEdge](sigs)
--- a/datanode/gateway/graphql/schema.graphql
+++ b/datanode/gateway/graphql/schema.graphql
@@ -4333,7 +4333,7 @@ type NodeSignatureEdge {
 "Connection type for retrieving cursor-based paginated node signature information"
 type NodeSignaturesConnection {
   "List of node signatures available for the connection"
-  edges: [NodeSignatureEdge]!
+  edges: [NodeSignatureEdge!]!
   "Page information for the connection"
   pageInfo: PageInfo!
 }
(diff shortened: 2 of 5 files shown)


wwestgarth (dismissed):

datanode/api/trading_data_v2.go
```diff
@@ -2218,7 +2218,7 @@ func (t *tradingDataServiceV2) ListNodeSignatures(ctx context.Context, req *v2.L
 
 	sigs, pageInfo, err := t.notaryService.GetByResourceID(ctx, req.Id, pagination)
 	if err != nil {
-		return nil, apiError(codes.NotFound, err)
+		return nil, fmt.Errorf("could not retrieve resource: %w", err)
```
edd, line 2221: I still don’t think this is correct, and isn’t what I said in DMs. It is not necessarily an error that there is no result for a resource request, so no error should be passed back. An empty list is sufficient.

What might be an error is if the resource ID could never match (e.g. if I search for ‘🍝’), in which case an error specifying that the ID is an invalid format *could* be an error.

edd (changes requested):

datanode/api/trading_data_v2.go
guoguojin: The resource ID checking takes place in the GetByResourceID method as it handles the conversion from string to resource ID, and returns an error if an invalid resource ID is provided. If you pass a valid Hex ID and no results are found, then you will not get an error, just an empty list. 

GetByResourceID has to return an error and it needs to be handled here because there are other database issues that may occur that could return an error (e.g. connectivity), and I believe it's appropriate for the API to return those errors which this should do.

edd: Thanks for taking the time to explain it to me @guoguojin - I got too big for my boots and tried to review some Go.

edd (dismissed):
Now that my misunderstanding of the Go is corrected, this looks great!

datanode/gateway/graphql/schema.graphql
```diff
@@ -4333,7 +4333,7 @@ type NodeSignatureEdge {
 "Connection type for retrieving cursor-based paginated node signature information"
 type NodeSignaturesConnection {
   "List of node signatures available for the connection"
-  edges: [NodeSignatureEdge]
+  edges: [NodeSignatureEdge]!
```
mattrussell36, line 4336: is it possible for this this be `edges: [NodeSignatureEdge!]!` ?

guoguojin: Sure, I'll make those updates for you.

mattrussell36 (approved):

wwestgarth (approved):

PR #6300, opened 2022-09-26
6291 use clef address from registry

### LINKED ISSUE #6291: Validators without clef wallets fail to start after checkpoint restore when clef used by other validators
test test_ethereum_key_rotation / branch feat_clef_ethereum_key_rotation

Steps to reproduce

- Setup network with 5 validators and two clef wallets on nodes 3/4 (test uses capsule_config_network_infra.hcl)
- Perform eth key rotation
- confirm events/multisig updates/datanode has new key for node
- Produce checkpoint
- Restore from checkpoint
- All nodes are restored apart from validator 1
logs 2022-09-23T17:36:32.657+0100	INFO	root	node/node.go:95	starting vega	{"version": "v0.56.0+dev", "commit-hash": "a7ce1fa22b1bd301753367e4798141c6ffc7ecb0-modified"}
- Validator 1 fails - stderr - couldn't get node wallets: failed to dial Clef daemon: dial unix: missing address

jenkins run demonstrating
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/10461/pipeline/293

wwestgarth:
closes #6291 

The `Nodewallet.ETH` section of core's config file has been removed, and the clef-address that you need to provide to `vega nodewallet import` and `vega nodewallet generate` commands and now passed in as direct options. As a result we now always use the clef-address that we save in the registry when a wallet is imported/generated and no longer read from the config file on startup. This means if someone wants to change the clef-address they will need to reimport which requires them to input the nodewallet passphrase.

A full system test run with these changes and the necessary capsule changes are here:
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/10150/pipeline

vegacapsule PR: https://github.com/vegaprotocol/vegacapsule/pull/274


cmd/vega/nodewallet/import.go
```diff
@@ -60,6 +64,10 @@ func (opts *importCmd) Execute(_ []string) error {
 	log := logging.NewLoggerFromConfig(logging.NewDefaultConfig())
 	defer log.AtExit()
 
+	if (opts.EthereumClefAccount != "") != (opts.EthereumClefAddress != "") {
+		return ErrClefOptionMissing
+	}
```
jeremyletang, line 69: Just to confirm, you are expecting that both clef settings are either specified, or not at all right?

wwestgarth: Yup, thats the one.

jeremyletang (commented):
Looks good. and much cleaner to not have the clef address both in the config and the nodewallet.

jeremyletang (approved):

PR #6303, opened 2022-09-26
fix: deposits returns empty list when no party is provided

### LINKED ISSUE #6273: Requesting list of all deposits for a Vega network returns empty list in data-node v2 REST API
If a user/developer attempts to query the data-node v2 APIs with the following query it returns an empty list:

`https://api.n08.testnet.vega.xyz/api/v2/deposits`

e.g. 

```
{
  "deposits": {
    "edges": [
    ],
    "pageInfo": {
      "hasNextPage": false,
      "hasPreviousPage": false,
      "startCursor": "",
      "endCursor": ""
    }
  }
}
```

The REST documentation gives the impression that the partyId filter is optional and that you can get a list (1000 paged) of all deposits (although this was not available in the v1 API). User is not presented with an error, we feel like this is a legitimate use case for the API - user might want to display all deposits on the network (or latest deposits etc)

If user queries for deposits for a partyId it does return deposits list in v2 API
If user queries for a deposit for valid deposit id in /deposit/id it does return the correct data in v2 API

guoguojin:
Closes #6273

### DIFF
--- a/datanode/gateway/graphql/schema.graphql
+++ b/datanode/gateway/graphql/schema.graphql
@@ -366,6 +366,14 @@ type Query {
   "Find a deposit using its ID"
   deposit("ID of the Deposit" id: ID!): Deposit
 
+  "Fetch all deposits"
+  deposits(
+    "Date range to fetch deposits between"
+    dateRange: DateRange,
+    "Pagination options"
+    pagination: Pagination,
+  ): DepositsConnection
+
   "Get data for a specific epoch, if ID omitted it gets the current epoch. If the string is 'next', fetch the next epoch"
   epoch(id: ID): Epoch!
 
@@ -714,6 +722,14 @@ type Query {
 
   "Find a withdrawal using its ID"
   withdrawal("ID of the withdrawal" id: ID!): Withdrawal
+
+  "Fetch all withdrawals"
+  withdrawals(
+    "Date range to fetch withdrawals between"
+    dateRange: DateRange,
+    "Pagination options"
+    pagination: Pagination,
+  ): WithdrawalsConnection
 }
 
 "All the states a transfer can transition between"
--- a/datanode/sqlstore/deposits.go
+++ b/datanode/sqlstore/deposits.go
@@ -145,8 +145,12 @@ func (d *Deposits) getByPartyCursorPagination(ctx context.Context, party string,
 func getDepositsByPartyQuery(party string, dateRange entities.DateRange) (string, []interface{}) {
 	var args []interface{}
 
-	query := fmt.Sprintf(`select id, status, party_id, asset, amount, foreign_tx_hash, credited_timestamp, created_timestamp, tx_hash, vega_time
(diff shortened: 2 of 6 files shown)


datanode/gateway/graphql/schema.graphql
```diff
@@ -366,6 +366,14 @@ type Query {
   "Find a deposit using its ID"
   deposit("ID of the Deposit" id: ID!): Deposit
 
+  "Fetch all deposits"
+  deposits(
+    "Date range to fetch deposits between"
+    dateRange: DateRange,
+    "Pagination options"
+    pagination: Pagination,
+  ): DepositsConnection
```
jeremyletang, line 375: Shouldn't this take an optional party?

datanode/gateway/graphql/schema.graphql
```diff
@@ -714,6 +722,14 @@ type Query {
...
   "Find a withdrawal using its ID"
   withdrawal("ID of the withdrawal" id: ID!): Withdrawal
+
+  "Fetch all withdrawals"
+  withdrawals(
+    "Date range to fetch withdrawals between"
+    dateRange: DateRange,
+    "Pagination options"
+    pagination: Pagination,
+  ): WithdrawalsConnection
```
jeremyletang, line 732: This too?

datanode/gateway/graphql/schema.graphql
guoguojin: No, we already have a depositsConnection subquery to return the deposits for a particular party defined in the `party` type. I added these deposits and withdrawals to allow the user to request list of all deposits/withdrawals as requested in the ticket, so that the GraphQL API was inline with REST API change requested.

datanode/gateway/graphql/schema.graphql
guoguojin: See above.

jeremyletang (approved):

datanode/sqlstore/deposits.go
```diff
@@ -145,8 +145,12 @@ func (d *Deposits) getByPartyCursorPagination(ctx context.Context, party string,
 func getDepositsByPartyQuery(party string, dateRange entities.DateRange) (string, []interface{}) {
 	var args []interface{}
 
-	query := fmt.Sprintf(`select id, status, party_id, asset, amount, foreign_tx_hash, credited_timestamp, created_timestamp, tx_hash, vega_time
-		from deposits_current where party_id = %s`, nextBindVar(&args, entities.PartyID(party)))
+	query := `select id, status, party_id, asset, amount, foreign_tx_hash, credited_timestamp, created_timestamp, tx_hash, vega_time
```
dexturr, line 148: I know I'm only supposed to look at the schema.gql but the SQL syntax between this and the next file is inconsistent and it bothers me 😂

dexturr (approved):

mattrussell36 (approved):

PR #6309, opened 2022-09-27
Improve wallet interactions framework

### LINKED ISSUE #6308: Support parallel requests on wallet API v2
Today we don't support parallel requests, which is fine from a CLI application point of view but doesn't help to get the most out of full fegled UI application like the desktop-wallet.

### LINKED ISSUE #6425: Introduce interaction for beginning and ending of request
We should warn the wallet front-end about the beginning and the ending of an interaction session, request wide, so it's easier for wallet front-end to respond to the request lifecycle.

### LINKED ISSUE #6426: Add a `name` field on interaction to know what they are when JSON
When the interactions are translated to json, understanding what the interaction is, is a pain.

We need to add a name.

### LINKED ISSUE #6427: Improve interactions documentation
There is little to no documentation on the interactions.

We need some to help developer integrate with the exisiting service.

### LINKED ISSUE #6428: Update the wallet connection decision for future work
Today we can only temporarily approve or reject a wallet connection using a boolean. In the future, we will need to support other options like permanently approve or reject a hostname. This must not be implemented right now, but the foundations must be laid.

So, we need to update the interaction to accept a "mode" instead of a simple boolean.

From:
```
true | false
```

To:
```
APPROVED_ONLY_THIS_TIME | REJECTED_ONLY_THIS_TIME
```

### LINKED ISSUE #6429: Rename pipeline to interactor for better understanding

### LINKED ISSUE #6430: Split the transaction status interaction depending on success and failure
Whatever the state of the sent transaction, we notify a `TransactionStatus`, with an error field filled on error, and a transaction hash field filled on success.

That's a bit of crap developer experience, let's split this in two distinct notification:
- `TrasanctionFailed` with `error` field
- `TransactionSucceeded` with `txHash` field

### LINKED ISSUE #6431: Pass the deserialized input data in `Transaction<Succeeded|Failed>` notifications
To help wallet fron-ends create a rich user experience, we need to return the input data of a transaction in its deserialised form.

Today, it's a byte array encoded using protobuf. It's hard for wallet front-ends to extract information about the transaction without having to embed language specific library to deserialized our proto messages.

Sending it in its deserialized form will help.

ValentinTrinque:
Close #6308 
Close #6428 
Close #6429 
Close #6430 
Close #6425 
Close #6308 
Close #6426 
Close #6427 
Close #6431


wwestgarth (dismissed):
Theres a lot going on, but it seems to all make sense!

jeremyletang (approved):

PR #6328, opened 2022-09-27
fix: use offset 0 to set postgres timezone

jeremyletang:
Set the offset of the timezone to 0 instead of UTC which will not be recognized if the tzdata is out of date


datanode/snapshot/service_test.go
```diff
@@ -87,14 +88,15 @@ func TestMain(t *testing.M) {
 		sqlBroker, err := setupSQLBroker(ctx, eventsFile, sqlConfig, snapshotService,
 			func(ctx context.Context, service *snapshot.Service, chainId string, lastCommittedBlockHeight int64) bool {
 				if lastCommittedBlockHeight > 0 && lastCommittedBlockHeight%snapshotInterval == 0 {
+					fmt.Printf("YEZ1\n")
```
ettec: remove this

datanode/snapshot/service_test.go
```diff
@@ -87,14 +88,15 @@ func TestMain(t *testing.M) {
...
 					lastSnapshot, err := service.CreateSnapshotAsync(ctx, chainId, snapshot.GetFromHeight(lastCommittedBlockHeight, snapshotInterval), lastCommittedBlockHeight)
 					if err != nil {
+						fmt.Printf("YEZ2\n")
 						panic(fmt.Errorf("failed to create snapshot:%w", err))
 					}
```
ettec, line 94: ...and this :)

jeremyletang: already done

datanode/snapshot/service_create_snapshot.go
```diff
@@ -89,7 +92,7 @@ func (b *Service) createSnapshot(ctx context.Context, chainID string, fromHeight
 			err = b.snapshotData(ctx, copyDataTx, dbMetaData, currentSnapshot, historySnapshot, chainID, toHeight)
 			b.snapshotInProgress.Store(false)
 			if err != nil {
-				b.log.Errorf("failed to snapshot data:%s", err)
+				b.log.Panic("failed to snapshot data", logging.Error(err))
 			} else {
 				err = os.Remove(snapshotInProgressFile)
```
ettec, line 97: this err should panic as well to make testing behave, not 100% we should panic in all other circumstances, but should be ok for now

jeremyletang: addressed.

peterbarrow (dismissed):

PR #6333, opened 2022-09-27
6327 checkpoints delegations to history

### LINKED ISSUE #6327: Checkpoints and Delegations should be history tables with synthetic PK
Getting PK failures on Checkpoints and delegations,  should be history tables with sythetic PK

ettec:
closes #6327 - checkpoints and delegation moved to history tables and PK (used for snapshot purposes) moved to synthetic key

### DIFF
--- a/datanode/sqlstore/migrations/0001_initial.sql
+++ b/datanode/sqlstore/migrations/0001_initial.sql
@@ -628,20 +628,52 @@ CREATE TABLE rewards(
 );
 
 CREATE TABLE delegations(
-  party_id         BYTEA NOT NULL, -- REFERENCES parties(id), TODO once parties table is populated
-  node_id          BYTEA NOT NULL REFERENCES nodes(id),
+  party_id         BYTEA NOT NULL,
+  node_id          BYTEA NOT NULL,
   epoch_id         BIGINT NOT NULL,
   amount           HUGEINT,
   tx_hash          BYTEA NOT NULL,
   vega_time        TIMESTAMP WITH TIME ZONE NOT NULL,
-  primary key (party_id, node_id, epoch_id, vega_time)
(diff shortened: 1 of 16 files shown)


datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -1324,8 +1359,11 @@ DROP TYPE IF EXISTS proposal_error;
 DROP TYPE IF EXISTS proposal_state;
 
 DROP TABLE IF EXISTS epochs;
-DROP VIEW IF EXISTS delegations_current;
+DROP TABLE IF EXISTS delegations_current;
 DROP TABLE IF EXISTS delegations;
+DROP TRIGGER IF EXISTS update_current_delegations ON delegations;
+DROP FUNCTION IF EXISTS update_current_delegations;
```
jeremyletang, line 1365: Probably not an issue, but could there be problem when dropping here, as we drop first the delegations_current, but it's used in the update_current_delegations function?

wwestgarth:
I don't want to hold this up, but can we throw in a quick unit test to check that we can add two checkpoints with the same block and hash? 

It will act as a reminder that this is a scenario that can happen, and stop us going round in circles, because we had a similar issue a while back with pkey constraints on checkpoints, which was fixed by removing the constraint. And it _also_ didn't add a unittest: https://github.com/vegaprotocol/vega/commit/a5492369b4b24951dfcbc6d32e53dbe3d26d7c3b

datanode/sqlstore/migrations/0001_initial.sql
ettec: don't think it does, but shifted it around anyways

pscott31 (dismissed):

wwestgarth (approved):

PR #6334, opened 2022-09-28
feat: add a simple rate limiter for graphql

jeremyletang:


wwestgarth (dismissed):

datanode/gateway/middleware.go
```diff
@@ -124,3 +132,92 @@ func (i *InjectableResponseWriter) Write(data []byte) (int, error) {
...
+func (s *SubscriptionRateLimiter) Dec(ip string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	cnt := s.m[ip]
+	s.m[ip] = cnt - 1
+}
+
+func (s *SubscriptionRateLimiter) WithSubscriptionRateLimiter(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// is that a subscription?
```
zale144, line 175: It's just to prevent panic going forward

datanode/gateway/middleware.go
```diff
@@ -124,3 +132,92 @@ func (i *InjectableResponseWriter) Write(data []byte) (int, error) {
...
+		if _, ok := w.(http.Hijacker); !ok {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		if ip, err := getIP(r); err != nil {
+			s.log.Debug("couldn't get client ip", logging.Error(err))
+		} else {
+			if err := s.Inc(ip); err != nil {
+				s.log.Error("couldn't client reached max sub", logging.Error(err))
```
zale144, line 185: This could be a bit clearer

datanode/gateway/middleware.go
```diff
@@ -124,3 +132,92 @@ func (i *InjectableResponseWriter) Write(data []byte) (int, error) {
...
+				w.Write([]byte(err.Error()))
+				// write error
+				return
+			}
+			defer func() {
+				s.Dec(ip)
+			}()
+		}
+
+		next.ServeHTTP(w, r)
```
zale144, line 196: Maybe if this was in the defer, the if/else above could be simplified a bit?

zale144 (commented):
Just some minor comments, not a show stopper

zale144:
> We could use the rate package instead of re-implementing this. What do you think ?
> 
> 
> 
> https://pkg.go.dev/golang.org/x/time/rate

That one doesn't have any grouping  (e.g. per ip) capability.

zale144:
> > We could use the rate package instead of re-implementing this. What do you think ?
> > 
> > 
> > 
> > https://pkg.go.dev/golang.org/x/time/rate
> 
> That one doesn't have any grouping  (e.g. per ip) capability.

Although, we could have a map of those Limiters, like `ip:limiter`. But it might be a bit heavy on the memory.
Better just take bits from it, like the token-based limiting, and add it to this one.

datanode/gateway/middleware.go
jeremyletang: haha I don't what I was trying to say here to be fair.

datanode/gateway/middleware.go
jeremyletang: hum not really as it makes weird to check if we've been limited. I'd rather keep like that just now I guess.

datanode/gateway/middleware.go
jeremyletang: ddressed

zale144 (approved):
Lgtm

pscott31 (approved):

PR #6342, opened 2022-09-28
fix: sql down script checks for view that are now tables

guoguojin:


ze97286 (dismissed):

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -1345,7 +1345,11 @@ DROP TABLE IF EXISTS oracle_data;
 DROP TABLE IF EXISTS oracle_specs;
 DROP TYPE IF EXISTS oracle_spec_status;
 
-DROP TABLE IF EXISTS positions_current;
```
ettec, line 1348: Is this necessary?, don't we just drop and recreate the database at the moment anyway?

guoguojin: No not always, but they have had to the last few times there's been a release, see the conversations on Slack https://vegaprotocol.slack.com/archives/C9E6XA8GK/p1664363321919129?thread_ts=1664360516.989949&cid=C9E6XA8GK

pscott31 (dismissed):

datanode/sqlstore/migrations/0001_initial.sql
ettec: yeah, we should change the release process to always nuke the DB.  At some point we will need to handle migration (which version scripts), but as we don't hold data across releases, not yet.

guoguojin:
Not required as deployments to environments will destroy existing database and migrate from clean.

PR #6347, opened 2022-09-28
fix: Do not take a snapshot before the interval has been set from the…

### LINKED ISSUE #6266: Snapshot: can't load from snapshot taken at height 1 after checkpoint reload
# Problem encountered
Which is ridiculous because there is no value in restoring from a snapshot at height 1 anyway. The reason it fails is because of the epoch-fast forwarding we do after restoring from a checkpoint which happens at the start of block 2 and we don't save in the snapshot that we need to fast forward.

I think what I'd rather do than handle this case, is to just not take a snapshot at height 1 because I have always believed this is a bug anyway. If you set the snapshot interval to 1000 you'd expect a snapshot at height 1000, 2000, etc and not 1001, 2001. 

Its just a consequence of the snapshot engine setting its `interval` field before the network parameters are propagated.

# Observed behaviour

# Expected behaviour

# System response

# Steps to reproduce

## Manual

## Automation

# Evidence

## Logs

## Additional context

# Definition of Done

**Before Merging**

**After Merging**

ze97286:
… network parameter

Closes #6266

### DIFF
--- a/core/snapshot/engine.go
+++ b/core/snapshot/engine.go
@@ -200,7 +200,7 @@ func New(ctx context.Context, vegapath paths.Paths, conf Config, log *logging.Lo
 		wrap:            appPL,
 		app:             appPL.AppState,
 		interval:        1, // default to every block
-		current:         1,
+		current:         -1,
 	}
 	return eng, nil
 }
@@ -1085,10 +1085,12 @@ func (e *Engine) Close() error {
 }
 
 func (e *Engine) OnSnapshotIntervalUpdate(ctx context.Context, interval int64) error {
-	e.interval = interval
-	if interval < e.current {
+	if interval < e.current || e.current < 0 {
 		e.current = interval
+	} else if interval > e.interval {
+		e.current += interval - e.interval
 	}
+	e.interval = interval
 	return nil
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,7 @@
 - [6348](https://github.com/vegaprotocol/vega/issues/6348) - Reduce pool size to leave more available `Postgres` connections
 
 ### 🐛 Fixes
-- [](https://github.com/vegaprotocol/vega/issues/xxxx) -
-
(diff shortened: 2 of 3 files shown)


jeremyletang (dismissed):

wwestgarth:
This is no way fixes the problem. If you look in the logs of the system test, with an interval of 12 we are still taking a snapshots at height 1, 13, 25 and not 12, 24.

ze97286:
> This is no way fixes the problem. If you look in the logs of the system test, with an interval of 12 we are still taking a snapshots at height 1, 13, 25 and not 12, 24.

you say in the ticket that the issue is that we're taking snapshot before the interval is propagated from the net param, this PR fixes that. why would it take a snapshot at block 1?

ValentinTrinque (dismissed):

wwestgarth:
I said in the ticket:
```
Its just a consequence of the snapshot engine setting its interval field before the network parameters are propagated.
```
and due to `interval` being set to `1` at initialisation, when the netparameter comes in we do not increase it due to the logic you just found.

Regardless, the ticket was to stop taking a snapshot at height 1, which you didn't fix because you didn't run a network to verify, or write a unittest.

core/snapshot/engine.go
```diff
@@ -1111,7 +1111,7 @@ func (e *Engine) Close() error {
 
 func (e *Engine) OnSnapshotIntervalUpdate(ctx context.Context, interval int64) error {
```
wwestgarth, line 1112: Can you write a unittest for this change please?

ze97286: does the system test evidence above not prove that it works?

wwestgarth: It proves it works right now. It does not prove it works tomorrow.

wwestgarth: If it were a more complicated fix where a unittest was extremely complicated, then providing a screen shot is sufficient, but when a unittest would be maybe 5 lines and trivial to write, then we should be writing them.

wwestgarth (approved):

PR #6385, opened 2022-10-03
feat: use new begin and end block events

### LINKED ISSUE #6211: Handle start and end event in datanode

### LINKED ISSUE #6414: Checkpoint and other events occur between EndBlock and BeginBlock events
# Problem encountered

As noted by @wwestgarth regarding core snapshots and `BeginBlock` and `EndBlock` events:

```
We do actually send out events for checkpoints during OnCommit in core, which I think will be after the EndBlock event, and I wonder if this will affect datanode-snapshot hashes.

I'm thinking about the case where we take a datanode snapshot on the block where a core checkpoint happens. The checkpoint event won't be included in datanode's snapshot because we'll commit the current block without it, but will (probably?) be included in the next snapshot as it'll get bundled up and commited to the DB as part of the next block.

If I then restore from that first datanode snapshot the checkpoint event will not be in there, but then core won't send that checkpoint event because it will restore from a point where it has already sent it. So that bit of data won't be in the next datanode snapshot and might give a different hash compared to a node that hasn't restored.

Its really edge-case-y so don't let it block this PR, but we might want to outlaw any events being sent in OnCommit in core.
```

# Observed behaviour
Checkpoint events are received after the `EndBlock` event and before the `BeginBlock` event suggesting those events might not be persisted in data-node.

```
2022-10-05T11:29:05.375+0100    DEBUG   root    broker/sqlstore_broker.go:228   received event  {"type": "EndBlockEvent", "trace-id": "C42F9FDB2C3573C7FAD8B530A3179F8B64D8D030571273913325B56D6FD89974"}
2022-10-05T11:29:05.380+0100    DEBUG   root    broker/sqlstore_broker.go:228   received event  {"type": "CheckpointEvent", "trace-id": "C42F9FDB2C3573C7FAD8B530A3179F8B64D8D030571273913325B56D6FD89974"}
2022-10-05T11:29:05.383+0100    DEBUG   root    broker/sqlstore_broker.go:228   received event  {"type": "BeginBlockEvent", "trace-id": "E9B6AC9DE4348A1C69C0658300BF273D3F4546BC9FEBCF2906A20F4EC5B74CFF"}

```

# Expected behaviour
No events should occur after `EndBlock` other than `BeginBlock`

guoguojin:
Closes #6211 
Closes #6414


datanode/sqlstore/node.go
```diff
@@ -43,6 +44,7 @@ func NewNode(connectionSource *ConnectionSource) *Node {
 
 func (store *Node) UpsertNode(ctx context.Context, node *entities.Node) error {
 	defer metrics.StartSQLQuery("Node", "UpsertNode")()
+	store.log.Info("Upserting node", zap.String("node", node.ID.String()))
```
pscott31, line 47: super minor but I think the convention here is to use logging.String()

core/processor/abci.go
```diff
@@ -625,15 +625,29 @@ func (app *App) OnInitChain(req tmtypes.RequestInitChain) tmtypes.ResponseInitCh
 	// let's assume genesis block is block 0
```
pscott31, line 625: Is this comment still true?

pscott31 (dismissed):

datanode/sqlstore/node.go
guoguojin: Sorry, that should have been removed. I will update it.

pscott31 (dismissed):

ettec (changes requested):

datanode/broker/sqlstore_broker.go
```diff
@@ -130,33 +130,35 @@ func (b *SQLStoreBroker) waitForFirstBlock(ctx context.Context, errCh <-chan err
 	} else if errors.Is(err, sqlstore.ErrNoLastBlock) {
 		lastProcessedBlock = entities.Block{
 			VegaTime: time.Time{},
-			Height:   -1,
```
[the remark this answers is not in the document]
ettec: the problem here is if its anything other than 0 and data node has no hsitory it won't work.  it will blow on the check below:  Handling non-0 first block should be a seperate PR IMO that should not hold up this, there are other implications.



`if beginBlock.BlockNr() > lastProcessedBlock.Height+1 {
					return nil, fmt.Errorf("block height on begin block, %d, is too high, the height of the last processed block is %d",
						beginBlock.BlockNr(), lastProcessedBlock.Height)
				}
`

wwestgarth:
Just want to bring up again a comment I left on the core side PR from before. We do actually send out events for checkpoints during `OnCommit` in core, which I think will be after the `EndBlock` event, and I wonder if this will affect datanode-snapshot hashes.

I'm thinking about the case where we take a datanode snapshot on the block where a core checkpoint happens. The checkpoint event won't be included in datanode's snapshot because we'll commit the current block without it, but will (probably?) be included in the next snapshot as it'll get bundled up and commited to the DB as part of the next block.

If I then restore from that first datanode snapshot the checkpoint event will not be in there, but then core won't send that checkpoint event because it will restore from a point where it has already sent it. So that bit of data _won't_ be in the next datanode snapshot and might give a different hash compared to a node that hasn't restored.

Its really edge-case so don't let it block this PR, but we might want to outlaw any events being sent in `OnCommit` in core.

datanode/broker/sqlstore_broker.go
```diff
@@ -214,15 +220,15 @@ func (b *SQLStoreBroker) processBlock(ctx context.Context, dbContext context.Con
...
 			b.log.Warningf("slow time update detected, time between checks %v, block height: %d, total block processing time: %v", slowTimeUpdateThreshold,
 				block.Height, blockTimer.duration)
 		case e := <-eventsCh:
 			metrics.EventCounterInc(e.Type().String())
 			blockTimer.startTimer()
-			if e.Type() == events.TimeUpdate {
-				timeUpdate := e.(entities.TimeUpdateEvent)
 
+			switch e.Type() {
+			case events.EndBlockEvent:
```
ettec, line 231: Can we have a check here to ensure that after end block event, the next event it a being block, anything else should cause a panic, reason is it will potentially result in inconsistent state across nodes (considering restore from snapshot)

ettec (changes requested):

guoguojin:
I've created this issue for the events that occur between EndBlock/BeginBlock https://github.com/vegaprotocol/vega/issues/6414

wwestgarth (approved):

jeremyletang (approved):

PR #6386, opened 2022-10-03
Cover new liquidity monitoring ACs + fix max OI not dropping (when expected) + new  🥒 step

### LINKED ISSUE #6387: Max OI not dropping
Max OI not dropping as expected do to incorrect timestamp when computing max from values recorded with same timestamp.

witgaw:
Closes https://github.com/vegaprotocol/system-tests/issues/1095
Closes  #6387

### DIFF
--- a/core/integration/steps/the_following_events_should_be_emitted.go
+++ b/core/integration/steps/the_following_events_should_be_emitted.go
@@ -57,4 +57,3 @@ func errEventNotEmitted(t events.Type) error {
 func errEventEmitted(t events.Type) error {
 	return fmt.Errorf("event of type %s has been emitted", t)
 }
-
--- a/core/integration/steps/the_following_events_should_be_emitted.go
+++ b/core/integration/steps/the_following_events_should_be_emitted.go
@@ -32,6 +32,18 @@ func TheFollowingEventsShouldBeEmitted(broker *stubs.BrokerStub, table *godog.Ta
 	return nil
 }
 
+func TheFollowingEventsShouldNotBeEmitted(broker *stubs.BrokerStub, table *godog.Table) error {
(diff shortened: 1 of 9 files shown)


witgaw (commented):
Ready for review.

core/integration/steps/the_following_events_should_be_emitted.go
```diff
@@ -32,6 +32,18 @@ func TheFollowingEventsShouldBeEmitted(broker *stubs.BrokerStub, table *godog.Ta
 	return nil
 }
 
+func TheFollowingEventsShouldNotBeEmitted(broker *stubs.BrokerStub, table *godog.Table) error {
```
EVODelavega, line 35: This is a bit of an iffy step/check IMO. If we were to use this to check, say, that after a trade we don't enter auction, we'd have an error, as the events related to monitoring auction will still be returned by `broker.GetBatch`. I do think this could be a useful check to have, but perhaps we should qualify that we expect events of type X not to be emitted in the current/last block instead of in general

witgaw: How is that different from the already existing `TheFollowingEventsShouldBeEmitted` step? It sort of requires manual work anyway (to clear events beforehand) as by default it contains all events from the beginning of the scenario.

witgaw: We could at some point properly implement the `given`, `when`, `then` keywords (now they can just be used interchangeably) and then have all the events only cached for lookup during the course of a single `given`/`when` (+ optional `and`) clause, but that's clearly not something for this PR.

witgaw (commented):
Ready for another review

jeremyletang (approved):

PR #6398, opened 2022-10-04
chore: add core transaction_result event

jeremyletang:
This adds a new event emited by the core: TransactionResult. It basically reuse most of the specifics of the TxErrEvent from earlier, but instead is emitted on every transaction processed by the chain.

This is build demands from the frontend team so they can open a stream and listen to these invents to know if the transaction was successful or not, and know if state have been update etc.

This is for now not availablke through the graphql API, because it's not be proven it will be of any use, and the implementation would take a lot of time (mapping of all commands which are embeded in the event).

For now this is available either from the grpc or REST endpoint. This is the example to use it through the websocket:
First call this API:
```
"wss://api.n07.testnet.vega.xyz/api/v2/stream/event/bus
```
Then the socket will be blocking until the details of what needs to be listened too is sent, the client should then send something like the following through the wire:
```
{"type": ["BUS_EVENT_TYPE_TRANSACTION_RESULT"], "partyId":"123456789abcdef..."}
```


pscott31:
I wonder if the rightful home for this is in the block explorer?

jeremyletang:
> I wonder if the rightful home for this is in the block explorer?

it cannot really. I'm not planning to store it or anything, and it will be in a near-ish future improve with information from the state of the core resulting from the processing of a transaction (e.g: and successful amend here would also embed the amended order as part of the payload). I'm not sure the block explorer would be able to get this kind of information? 

Also that would mean the block explorer would need listen to the event bus from the core? Seeing that it access only the database which only knows if the transaction was put in a block really, but not much more about it (apart from the event attached to it).

guoguojin (approved):

PR #6433, opened 2022-10-06
feat: migrate remaining views of current data to tables

### LINKED ISSUE #6378: Migrate remaining 'current' views to 'current' tables
In datanode, we split data into history data (which are hypertables) and current data (all non-hypertables).    In some cases we still have views to get current values from historical data kept in non-history tables, this means the current state snapshot will grow as historical data is added to these types.   For consistency with other types and to ensure snapshots don't grow in size unnecessarily we should migrate these views to tables with a corresponding trigger to update them (see e.g. positions).
Views to migrate:

markets_current
deposits_current
withdrawals_current
network_parameters_current
oracle_data_current
stake_linking_current

guoguojin:
Close #6378


jeremyletang (approved):
Looks goot to me. Might want a review from @pscott31 and @ettec before merging though. 

Just one comment, we really need to start using one style for all the sql syntaxes, uppercase / lowercase, but not a blocker for now.

pscott31 (approved):
Looks OK to me. The only thing to maybe consider is that each of these triggers means a bit of extra delay when inserting into these tables, but I think these aren't particularly busy ones so it should be fine.

PR #6446, opened 2022-10-07
fix: restore data-node snapshot tests

### LINKED ISSUE #6442: Re-enable data-node snapshot tests
Data-node snapshot tests were disable during the PR for BeginBlock and EndBlock message processing. The tests need to be re-enabled using a new event file with the corresponding event updates.

guoguojin:
Closes #6442


datanode/snapshot/service.go
```diff
@@ -128,9 +128,9 @@ func (b *Service) Types() []events.Type {
 
 func GetFromHeight(toHeight int64, snapshotInterval int64) int64 {
 	fromHeight := int64(0)
-	if toHeight-snapshotInterval != 0 {
-		fromHeight = toHeight - (snapshotInterval - 1)
-	}
+	// if toHeight-snapshotInterval != 0 {
+	fromHeight = toHeight - (snapshotInterval - 1)
+	//}
```
pscott31, line 133: Seems a bit odd to leave the commented code in - plus it makes the first line on the function redundant?

guoguojin: Sorry, that should have been deleted.

pscott31 (approved):

PR #6447, opened 2022-10-07
fix: a couple of gql bugs related to erc20 bundles

### LINKED ISSUE #6236: [datanode] graphql erc20ListAssetBundle schema incorrect for assets that are not found
Testing on stagnet3 and testnet:

https://github.com/vegaprotocol/vega/blob/f0da7580017164aec9ad9033b00bbb42075cd3e4/datanode/gateway/graphql/schema.graphql#L372-L376

# API incorrectly returns `null` if asset is not found
## Null on an asset that can never exist
```graphql
{
  erc20ListAssetBundle(assetId:"test") {
    signatures
  }
}
```

## Null on a built in asset that does exist

```graphql
{
  erc20ListAssetBundle(assetId:"VOTE") {
    signatures
  }
}
```

Both return:
```json

{
  "errors": [
    {
      "message": "rpc error: code = NotFound desc = NotFound error",
      "path": [
        "erc20ListAssetBundle"
      ]
    }
  ],
  "data": null
}
```

# Expected result
- Schema lists these fields as non-nullable. The quick fix is to make the field nullable. Not finding data for something is not an error.

~This endpoint no longer exists on devnet, if it’s replaced by #6230 we have similar problems there~ This is apparently a deployment error

Split #6238 out to a separate ticket

### LINKED ISSUE #6238: [datanode] erc20listAssetBundle returns 0x for assets with no signatures
Testing on stagnet3 and testnet:

https://github.com/vegaprotocol/vega/blob/f0da7580017164aec9ad9033b00bbb42075cd3e4/datanode/gateway/graphql/schema.graphql#L372-L376

# No signatures return as `0x`
```graphql
{
  erc20ListAssetBundle(assetId:"98032ba34576f8012de9b822e1da3ed4b6223a4f4e05f573002d441ffb4bf314") {
    signatures
  }
}
```

```json
{
  "data": {
    "erc20ListAssetBundle": {
      "signatures": "0x"
    }
  }
}
```

# Expected result
- `0x` is not a valid signature and should not be returned

Split out from #6236

pscott31:
Closes #6236 
Closes #6238

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -700,11 +700,6 @@ func (t *tradingDataServiceV2) GetERC20MultiSigSignerAddedBundles(ctx context.Co
 			return nil, apiError(codes.Internal, err)
 		}
 
-		pack := "0x"
-		for _, v := range signatures {
-			pack = fmt.Sprintf("%v%v", pack, hex.EncodeToString(v.Sig))
-		}
-
 		edges = append(edges,
 			&v2.ERC20MultiSigSignerAddedBundleEdge{
 				Node: &v2.ERC20MultiSigSignerAddedBundle{
@@ -712,7 +707,7 @@ func (t *tradingDataServiceV2) GetERC20MultiSigSignerAddedBundles(ctx context.Co
 					Submitter:  b.Submitter.String(),
 					Nonce:      b.Nonce,
 					Timestamp:  b.VegaTime.UnixNano(),
-					Signatures: pack,
+					Signatures: packNodeSignatures(signatures),
 					EpochSeq:   strconv.FormatInt(b.EpochID, 10),
 				},
 				Cursor: b.Cursor().Encode(),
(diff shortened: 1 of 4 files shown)


datanode/api/trading_data_v2.go
```diff
@@ -907,7 +907,11 @@ func (t *tradingDataServiceV2) GetERC20ListAssetBundle(ctx context.Context, req
 	}
 
 	// now we pack them
-	pack := "0x"
+	pack := ""
+	if len(signatures) > 0 {
```
wwestgarth, line 911: Wondering if in the case when there are no signatures, then we should treat it as not-found (or null or whatever the conclusion was to return in not found cases)? Because the query is "can I have the ListAssetBundle for this asset please" and if there are no signatures, then it means no bundle has been sent out from core.

wwestgarth: I really don't have any strong feelings eitherway.

For `GetERC20MultiSigSignerAddedBundles` and `GetERC20MultiSigSignerRemovedBundles` you are right, they should probably be `ListXXX` because it does return multiple bundles, and in this case if there is an `GetAddedEvent()` then the signatures will (should) follow, but could still just be omitted from the list until they are through.

For `GetERC20ListAssetBundle` its a little different because the asset could exist, have been rejected at the proposal stage and so no signatures will come through (same with `GetERC20WithdrawalApproval`, a withdrawal can be rejected). I guess I was just thinking it might give false hope if the API returns _something_ even though the bit they really need from it is never coming.

Maybe @vegaprotocol/frontend or @edd have an opinion?

wwestgarth (approved):

dexturr (approved):

PR #6479, opened 2022-10-12
feat: handle upgrade event

### LINKED ISSUE #6080: datanode to wait for upgrade block
When validators decides to schedule an upgrade, and event is sent to the data-node. The datanode should listen to these events and to keep track of scheduled upgrades.

Once the time of the upgrade is reached, the visor will then ask the data-node to shutdown. In order to ensure that the whole state of the data-node is correct regarding to the validator nodes, the data-node should ensure that it processed all blocks up to the block of the  upgrade.

Respectively the core node shouldn't shutdown as well until the data-node have consumed all blocks in the broker queue (pending to be sent).

Once all events have been saved in the database, both nodes should stop gracefully.

guoguojin:
Closes #6080


datanode/broker/sqlstore_broker.go
```diff
@@ -119,8 +130,16 @@ func (b *SQLStoreBroker) Receive(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("failed to receive events, unable to get chain id:%w", err)
 		}
-		lastCommittedBlockHeight := nextBlock.Height - 1
-		if b.onBlockCommitted(ctx, chainID, lastCommittedBlockHeight) {
+
+		if b.onBlockCommitted(ctx, chainID, b.lastBlock.Height) {
+			if b.receivedProtocolUpgradeEvent {
```
pscott31, line 135: Is this code ever called? It looks like snapshotService.OnBlockCommitted only ever returns false.. and even if that wasn't the case, don't we get equivalent behaviour by  just moving

```
		if b.receivedProtocolUpgradeEvent {
			b.protocolUpgradeService.SetProtocolUpgradeStarted()
			return nil
		}
``` 
above this block?

pscott31 (dismissed):
Looks good, apart from the oddness of the return value in onBlockCommitted but that pre-dates this PR :)

dexturr (approved):

pscott31 (approved):

PR #6500, opened 2022-10-13
feat: datanode decentralized history

### LINKED ISSUE #6227: Datanode to share snapshots via the network
The Readme file in datanode/dehistory has a full description of the why/what and how of this change.

Breaking change:  the command data-node init command now needs to be passed the chain id, e.g.

data-node init  test-chain-7cevW

In addition any existing data nodes that are not init'ed from fresh each release (e.g mainnet), will need to add the following to the data-nodes config.toml:

`ChainID = "<the networks chain ID here"`

ettec:
Closes #6227 - Readme in datanode/dehistory contains full explanation of how the decentralized history works

Breaking change:  data-node init command now requires that the chain id be passed as a parameter e.g:

data-node init test-chain-7cevW2

In addition any exisiting data nodes that are not init'ed from fresh each release (e.g mainnet), will need to add the following to the data-nodes config.toml:

`ChainID = "<the networks chain ID here"`


datanode/dehistory/README.md
```diff
@@ -0,0 +1,12 @@
+# Decentralized History
```
jeremyletang, line 1: Nice docs haha

datanode/dehistory/fsutil/fsutil.go
```diff
@@ -0,0 +1,203 @@
...
+	"os"
+	"path/filepath"
+
+	vgfs "code.vegaprotocol.io/vega/libs/fs"
+)
+
+// TODO File containing common file manipulation logic used across dehistory - exists elsewhere?
+
+// RemoveAllFromDirectory used in place of os.RemoveAll when the directory should be emptied but not removed.
+func RemoveAllFromDirectory(dir string) error {
```
jeremyletang, line 19: Using https://pkg.go.dev/path/filepath#WalkDir here would probably replace most of this function content.

datanode/dehistory/fsutil/fsutil.go
```diff
@@ -0,0 +1,203 @@
...
+
+	deviceMajVersion, err = UntarFile(zr, decompressedFilesDestination)
+	if err != nil {
+		return 0, err
+	}
+
+	return deviceMajVersion, nil
+}
+
+var ErrMismatchedDeviceVersion = errors.New("mismatched device versions")
```
jeremyletang, line 164: we usually group all errors at the top of the file.

datanode/dehistory/fsutil/fsutil.go
```diff
@@ -0,0 +1,203 @@
...
+
+		targetFilePath := filepath.Join(decompressedFilesDestination, header.Name)
+		fileDeviceVersion := header.Devmajor
+		if deviceVersion == -1 {
+			deviceVersion = fileDeviceVersion
+		} else {
+			if fileDeviceVersion != deviceVersion {
+				return 0, ErrMismatchedDeviceVersion
+			}
+		}
```
jeremyletang, line 187:
```suggestion
		if deviceVersion == -1 {
			deviceVersion = header.Devmajor
		} else if header.Devmajor != deviceVersion {
			return 0, ErrMismatchedDeviceVersion
		}
```

datanode/dehistory/initialise/initialise.go
```diff
@@ -0,0 +1,300 @@
...
+	from, to, err := deHistoryService.LoadAllAvailableHistoryIntoDatanode(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to load history into the datanode%w", err)
+	}
+	log.Infof("loaded history from height %d to %d into the datanode", from, to)
+
+	return nil
+}
+
+var ErrFailedToGetSegment = errors.New("no history segment found")
```
jeremyletang, line 80: same with this error, usually at the top of the fiel

datanode/dehistory/initialise/initialise.go
```diff
@@ -0,0 +1,300 @@
...
+
+	segmentsList := make([]*v2.HistorySegment, 0, len(peerToSegment))
+
+	segmentsInfo := "Most Recent History Segments:\n\n"
+	for peer, segment := range peerToSegment {
+		segmentsList = append(segmentsList, segment)
+		segmentsInfo += fmt.Sprintf("Peer:%-39s Segment{%s}\n\n", peer, segment)
+	}
+
+	logInfof(segmentsInfo)
```
jeremyletang, line 111: Why do this formatting here? If can we just log info the type directly? adding \n and stuff in a log will be not very readable / practical to use seeing that any other logs are meant to be jsion structured logs

ettec: good point, sorted that now

candida-d (commented):
I'd love if this included some more documentation, particularly in the protos.

datanode/dehistory/fsutil/fsutil.go
```diff
@@ -0,0 +1,200 @@
...
+var ErrMismatchedDeviceVersion = errors.New("mismatched device versions")
+
+// RemoveAllFromDirectory used in place of os.RemoveAll when the directory should be emptied but not removed.
+func RemoveAllFromDirectory(dir string) error {
+	exists, err := vgfs.PathExists(dir)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return nil
```
pscott31, line 25: super minor but maybe it should be an error if the dir doesn't exist?

datanode/dehistory/snapshot/block_commit_handler.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+	}
+
+	if b.snapshotRequiredAtBlockHeight(blockHeight) {
+		fromHeight := GetFromHeight(blockHeight, b.blockInterval)
+		b.snapshotData(ctx, chainID, blockHeight, fromHeight)
+	}
+
+	return false
```
pscott31, line 70: This seems to be the only way out of this function, perhaps we should remove the return arg?

pscott31 (dismissed):
A lot of code to grok at once but I've gone through it all as best I can and have made a few minor comments; in general it looks sane to me!

datanode/dehistory/store/store.go
```diff
@@ -0,0 +1,616 @@
...
+
+	for _, peer := range p.ipfsNode.PeerHost.Network().Peers() {
+		if peer.String() == peerIDStr {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (p *Store) AddSnapshotData(ctx context.Context, historySnapshot snapshot.History, currentState snapshot.CurrentState,
```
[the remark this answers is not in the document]
ettec: its possible, sort of, currently only one goroutine calls this thread, so in that sense its not possible. I would have this method private but the caller (dehistory service) sits in a seperate parent package, the caller is however the only one with a ref to the instance of the store and as long as it doesn't  expose that store them there wouldn't  be an issue.   Thoughts?

datanode/gateway/graphql/schema.graphql
```diff
@@ -4074,6 +4077,18 @@ type NetworkLimits {
   proposeAssetEnabledFrom: Timestamp!
 }
 
+"A segment of datanode history"
+type HistorySegment {
+    "From height of the history segment"
+    fromHeight:       Int!
+    "To height of the history segment"
```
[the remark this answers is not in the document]
ettec: I've updated it to explicitly say block height, this is returned data, not a request for data

datanode/gateway/graphql/schema.graphql
```diff
@@ -4074,6 +4077,18 @@ type NetworkLimits {
   proposeAssetEnabledFrom: Timestamp!
 }
 
+"A segment of datanode history"
+type HistorySegment {
+    "From height of the history segment"
```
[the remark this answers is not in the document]
ettec: I've updated it to explicitly say block height, this is returned data, not a request for data

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1999,3 +2013,50 @@ message GetProtocolUpgradeStatusRequest {}
...
+
+message HistorySegment {
+    int64 from_height = 1;
+    int64 to_height = 2;
+    string chain_id = 3;
+    string history_segment_id = 4;
+    string previous_history_segment_id = 5;
+}
+
+// -- DeHistory --
```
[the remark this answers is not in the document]
ettec: pdated with the summary as suggested and a link to the Decentalized History readme

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -373,6 +373,20 @@ service TradingDataService {
 
     // Protocol Upgrade status
     rpc GetProtocolUpgradeStatus(GetProtocolUpgradeStatusRequest) returns (GetProtocolUpgradeStatusResponse);
+
+    // Decentralised History
+    //
+    // Get the most recent history segment
```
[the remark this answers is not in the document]
ettec: updated with the summary as suggested and a link to the Decentalized History readme

datanode/dehistory/fsutil/fsutil.go
ettec: ah yeah, thats nto clear, renamed the method to RemoveAllFromDirectoryIfExists

datanode/dehistory/fsutil/fsutil.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to walk directory:%w", err)
+	}
+
+	return nil
+}
+
+func MkdirAllIgnoringUMask(dir string) error {
```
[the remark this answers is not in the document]
ettec: good point, that was a refactor overlook, I've updated the error messages   (and here we see why error wrapping is good ;-)  but I'm not going there again :0)

datanode/dehistory/store/store_ipfs.go
```diff
@@ -0,0 +1,303 @@
+package store
```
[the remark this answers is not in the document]
ettec: the file was getting unwieldy so I split it into 2, with ipfs related stuff in the seconds, however, I think theres been a fair bit of refactrogin since then and the split is no longer adhered to and the files are short so I've compressed back into the one file store.go

datanode/dehistory/snapshot/block_commit_handler.go
ettec: yeah, sorted this and remove that bool return val, seems it was only being used in tests.  now uses the context instead to cause the broker to exit

datanode/dehistory/store/store_ipfs.go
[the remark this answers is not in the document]
pscott31: fair enough, not a big deal

pscott31 (dismissed):

pscott31 (dismissed):

cmd/data-node/commands/init.go
```diff
@@ -31,10 +32,16 @@ type InitCmd struct {
 
 var initCmd InitCmd
 
-func (opts *InitCmd) Execute(_ []string) error {
+func (opts *InitCmd) Execute(args []string) error {
 	logger := logging.NewLoggerFromConfig(logging.NewDefaultConfig())
 	defer logger.AtExit()
 
+	if len(args) != 1 {
```
wwestgarth, line 39: Do we really need to provide this during init, and not just left for the user to fill in the config file afterwards (or they can do `datanode dehistory load --chainID=BLAH`? Feels odd that it needs to be supplied at init like its tied to this initialisation of the datanode, but then you can just change it at any time in the config?

ettec: the datanode uses the chain id to connect to the correct swarm at startup, such that when the user does `data-node node` it can automatically determine its peer set, get the latest history segment, load it, and start.  Without this the alternative would mean either passing it in on `data-node node` each time, or having the user do a seperate process to load the data.  In order to make the experience of starting up a new node as seemless as possible we want to avoid the latter and have it basically get its self into the correct state to be able to process the latest blocks without any intervention.

wwestgarth: Sure, but its in a config file so I can change it at any point right? What happens if I do that, stop the data-node then start it again with a different chainID? I hoping something with tell me I'm being silly.

I just think anything that gets set at the `init` call should only be able to be changed if I reinitialise with `init --force`.

We also have a check in the broker that makes sure that the chainID sent by the core node never changes by saving the first chainID it receives, and thats saved into the database (I think):
```
func checkChainID(expectedChainInfo ChainInfoI, chainID string) error {}
```
I think this now probably needs to be extended to make sure that the core node we're connected to is from the same chainID we initialised the datanode to be with.

ettec: rgds the broker it already has a check for that, if data is loaded from decentralized history and then the node starts to consume data from a different chain it will error as loading decentralized history sets the chain id that is used by the broker to verify the events chain.   I agree it would be preferable to use the chain set in config for this check as well though, I'll make that change.

It doesn't explicitly check the chain id in config matches that in data at startup at the moment, it probably should this will prevent any unexpected behaviour, I'll add that.

The main thing here is are we happy with it being passed at init?  I can think of other ways of doing this, but they seem a bit brittle and over-complicated and make it harder to start a data-node in isolation, I purposely raised this at the demo and there was silence, which was sort of agreement I guess ;-)

wwestgarth: Haha sure keep it at during `init`, you're right that no one else complained at the demo.

AnExsomnis (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -428,6 +428,22 @@ service TradingDataService {
 
     // Protocol Upgrade status
     rpc GetProtocolUpgradeStatus(GetProtocolUpgradeStatusRequest) returns (GetProtocolUpgradeStatusResponse);
+
+    // -- Decentralized History --
+    // Decentralised History allows the data-node to reach out to peer nodes to fetch the most recent history, as well as
+    // older history if desired, such that it can quickly get itself up to the latest block height of the network and start
+    // to consume events for the latest block from the Vega core.
+    // See https://github.com/vegaprotocol/vega/tree/develop/datanode/dehistory/README.md for a full description of Decentralised History
```
candida-d, line 436:
```suggestion
    // Decentralized History
    //
    // Decentralised History allows the data-node to reach out to peer nodes to fetch the most recent history, as well as
    // older history if desired, such that it can quickly get itself up to the latest block height of the network and start
    // to consume events for the latest block from the Vega core.
    // See https://github.com/vegaprotocol/vega/tree/develop/datanode/dehistory/README.md for a full description of Decentralised History
```

candida-d: The line break here means that "Decentralized History" works as a title in the docs, rather than just some text. It will make things look a lot nicer/more organised when generated

AnExsomnis (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2072,3 +2105,53 @@ message GetProtocolUpgradeStatusRequest {}
...
+    string chain_id = 3;
+    string history_segment_id = 4;
+    string previous_history_segment_id = 5;
+}
+
+// -- Decentralized History --
+// Decentralised History allows the data-node to reach out to peer nodes to fetch the most recent history, as well as
+// older history if desired, such that it can quickly get itself up to the latest block height of the network and start
+// to consume events for the latest block from the Vega core.
+// See https://github.com/vegaprotocol/vega/tree/develop/datanode/dehistory/README.md for a full description of Decentralised History
```
candida-d, line 2123:
```suggestion
// Decentralized History
//
// Decentralised History allows the data-node to reach out to peer nodes to fetch the most recent history, as well as
// older history if desired, such that it can quickly get itself up to the latest block height of the network and start
// to consume events for the latest block from the Vega core.
// See https://github.com/vegaprotocol/vega/tree/develop/datanode/dehistory/README.md for a full description of Decentralised History
```

candida-d (dismissed):
Approved, but still needs another FE reviewer to sign off on the GQL changes

pscott31 (dismissed):

AnExsomnis (dismissed):

MadalinaRaicu (dismissed):

pscott31 (approved):

candida-d (approved):

MadalinaRaicu (approved):

PR #6514, opened 2022-10-14
Improve and document balance history API

### LINKED ISSUE #6513: GetBalanceHistory document/refactor/rename
GetBalanceHistory should be ListBalanceHistory
The current API is confusing, no one gets the 'grouping' feature
It currently aggregates everything by default (including balances of different assets)
It also leaks account IDs

pscott31:
Closes #6513


datanode/gateway/graphql/resolvers.go
```diff
@@ -1291,28 +1291,36 @@ func (r *myQueryResolver) Statistics(ctx context.Context) (*vegaprotoapi.Statist
...
 	}
 	return resp.GetBalances(), nil
 }
 
+func unboxBool(b *bool) bool {
+	if b == nil {
+		return false
+	}
+	return *b
+}
```
jeremyletang, line 1322: I guess we could add this as some function that we could reuse eventually in the ptr package?
Something like so I suppose:
```
func UnBox[T any](v *T) (out T) {
	if v != nil {
		return *v
	}
	return
}
```

pscott31: Yeah that's quite nice, I'll do that.. need to fix some linty stuff anyway

pscott31: Added this to the 'ptr' package

datanode/gateway/graphql/schema.graphql
```diff
@@ -509,12 +509,16 @@ type Query {
...
-  historicBalances(
-    "Optional filter to restrict the result to a specific account or asset"
+  "Get historical balances for an account within the given date range, with optional netting over party, market and account type"
+  balanceHistory(
+    "Optional filter to restrict queried accounts to those of a specific asset, party, market or type"
     filter: AccountFilter,
-    "Optional grouping clause for the results"
-    groupBy: [AccountField],
+    "Whether balances for different parties should be netted together. If so, party_id of results will be null"
+    sum_across_parties: Boolean,
```
mattrussell36, line 517: chore (blocking): Can we use camel case for consistentcy

pscott31: fixed

AnExsomnis (commented):
Seems all right to me.
But I have some concerns about `ListBalanceHistory` name - is `list` more suitable to countables?
And balance and history are uncountable? Or maybe I misinterpret it :thinking:

mattrussell36 (dismissed):

pscott31:
> Seems all right to me. But I have some concerns about `ListBalanceHistory` name - is `list` more suitable to countables? And balance and history are uncountable? Or maybe I misinterpret it 🤔

I'm not sure what you mean by uncountable - I mean there are a finite number of them so they aren't uncountable in the [mathematical ](https://en.wikipedia.org/wiki/Uncountable_set)sense. The convention we started (for the grpc api at least) is that things that return a list are called 'ListXXX' and things that return a single item are 'GetXXX', IIRC this comes from some [google naming](https://cloud.google.com/apis/design/naming_convention#method_names) convention that ways you should name things VerbNoun.

That said, conventions aren't stuck in stone - what would you suggest instead?

AnExsomnis:
> > Seems all right to me. But I have some concerns about `ListBalanceHistory` name - is `list` more suitable to countables? And balance and history are uncountable? Or maybe I misinterpret it thinking
> 
> I'm not sure what you mean by uncountable - I mean there are a finite number of them so they aren't uncountable in the [mathematical ](https://en.wikipedia.org/wiki/Uncountable_set)sense. The convention we started (for the grpc api at least) is that things that return a list are called 'ListXXX' and things that return a single item are 'GetXXX', IIRC this comes from some [google naming](https://cloud.google.com/apis/design/naming_convention#method_names) convention that ways you should name things VerbNoun.
> 
> That said, conventions aren't stuck in stone - what would you suggest instead?

Yes I meant countable/uncountable in mathematical sense; the balance history _content_ contains a finite number of objects, that makes sense to be _listed_. My comment was about having "history" and "balance" with list in the naming; I would have it something like "ListBalanceHistoryRecords" for example or something like that. I would not insist though, since I am not a native speaker, but seems a little better to me.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
     //   - listing ledger entries with filtering on the receiving account (party_id, market_id, asset_id, account_type)
     //   - listing ledger entries with filtering on the sending AND receiving account
     //   - listing ledger entries with filtering on the transfer type (on top of above filters or as a standalone option)
-    rpc GetBalanceHistory(GetBalanceHistoryRequest) returns (GetBalanceHistoryResponse) {}
+    rpc ListLedgerEntries(ListLedgerEntriesRequest) returns (ListLedgerEntriesResponse) {}
+
+    // Balances
+    //
+    // ListBalanceHistory is for querying the change in balances of accounts over a period of time.
```
candida-d, line 84:
```suggestion
    // ListBalanceHistory is for querying the change in account balances over a period of time.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    //
+    // An account is defined a 4-tuple of (asset_id, type, party_id, market_id).
+    //   - Every acount has associated asset and type.
```
candida-d, line 87:
```suggestion
    //   - Every account has associated asset and type.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    //   - Certain types of account (for example, the global reward pool) do not have an associated party.
```
candida-d, line 88:
```suggestion
    //   - Certain account types (for example, the global reward pool) do not have an associated party.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    //     These are denoted by the special party identifier '*'
+    //   - Certain types of account do not have an associated market (for example general party accounts)
```
candida-d, line 90:
```suggestion
    //   - Certain account types do not have an associated market (for example general accounts)
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    //
+    // For example, suppose that we had stored in our database the following balance history:
+    //
+    // vega_time  party_id    asset_id    market_id   account_type   balance
+    // 1          dave        eth         !           general        1
+    // 2          dave        eth         ethusd      margin         10
+    // 2          jim         eth         !           general        100
+    // 3          dave        eth         !           general        2
+    //
+    // NB: for clarity I am using example names for party, market and asset id; they are acutally 256bit hex encoded hashes.
```
candida-d, line 108:
```suggestion
    // NB: for clarity this uses example names for party, market and asset id; they are actually 256bit hex encoded hashes.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    //     These are denoted by the special market identifier '!'
+    //
+    // In its default usage, ListBalanceHistoryRequest will return a list of:
+    // (vega_time, asset_id, account_type, party_id, market_id, balance)
+    // With a set of entries for each block in which the balance of any account changes.
+    //
+    // Balances are 'filled forward', so for a given a set of results from ListBalanceHistory, looking
+    // at rows for a given vega_time will give you the complete set of balances at that time.
+    //
+    // For example, suppose that we had stored in our database the following balance history:
```
candida-d, line 100:
```suggestion
    // For example, suppose that stored in the database was the following balance history:
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    // 2          dave        eth         ethusd      margin         10
+    // 2          jim         eth         !           general        100
+    // 3          dave        eth         !           general        2
+    // 3          dave        eth         ethusd      general        10
+    // 3          jim         eth         !           general        100
+    //
+    // You may optionally filter which accounts are queried using the `filter` argument.
+    //
+    // You may also optionally aggregate balances using the `sum_across_party`, `sum_across_markets`,
+    // and `sum_across_types` parameters. For example, lets say you want to see the balances of
```
candida-d, line 124:
```suggestion
    // and `sum_across_types` parameters. For example, let's say you want to see the balances of
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    // You can query that by passing `sum_accross_markets=true` and `sum_across_types=true` which would
+    // return
+    //
+    // vega_time  party_id    asset_id    market_id   account_type   balance
+    // 1          dave        eth                                    1
+    // 2          dave        eth                                    11
+    // 2          jim         eth                                    100
+    // 3          dave        eth                                    12
+    // 3          jim         eth                                    100
+    rpc ListBalanceHistory(ListBalanceHistoryRequest) returns (ListBalanceHistoryResponse) {}
```
candida-d, line 136: @pscott31 Would you mind also fixing the spelling error on line 148 - 'market detph' while you're in here?

candida-d (commented):
So much documentation! :D 

I left suggestions for typo fixes & other small things - and I hope you can find it in your heart to fix an unrelated spelling error in a description field

pscott31:
> Yes I meant countable/uncountable in mathematical sense; the balance history _content_ contains a finite number of objects, that makes sense to be _listed_. My comment was about having "history" and "balance" with list in the naming; I would have it something like "ListBalanceHistoryRecords" for example or something like that. I would not insist though, since I am not a native speaker, but seems a little better to me.

hm, yes, I see your point - though 'Records' seems somewhat redundant, e.g. if you had a API point which listed books in a library it would be `ListBooks`, not `ListBookRecords`; but then `ListBalances` to me feels like it should be showing you _current_
balances rather than changes in them over time.

How about `ListBalanceChanges`

mattrussell36 (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -81,7 +77,63 @@ service TradingDataService {
...
+    //
+    // For example, suppose that stored in the database was the following balance history:
+    //
+    // vega_time  party_id    asset_id    market_id   account_type   balance
+    // 1          dave        eth         !           general        1
+    // 2          dave        eth         ethusd      margin         10
+    // 2          jim         eth         !           general        100
+    // 3          dave        eth         !           general        2
+    //
+    // NB: for clarity this uses example names for party, market and asset id; they are acutally 256bit hex encoded hashes.
```
edd, line 108:
```suggestion
    // NB: for clarity this uses example names for party, market and asset id; they are actually 256bit hex encoded hashes.
```

typo fix

jeremyletang (dismissed):

AnExsomnis (dismissed):

candida-d (approved):

jeremyletang (approved):

PR #6518, opened 2022-10-17
fix: ignore parties with zero delegation in reward payout event

### LINKED ISSUE #6510: Reward payments produced with zero payment amount
Where the reward pool is exhausted during a period of POS calculations for rewards we do sometimes generate zero payment events.

This should not happen to reduce datanode work.

ze97286:
Closes #6510

### DIFF
--- a/core/rewards/staking_reward_calculator_test.go
+++ b/core/rewards/staking_reward_calculator_test.go
@@ -63,6 +63,27 @@ func testCalcRewardsZeroScores(t *testing.T) {
 	require.Equal(t, 0, len(res.partyToAmount))
 }
 
+func TestFilterZeros(t *testing.T) {
+	delegatorShare, _ := num.DecimalFromString("0.3")
+	scores := map[string]num.Decimal{}
+	scores["node1"] = num.NewDecimalFromFloat(1)
+	scores["node2"] = num.DecimalZero()
+	scores["node3"] = num.DecimalZero()
+	scores["node4"] = num.DecimalZero()
+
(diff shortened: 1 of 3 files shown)


jeremyletang (dismissed):

core/rewards/staking_reward_calculator_test.go
```diff
@@ -63,6 +63,19 @@ func testCalcRewardsZeroScores(t *testing.T) {
...
+	delegatorShare, _ := num.DecimalFromString("0.3")
+	scores := map[string]num.Decimal{}
+	scores["node1"] = num.NewDecimalFromFloat(1)
+	scores["node2"] = num.DecimalZero()
+	scores["node3"] = num.DecimalZero()
+	scores["node4"] = num.DecimalZero()
+
+	res := calculateRewardsByStake("1", "asset", "rewardsAccountID", num.NewUint(100000), scores, []*types.ValidatorData{{NodeID: "node1", StakeByDelegators: num.NewUint(500), SelfStake: num.NewUint(1000), Delegators: map[string]*num.Uint{"zohar": num.UintZero(), "jeremy": num.NewUint(500)}}}, delegatorShare, num.UintZero(), rng, logging.NewTestLogger())
+	require.Equal(t, num.NewUint(100000), res.totalReward)
+	require.Equal(t, 2, len(res.partyToAmount))
```
ValentinTrinque, line 76: Can we add a party with 0 delegation there and  verify that there are `zohar` and `jeremy` specifically ?

With this verification, I don't think we can't ensure it's correct.

ValentinTrinque (changes requested):

core/rewards/staking_reward_calculator_test.go
ze97286: I disagree. without this fix the count would return 3, now it returns 2. I am already a party with 0 delegation.

ze97286: but sure, can add verification by party to demonstrate that "zohar" is not there.

ValentinTrinque (dismissed):

pscott31 (approved):

PR #6533, opened 2022-10-18
feat: add last-block sub-command to datanode CLI

### LINKED ISSUE #6527: Add a datanode cmd to get the last block
Requirment to make using visor easier to restart network, add a command to datanode that will return the height of the last block committed, this will make it easier for Visor to figure at what snapshot height is should start the core.  (the same or less than last comitted block height basically)

guoguojin:
Closes #6527

### DIFF
--- a/cmd/data-node/commands/last-block.go
+++ b/cmd/data-node/commands/last-block.go
@@ -38,6 +38,11 @@ func (cmd *LastBlockCmd) Execute(_ []string) error {
 	}
 
 	cmd.Config, err = cfgLoader.Get()
+	if err != nil {
+		handleErr(log, cmd.Output.IsJSON(), "couldn't load configuration", err)
+		os.Exit(1)
+	}
+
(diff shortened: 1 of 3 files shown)


cmd/data-node/commands/last-block.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	var lastBlock int64
+	err = conn.QueryRow(ctx, "select max(height) from blocks").Scan(&lastBlock)
+	if err != nil {
+		log.Error("Unable to retrieve last block", logging.Error(err))
+		os.Exit(1)
+	}
+
+	log.Info("Last block", logging.Int64("height", lastBlock))
```
wwestgarth, line 72: Could it also (optionally) print the answer out in JSON like we do with some of out other commands e.g `vega version --output=json`.

I can imagine this would be something useful for someone to script against?

guoguojin: Sure, I will add that

wwestgarth (approved):

PR #6534, opened 2022-10-18
Feature/6496

### LINKED ISSUE #6496: Improve LP and Pegged orders repricing
As of now when the system reprices pegged orders or liquidity orders, it behave like so:
- remove all pegs/lp from the book
- remove them from the position
- reprice them
- resubmit them one by one in order throught the submitOrder flow

This causes the system to do a margin calculation for every re-submission. Because of this, as the margin is recalculated on every order re submission, the following pattern appears on the ledger movements:
- submit first order, the users have too much margin, instruct collateral to release margin
- submit N next orders, the margin is too low, so re top up all funds because of margin low call.

This basically reduce a lot the margin with the first order, to get it back to the exact level it was before repricing.

We want now to change the flow a little to:
- remove all orders from the book
- remove all potential orders from the positions
- reprice all order
- register all order in the new position
- calculate the expected new position / potential position
- calculate the margin
- if the margin is OK, we can re submit the orders and do 1 margin transfer if necessary
- if the margin is not OK, we go through the distressed traders flow

jeremyletang:
close #6496


Jiajia-Cui (dismissed):
all the feature tests look good to me

core/execution/special_orders.go
```diff
@@ -268,57 +248,183 @@ func (m *Market) updateLPOrders(
...
+
+	// send cancel events
+	m.broker.SendBatch(orderEvts)
+
+	// an ordered list of positions
+	var (
+		positions     = make([]events.MarketPosition, 0, len(partiesPos))
+		marginsBefore = map[string]*num.Uint{}
+		id            = m.GetID()
+		assetID, _    = m.mkt.GetAsset()
```
zale144, line 293: we sure this won't err?

core/execution/special_orders.go
```diff
@@ -268,57 +248,183 @@ func (m *Market) updateLPOrders(
...
+	)
+	// now we can check parties positions
+	for party, pos := range partiesPos {
+		positions = append(positions, pos)
+		mar, _ := m.collateral.GetPartyMarginAccount(id, party, assetID)
```
zale144, line 298: and this?

core/execution/special_orders.go
```diff
@@ -268,57 +248,183 @@ func (m *Market) updateLPOrders(
...
 			}
 		}
 
-		// if an auction has been started, we just break now
-		if m.as.InAuction() {
-			// enteredAuction = true
-			// auctionFrom = i
-			// break
-			return nil
+		_ = m.equityShares.SharesExcept(m.liquidity.GetInactiveParties())
```
zale144, line 331: does this err?

core/execution/special_orders.go
```diff
@@ -268,57 +248,183 @@ func (m *Market) updateLPOrders(
...
+	// alright, here we need to go weird over things because we want to find what
+	// parties have been considered distressed by the risk / collateral engine BUT
+	// for which the LP submission where still pending a first deployment.
+	// In which case no bond slashing is being taken, and they shall not be
+	// closed as well but the lp submission should only be cancelled.
+
+	// so first we will find all pending which would be closed
+	for _, v := range closed {
+		if m.liquidity.IsPending(v.Party()) {
+			_ = m.cancelPendingLiquidityProvision(
```
zale144, line 359: err here? does it matter?

zale144 (dismissed):
a couple of minor comments, otherwise lgtm

core/execution/special_orders.go
jeremyletang: nope doesnt' mattert here, it cannot fail, API needs to be fixed later to reflect this.

core/execution/special_orders.go
jeremyletang: yep, couldn't even instantiate a market without this, it's a bit of a annoying api as well a panic might somehow be better

core/execution/2965_refresh_lp_orders_size_test.go
```diff
@@ -233,32 +234,33 @@ func TestRefreshLiquidityProvisionOrdersSizes(t *testing.T) {
...
 				types.OrderStatusActive,
-				0x341, // cancelled
+				833, // cancelled
 			},
 		}
 
 		require.Len(t, found, len(expectedStatus))
 
 		for i, expect := range expectedStatus {
+			fmt.Printf("FOUND: %s\n\n", found[i].String())
```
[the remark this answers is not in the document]
jeremyletang: that's debug, I'll remove.

core/execution/special_orders.go
jeremyletang: Nope, it returns a value that we don't need here.

core/execution/special_orders.go
jeremyletang: It could if something went very wrong down the line, but seeing that we have a position for this party, it MUST have a margin account, happy to had a Panic there if you'd like?

zale144: I think we should always handle errors, even if we are sure that they won't happen. Sometimes we are sure that they cannot happen but they still do. If an error does somehow happen, we will have an easier time tracking it down.  But if anything, we should do it for consistency :)

jeremyletang: fair enough, here a panic would do then I suppose.

core/execution/order_test.go
```diff
@@ -1893,36 +1893,36 @@ func testPeggedOrderOutputMessages(t *testing.T) {
 	confirmation, err := tm.market.SubmitOrder(ctx, &order)
 	require.NoError(t, err)
 	assert.NotNil(t, confirmation)
-	assert.Equal(t, uint64(13), tm.orderEventCount)
+	assert.Equal(t, int(11), int(tm.orderEventCount))
```
zale144, line 1896: no need to cast `11` into `int`. Same for all the ones below

zale144 (dismissed):

zale144 (dismissed):
Lgtm

witgaw (approved):
`.feature` file changes LGTM

PR #6537, opened 2022-10-18
fix: fix oracleDatabySpecConnectionId response

### LINKED ISSUE #6234: [datanode] graphql query for oracleDataBySpecConnection
In testnet and stagnet3:

```graphql
{
  oracleDataBySpecConnection(oracleSpecId:"test") {
    pageInfo {
      hasNextPage
    }
  }
}
```

# Expected result
```json
{
  "data": {
    "oracleDataBySpecConnection": {
      "pageInfo": {
        "hasNextPage": false
      }
    }
  }
}
```

# Actual result
```json

{
  "errors": [
    {
      "message": "rpc error: code = InvalidArgument desc = InvalidArgument error",
      "path": [
        "oracleDataBySpecConnection"
      ]
    }
  ],
  "data": null
}

```

See also #6230

AnExsomnis:
close: #6234

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -1728,7 +1728,7 @@ func (t *tradingDataServiceV2) ListOracleData(ctx context.Context, req *v2.ListO
 	}
 
 	if err != nil {
-		return nil, apiError(codes.Internal, err)
+		apiError(codes.Internal, ErrOracleServiceGetSpec, fmt.Errorf("could not retrieve data for OracleSpecID: %s %w", *req.OracleSpecId, err))
 	}
 
 	edges, err := makeEdges[*v2.OracleDataEdge](data)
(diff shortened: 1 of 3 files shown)


datanode/api/trading_data_v2.go
```diff
@@ -1728,7 +1728,7 @@ func (t *tradingDataServiceV2) ListOracleData(ctx context.Context, req *v2.ListO
 	}
 
 	if err != nil {
-		return nil, apiError(codes.Internal, err)
+		return nil, fmt.Errorf("could not retrieve data for OracleSpecId: %s %w", *req.OracleSpecId, err)
```
pscott31, line 1731: I appreciate we haven't been consistently doing this, but my understanding that it was best practice to return apiErrors() from GRPC endpoints? 

e.g. maybe in api/errors.go 

```
ErrOracleServiceGetSpec         = errors.New("failed retrieve data for oracle spec")
```
then in trading_data_v2.go something like :
```
apiError(codes.Internal, ErrOracleServiceGetSpec, fmt.Errorf("could not retrieve data for OracleSpecId: %s %w", *req.OracleSpecId, err) )
```

Maybe we should have a README or something where we specify what the best practice is here, as I know that personally I haven't been that consistent about it.

AnExsomnis: @pscott31 , This has been afressed here -> https://github.com/vegaprotocol/vega/pull/6537/commits/b11087201a7cc8434141599089f4f2ddb50774c5#diff-a493a31fca9e62f51a5f1c9971fa04a87210e5ff399f4737d5f8a610c5eb17c8R1731
I think it would be good to have a README in the datanode/api, was that what you meant?

pscott31 (dismissed):

pscott31 (approved):

PR #6541, opened 2022-10-19
chore: remove conflation

### LINKED ISSUE #6542: Remove conflation of data - now have decentralized history
Original purpose of the conflation was to allow a course grained sample of historical data to kept whilst still being able to have a reasonably sized database.  With decentralised it is possible to load up history from any point so in time so there is no real need for this behaviour.  The conflation also means that databases with the same block span of data could have potentially different data as the generation of the conflated data is done as a background task.

ettec:
closes #6542

### DIFF
--- a/datanode/sqlstore/balances.go
+++ b/datanode/sqlstore/balances.go
@@ -100,14 +100,14 @@ func (bs *Balances) Query(filter entities.AccountFilter, groupBy []entities.Acco
 	// This query is the one that gives us our results
 	query := `
         WITH our_accounts AS (%s),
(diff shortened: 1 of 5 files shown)


pscott31 (commented):
I love to see simplification and deleted code but I worry we're throwing the baby out with the bathwater a bit here.. I think part of the initial data-node spec from Barney was that we be able to keep 'conflated' (e.g. hourly/daily) summaries of data for longer than the full raw set.. I can see that this gets rather fiddly in conjunction with decentralised history and nodes not having all the raw data..

I guess we need to carefully define what is required, and maybe have something in-between 'current' and 'full history' notions of data

PR #6560, opened 2022-10-21
chore: require filtering on party for balance changes

pscott31:


datanode/api/trading_data_v2.go
```diff
@@ -282,6 +282,10 @@ func (t *tradingDataServiceV2) ListBalanceChanges(ctx context.Context, req *v2.L
 		return nil, fmt.Errorf("parsing filter: %w", err)
 	}
 
+	if len(filter.PartyIDs) == 0 {
+		return nil, apiError(codes.InvalidArgument, fmt.Errorf("You filter by at least one party ID"))
+	}
+
```
AnExsomnis, line 288: I think the idea was to limit the querying to one party only, so you may want to change that to 
if len(filter.PartyIds) != 1
becase in that way they can give large number of parties.

Please correct me if I am wrong.

AnExsomnis: I have added a separate error for that in my PR: 
```var ErrLedgerEntryFilterForParty = fmt.Errorf("filtering ledger entries should be limited to a single party")```
You can use it (needs an adjustment for that since my PR is not merged at the moment), or if you have better idea - I can change it from my side also.

jeremyletang: correct @AnExsomnis

pscott31:
I think this is no longer needed now we have the simplified API without grouping

PR #6573, opened 2022-10-22
Restructure logger used in wallet for reusability

ValentinTrinque:


libs/zap/config_windows.go
```diff
@@ -1,4 +1,6 @@
-package cmd
+//go:build !windows
```
jeremyletang, line 1: _windows file with go:build !windows ?

ValentinTrinque: yes it's a mistake

jeremyletang (dismissed):

jeremyletang (approved):

PR #6578, opened 2022-10-24
chore: dns remove data merge from dehistory

### LINKED ISSUE #6577: Remove datamerge from dehistory
Original datanode snapshot behaviour had been to snapshot datanode data before restoring it from snapshots.  With the way decentralized history works this state is local anyway and so there is no need to snapshot the datanode data before restoring from decentralized history, this behaivour should be removed.

ettec:
closes #6577 

Original datanode snapshot behaviour had been to snapshot datanode data before restoring it from snapshots.  With the way decentralized history works this state is local anyway and so there is no need to snapshot the datanode data before restoring from decentralized history, this behaivour should be removed.

### DIFF
--- a/cmd/data-node/commands/dehistory/show.go
+++ b/cmd/data-node/commands/dehistory/show.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sort"
 
+	"code.vegaprotocol.io/vega/datanode/dehistory/aggregation"
+
 	"code.vegaprotocol.io/vega/datanode/dehistory/initialise"
 	"code.vegaprotocol.io/vega/logging"
 	v2 "code.vegaprotocol.io/vega/protos/data-node/api/v2"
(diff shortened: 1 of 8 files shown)


datanode/dehistory/snapshot/service_create_snapshot.go
```diff
@@ -84,26 +76,18 @@ func (b *Service) createSnapshot(ctx context.Context, chainID string, fromHeight
...
+	cleanUp = append(cleanUp, func() { _ = os.Remove(snapshotInProgressFile) })
+
+	go func() {
 		defer func() { runAllInReverseOrder(cleanUp) }()
 		err = b.snapshotData(ctx, copyDataTx, dbMetaData, currentSnapshot, historySnapshot)
 		if err != nil {
-			return CreateSnapshotResult{}, fmt.Errorf("failed to snapshot data:%w", err)
+			b.log.Error("failed to snapshot data", logging.Error(err))
+			if b.config.PanicOnSnapshotCreationError {
+				panic(fmt.Sprintf("failed to snapshot data:%s", err))
```
jeremyletang, line 87: why not use b.log.Panic() here instead?

cmd/data-node/commands/dehistory/show.go
```diff
@@ -65,6 +67,19 @@ func (cmd *showCmd) Execute(_ []string) error {
 		fmt.Printf("%s\n", segment)
 	}
 
+	contiguousHistory := GetHighestContiguousHistoryFromHistorySegments(segments)
+	// Sort history oldest first
+	sort.Slice(contiguousHistory, func(i, j int) bool {
```
wwestgarth, line 72: Isn't it already sorted by now? `GetHighestContiguousHistoryFromHistorySegments` returns it sorted (via `GetHighestContiguousHistory`)

datanode/dehistory/snapshot/service_create_snapshot.go
ettec: it did originally, but got an error using that and it did not log the original error, so I change it to this instead so that we always get the cause

cmd/data-node/commands/dehistory/show.go
ettec: no, the method returns latest first

jeremyletang (approved):

PR #6586, opened 2022-10-24
chore: improve TransferType mapping

### LINKED ISSUE #6482: Improve TransferType handling in datanode models
Currently transfer type consts are manually defined inside datanode/gateway/graphql/models.go
Replace the definitions with auto generated from TransferType_name / TransferType_value maps that protobufs makes.
For reference see AccountType

Purpose is to decrease the number of places that need editing when a new type is added/removed.

AnExsomnis:
close: #6482


pscott31:
While were at it, I noticed the other day that transfer types are stored in the ledger table as strings not enums; we should probably change that as it's a fair few bytes for for each of the very many ledger entries we have.. but maybe that's another ticket :)

jeremyletang (approved):

PR #6588, opened 2022-10-24
feat: match validation to the required ranges

### LINKED ISSUE #6582: Check / Update Network Parameter Validation Ranges
# Feature Overview

**In order to** ensure that the network parameters are within the specified ranges
**We will** add the ranges into the code for each network parameter (somewhere "next to" keys.go ?)
**So that** we can test that the parameters are within range and link to these files from the spec so the spec does not go out-of-date

## Specs
- [Ranges to be added](https://docs.google.com/spreadsheets/d/1tz9t2wYJ5kEFkngfYowPWKHhN4aTgRFM2O2rSOAoyvw/edit#gid=453838910)

# Tasks
- [ ] add the ranges (if not added) or check they match those defined in the spreadsheet

Once done @gordsport to open a specs issue to:
- [ ] link the files in the specs
- [ ] create acceptance criteria (and codes) for the net param ranges

# Definition of Done

**Before Merging**

**After Merging**

ze97286:
Closes #6582


davidsiska-vega:
@ze97286 Question: one thing I cannot figure out looking at the PR is: 
1) does the value in a network parameter change proposal that's submitted to the network get checked against these ranges? 

2) Why can't we (as some additional step somewhere) also do the additional validations (like scaling factors being sorted right)?

ze97286:
> @ze97286 Question: one thing I cannot figure out looking at the PR is:
> 
> 1. does the value in a network parameter change proposal that's submitted to the network get checked against these ranges?
> 2. Why can't we (as some additional step somewhere) also do the additional validations (like scaling factors being sorted right)?

1. The ranges mean that it is infeasible to set the network parameter into a value that is not in the range. the network parameter will refuse to accept a value that does not pass the check whether it's coming from an update or genesis or checkpoint or whatever. 
2. it's just not structured like this. The validator of network parameter a has no access to the value of network parameter b. I think you're thinking about it as a function that is applied at the time we're modifying some parameter, but it's not exactly. As in (1) here above, regardless of the source of the value, genesis, checkpoint, update request, it would make it infeasible for a value to be out of the defined range. I'm not saying it can't be done, but I am saying it's not written this way. A [dirty] workaround could be to wrap dependent parameters together like we do with the scaling factors.

davidsiska-vega:
> 1. it is infeasible to set the network parameter into a value that is not in the range.

But would the check be made at the time the network parameter change is submitted (before voting starts)? Or would the check only happen once the voting ends and the system tries to set it?

ze97286:
> > 1. it is infeasible to set the network parameter into a value that is not in the range.
> 
> But would the check be made at the time the network parameter change is submitted (before voting starts)? Or would the check only happen once the voting ends and the system tries to set it?

upon submission, not upon enactment.

davidsiska-vega:
> 2\. I think you're thinking about it as a function that is applied at the time we're modifying some parameter, but it's not exactly.

Yes, if I was writing the code(*) I'd do another pass after all the values have been set (maybe in some temporary copy that can be discarded or whatever) to check the "additional constraints" are also satisfied and if not the code should refuse to start / restore / start the voting process on the network parameter; whichever applies. Also another check should run at the end of the voting process (if the vote was sucessful) and check that the new value is still valid (because other params may have changed in the meantime).

jeremyletang (approved):

Vegaklaus:
> > > 1. it is infeasible to set the network parameter into a value that is not in the range.
> > 
> > 
> > But would the check be made at the time the network parameter change is submitted (before voting starts)? Or would the check only happen once the voting ends and the system tries to set it?
> 
> upon submission, not upon enactment.

Could we do both ? Once we add parameter dependencies, a parameter value can be valid on submission, and invalid on enactment, so some range check needs to be done on both sides anyhow

ze97286:
once we do, it's a requirement. It's specified on the ticket.

PR #6592, opened 2022-10-24
fix: Return vote data when querying for proposals

### LINKED ISSUE #6356: Querying for Proposals doesn't return vote data
Running a proposals query doe not return vote data.

**Example Query:**

```
query {
  proposalsConnection {
    edges {
      node {
        id
        state
        errorDetails
        rejectionReason
        
        votes {
         yes {
          totalNumber
          totalWeight
          totalTokens
          votes {
            value
            party {
              id
            }
            proposalId
          }
        }
          no {
            totalNumber
            totalWeight
            totalTokens
          votes {
            value
            party {
              id
            }
            proposalId
          }
        }
        }
      }
    }
  }
}

```
**Returns**

```
{
  “data”: {
    “proposalsConnection”: {
      “edges”: [
        {
          “node”: {
            “id”: “99cdd22a9c98650d315b783976d199d338d4987486fb8bb1c80e8b5c172a6647",
            “state”: “STATE_ENACTED”,
            “errorDetails”: null,
            “rejectionReason”: null,
            “votes”: {
              “yes”: {
                “totalNumber”: “0",
                “totalWeight”: “0",
                “totalTokens”: “0",
                “votes”: null
              },
              “no”: {
                “totalNumber”: “0",
                “totalWeight”: “0",
                “totalTokens”: “0",
                “votes”: null
              }
            }
          }
        },
        {
          “node”: {
            “id”: “2fddd551bbf190bc00c796432c133da46cf59268912d4b23bea36de1715a2eb9",
            “state”: “STATE_ENACTED”,
            “errorDetails”: null,
            “rejectionReason”: null,
            “votes”: {
              “yes”: {
                “totalNumber”: “0",
                “totalWeight”: “0",
                “totalTokens”: “0",
                “votes”: null
              },
              “no”: {
                “totalNumber”: “0",
                “totalWeight”: “0",
                “totalTokens”: “0",
                “votes”: null
              }
            }
          }
        },
        {
          “node”: {
            “id”: “f717e85e90eeb22e600f8655b9159e99a39476e67e775282080883d40ae733b7",
            “state”: “STATE_ENACTED”,
            “errorDetails”: null,
            “rejectionReason”: null,
            “votes”: {
              “yes”: {
                “totalNumber”: “0",
                “totalWeight”: “0",
                “totalTokens”: “0",
                “votes”: null
              },
              “no”: {
                “totalNumber”: “0",
                “totalWeight”: “0",
                “totalTokens”: “0",
                “votes”: null
              }
            }
          }
        }
      ]
    }
  }
}
```

zale144:
Closes #6356


datanode/gateway/graphql/proposal_resolver.go
```diff
@@ -88,11 +89,20 @@ func (r *proposalResolver) Terms(_ context.Context, data *types.GovernanceData)
 	return data.Proposal.Terms, nil
 }
 
-func (r *proposalResolver) Votes(_ context.Context, obj *types.GovernanceData) (*ProposalVotes, error) {
-	if obj == nil {
+func (r *proposalResolver) Votes(ctx context.Context, obj *types.GovernanceData) (*ProposalVotes, error) {
+	if obj == nil || obj.Proposal == nil {
```
wwestgarth, line 93: I don't think this is right as it only solves it for the graphQL endpoint and not the call rpc call `ListGoveranceData`. We should already have the vote data in `obj` by now.

If you look at the call to `func (t *tradingDataServiceV2) GetGovernanceData()` we end up calling `t.governanceService.GetNoVotesForProposal(ctx, proposal.ID.String())` to populate the votes.

I believe we just need a similar call in the `ListGoveranceData()` probably in `func (p Proposal) ToProtoEdge()`.

datanode/api/trading_data_v2.go
```diff
@@ -2644,6 +2644,31 @@ func (t *tradingDataServiceV2) proposalToGovernanceData(ctx context.Context, pro
 	return &gd, nil
 }
 
+func (t *tradingDataServiceV2) GetVotesByProposal(ctx context.Context, req *v2.GetVotesByProposalRequest) (*v2.GetVotesByProposalResponse, error) {
```
wwestgarth, line 2647: I'm not sure if this endpoint has any value, you can get the same data returned by calling `GetGovernanceData()` and passing in a proposal ID?

datanode/gateway/graphql/proposal_resolver.go
zale144: @wwestgarth Yeah, that was my initial idea, to just do the same thing as with `GetGovernanceData()`, however the problem I had with that is that the query can end up being a bit "expensive". For every single proposal we would be querying a list of votes, even worse separate query for "yes" and for "no" votes.

So I thought maybe just pull the votes when GQL asks for them. Another way would be to pass a flag down to rpc to ask to include the votes (or not), but that seemed hacky and not consistent with the rest of the endpoints.

wwestgarth: Yea I don't really know the perf impact (although I suspect the total proposal/vote data will never grow that large really) -- I'll delegate to the postgres-optimisation-pro @pscott31. Any ideas why we do a separate query to get the votes, and not just one query that looks at two tables?

zale144: @pscott31 Any ideas on this?

pscott31: I think we should do it in the GRPC endpoint like we do in `GetGovernanceData` (I think I suggested this on slack IIRC). If it turns out to be too expensive we can push it down and do it all in one query, but I think there are sufficiently few of these that it doesn't really matter

zale144:
Moved to #6658

PR #6594, opened 2022-10-25
fix: ensure liquidity order book remains in sync when amendng an order

### LINKED ISSUE #6563: Snapshots: failure seen on stagnet1 pipeline
# Problem encountered
Snapshot failures are being seen on the stagnet1 snapshot-pipeline

We need to investigate and fix.

# Observed behaviour
Snapshots created after block ~40000 fall out of consensus with the following error:
```
[2022-10-21T10:34:00.773Z] 2022-10-21T10:34:00.603Z	ERROR	tendermint	v0/reactor.go:379	Error in validation	{"err": "wrong Block.Header.LastResultsHash.  Expected 2929F9ECA252194F20D5846E6DD23A4B69EEA006433A0A8993F0D0740C3C1D76, got 18E072A5DCBF652CED0043A929675258C5B243533EC043302B5001C48ED0B4BA"}
```

# Expected behaviour
All snapshots should work.

# Definition of Done

**Before Merging**

**After Merging**

wwestgarth:
closes #6563

So the symptom in this one was that after loading from a snapshot an attempt to amend a LP order would result in different margin levels being calculated, then we'd fall out of consensus because account balances would differ.

Tracing it back and looking at the data stored in the snapshot it turned out there was a limit order by the same party, stored in the order-book and the liquidity engine, but in each place they had different values for their price and remaining. This limit order had at some point been amended and in the order-book we had the correct amended order but in the liquidity engine we had the old cancelled order. So it looks like when _not_ loading from a snapshot we were using the information from the unamended order to calculate margin levels.

When we load from a snapshot we would ensure that the pointer to the order in the order-book is reused in the liquidity engine which was "accidently" synchronising the orders like they should have been, and this caused the difference.

To fix I have changed the liquidity engine to not longer manage a parallel set of orders, and it now reads directly from the orderbook when it needs to know about existing orders. This has caused a few difference in numbers in the integrations tests which were essentially down to cases where previously the liquidity engine/matchng engine had gotten out of sync.

new system-test for this specific snapshot case: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/20731/pipeline/

full system test run: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/12114/pipeline


core/execution/market.go
```diff
@@ -2456,6 +2456,9 @@ func (m *Market) AmendOrderWithIDGenerator(
 
 	if !m.as.InAuction() {
 		m.checkForReferenceMoves(ctx, allUpdatedOrders, false)
+	} else {
+		// make sure that the amended orders are updated in the liquidity engine too
+		m.liquidity.SynchronisePartyOrders(party, allUpdatedOrders)
```
jeremyletang, line 2461: Isn’t the Update method on liquidity sufficient for that? Can’t remember what parameters it takes tho.

jeremyletang: Actually yea I think we should ideally use the update method or so, which will also recreate order if needed?

wwestgarth: Its the same parameter but it calls stuff like `e.createOrUpdateForParty()` which I wasn't sure if we wanted to do when we literally just want to make the orders the same. 

Theres was also `updatePartyOrders()` which I tried but there is some check in there which we didn't satisfy and then didn't update.

jeremyletang: Alright, I’ll think a little bit over it if you don’t mind before we merge.

wwestgarth: Yea sure thing, rather it be done properly than something that just works.

ze97286: @wwestgarth @jeremyletang I wonder if we're not creating a problem to ourselves, i.e. in trying to keep two instances in sync instead of just having one source of truth being the order book and not keeping another copy elsewhere but querying the order book for it when needed.

ze97286: FYI, https://github.com/vegaprotocol/vega/issues/6612 is the same issue. Snapshot 82000 which is the one we try to restore from has the order in inconsistent state between the order book and the liquidity party orders. When the node is loading this snapshot, the inconsistency is corrected via the reconcile() but now it's inconsistent with all other nodes. 

on the left is the liquidityPartiesOrders party orders from snapshot 82000. On the right is the same from a snapshot taken right after the loading to memory of 82000.  
snapshot_lp_party_orders_vs_after_loading

on the left is the order book order in snapshot 82000 and on the right, the same after loading. 
snapshot_orderbook_vs_after_loading

You can see that in the snapshot the order in party orders is in incorrect state, which is corrected when loaded in the snapshot. When running with this branch it leads to the expected panic.

wwestgarth: Agree that we should just be asking the orderbook (unless there was a good reason why it wasn't originally done this way?). At a glance it doesn't look to be a hard change to make.

PR #6609, opened 2022-10-26
chore: panic if events are received in an unexpected order

### LINKED ISSUE #6602: Datanode should confirm event order
Data-node should check events are received in the expected order as close to the source as possible  (i.e the broker that used in the socker server) to catch early any issues with out of order events which can lead to hard to diagnose errors upstream if they get through.

pscott31:
Closes #6602

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### 🛠 Improvements
 - [6613](https://github.com/vegaprotocol/vega/issues/6613) - Add file buffering to datanode
+- [6602](https://github.com/vegaprotocol/vega/issues/6602) - Panic if data node receives events in unexpected order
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 ### 🛠 Improvements
 - [6530](https://github.com/vegaprotocol/vega/issues/6530) - Add command to rename a wallet
 - [6531](https://github.com/vegaprotocol/vega/issues/6531) - Add command to update the passphrase of a wallet
+- [6602](https://github.com/vegaprotocol/vega/issues/6602) - Panic if data node receives events in unexpected order
```
jeremyletang, line 26: shouldn't this be in the .61 section?

guoguojin (approved):

PR #6618, opened 2022-10-27
fix: emit event for stopping LP upon settlement

### LINKED ISSUE #6617: Liquidity provisions are not stopped when market is settled
# Problem encountered
When the core settles a market it doesn't stop liquidity provisions which is causing datanode still showing those LPs for the market.

For example: 
image

image

ze97286:
Closes #6617

(unit tests are coming)


core/execution/market.go
```diff
@@ -837,6 +837,14 @@ func (m *Market) closeMarket(ctx context.Context, t time.Time) error {
 
 	m.removeOrders(ctx)
 
+	for _, party := range m.liquidity.ProvisionsPerParty().Slice() {
+		// we don't care about the actual orders as they will be cancelled in the book as part of settlement anyways.
+		_, err := m.liquidity.StopLiquidityProvision(ctx, party.Party)
```
jeremyletang, line 842: This returns a list of orders, at this point they most likely are already cancelled right?

ze97286: that's literally what the comment in line 841 says...

jeremyletang: oh boi... great proof comments are not so useful I guess, or that I'm an idiot.

jeremyletang (approved):

PR #6621, opened 2022-10-27
feat: check if risk factor is negative

Jiajia-Cui:
when mu=-20, tau=0.1, sigma=1 in risk model, the risk factor short becomes negative.

Currently, it seems that core is not able to handle this situation


Jiajia-Cui:
close as range validation has been added [PR](https://github.com/vegaprotocol/vega/pull/6631)

PR #6626, opened 2022-10-27
fix: Fix panic in recvEventRequest

### LINKED ISSUE #6623: Datanode panic due to nil pointer dereference in GRPC status handling
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa3a600]
goroutine 7234144 [running]:
google.golang.org/grpc/internal/status.(*Error).GRPCStatus(0xbba?)
        /home/runner/go/pkg/mod/google.golang.org/grpc@v1.50.0/internal/status/status.go:155
google.golang.org/grpc/status.FromError({0x3b7e820, 0x0})
        /home/runner/go/pkg/mod/google.golang.org/grpc@v1.50.0/status/status.go:94 +0xea
```

We're pretty sure this is due to memory corruption due to sharing an error value between threads involved in subscriptions

zale144:
Closes #6623


datanode/api/event_observer.go
```diff
@@ -16,13 +16,14 @@ import (
 	"context"
 	"time"
 
+	"google.golang.org/grpc/codes"
+
```
jeremyletang, line 20: Ususally we order the package like so:
- std
- external
- vega

(although while I've always done that, I realised recently that goimports does like you do? which is weird as I've always used goimport and it used to be the behaviour I describe I'm sure???)

datanode/api/event_observer.go
```diff
@@ -186,31 +188,33 @@ func observeEventsWithAck(
 	}
 }
 
-func recvEventRequest(
-	stream eventBusServer,
-) (*protoapi.ObserveEventBusRequest, error) {
-	readCtx, cfunc := context.WithTimeout(stream.Context(), 5*time.Second)
-	oebCh := make(chan protoapi.ObserveEventBusRequest)
-	var err error
+const defaultReqTimeout = time.Second * 5
```
jeremyletang, line 191: Const are usually at the top of the file, also would it make sense for this to be a config in the config file not saying it should, probably doesn't need to be.

datanode/api/trading_data_v2.go
```diff
@@ -1723,7 +1723,7 @@ func (t *tradingDataServiceV2) ListOracleData(ctx context.Context, req *v2.ListO
 	}
 
 	if err != nil {
-		apiError(codes.Internal, ErrOracleServiceGetSpec, fmt.Errorf("could not retrieve data for OracleSpecID: %s %w", *req.OracleSpecId, err))
+		return nil, apiError(codes.Internal, ErrOracleServiceGetSpec, fmt.Errorf("could not retrieve data for OracleSpecID: %s %w", *req.OracleSpecId, err))
```
jeremyletang, line 1726: ah nice catch, this was useless before.

jeremyletang (approved):

datanode/api/event_observer.go
zale144: but isn't it like that?

```
- std:
"context"
 "time"

- external:
"google.golang.org/grpc/codes"

- vega:
"code.vegaprotocol.io/vega/core/events"
"code.vegaprotocol.io/vega/datanode/metrics"
"code.vegaprotocol.io/vega/datanode/subscribers"
"code.vegaprotocol.io/vega/logging"
protoapi "code.vegaprotocol.io/vega/protos/vega/api/v1"
eventspb "code.vegaprotocol.io/vega/protos/vega/events/v1"
```

datanode/api/event_observer.go
zale144: I'd move it into config only if that config would be kind of for general stream timeouts or whatever, and not just for this one case. Otherwise it would be overkill imo

datanode/api/event_observer.go
jeremyletang: oh I was misraken in what I said:
-std
-vega
-external

PR #6636, opened 2022-10-28
feat: dehistory data retention

### LINKED ISSUE #6635: Dehistory Data Retention
Should be able to specify how much dehistory data to retain locally on the node to avoid filling up the disk.

ettec:
closes #6635  Feature to specify a data retention period for dehistory data stored locally on the node.

### DIFF
--- a/datanode/dehistory/service.go
+++ b/datanode/dehistory/service.go
@@ -103,7 +103,7 @@ func (d *Service) GetHighestBlockHeightHistorySegment() (store.SegmentIndexEntry
 }
 
 func (d *Service) ListAllHistorySegments() ([]store.SegmentIndexEntry, error) {
-	return d.store.ListAllHistorySegments()
+	return d.store.ListAllHistorySegmentsOldestFirst()
 }
 
 func (d *Service) FetchHistorySegment(ctx context.Context, historySegmentID string) (store.SegmentIndexEntry, error) {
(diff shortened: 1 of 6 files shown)


pscott31 (commented):
The code looks fine, but does this mean that by default we won't have more than a week of history available in IPFS, or do the files live on in there?

ettec:
> The code looks fine, but does this mean that by default we won't have more than a week of history available in IPFS, or do the files live on in there?

Correct, default is one week, there needs to be some limit otherwise users disk will fill up.  Now it would be possible for some power user(s) to hold years worth of data if they have disk big enough and make that available to the community via dehistoty, how we would incentivise that is another question.

ettec:
> > > The code looks fine, but does this mean that by default we won't have more than a week of history available in IPFS, or do the files live on in there?
> > 
> > 
> > Correct, default is one week, there needs to be some limit otherwise users disk will fill up. Now it would be possible for some power node(s) to hold years worth of data if they have disk big enough and make that available to the community via dehistory, how we would incentivise that is another question. Basically if one or more nodes holds the data it is available, not every node needs to hold years of data (so casual users running on their laptop for example) , however there is a cost to provider(s) in terms of disk space and network usage, so we would need some way to recompense them if they do this.
> 
> We could also just have actual ipfs node configured to join the same group/namespace/whatever this is called again, and save forever the files there.

Yeah, for sure, we would even stick the data on filecoin or similar if can figure out a way to pay for it.

pscott31 (approved):

PR #6638, opened 2022-10-30
fix: pointer issues with entity types

### LINKED ISSUE #6637: Datanode entities should copy values when converting to/from proto
A number of entity types in the datanode/entities package are assigning raw pointers to certain fields, without cloning/copying the values. This can lead to bugs, and should be replaced with `DeepClone()` or copies of the values.

EVODelavega:
Closes #6637: fix potential bugs in entities package due to pointer assignments

### DIFF
--- a/datanode/entities/aggregated_balance.go
+++ b/datanode/entities/aggregated_balance.go
@@ -138,6 +138,7 @@ func (balance *AggregatedBalance) ToProto() *v2.AggregatedBalance {
 }
 
 func (balance AggregatedBalance) Cursor() *Cursor {
+	// though this should be OK, it's not exactly safe for concurrent use
 	return NewCursor(AggregatedBalanceCursor{
 		VegaTime:  balance.VegaTime,
 		AccountID: balance.AccountID,
--- a/datanode/entities/data_spec.go
+++ b/datanode/entities/data_spec.go
@@ -34,7 +34,7 @@ func (s *ExternalDataSourceSpec) ToProto() *datapb.ExternalDataSourceSpec {
 
 func ExternalDataSourceSpecFromProto(spec *datapb.ExternalDataSourceSpec, txHash TxHash, vegaTime time.Time) (*ExternalDataSourceSpec, error) {
 	if spec.Spec != nil {
-		ds, err := DataSourceSpecFromProto(spec.Spec.DeepClone(), txHash, vegaTime)
+		ds, err := DataSourceSpecFromProto(spec.Spec, txHash, vegaTime)
 		if err != nil {
 			return nil, err
 		}
--- a/datanode/entities/data_spec.go
+++ b/datanode/entities/data_spec.go
@@ -34,7 +34,7 @@ func (s *ExternalDataSourceSpec) ToProto() *datapb.ExternalDataSourceSpec {
 
 func ExternalDataSourceSpecFromProto(spec *datapb.ExternalDataSourceSpec, txHash TxHash, vegaTime time.Time) (*ExternalDataSourceSpec, error) {
 	if spec.Spec != nil {
-		ds, err := DataSourceSpecFromProto(spec.Spec, txHash, vegaTime)
+		ds, err := DataSourceSpecFromProto(spec.Spec.DeepClone(), txHash, vegaTime)
 		if err != nil {
 			return nil, err
 		}
--- a/datanode/entities/liquidity_provision.go
+++ b/datanode/entities/liquidity_provision.go
@@ -80,11 +80,11 @@ func LiquidityProvisionFromProto(lpProto *vega.LiquidityProvision, txHash TxHash
 	buys := make([]LiquidityOrderReference, 0, len(lpProto.Buys))
 
 	for _, sell := range lpProto.Sells {
-		sells = append(sells, LiquidityOrderReference{sell})
+		sells = append(sells, LiquidityOrderReference{sell.DeepClone()})
 	}
 
 	for _, buy := range lpProto.Buys {
-		buys = append(buys, LiquidityOrderReference{buy})
+		buys = append(buys, LiquidityOrderReference{buy.DeepClone()})
 	}
 
 	return LiquidityProvision{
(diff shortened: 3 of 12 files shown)


pscott31 (commented):
I'm not sure why this is needed? Perhaps there is something I'm missing :)

datanode/entities/aggregated_balance.go
```diff
@@ -138,6 +138,7 @@ func (balance *AggregatedBalance) ToProto() *v2.AggregatedBalance {
 }
 
 func (balance AggregatedBalance) Cursor() *Cursor {
+	// though this should be OK, it's not exactly safe for concurrent use
```
pscott31, line 141: What do you mean?

datanode/entities/withdrawal.go
```diff
@@ -50,6 +50,15 @@ func WithdrawalFromProto(withdrawal *vega.Withdrawal, txHash TxHash, vegaTime ti
 	if amount, err = decimal.NewFromString(withdrawal.Amount); err != nil {
 		return nil, fmt.Errorf("invalid amount: %w", err)
 	}
+	ext := WithdrawExt{}
+	if withdrawal.Ext != nil {
+		cpy := *withdrawal.Ext
+		// if erc20 := withdrawal.Ext.GetErc20(); erc20 != nil {
+		// cpy20 := *erc20
+		// cpy.Ext = &cpy20
+		// }
```
pscott31, line 59: did you mean to leave this comment in?

datanode/entities/withdrawal.go
```diff
@@ -63,13 +72,22 @@ func WithdrawalFromProto(withdrawal *vega.Withdrawal, txHash TxHash, vegaTime ti
...
 }
 
 func (w Withdrawal) ToProto() *vega.Withdrawal {
+	var pbExt *vega.WithdrawExt
+	if w.Ext.WithdrawExt != nil {
+		cpy := *w.Ext.WithdrawExt
+		// if erc20 := w.Ext.WithdrawExt.GetErc20(); erc20 != nil {
+		// cpy20 := *erc20
+		// cpy.Ext = &cpy20
+		// }
```
pscott31, line 88: Ditto

datanode/entities/data_spec.go
```diff
@@ -34,7 +34,7 @@ func (s *ExternalDataSourceSpec) ToProto() *datapb.ExternalDataSourceSpec {
 
 func ExternalDataSourceSpecFromProto(spec *datapb.ExternalDataSourceSpec, txHash TxHash, vegaTime time.Time) (*ExternalDataSourceSpec, error) {
 	if spec.Spec != nil {
-		ds, err := DataSourceSpecFromProto(spec.Spec, txHash, vegaTime)
+		ds, err := DataSourceSpecFromProto(spec.Spec.DeepClone(), txHash, vegaTime)
```
ettec, line 37: Not clear on why this change is necessary? this data is not cached and updated afaik

ettec (changes requested):

datanode/entities/aggregated_balance.go
EVODelavega: That comment was more for my own sanity while going through the entities package. Technically, because the ID fields in the `AggregateBalance` type are pointers, and creating an `AggregatedBalanceCursor`, like we are doing here, is not an atomic operation, so _in theory_ this could be a problem if we were to alter the values we point to concurrently to calling this `Cursor` function. In practice, however, this is a non-issue. Just left the comment there to indicate (to myself) that I'd already gone through this file

datanode/entities/data_spec.go
EVODelavega: In general, when taking a proto variable and getting the entity equivalent thereof, it's just all round preferable for the two to not have any shared state. Seeing as we're passing this `spec.Spec` to a `FomrProto` function, the clone here is not needed. In other places (most notably liquidity provisions) we were assigning a pointer straight from the proto variable, which we shouldn't be doing.

datanode/entities/aggregated_balance.go
pscott31: Maybe I'm missing a something here - but I don't we what using a pointer has to do with it; I agree that there's no synchronisation here so it's not safe to make a cursor while you're altering the entity itself; but that that's the case with all the entities and indeed, most go structs?

If I were wanting to share these between threads (which I'm pretty sure we don't) I'd probably expect the caller to manage the synchronisation.

I guess my question is what would 'OK' look like for you here? If it's not OK in some way I don't understand we should fix it, otherwise I think we should remove the comment because as-is when reading it it sows Fear, Uncertainty and Doubt in the heart of the reader ;)

datanode/entities/data_spec.go
ettec: this assumes that the proto message will be updated, this is not the case, it is discarded after the entity is created, this is the same pattern throughout datanode, and it effectively achieves the same as clone, adding a clone is unnecessary, for example:
`func DataSourceSpecFromProto(specProto *datapb.DataSourceSpec) *DataSourceSpec {
	return &DataSourceSpec{
		ID:        specProto.Id,
		CreatedAt: specProto.CreatedAt,
		UpdatedAt: specProto.UpdatedAt,
		Config:    DataSourceSpecConfigurationFromProto(specProto.Config),
		Status:    specProto.Status,
	}
}`

EVODelavega: As I found out while fixing the unit tests for markets_test.go in sqlstore, because entities and protos both pointed to the same `TradableInstrument`, a number of test cases in there essentially boiled down to `assert.Equal(t, market.TradableInstrument, market.TradableInstrument)`. Even if we don't alter any fields on the proto (or entities) that are shared between multiple variables in the code itself, we were clearly relying on flawed unit tests because of the pointer sharing.

If nothing else, these changes have brought this issue to light, and offer a fix so our unit tests are more solid/reliable.

ettec: ic, yeah in the case of market specifically  it is not cloning the TradableInstrument and holds a pointer.  However, the entity can still be considered immutable (as is the case with all entities), hence why it is acceptable to share the cached instance of entity (and its corresponding TradableInstrument) in this case without the need to clone it.  Having the tests rely on that impl detail is a bit shoddy as it assumes we don't have cache eviction (which we don't currently), the tests would fail if this were to change and would need to do a value check on tradableinstrument instead.

datanode/entities/liquidity_provision.go
```diff
@@ -110,10 +110,10 @@ func (lp *LiquidityProvision) ToProto() *vega.LiquidityProvision {
 	buys := make([]*vega.LiquidityOrderReference, 0, len(lp.Buys))
 
 	for _, sell := range lp.Sells {
-		sells = append(sells, sell.LiquidityOrderReference)
+		sells = append(sells, sell.LiquidityOrderReference.DeepClone())
```
ettec, line 113: So this should be removed?

ettec (changes requested):

ettec (changes requested):

datanode/entities/aggregated_balance.go
ettec: I don't think this is resolved.

PR #6684, opened 2022-11-02
feat: update validation range for sigma in risk

Jiajia-Cui:
I would like to update validation range for sigma in risk model from 1e-3<=x<=100 to 1e-3<=x<=50


commands/proposal_submission_new_market_test.go
```diff
@@ -2794,7 +2794,7 @@ func testNewLogNormalRiskParametersChangeSubmissionInvalidSigma(t *testing.T) {
 								Tau:                   0.2,
 								Params: &protoTypes.LogNormalModelParams{
 									Mu:    0.0,
-									Sigma: 100,
+									Sigma: 50,
```
ze97286, line 2797: you also need to update the error message for the range on line 2736 and 2760

ze97286 (approved):

PR #6700, opened 2022-11-04
feat: implement chain replying and snapshot restore for nullblockchain

### LINKED ISSUE #6463: Null chain support for recording and replaying blocks
In order to reproduce crashes observed via https://github.com/vegaprotocol/vega-market-sim/ we would like functionality to replay transactions (just like with Tendermint based chain). 

We've had a go at implementing it at the `vega-market-sim` level https://github.com/vegaprotocol/vega-market-sim/pull/202 but it there are issues with deterministic replay which may be easier to handle at the null chain level. The other bonus of doing at null-chain level is that it should be possible to load a snapshot (which null chain produces) and then just replay a few (thousand) blocks. 

Asks:
- null chain records transactions
- null chain support replay of transactions from a given snapshot

wwestgarth:
closes #6463 

A core node running with the nullblockchain can now optionally record per-block-transactions to a file, allowing the chain to be replayed if stopped and then started. The replay contains an app-hash comparison to ensure an core's state is identical. The nullblockchain can also now handle when a core node is started from a snapshot. 

The result is that by setting the following CLI options (or changing them in the core config)
```
--blockchain.nullchain.replay-file=SOME_PATH/replay-file
--blockchain.nullchain.replay
--blockchain.nullchain.record
```
you get behaviour very similar to that of Tendermint when starting/stopping a chain.


ze97286:
random thought: it might be nice to save the blocks and transactions in the same format tendermint does internally - then you can support replaying a real chain on null blockchain by just providing it the state directory or the block/tx databases.

jeremyletang (approved):

PR #6702, opened 2022-11-04
fix: remove unions from orders views

### LINKED ISSUE #6699: Remove unions and triggers on orders views
The `orders` and `orders_current` uses a union on `orders_live` and `orders_history` table, however the `orders_history` table should now contain the data from `orders_live` also and the unions are no longer needed.

As the `orders` and `orders_current` views are queried by the API, the `union` queries take a long time to complete and therefore make the API appears as though they are not returning data.

guoguojin:
Closes #6699

### DIFF
--- a/datanode/dehistory/service_test.go
+++ b/datanode/dehistory/service_test.go
@@ -15,16 +15,15 @@ import (
 	"testing"
 	"time"
 
-	config2 "code.vegaprotocol.io/vega/datanode/config"
-	"code.vegaprotocol.io/vega/datanode/dehistory"
-	"code.vegaprotocol.io/vega/datanode/dehistory/initialise"
-	"code.vegaprotocol.io/vega/datanode/dehistory/store"
-
 	"code.vegaprotocol.io/vega/cmd/data-node/commands/start"
 	"code.vegaprotocol.io/vega/datanode/broker"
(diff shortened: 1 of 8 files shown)


datanode/dehistory/service_test.go
```diff
@@ -262,6 +272,10 @@ func TestMain(t *testing.M) {
 }
 
 func TestRestoringFromDifferentHeightsWithFullHistory(t *testing.T) {
+	if schemaChanged {
```
ettec, line 275: as discussed, just update the hash, no need for the schemaChanged flag

ettec (changes requested):

datanode/dehistory/service_test.go
guoguojin: Thanks, my mistake. Fixed https://github.com/vegaprotocol/vega/pull/6702/commits/05f9ce2b83c1e6b5fa485bb137ba0bd202c05e65

jeremyletang:
@guoguojin @ettec I removed the `drop view if exists orders;` statement here. that can't work if orders was a table and we are trying to drop it, this results into an error.

I think that ongoing we should suffix all views with `_view` or something like so just to prevent this kind of issues in future migrations.

pscott31 (approved):

PR #6723, opened 2022-11-07
6563 keep orders in sync

### LINKED ISSUE #6563: Snapshots: failure seen on stagnet1 pipeline

wwestgarth:
closes #6563

This PR changes the liquidity engineso that it asks the order-book for orders instead of keeping its own copy of orders-per-liquidity-provider. This will remove a class of bugs that could cause snapshot-restores to fail if the orders did become out-of-sync.

The change has caused some numbers in the integration tests to change slightly, with the difference being traced down to instances were the order-book and the orders in the liquidity engine did become out of sync.

Full system test run shows failures that match with existing overnight run failures: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/12435/tests


core/execution/liquidity_provision.go
```diff
@@ -707,6 +709,14 @@ func (m *Market) cancelLiquidityProvision(
 		return err
 	}
 
+	// FIXME(JEREMY): if sorting them is the actual solution
+	// review the implementation to write some eventually more efficient
+	// way to sort this here and make sure that all orders are always
+	// cancelled in the same order
```
jeremyletang, line 715: that probably can go now haha

core/liquidity/engine.go
```diff
@@ -306,6 +261,18 @@ func (e *Engine) ProvisionsPerParty() ProvisionsPerParty {
 	return e.provisions.ProvisionsPerParty
 }
 
+// SaveLPOrders sets LP orders that have been cancelled from the book but we need to know
+// what they were to recalculate and redeploy.
+func (e *Engine) SaveLPOrders() {
+	for _, p := range e.provisions.ProvisionsPerParty {
+		e.lpPartyOrders[p.Party] = e.orderBook.GetOrdersPerParty(p.Party)
```
jeremyletang, line 268: you will have all the order of the LP here, not only those created by the liquidity engine, is that OK?

jeremyletang: Would you not want to use GetLiquidityOrders(party) here?

core/matching/snapshot.go
```diff
@@ -130,13 +122,10 @@ func (b *OrderBook) LoadState(_ context.Context, payload *types.Payload) ([]type
...
 
-	if orders, ok := b.ordersPerParty[o.Party]; !ok {
-		b.ordersPerParty[o.Party] = map[string]struct{}{
-			o.ID: {},
-		}
-	} else {
-		orders[o.ID] = struct{}{}
+	// reconstruct pegged orders state
+	if o.PeggedOrder != nil {
+		b.peggedOrders[o.ID] = struct{}{}
```
jeremyletang, line 129: why not add this in the add(order) function?

jeremyletang:
> I've done some perf testing on this branch and the number of events it produces and the peak order rate for the number of LP orders is exactly the same as before, no better and no worse. Here is a quick update of the table compared to the previous branch with the LP optimisations in: image

thanks @peterbarrow

core/liquidity/engine.go
wwestgarth: Yea good spot, changed it.

core/matching/snapshot.go
wwestgarth: Moved it into `add()`.

davidsiska-vega (approved):
I've looked at the `*.feature` files and numbers within and I don't see anything too troublesome.

ze97286 (approved):

jeremyletang (approved):

Issue #6676, opened 2022-11-02
Update feature tests for 0005-COLL-001 for additional events


gordsport:
As part of the spec reviews it was mentioned that for the AC 0005-COLL-001 we would add additional tests for other events as per:

- https://github.com/vegaprotocol/specs/issues/1321

gordsport:
closing as this has to be done in a ST
- https://github.com/vegaprotocol/system-tests/issues/563

PR #6739, opened 2022-11-08
Cover new AC

witgaw:
Cover [0012-POSR-007](https://github.com/vegaprotocol/vega/issues/6676):  When a party is distressed at the point of leaving an auction it should get closed out immediately.


Jiajia-Cui (approved):
LGTM

PR #6742, opened 2022-11-08
fix: new nodes appear in the datanode from when they are announced an…

### LINKED ISSUE #6696: Announce node: Data node API should show pending validators in same epoch they're announced in
Currently, a prospective validator node is only shown by the API in the epoch following their "announcement". That means that, on the token site, a validator can't self stake right away, which can delay the whole process by a day in a 24-hour epoch network.

wwestgarth:
closes #6696 

We now send the current epoch in the `ValidatorUpdate` event and use that value instead of `from_epoch` in the datanode as the epoch which the node started to "exist".

Branch with system-test that announces a node with `from_epoch` 10 epochs in the future, but the data-node API now shows it straight away: https://github.com/vegaprotocol/system-tests/compare/announce_in_same_epoch?expand=1

Test passing: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/23203/pipeline/


pscott31 (approved):
ah well done, I didn't consider this - I was testing it by sending events to the datanode assuming from_epoch meant what epoch_seq does here and couldn't find anything wrong :)

wwestgarth:
> ah well done, I didn't consider this - I was testing it by sending events to the datanode assuming from_epoch meant what epoch_seq does here and couldn't find anything wrong :)

It crossed my mind when you mentioned it at standup, but for some reason I read the ticket completelt wrong and thought the bug was specifically just about if someone announced and set their `from_epoch` to be epoch-now.

PR #6752, opened 2022-11-09
chore: set all release as pre-release

jeremyletang:


daniel1302 (dismissed):
It means that even if we create the standard release(non-pre-release), the GitHub action marks it as a pre-release when publishing binaries and we need to update status after all binaries are published to the release. As long as we are fine with the above, it’s good to merge.

As per docs:

💡 When the release info keys (such as name, body, draft, prerelease, etc.) are not explicitly set and there is already an existing release for the tag, the release will retain its original info.

jeremyletang:
> It means that even if we create the standard release(non-pre-release), the GitHub action marks it as a pre-release when publishing binaries and we need to update status after all binaries are published to the release. As long as we are fine with the above, it’s good to merge.
> 
> As per docs:
> 
> 💡 When the release info keys (such as name, body, draft, prerelease, etc.) are not explicitly set and there is already an existing release for the tag, the release will retain its original info.

I think that's fine, the goal here is to mark every single release as pre-release, until we are ready to push them to mainnet, and then only the status of the release would be changed manually. So I would think this is fine?

ValentinTrinque (approved):

PR #6756, opened 2022-11-09
6640 time based mtm

### LINKED ISSUE #6640: Update mark price only once per block
The following spec change needs to be implemented in core:

- https://github.com/vegaprotocol/specs/pull/1356

This changes the frequency of the mark price calculation from once every trade to once every block to reduce the number of events being created and thus passed to the data node

EVODelavega:
Fixes #6640 

MTM settlement should only happen at set interval


core/execution/engine.go
```diff
@@ -794,6 +799,19 @@ func (e *Engine) OnMarketAuctionMinimumDurationUpdate(ctx context.Context, d tim
...
 }
 
+func (e *Engine) OnMarkPriceUpdateMaximumFrequency(ctx context.Context, d time.Duration) error {
+	// we make sure to update both the copy and the actual markets for snapshots
+	// although we can most likely just update the market because we're already ensuring the nextMTM value
+	// is set correctly when getting state
+	for _, cpMkt := range e.marketsCpy {
+		mkt := e.markets[cpMkt.mkt.ID]
+		mkt.OnMarkPriceUpdateMaximumFrequency(ctx, d)
+		cpMkt.nextMTM = mkt.nextMTM
```
wwestgarth, line 809: I don't think we need to worry about making sure `e.marketsCpy` is also updated, because its just a pointer to the same market?

EVODelavega: Unfortunately, not pointers to the same object.

wwestgarth: Are you sure? in the engine they're defined like his:
```
markets    map[string]*Market
marketsCpy []*Market
```

and when we add a market we do this:
```
e.markets[marketConfig.ID] = mkt
e.marketsCpy = append(e.marketsCpy, mkt)
```
?

jeremyletang: marketsCpy definitely points to the same objects, they just exists to iterate deterministically over the markets without sorting them everytime.

EVODelavega: alright, don't know why I thought otherwise, I'll fix that up && push

core/netparams/defaults.go
```diff
@@ -211,11 +211,14 @@ func defaultNetParams() map[string]value {
...
 
 		SpamProtectionMaxBatchSize: NewUint(UintGTE(num.NewUint(2)), UintLTE(num.NewUint(200))).Mutable(true).MustUpdate("15"),
+
+		MarkPriceUpdateMaximumFrequency: NewDuration().Mutable(true).MustUpdate("5s"),
 	}
 
 	// add additional cross net param rules
 	m[MarketAuctionMinimumDuration].AddRules(DurationDependentLT(MarketAuctionMaximumDuration, m[MarketAuctionMaximumDuration].(*Duration)))
 	m[MarketAuctionMaximumDuration].AddRules(DurationDependentGT(MarketAuctionMinimumDuration, m[MarketAuctionMinimumDuration].(*Duration)))
+	m[MarkPriceUpdateMaximumFrequency].AddRules(DurationGTE(time.Duration(0)))
```
jeremyletang, line 221: we usuyally add them here when they depends on others I suppose, your's should be in the map literal. Also I think there's a max value of 24h

EVODelavega: Yeah, I know we usually add it when they're interdependent, but seeing as durations can technically be negative (`time.Add(-5)`), and we instantiate the `nextMTM` with a zero time, I added this to avoid cases where we subtract a negative duration from a zero time. I'll add the max though

datanode/gateway/graphql/schema.graphql
```diff
@@ -252,6 +252,8 @@ type MarketData {
   marketValueProxy: String!
   "The equity like share of liquidity fee for each liquidity provider"
   liquidityProviderFeeShare: [LiquidityProviderFeeShare!]
+  "RFC3339Nano time indicating when the next time positions will be marked to market"
```
candida-d, line 255:
```suggestion
  "RFC3339Nano time indicating the next time positions will be marked to market"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -310,6 +312,8 @@ type ObservableMarketData {
   marketValueProxy: String!
   "The equity like share of liquidity fee for each liquidity provider"
   liquidityProviderFeeShare: [ObservableLiquidityProviderFeeShare!]
+  "RFC3339Nano time indicating when the next time positions will be marked to market"
```
candida-d, line 315:
```suggestion
  "RFC3339Nano time indicating the next time positions will be marked to market"
```

candida-d (dismissed):
Small suggested change to the gql description. Might need to be reviewed by someone else in front end

core/execution/engine.go
```diff
@@ -810,6 +815,17 @@ func (e *Engine) OnMarketAuctionMinimumDurationUpdate(ctx context.Context, d tim
 	return nil
 }
 
+func (e *Engine) OnMarkPriceUpdateMaximumFrequency(ctx context.Context, d time.Duration) error {
+	// we make sure to update both the copy and the actual markets for snapshots
```
ze97286, line 819: is this comment still correct?

EVODelavega: Will remove

core/execution/engine_snapshot.go
```diff
@@ -104,12 +109,15 @@ func (e *Engine) restoreMarket(ctx context.Context, em *types.ExecMarket) (*Mark
...
 
+	mkt.nextMTM = nextMTM
 	e.markets[marketConfig.ID] = mkt
 	e.marketsCpy = append(e.marketsCpy, mkt)
 
 	if err := e.propagateInitialNetParams(ctx, mkt); err != nil {
 		return nil, err
 	}
+	// ensure this is set correctly
+	mkt.nextMTM = nextMTM
```
ze97286, line 120: is line 112 not sufficiently set correctly?

EVODelavega: I don't think so. I think line 112 is redundant, if anything. By calling `propagateInitialNetParams` the `nextMTM` time will get reset to be N seconds in the future. When restoring from a snapshot we have to make sure we'll MTM the exact same time as other nodes did, so the snapshot could have markets with a next MTM 2 seconds in the future, but the `propagateIntialNetParams` call would override that to be N seconds in the future (whatever the net param is)

datanode/gateway/graphql/schema.graphql
EVODelavega: removed

core/execution/engine_snapshot.go
ze97286: well, you the earlier assignment so now it all makes sense.

core/execution/liquidity_provision_test.go
```diff
@@ -435,7 +437,8 @@ func TestSubmit(t *testing.T) {
 				}
 			}
 
-			assert.Len(t, found, 1)
+			// @TODO figure out why this doesn't happen anymore
```
wwestgarth, line 440: Is this TODO still relevant?

core/integration/features/margin/6202-release-excess-on-amend.feature
```diff
@@ -57,23 +60,28 @@ Feature: test margin during amending orders
...
+      # | party4 | USD   | ETH/MAR22 | 85366  | 99914634 | 0     |
     #check the margin levels
     Then the parties should have the following margin levels:
       | party  | market id | maintenance | search | initial | release |
-      | party1 | ETH/MAR22 | 16015       | 17616  | 19218   | 22421   |
-      | party4 | ETH/MAR22 | 71139       | 78252  | 85366   | 99594   |
+      | party1 | ETH/MAR22 | 30429       | 33471  | 36514   | 79114   |
+      | party4 | ETH/MAR22 | 149391      | 164329 | 179268  | 209146  |
+      # | party1 | ETH/MAR22 | 16015       | 17616  | 19218   | 22421   |
+      # | party4 | ETH/MAR22 | 71139       | 78252  | 85366   | 99594   |
```
Jiajia-Cui: I think the margin for party4 should be the old value: 71139, the same as margin for party1

core/execution/market.go
```diff
@@ -1602,57 +1636,72 @@ func (m *Market) handleConfirmation(ctx context.Context, conf *types.OrderConfir
...
+	// orderUpdates = m.confirmMTM(ctx, conf.Order)
+	// }
+	// } else {
+	// we had no trade, but still want to register this position in the settlement
+	// engine I guess
+	// party := conf.Order.Party
+	// if pos, ok := m.position.GetPositionByPartyID(party); ok {
+	// m.settlement.AddPosition(party, pos)
+	// }
+
```
witgaw, line 1690:
```suggestion
```

jeremyletang (dismissed):

core/execution/market.go
```diff
@@ -721,7 +726,23 @@ func (m *Market) OnTick(ctx context.Context, t time.Time) bool {
 		}
 	}
 
-	// check auction, if any
+	// MTM if enough time has elapsed, we are not in auction, and we have a non-zero mark price.
+	// we MTM in leaveAuction before deploying LP orders like we did before, but we do update nextMTM there
+	if mp := m.getCurrentMarkPrice(); mp != nil && !mp.IsZero() && !m.nextMTM.After(t) && !m.as.InAuction() {
+		m.nextMTM = t.Add(m.mtmDelta) // add delta here
```
witgaw, line 732: Super nitpicky/paranoid, but I think when exiting the auction we first do MTM and then update `m.nextMTM` wouldn't it be safer to have the same order of operations in both cases in cases there's some future change that makes `confirmMTM` somehow dependent on `m.nextMTM`?

witgaw (commented):
Looking really good at first glance, I'll now look closer into feature test changes and try to understand why we get a slight inflation in the number of events emitted in some tests.

Some comments below:

core/execution/market.go
EVODelavega: It shouldn't really matter. The reason why I set the nextMTM at the end when leaving auction is because I wanted to change as little as possible around the `leaveAuction` flow. It's kind of a self-contained thing, and setting the MTM time is kind of the same as `m.as.Left()`  (ie confirming we're done)

core/integration/features/verified/closeout-cascade.feature
```diff
@@ -37,7 +37,7 @@ Feature: Closeout-cascades
     When the parties submit the following liquidity provision:
       | id  | party  | market id | commitment amount | fee   | side | pegged reference | proportion | offset | lp type    |
       | lp1 | lpprov | ETH/DEC19 | 100000            | 0.001 | sell | ASK              | 100        | 55     | submission |
-      | lp1 | party0 | ETH/DEC19 | 100000            | 0.001 | buy  | BID              | 100        | -55    | submission |
+      | lp1 | party0 | ETH/DEC19 | 100000            | 0.001 | buy  | BID              | 100        | 55     | amendment  |
```
EVODelavega, line 40: This makes no sense, both LPs share the same ID column, but are different parties, and the second one is an amendment?? I don't get what's going on here.

The fact that the parties are different makes even less sense considering both submissions/amendments are incomplete, only accounting for one side (sell/buy)

Jiajia-Cui: agree, updated the LPs to the same party now

daniel1302 (commented):
The Jenkins PR is failing due to conflicts

https://jenkins.ops.vega.xyz/job/vega/job/PR-6756/54/console

image

core/execution/market.go
witgaw: Cool, fair enough 👍

core/execution/market_callbacks.go
```diff
@@ -107,3 +107,11 @@ func (m *Market) OnMarketAuctionMinimumDurationUpdate(ctx context.Context, d tim
 		m.broker.Send(evt)
 	}
 }
+
+func (m *Market) OnMarkPriceUpdateMaximumFrequency(ctx context.Context, d time.Duration) {
+	if !m.nextMTM.IsZero() {
+		m.nextMTM = m.nextMTM.Add(-m.mtmDelta)
+	}
+	m.nextMTM = m.nextMTM.Add(d)
```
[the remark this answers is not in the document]
EVODelavega: The current code will shorten that all the same. If the `nextMTM` is set to a given time, we subtract the old delta, and add the new. If the new delta is larger, the `nextMTM` time will be increased by the difference between old and new, if we decrease the delta, we update `nextMTM` to be `last_mtm + new_delta`. If the new delta places the next MTM in the past, then next block we will mark to market

jeremyletang (approved):

PR #6796, opened 2022-11-11
feat: max gas enforcement implementation

### LINKED ISSUE #6795: MaxGas enforcement
# Feature Overview

See spec ticket https://github.com/vegaprotocol/specs/pull/1397

ze97286:
Closes #6795

### DIFF
--- a/core/matching/orderbook.go
+++ b/core/matching/orderbook.go
@@ -126,6 +126,11 @@ func (b *OrderBook) ReloadConf(cfg Config) {
 	b.cfgMu.Unlock()
 }
 
+// GetOrderBookLevelCount returns the number of levels in the book.
+func (b *OrderBook) GetOrderBookLevelCount() uint64 {
+	return uint64(len(b.buy.levels) + len(b.sell.levels))
+}
+
 func (b *OrderBook) GetPeggedOrdersCount() uint64 {
 	return b.peggedOrdersCount
 }
--- a/core/netparams/defaults.go
+++ b/core/netparams/defaults.go
@@ -211,8 +211,9 @@ func defaultNetParams() map[string]value {
 		LimitsProposeAssetEnabledFrom:  NewString(checkOptionalRFC3339Date).Mutable(true).MustUpdate(""), // none by default
 
 		SpamProtectionMaxBatchSize: NewUint(UintGTE(num.NewUint(2)), UintLTE(num.NewUint(200))).Mutable(true).MustUpdate("15"),
-
-		MaxPeggedOrders: NewUint(UintGTE(num.NewUint(0)), UintLTE(num.NewUint(10000))).Mutable(true).MustUpdate("1500"),
+		MaxGasPerBlock:             NewUint(UintGTE(num.NewUint(100)), UintLTE(num.NewUint(10000000))).Mutable(true).MustUpdate("100000"),
+		DefaultGas:                 NewUint(UintGTE(num.NewUint(1)), UintLTE(num.NewUint(99))).Mutable(true).MustUpdate("1"),
+		MaxPeggedOrders:            NewUint(UintGTE(num.NewUint(0)), UintLTE(num.NewUint(10000))).Mutable(true).MustUpdate("1500"),
 	}
 
 	// add additional cross net param rules
--- a/core/positions/engine.go
+++ b/core/positions/engine.go
@@ -423,3 +423,14 @@ func I64MaxAbs(vals ...int64) int64 {
 	}
 	return r
 }
+
+// GetOpenPositionCount returns the number of open positions in the market.
+func (e *Engine) GetOpenPositionCount() uint64 {
+	var total uint64
+	for _, mp := range e.positionsCpy {
+		if mp.Size() != 0 {
+			total++
+		}
+	}
+	return total
+}
--- a/core/processor/abci.go
+++ b/core/processor/abci.go
@@ -48,8 +48,11 @@ import (
 	eventspb "code.vegaprotocol.io/vega/protos/vega/events/v1"
 
 	tmtypes "github.com/tendermint/tendermint/abci/types"
+	tmtypes1 "github.com/tendermint/tendermint/proto/tendermint/types"
 	tmtypesint "github.com/tendermint/tendermint/types"
 )
+
+const AppVersion = 1
 
 var (
 	ErrPublicKeyCannotSubmitTransactionWithNoBalance  = errors.New("public key cannot submit transaction without balance")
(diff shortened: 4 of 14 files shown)


core/netparams/defaults.go
```diff
@@ -211,6 +211,8 @@ func defaultNetParams() map[string]value {
 		LimitsProposeAssetEnabledFrom:  NewString(checkOptionalRFC3339Date).Mutable(true).MustUpdate(""), // none by default
 
 		SpamProtectionMaxBatchSize: NewUint(UintGTE(num.NewUint(2)), UintLTE(num.NewUint(200))).Mutable(true).MustUpdate("15"),
+		MaxGasPerBlock:             NewUint(UintGTE(num.NewUint(100)), UintLTE(num.NewUint(10000000))).Mutable(true).MustUpdate("10000"),
```
davidsiska-vega, line 214: I think given the other constants I'd make the default to be 10^5 rather than 10^4 but of course probably best to set in genesis explicitly in all environments anyway.

davidsiska-vega (dismissed):
LGTM; if we end up changing the spec then the code edits should be relatively minimal and it's probably worth merging this just so we see how it looks on dev / stag.

davidsiska-vega (dismissed):

core/matching/orderbook.go
```diff
@@ -126,6 +126,11 @@ func (b *OrderBook) ReloadConf(cfg Config) {
 	b.cfgMu.Unlock()
 }
 
+// GetOrderBookLevelCount returns the number of levels in the book.
+func (b *OrderBook) GetOrderBookLevelCount() uint64 {
+	return uint64(len(b.ordersByID))
```
peterbarrow, line 131: Sanity check, this looks like returning the number of orders, not the number of levels?

peterbarrow (dismissed):

core/processor/abci.go
```diff
@@ -718,10 +722,21 @@ func (app *App) OnEndBlock(req tmtypes.RequestEndBlock) (ctx context.Context, re
...
-		}
+		resp.ValidatorUpdates = powerUpdates
+	}
+
+	// update max gas based on the network parameter
+
+	resp.ConsensusParamUpdates = &tmtypes.ConsensusParams{
+		Block: &tmtypes.BlockParams{
+			MaxGas:   int64(app.gastimator.OnBlockEnd()),
+			MaxBytes: 22020096,
```
jeremyletang, line 735: Why that hardcode default?

core/processor/gastimator.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// OnMaxGasUpdate updates the max gas from the network parameter.
+func (g *Gastimator) OnMaxGasUpdate(ctx context.Context, max *num.Uint) error {
+	g.maxGas = max.Uint64()
+	return nil
+}
+
+// OnDefaultGasnUpdate updates the default gas wanted per transaction.
+func (g *Gastimator) OnDefaultGasnUpdate(ctx context.Context, def *num.Uint) error {
```
jeremyletang, line 53: is Gasn a typo here?

core/matching/orderbook.go
davidsiska-vega: For me that's not a killer by the way - if number of orders is much easier - I would be happy to change the spec draft. 

I don't know what causes the cost - whether it's more the number of levels or more the number of orders.

core/positions/engine.go
```diff
@@ -423,3 +423,14 @@ func I64MaxAbs(vals ...int64) int64 {
 	}
 	return r
 }
+
+// GetOpenPositionCount returns the number of open positions in the market.
+func (e *Engine) GetOpenPositionCount() uint64 {
+	var total uint64
+	for _, mp := range e.positionsCpy {
+		if mp.Size() != 0 {
```
jeremyletang, line 431: Do we not account for potential positions here?

core/processor/abci.go
ze97286: it has to be something, what do you recommend?

core/processor/gastimator.go
ze97286: it is

jeremyletang (commented):
I don't think it should matter, but using floats in the gas estimations things would be alright?

Also wondering if we should add gas estimation to transaction which require the validator to sign payloads (emit signature, withdraw, etc) as these could really become quite long to process

core/matching/orderbook.go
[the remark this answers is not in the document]
ze97286: I'll change to that

core/processor/abci.go
ze97286: well, it doesn't matter so much what the average size of blocks in testnet is, but the consensus params are: 

image

jeremyletang: ah fair enough then.

core/processor/gastimator.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return uint64(math.Min(float64(uint64(totalBatchGas)), float64(g.maxGas-1))), nil
+}
+
+// gasOrder = network.transaction.defaultgas + peg cost factor x pegs
+// + LP shape cost factor x shapes
+// + position factor x positions
+// + level factor x levels
+// gasOrder = min(maxGas-1,gasOrder).
+func (g *Gastimator) orderGastimate(marketID string) (uint64, error) {
```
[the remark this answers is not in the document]
ze97286: sure

ze97286:
> I don't think it should matter, but using floats in the gas estimations things would be alright?
> 
It's a local decision of the validator so shouldn't matter.

core/processor/abci.go
wwestgarth: Could you use `tmtypesint.DefaultBlockParams().MaxBytes` to get the default value that tendermint sets? It'll avoid the hard-coding and make it clearer its just a default.

core/processor/abci.go
```diff
@@ -718,10 +722,21 @@ func (app *App) OnEndBlock(req tmtypes.RequestEndBlock) (ctx context.Context, re
...
+		},
+		Version: &tmtypes1.VersionParams{
+			AppVersion: 1,
+		},
```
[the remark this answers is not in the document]
ze97286: what do you mean?

core/processor/abci.go
ze97286: yeah, that makes sense

core/processor/abci.go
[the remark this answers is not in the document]
jeremyletang: I'm just thinking that this should be either not hardcode or we need probably to handle the app version properly someday. no real action just here and now.

jeremyletang:
> It's a local decision of the validator so shouldn't matter.

fair enough

jeremyletang (approved):

davidsiska-vega (approved):

PR #6805, opened 2022-11-14
chore: use Timestamp in graphql

### LINKED ISSUE #6716: Inconsistent use of timestamps in GQL - some are RFC3339 strings and some are number-of-nanoseconds timestamps
# Problem encountered
Most of GQL API uses RFC3339Nano timestamps, but some (notably 'daterange') are expecting a integer number of nanoseconds wrapped in as string.

NB: In GRPC (and by extension REST), all use number-of-nanoseconds and that is fine.

The precipitating case was `dateRange` in `ordersConnection`, but we should go through and check for other violations.

jeremyletang:
close #6716


CHANGELOG.md
```diff
@@ -9,12 +9,13 @@
...
 
 ### 🛠 Improvements
-- [6795]https://github.com/vegaprotocol/vega/issues/6795 - max gas implementation
+- [6795](https://github.com/vegaprotocol/vega/issues/6795 - max gas implementation
 - [6641](https://github.com/vegaprotocol/vega/issues/6641) - network wide limits
 
 ### 🐛 Fixes
 - [6801](https://github.com/vegaprotocol/vega/issues/6801) - Fix internal data source validations
 - [6799](https://github.com/vegaprotocol/vega/issues/6799) - Move LP fees in transit to the network treasury
+- [6716](https://github.com/vegaprotocol/vega/issues/6716) - Use timestamp on all times fields
```
wwestgarth, line 18: Should this be in with breaking changes?

jeremyletang: Hum, maybe, will make it so.

wwestgarth (approved):

dexturr (approved):

PR #6818, opened 2022-11-15
fix: level db throwing resource temporarily unavailable error

### LINKED ISSUE #6813: Level DB failure in PUP tests
Intermittently seeing this error in PUP testing:

failed to initialise decentralized history:failed to create deHistory service:failed to create decentralized history store:failed to create index:failed to open level db file:resource temporarily unavailable

ettec:
closes #6813

### DIFF
--- a/datanode/dehistory/store/index.go
+++ b/datanode/dehistory/store/index.go
@@ -8,97 +8,125 @@ import (
 	"path/filepath"
 	"sort"
 
-	"github.com/syndtr/goleveldb/leveldb"
-	"github.com/syndtr/goleveldb/leveldb/opt"
-	"github.com/syndtr/goleveldb/leveldb/util"
+	bolt "go.etcd.io/bbolt"
 )
 
 var ErrIndexEntryNotFound = errors.New("index entry not found")
 
-type LevelDbBackedIndex struct {
-	db *leveldb.DB
+type BBoltBackedIndex struct {
+	db *bolt.DB
 }
 
-func NewIndex(dataDir string) (*LevelDbBackedIndex, error) {
-	db, err := leveldb.OpenFile(filepath.Join(dataDir, "index.db"), nil)
+const indexBucket = "index"
+
+func NewIndex(dataDir string) (*BBoltBackedIndex, error) {
+	db, err := bolt.Open(filepath.Join(dataDir, "index.db"), 0o666, nil)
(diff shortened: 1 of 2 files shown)


datanode/dehistory/store/index.go
```diff
@@ -8,97 +8,125 @@ import (
...
-	"github.com/syndtr/goleveldb/leveldb/util"
+	bolt "go.etcd.io/bbolt"
 )
 
 var ErrIndexEntryNotFound = errors.New("index entry not found")
 
-type LevelDbBackedIndex struct {
-	db *leveldb.DB
+type BBoltBackedIndex struct {
+	db *bolt.DB
```
wwestgarth, line 17: Why the change away from levelDB? `resource temporarily unavailable` usually just means that `db.Close()` wasn't called properly when shutting down, then the next process trying to open it gets that error.

Also I'm sure on a previous PR it was decided to use levelDB to keep the tools consistent across all of core?

jeremyletang (changes requested):
Wait, are we changing the database backend because of an error? that seems wrong, there must be a way to fix the problem without going with something so radical.
We use leveldb in the core for snapshot and tendermint does it as well without any issue. We shall use level db here as well as it makes easier to resonnate about 1 DB backend instead of multiple.

datanode/dehistory/store/index.go
ettec: ok, thats interesting, digging into the code it was not possible to get a handle on the cause as the error handling in code is terrible and adds no context to be able to diagnose the underlying cause - doesn't inspire confidence in the lib.

If we think that's the cause I'll pull the PR see if can figure out why close not being called.

wwestgarth: Where I've seen that error before was when doing the core-snapshot work -- there was a system-test which would stop the node, then call `vegatools snapshotdb` to look how many snapshots were saved. The call to `vegatools snapshotdb` would sometime fail with that resource unavailable error, and making sure we closed the db both in core on shutdown and in the tool fixed it. So it sounds very similar to this!

ettec:
Previously encounters of the error 'resource temporarily unavailable' has been found to imply that LevelDB was open elsewhere (thanks for pointing out Will), turns out this was indeed the case as Visor wasn't shutting down the old datanode before starting the new, hence the issue.

PR #6833, opened 2022-11-15
feat: unified errors

### LINKED ISSUE #6332: Standardise GraphQL 'no result' results
(wip ticket)

# Feature Overview

**In order to** provide consistent information to API users
**We will** make sure all queries report 'no result's consistently
**So that** it's less surprising

# Background
**Note: party here is an illustration, a stand in for all queries**

We have inconsistent nullability and errors with return types on our queries. The way we return a result for a specific party, if the party doesn't exist, breaks the schema by returning null and an error field, where the schema says it will never be null. 

# Tasks
- [ ] Ensure the result of a query matches the schema
  - [ ] `party` should *not* return an `error` field if the specified ID does not exist. It should return party as null
  - [ ] `partyConnection` should always return an array, even if it is empty (i.e. `partyConnection: [Party!]!` - there will always be an array and any item in the array will be a Party)
- [ ] Do the same for markets
- [ ] And assets
- [ ] And proposals
- [ ] And so on

jeremyletang:
Add some helper and cleanups for graphql error when a resource is not fund. this is just kinda a proposal for now.

Screenshot 2022-11-15 at 17 14 07

close #6332

### DIFF
--- a/datanode/sqlstore/blocks.go
+++ b/datanode/sqlstore/blocks.go
@@ -93,7 +93,7 @@ func (bs *Blocks) GetLastBlock(ctx context.Context) (entities.Block, error) {
 	}
 	defer metrics.StartSQLQuery("Blocks", "GetLastBlock")()
 
-	lastBlock, err := GetLastBlockUsingConnection(ctx, bs.Connection)
+	lastBlock, err := bs.getLastBlockUsingConnection(ctx, bs.Connection)
 	// FIXME(woot?): why do we set that before checking for error, that would clearly fuckup the cache or something innit?
 	bs.lastBlock = lastBlock
 	if err != nil {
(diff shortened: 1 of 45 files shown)


datanode/sqlstore/blocks.go
```diff
@@ -112,7 +112,17 @@ func (bs *Blocks) setLastBlock(b entities.Block) {
...
+	block := &entities.Block{}
+	if err := pgxscan.Get(ctx, connection, block, `SELECT vega_time, height, hash
+		FROM blocks order by height asc limit 1`); err != nil {
+		return entities.Block{}, bs.wrapE(err)
+	}
+
+	return *block, nil
 }
 
 func GetOldestHistoryBlockUsingConnection(ctx context.Context, connection Connection) (entities.Block, error) {
```
guoguojin, line 128: Do we need to keep the old `GetOldestHistoryBlockUsingConnection` and `GetLastBlockUsingConnection` below? That code duplication could be confusing and cause problems if we need to change anything in the future and forget to do both.

jeremyletang: yea it might not be needed, just didn't handle error exactly the same now.

guoguojin (approved):

PR #6838, opened 2022-11-16
feat: feature test to cover LP shape limits

ze97286:
Feature test covering network limit ACs for max LP shape limit.


Jiajia-Cui (approved):
the feature test looks good to me

jeremyletang (approved):

Issue #6831, opened 2022-11-15
Snapshot: settlement engine position differ during SettleMTM


wwestgarth:
# Problem encountered
A snapshot failure was observed on fairground, and as always it comes with a long backstory.

We have a case where during `settlement.SettleMTM()` a party hits the following piece of code:
```
// we don't want to accidentally MTM a party who closed out completely when they open
// a new position at a later point, so remove if size == 0
if current.size == 0 && current.Buy() == 0 && current.Sell() == 0 {
	// broke this up into its own func for symmetry
	e.rmPosition(party)
}
```

and their position is removed from the settlement engine, _but still exists in the positions engine_.

We then take a snapshot, but we only snapshot the position engine. When we restore the market we naively push all restored positions into the settlement engine:
```
func (m *Market) PostRestore(ctx context.Context) error {
	m.settlement.Update(m.position.Positions())
```

which leads to a mismatch in restored state. The mismatch causes a consensus failure the next time a trade occurs and we enter `settlement.SettleMTM()` again. We make this call to `getCurrentPositions()`:
```
for _, evt := range positions {
	party := evt.Party()

	// get the current position, and all (if any) position changes because of trades
	current := e.getCurrentPosition(party, evt)
```

For the node that _hasn't restored_ and _doesn't_ have a position for the party in the settlement engine, this call creates a position with the updated post-trade position from the positions engine and we do nothing:
```
if !hasTraded && current.price.EQ(markPrice) {
	// no changes in position and markPrice hasn't changed -> nothing needs to be marked
	continue
}
```

for the node that has restored from a snapshot, `getCurrentPosition()` returns the old position we just pushed in from the positions engine, which has different price and then we go an calculate a mtm-share. Accounts then differ, consensus failure panic.

An obvious fix would be to save in the snapshot which positions weren't in the settlement engine and to not restore them, but first I want to be sure that getting in a state where we have a position in the positions engine, but not in the settlement engine is expected.

# Observed behaviour
A clear and concise description of how the system is behaving.

# Expected behaviour

# System response
Describe what the system response was, include the output from the command, automation, or else.

# Steps to reproduce

## Manual
Steps to reproduce the behaviour manually:

## Automation
Link to automation and explanation on how to run it to reproduce the problem/bug

# Evidence

## Logs
If applicable, add logs and/or screenshots to help explain your problem.

## Additional context
Add any other context about the problem here including; system version numbers, components affected.

# Definition of Done
[...]

PR #6850, opened 2022-11-17
fix: remove position from position engine when closed

### LINKED ISSUE #6849: Remove positions from position engine when party close positions
As of now we keep position with a 0 size and potential buy / sell when a position is closed. this needs fixed.

jeremyletang:
close #6849

### DIFF
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -43,7 +43,6 @@ import (
 	"code.vegaprotocol.io/vega/libs/num"
 	"code.vegaprotocol.io/vega/libs/ptr"
 	"code.vegaprotocol.io/vega/logging"
-	"golang.org/x/exp/maps"
 )
 
 // InitialOrderVersion is set on `Version` field for every new order submission read from the network.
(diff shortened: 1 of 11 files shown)


ze97286:
isn't this fixing the snapshot issue that @wwestgarth found?

jeremyletang:
> isn't this fixing the snapshot issue that @wwestgarth found?

Nope @EVODelavega is on that one

ze97286:
> > isn't this fixing the snapshot issue that @wwestgarth found?
> 
> Nope @EVODelavega is on that one

I don't know what @EVODelavega is fixing but it seems very relevant to the description [here](https://github.com/vegaprotocol/vega/issues/6831#issue-1449951013) - i.e. the position engine was keeping zero position while the settlement engine was removing them, which would cause the restored node to have them back in the settlement engine. May be other things in play but I think it may be relevant to what you're fixing here no?

jeremyletang:
> > > isn't this fixing the snapshot issue that @wwestgarth found?
> > 
> > 
> > Nope @EVODelavega is on that one
> 
> I don't know what @EVODelavega is fixing but it seems very relevant to the description [here](https://github.com/vegaprotocol/vega/issues/6831#issue-1449951013) - i.e. the position engine was keeping zero position while the settlement engine was removing them, which would cause the restored node to have them back in the settlement engine. May be other things in play but I think it may be relevant to what you're fixing here no?

Alright, I see. So I think reading the ticket both our tickets are two side of a same coin. This branch is actually making sure that the positions in the position engine, which are being closed, are not staying in the position engine. Before that the position will stay with a size of zero and would be used during MTM and would send empty events. Elias PR (https://github.com/vegaprotocol/vega/pull/6890) Is actually changing the storage of the settlement engine which had been broken since the last changes in the MTM.

core/settlement/engine.go
```diff
@@ -476,6 +476,13 @@ func (e *Engine) AddPosition(party string, evt events.MarketPosition) {
 	_ = e.getCurrentPosition(party, evt)
 }
 
+func (e *Engine) HasPosition(party string) bool {
+	_, okPos := e.pos[party]
+	_, okTrades := e.trades[party]
```
ze97286, line 481: I don't understand the okTrades - this would return true in all cases you ever had a position and were not distressed, even if you don't hold it anymore. what is the purpose of it?

core/execution/market_test.go
```diff
@@ -6412,6 +6412,7 @@ func Test3008And3007CancelLiquidityProvision(t *testing.T) {
 				found = append(found, evt.LedgerMovements()...)
 			}
 		}
+		// fmt.Printf("%#v\n", found[1].Entries[0].Type)
```
ze97286, line 6415: might as well remove

core/settlement/engine.go
jeremyletang: well it's not always that there's a position AND trades during the MTM window. you could have a position (which in that case was registered with a potential position at the time) but 0 trades. but at the end of the window, the party might have cancelled their orders and no trades happened. At that point the position engine might have deleted the position as it's of size 0, and no orders are left in the book.

core/execution/market_test.go
jeremyletang: correct.

core/execution/market.go
```diff
@@ -1271,22 +1286,33 @@ func (m *Market) releaseExcessMargin(ctx context.Context, positions ...events.Ma
...
+		// do not remove them for now
+		if m.settlement.HasPosition(pos.Party()) {
+			continue
+		}
+
 		// now check if all buy/sell/size are 0
 		if pos.Buy() != 0 || pos.Sell() != 0 || pos.Size() != 0 || !pos.VWBuy().IsZero() || !pos.VWSell().IsZero() {
 			// position is not 0, nothing to release surely
-			return
+			continue
```
ze97286, line 1298: good catch.

ze97286 (dismissed):

EVODelavega (dismissed):

ValentinTrinque (approved):

ze97286 (approved):

PR #6866, opened 2022-11-18
feat: dehistory handle protocol upgrade

### LINKED ISSUE #6770: Dehistory segment should be created on protocol upgrade
Dehistory segment should be created on protocol upgrade.   Right now the segments are created in datanode according to the snapshot.interval network param.  However core also snapshots on protocol upgrade, for consistency and to ensure upgrades across boundaries work the datanode should also create a segment on protocol upgrade boundary.

ettec:
closes #6770

### DIFF
--- a/cmd/data-node/commands/start/node_pre.go
+++ b/cmd/data-node/commands/start/node_pre.go
@@ -195,7 +195,7 @@ func (l *NodeCommand) initialiseDatabase() error {
 		}
 	}
 
-	err = sqlstore.MigrateSchema(l.Log, l.conf.SQLStore.ConnectionConfig)
+	err = sqlstore.MigrateToLatestSchema(l.Log, l.conf.SQLStore)
 	if err != nil {
 		return fmt.Errorf("failed to migrate to latest schema:%w", err)
 	}
@@ -233,21 +233,26 @@ func (l *NodeCommand) preRun([]string) (err error) {
 
 	eventSource = broker.NewFanOutEventSource(eventSource, l.conf.SQLStore.FanOutBufferSize, 2)
 
-	var onBlockCommittedFn func(ctx context.Context, chainId string, lastCommittedBlockHeight int64)
+	var onBlockCommittedHandler func(ctx context.Context, chainId string, lastCommittedBlockHeight int64)
+	var protocolUpgradeHandler broker.ProtocolUpgradeHandler
 
-	if l.snapshotService != nil {
+	if l.conf.DeHistory.Enabled {
 		blockCommitHandler := snapshot.NewBlockCommitHandler(l.Log, l.snapshotService.SnapshotData, l.networkParameterService.GetByKey,
-			l.conf.Broker)
-		onBlockCommittedFn = blockCommitHandler.OnBlockCommitted
+			bool(l.conf.Broker.UseEventFile), l.conf.Broker.FileEventSourceConfig.TimeBetweenBlocks.Duration)
+		onBlockCommittedHandler = blockCommitHandler.OnBlockCommitted
+		protocolUpgradeHandler = dehistory.NewProtocolUpgradeHandler(l.Log, l.protocolUpgradeService,
+			l.deHistoryService.CreateAndPublishSegment)
 	} else {
-		onBlockCommittedFn = func(ctx context.Context, chainId string, lastCommittedBlockHeight int64) {}
+		onBlockCommittedHandler = func(ctx context.Context, chainId string, lastCommittedBlockHeight int64) {}
+		protocolUpgradeHandler = dehistory.NewProtocolUpgradeHandler(l.Log, l.protocolUpgradeService,
+			func(ctx context.Context, chainID string, toHeight int64) error { return nil })
 	}
 
 	l.sqlBroker = broker.NewSQLStoreBroker(l.Log, l.conf.Broker, l.conf.ChainID, eventSource,
 		l.transactionalConnectionSource,
 		l.blockStore,
-		l.protocolUpgradeService,
-		onBlockCommittedFn,
+		onBlockCommittedHandler,
+		protocolUpgradeHandler,
 		l.GetSQLSubscribers(),
 	)
 
@@ -281,7 +286,12 @@ func (l *NodeCommand) initialiseDecentralizedHistory() error {
 	var err error
 	l.snapshotService, err = snapshot.NewSnapshotService(snapshotServiceLog, l.conf.DeHistory.Snapshot,
 		l.conf.SQLStore.ConnectionConfig, l.vegaPaths.StatePathFor(paths.DataNodeDeHistorySnapshotCopyFrom),
-		l.vegaPaths.StatePathFor(paths.DataNodeDeHistorySnapshotCopyTo))
+		l.vegaPaths.StatePathFor(paths.DataNodeDeHistorySnapshotCopyTo), func(version int64) error {
+			if err = sqlstore.MigrateToSchemaVersion(deHistoryLog, l.conf.SQLStore, version); err != nil {
+				return fmt.Errorf("failed to migrate to schema version %d: %w", version, err)
+			}
+			return nil
+		})
 	if err != nil {
 		return fmt.Errorf("failed to create snapshot service:%w", err)
 	}
--- a/datanode/dehistory/protocol_upgrade_handler.go
+++ b/datanode/dehistory/protocol_upgrade_handler.go
@@ -1,0 +1,41 @@
+package dehistory
+
+import (
+	"context"
+	"fmt"
+
+	"code.vegaprotocol.io/vega/datanode/service"
+	"code.vegaprotocol.io/vega/logging"
+)
+
+type ProtocolUpgradeHandler struct {
+	log                     *logging.Logger
+	protocolUpgradeService  *service.ProtocolUpgrade
+	createAndPublishSegment func(ctx context.Context, chainID string, toHeight int64) error
+}
+
+func NewProtocolUpgradeHandler(log *logging.Logger, protocolUpgradeService *service.ProtocolUpgrade,
+	createAndPublishSegment func(ctx context.Context, chainID string, toHeight int64) error,
+) *ProtocolUpgradeHandler {
+	return &ProtocolUpgradeHandler{
+		log:                     log.Named("protocol-upgrade-handler"),
+		protocolUpgradeService:  protocolUpgradeService,
+		createAndPublishSegment: createAndPublishSegment,
+	}
+}
+
+func (t *ProtocolUpgradeHandler) OnProtocolUpgradeEvent(ctx context.Context, chainID string,
+	lastCommittedBlockHeight int64,
+) error {
+	if err := t.createAndPublishSegment(ctx, chainID, lastCommittedBlockHeight); err != nil {
+		return fmt.Errorf("failed to create and publish segment: %w", err)
+	}
+
+	t.protocolUpgradeService.SetProtocolUpgradeStarted()
+
+	return nil
+}
+
+func (t *ProtocolUpgradeHandler) GetProtocolUpgradeStarted() bool {
+	return t.protocolUpgradeService.GetProtocolUpgradeStarted()
+}
--- a/datanode/dehistory/snapshot/service_create_snapshot.go
+++ b/datanode/dehistory/snapshot/service_create_snapshot.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"hash"
 	"io"
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/georgysavva/scany/pgxscan"
 
 	"code.vegaprotocol.io/vega/datanode/sqlstore"
 
@@ -21,7 +24,7 @@ import (
 	"github.com/jackc/pgx/v4/pgxpool"
 )
 
-type CreateSnapshotResult struct {
+type MetaData struct {
 	CurrentStateSnapshot     CurrentState
 	HistorySnapshot          History
 	CurrentStateSnapshotPath string
@@ -29,19 +32,30 @@ type CreateSnapshotResult struct {
 	DatabaseVersion          int64
 }
 
-func (b *Service) CreateSnapshot(ctx context.Context, chainID string, fromHeight int64, toHeight int64) (CreateSnapshotResult, error) {
+var (
+	ErrSnapshotExists = errors.New("Snapshot exists")
+	ErrNoLastSnapshot = errors.New("No last snapshot")
+)
+
+func (b *Service) CreateSnapshot(ctx context.Context, chainID string, toHeight int64) (MetaData, error) {
+	return b.createNewSnapshot(ctx, chainID, toHeight, false)
+}
+
+func (b *Service) CreateSnapshotAsynchronously(ctx context.Context, chainID string, toHeight int64) (MetaData, error) {
+	return b.createNewSnapshot(ctx, chainID, toHeight, true)
+}
+
+func (b *Service) createNewSnapshot(ctx context.Context, chainID string, toHeight int64,
+	async bool,
+) (MetaData, error) {
 	var err error
 	if len(chainID) == 0 {
-		return CreateSnapshotResult{}, fmt.Errorf("chain id is required")
+		return MetaData{}, fmt.Errorf("chain id is required")
 	}
-
-	currentSnapshot := NewCurrentSnapshot(chainID, toHeight)
-	historySnapshot := NewHistorySnapshot(chainID, fromHeight, toHeight)
-	b.log.Infof("creating snapshot for %+v", historySnapshot)
 
 	dbMetaData, err := NewDatabaseMetaData(ctx, b.connConfig)
 	if err != nil {
-		return CreateSnapshotResult{}, fmt.Errorf("failed to get data dump metadata: %w", err)
+		return MetaData{}, fmt.Errorf("failed to get data dump metadata: %w", err)
 	}
 
 	var cleanUp []func()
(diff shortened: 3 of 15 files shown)


datanode/dehistory/snapshot/service_create_snapshot.go
```diff
@@ -61,30 +65,44 @@ func (b *Service) CreateSnapshot(ctx context.Context, chainID string, fromHeight
...
-		return CreateSnapshotResult{}, fmt.Errorf("failed to begin copy table data transaction: %w", err)
+		return MetaData{}, fmt.Errorf("failed to begin copy table data transaction: %w", err)
 	}
 	// Rolling back a committed transaction does nothing
 	cleanUp = append(cleanUp, func() { _ = copyDataTx.Rollback(ctx) })
 
 	if _, err = copyDataTx.Exec(ctx, "SET TRANSACTION ISOLATION LEVEL SERIALIZABLE"); err != nil {
 		runAllInReverseOrder(cleanUp)
-		return CreateSnapshotResult{}, fmt.Errorf("failed to set transaction isolation level to serilizable:%w", err)
+		return MetaData{}, fmt.Errorf("failed to set transaction isolation level to serilizable:%w", err)
```
wwestgarth, line 82: Can we just make sure we have a space before the `%w` to make the errors more readable? i.e `: %w`?

datanode/sqlstore/sqlstore.go
```diff
@@ -38,21 +38,57 @@ var EmbedMigrations embed.FS
...
+	err := runMigration(log, config, migrate)
+	if err != nil {
+		return fmt.Errorf("failed to run migration to latest schema:%w", err)
+	}
+
+	return nil
+}
+
+func MigrateToSchemaVersion(log *logging.Logger, config Config, version int64) error {
+	migrate := func(db *sql.DB) error {
```
wwestgarth, line 59: I don't think we need to create these inline functions everywhere, can they not just look like this:
```
func MigrateToSchemaVersion() {
    db := getDbThing()
    defer db.Close()

    if err := goose.UpTo(db, SQLMigrationsDir, version); err != nil {
        return fmt.Errorf("error migrating sql schema: %w", err)
    }
}
```
and

```
func MigrateToLatestSchema() {
    db := getDbThing()
    defer db.Close()

    if err := goose.Up(db, SQLMigrationsDir); err != nil {
        return fmt.Errorf("error migrating sql schema: %w", err)
    }
}
```
where `getDbThing()` is the start of `runMigration`

datanode/dehistory/store/store.go
```diff
@@ -521,6 +521,19 @@ func (p *Store) extractHistorySegmentToStagingArea(ctx context.Context, toHeight
 	return nil
 }
 
+func (p *Store) HasHistorySegmentForHeight(toHeight int64) (bool, error) {
+	_, err := p.index.Get(toHeight)
```
wwestgarth, line 525: Could we invert the logic and not have so much indentation?
```
func (p *Store) HasHistorySegmentForHeight() {
    _, err := p.index.Get(toHeight)
    if err == nil {
        return true, nil
    }

    if errors.Is(err, ErrIndexEntryNotFound) {
        return false, nil
    }

    return false, fmt.Errorf("failed to get index entry for height:%d:%w", toHeight, err)
}
```

cmd/data-node/commands/start/node_pre.go
```diff
@@ -233,21 +233,28 @@ func (l *NodeCommand) preRun([]string) (err error) {
...
-			l.conf.Broker)
-		onBlockCommittedFn = blockCommitHandler.OnBlockCommitted
+			bool(l.conf.Broker.UseEventFile), l.conf.Broker.FileEventSourceConfig.TimeBetweenBlocks.Duration)
+		onBlockCommittedHandler = blockCommitHandler.OnBlockCommitted
+		protocolUpgradeHandler = dehistory.NewProtocolUpgradeHandler(l.Log, l.conf.DeHistory, l.protocolUpgradeService,
+			l.snapshotService.CreateSnapshotAsynchronously, l.deHistoryService.SnapshotPublished)
 	} else {
-		onBlockCommittedFn = func(ctx context.Context, chainId string, lastCommittedBlockHeight int64) {}
+		onBlockCommittedHandler = func(ctx context.Context, chainId string, lastCommittedBlockHeight int64) {}
+		protocolUpgradeHandler = dehistory.NewProtocolUpgradeHandler(l.Log, l.conf.DeHistory, l.protocolUpgradeService,
```
wwestgarth, line 247: Why can't `NewProtocolUpgradeHandler()` just take in `l.snapshotService` and `l.dehistoryService` to call those functions? Its all a bit convoluted passing their functions around. The config is save in it already so you just need to do `l.conf.DeHistory.Enabled` as a condition on whether to call them:
```
func (t *ProtocolUpgradeHandler) OnProtocolUpgradeEvent() {
    if !t.cfg.Enabled {
        t.protocolUpgradeService.SetProtocolUpgradeStarted()
        return
   }

    // do the snapshot stuff
    t.snapshotService.createSnapshotAsynchronously()
}

File: datanode/dehistory/snapshot/service_load_snapshot.go [hunk 5]
```diff
@@ -104,10 +104,11 @@ func (b *Service) loadSnapshot(ctx context.Context, snapshotData compressedFileM
 		return 0, fmt.Errorf("failed to decompress and untar data: %w", err)
 	}
 
-	// TODO handle loading of data from older database version by upgrading database as snapshots
-	// TODO are loaded, for now it is expected that all snapshots are from the same database version
 	if currentDbVersion != snapshotDbVersion {
-		return 0, fmt.Errorf("snapshot database version %d does not match current database version %d", snapshotDbVersion, currentDbVersion)
+		err = b.migrateDatabaseToVersion(snapshotDbVersion)
```
wwestgarth, line 108: The caller calls this in a loop like this:
```
for _, history := range contiguousHistory {
	rowsCopied, err := b.loadSnapshot(ctx, history, sourceDir, dbVersion, vegaDbConn)
```
but we never update `dbVersion` to be the newer one. So if `dbVersion` is 1 so if we hit segment N with version 2 we'll call to migrate. Then we'll roll to segment N+1 which is still version 2 so we don't need to migrate, but we will because `dbVersion` is still 1.

datanode/dehistory/protocol_upgrade_handler.go
```diff
@@ -0,0 +1,81 @@
...
+) (snapshot.MetaData, error) {
+	snapshotMetaData, err := t.createSnapshotAsynchronously(ctx, chainID, lastCommittedBlockHeight)
+	if err != nil && !errors.Is(err, snapshot.ErrSnapshotExists) {
+		return snapshot.MetaData{}, fmt.Errorf("failed to create snapshot:%w", err)
+	}
+
+	t.log.Infof("waiting for snapshot to be published, maximum wait time %s, snapshot %+v",
+		t.cfg.ProtocolUpgradeTimeout.Duration, snapshotMetaData)
+
+	ctxWithTimeout, cancelFn := context.WithTimeout(ctx, t.cfg.ProtocolUpgradeTimeout.Duration)
```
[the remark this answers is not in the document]
ettec: I took the view that there is a meaningfully higher probability of issues occurring during protocol upgrade and therefore it would be prudent to have the last pre upgrade segment both created AND published to give us the most recovery options (for example you can restore from dehistory if that were needed in some unforeseen scenario)  But maybe being overzealous you think?

datanode/dehistory/snapshot/service_load_snapshot.go
ettec: so we do, guess goose must be doing something sensible here to avoid multiple upgrade, will sort this

datanode/dehistory/protocol_upgrade_handler.go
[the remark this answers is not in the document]
wwestgarth: I don't really mind whether it gets published or not, I just thought maybe not publishing it would make it a bit more simple. 

I think as long as its all synchronous it'll be fine because it'll help minimise the downtime during the upgrade. For example if a node operator sets their `AddSnapshotsInterval` to `1hr` for whatever reason, during upgrade the datanode could just be sitting their not shutting down for an hour waiting just waiting for that publish tick to happen.

datanode/dehistory/snapshot/service_load_snapshot.go
ettec: sorted now

cmd/data-node/commands/start/node_pre.go
ettec: Protocol handler has changed, this comment now out of date

datanode/dehistory/store/store.go
ettec: code removed - no longer needed

datanode/sqlstore/sqlstore.go
ettec: done this was to remove a bit of duplicaton of db creation and goose setup, but arguagbly the complexity not worth the biscuit, just duplicated it instead, encapuslation of db creation saves one line so not really worth it.

datanode/dehistory/protocol_upgrade_handler.go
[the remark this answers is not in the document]
ettec: refactored the code to allow synchronous creation and publication

wwestgarth:
Looks good, just two things which are maybe non-blocking:
- might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
- The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?

ettec:
> Looks good, just two things which are maybe non-blocking:
> 
> * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?

True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!).  Will check with jake.

wwestgarth:
> > Looks good, just two things which are maybe non-blocking:
> > 
> > * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> > * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?
> 
> True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!). Will check with jake.

Well its a config option so we don't know (and have no control over) what node operators will set the interval to. And if it turns out on testnet or whatever the snapshots do take longer with more data, we can't just keep slowly increasing the timeout.
So what is the reason for this timeout to exist? what happens if we wait too long, and can we not just test for that condition instead?

ettec:
> > > Looks good, just two things which are maybe non-blocking:
> > > 
> > > * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> > > * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?
> > 
> > 
> > True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!). Will check with jake.
> 
> Well its a config option so we don't know (and have no control over) what node operators will set the interval to (for example on testnet I think its still set to 5s). And if it turns out the snapshots do take longer with more data, we can't just keep slowly increasing the timeout.
> 
> So what is the reason for this timeout to exist? what happens if we wait too long, and can we not just test for that condition instead?

On non real life networks with very short intervals its possible for the creation time to take longer than block processing time between intervals, this lock is to prevent that happening and ensure the segments are produced sequentially.  There is no need for a timeout necessarily, could just be a bog standard lock,  my thinking was that after a reasoable period it should timeout and kill the node rather than just sit and hold up the processing, maybe not have the config perhaps, but by having a timeout we are in effect testing for that condition, though we also producing stats to monitor this production time, there is nothing that can be done if this processing is blocked for whatever reason, the datanode cannot continue as it cannot produce new segments.

wwestgarth:
> > > > Looks good, just two things which are maybe non-blocking:
> > > > 
> > > > * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> > > > * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?
> > > 
> > > 
> > > True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!). Will check with jake.
> > 
> > 
> > Well its a config option so we don't know (and have no control over) what node operators will set the interval to (for example on testnet I think its still set to 5s). And if it turns out the snapshots do take longer with more data, we can't just keep slowly increasing the timeout.
> > So what is the reason for this timeout to exist? what happens if we wait too long, and can we not just test for that condition instead?
> 
> On non real life networks with very short intervals its possible for the creation time to take longer than block processing time between intervals, this lock is to prevent that happening and ensure the segments are produced sequentially. There is no need for a timeout necessarily, could just be a bog standard lock, my thinking was that after a reasoable period it should timeout and kill the node rather than just sit and hold up the processing, maybe not have the config perhaps, but by having a timeout we are in effect testing for that condition, though we also producing stats to monitor this production time, there is nothing that can be done if this processing is blocked for whatever reason, the datanode cannot continue as it cannot produce new segments.

I don't think we can sweep it away as being only a problem on "non-real" networks though. With this protocol upgrade change its entirely possible to have a snapshot taken on two subsequent blocks.

To me it sounds like the condition to panic is if we have 2 pending snapshots waiting.

ettec:
> > > > > Looks good, just two things which are maybe non-blocking:
> > > > > 
> > > > > * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> > > > > * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?
> > > > 
> > > > 
> > > > True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!). Will check with jake.
> > > 
> > > 
> > > Well its a config option so we don't know (and have no control over) what node operators will set the interval to (for example on testnet I think its still set to 5s). And if it turns out the snapshots do take longer with more data, we can't just keep slowly increasing the timeout.
> > > So what is the reason for this timeout to exist? what happens if we wait too long, and can we not just test for that condition instead?
> > 
> > 
> > On non real life networks with very short intervals its possible for the creation time to take longer than block processing time between intervals, this lock is to prevent that happening and ensure the segments are produced sequentially. There is no need for a timeout necessarily, could just be a bog standard lock, my thinking was that after a reasoable period it should timeout and kill the node rather than just sit and hold up the processing, maybe not have the config perhaps, but by having a timeout we are in effect testing for that condition, though we also producing stats to monitor this production time, there is nothing that can be done if this processing is blocked for whatever reason, the datanode cannot continue as it cannot produce new segments.
> 
> I don't think we can sweep it away as being only a problem on "non-real" networks though. With this protocol upgrade change its entirely possible to have a snapshot taken on two subsequent blocks.
> 
> To me it sounds like the condition to panic is if we have 2 pending snapshots waiting.

yes, possible to do that, so we do need a lock to ensure that doesn't happen on real network as well in that scenario.  But maybe the timeout is an unneccessary complication? as long as we log that its waiting - one less bit of a config and I don't see any downside really.

wwestgarth:
> > > > > > Looks good, just two things which are maybe non-blocking:
> > > > > > 
> > > > > > * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> > > > > > * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?
> > > > > 
> > > > > 
> > > > > True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!). Will check with jake.
> > > > 
> > > > 
> > > > Well its a config option so we don't know (and have no control over) what node operators will set the interval to (for example on testnet I think its still set to 5s). And if it turns out the snapshots do take longer with more data, we can't just keep slowly increasing the timeout.
> > > > So what is the reason for this timeout to exist? what happens if we wait too long, and can we not just test for that condition instead?
> > > 
> > > 
> > > On non real life networks with very short intervals its possible for the creation time to take longer than block processing time between intervals, this lock is to prevent that happening and ensure the segments are produced sequentially. There is no need for a timeout necessarily, could just be a bog standard lock, my thinking was that after a reasoable period it should timeout and kill the node rather than just sit and hold up the processing, maybe not have the config perhaps, but by having a timeout we are in effect testing for that condition, though we also producing stats to monitor this production time, there is nothing that can be done if this processing is blocked for whatever reason, the datanode cannot continue as it cannot produce new segments.
> > 
> > 
> > I don't think we can sweep it away as being only a problem on "non-real" networks though. With this protocol upgrade change its entirely possible to have a snapshot taken on two subsequent blocks.
> > To me it sounds like the condition to panic is if we have 2 pending snapshots waiting.
> 
> yes, possible to do that, so we do need a lock to ensure that doesn't happen on real network as well in that scenario. But maybe the timeout is an unneccessary complication? as long as we log that its waiting - one less bit of a config and I don't see any downside really.

I'm happy for this to be addressed separately as I think its wider issue eg. is also affects replay.

wwestgarth (dismissed):

ettec:
> > > > > > > Looks good, just two things which are maybe non-blocking:
> > > > > > > 
> > > > > > > * might be worth @jgsbennett running one of his upgrade tests (or advise which test would work on the CI) first. I know hes trying to make progress on his PR and it would be sad if things started failing under his feet
> > > > > > > * The upgrade "forcing" a snapshot means that it is possible for two snapshots to happen very close to each other and I wonder what that means for the lock-timeout we have on snapshot creation. What is the upper-bound on the time we should wait for the lock? If its the case where it'll go all wrong if we have 2 pending snapshot queued up then would it be worth just binning the timeout on the lock, and just panicking if when get to the place where we have 2 pending snapshots?
> > > > > > 
> > > > > > 
> > > > > > True, it would be possible for close together snapshots, however the lock timeout is 5 mins, which should be plenty of time to produce both (until bitter experience says otherwise!). Will check with jake.
> > > > > 
> > > > > 
> > > > > Well its a config option so we don't know (and have no control over) what node operators will set the interval to (for example on testnet I think its still set to 5s). And if it turns out the snapshots do take longer with more data, we can't just keep slowly increasing the timeout.
> > > > > So what is the reason for this timeout to exist? what happens if we wait too long, and can we not just test for that condition instead?
> > > > 
> > > > 
> > > > On non real life networks with very short intervals its possible for the creation time to take longer than block processing time between intervals, this lock is to prevent that happening and ensure the segments are produced sequentially. There is no need for a timeout necessarily, could just be a bog standard lock, my thinking was that after a reasoable period it should timeout and kill the node rather than just sit and hold up the processing, maybe not have the config perhaps, but by having a timeout we are in effect testing for that condition, though we also producing stats to monitor this production time, there is nothing that can be done if this processing is blocked for whatever reason, the datanode cannot continue as it cannot produce new segments.
> > > 
> > > 
> > > I don't think we can sweep it away as being only a problem on "non-real" networks though. With this protocol upgrade change its entirely possible to have a snapshot taken on two subsequent blocks.
> > > To me it sounds like the condition to panic is if we have 2 pending snapshots waiting.
> > 
> > 
> > yes, possible to do that, so we do need a lock to ensure that doesn't happen on real network as well in that scenario. But maybe the timeout is an unneccessary complication? as long as we log that its waiting - one less bit of a config and I don't see any downside really.
> 
> I'm happy for this to be addressed separately as I think its wider issue eg. is also affects replay.

Actually it shouldn't affect replay, usually at an interval of 1000, the snapshot creation time takes less time than processing a 1000 blocks.  Currently due to a bug on the copy filter (90% sure at least - tbc) its taking a loooong time to snapshot data.

jeremyletang (approved):

PR #6883, opened 2022-11-21
WIP: rationalise when we check for reference moves

witgaw:


gordsport:
Hey @witgaw 
Is this PR still required?

witgaw:
It's not strictly required for anything (there are no issues corresponding to it), but it's something I'd like to explore. I'll close the PR and keep the branch alive to reduce the clutter.

PR #6890, opened 2022-11-21
fix: add MTM before final settlement and include settlement engine in…

### LINKED ISSUE #6831: Snapshot: settlement engine position differ during SettleMTM

### LINKED ISSUE #6892: market_data event to determine next MTM recalc always "nextMarkToMarket":"-6795364573871345152"}
MTM rework has been merged and a new field has been added to market_data event to determine a time when the MTM calculation is going to be run at market level.

This timestamp is always the same value.

"nextMarkToMarket":"-6795364573871345152"}

EVODelavega:
Settlement engine now is its own entity in snapshots + sort out issue with MTM settlement (in particular trades) not being part of the final settlement when closing a market

Closes #6831
closes #6892


wwestgarth:
I've been looking at the failing snapshot test, and the issue is that there is a pointer to a position that is shared between the positions engine and the settlement engine. Upon restore that pointer-link is not restored.

The pointers get shared in the call to `SettleMTM()` which recieves a slice of interfaces `[]events.MarketPosition` which underneath are actually a pointer to a postion. When we call `current := e.getCurrentPosition(party, evt)` we store in the settlement engine the pointer to the position wrapped in the settlement engine's view of a position:
```
func newPos(evt events.MarketPosition) *pos {
	return &pos{
		MarketPosition: evt, <-- this is really a pointer to the positions engines position
		party:          evt.Party(),
		size:           evt.Size(),
		price:          evt.Price().Clone(),
	}
}
```

later during a market `OnTick()` we call `m.removeExpiredOrders()` which removes an order and updates the party's position. In the _unrestored node_ this also updates the position in the settlement engine, in the _restored node_ it doesn't. We then fall out of consensus.

So we need to either (quick, dirty, and hack) reshare the pointer when we restore, or do it properly and not share the pointers.

jeremyletang:
> I've been looking at the failing snapshot test, and the issue is that there is a pointer to a position that is shared between the positions engine and the settlement engine. Upon restore that pointer-link is not restored.
> 
> The pointers get shared in the call to `SettleMTM()` which recieves a slice of interfaces `[]events.MarketPosition` which underneath are actually a pointer to a postion. When we call `current := e.getCurrentPosition(party, evt)` we store in the settlement engine the pointer to the position wrapped in the settlement engine's view of a position:
> 
> ```
> func newPos(evt events.MarketPosition) *pos {
> 	return &pos{
> 		MarketPosition: evt, <-- this is really a pointer to the positions engines position
> 		party:          evt.Party(),
> 		size:           evt.Size(),
> 		price:          evt.Price().Clone(),
> 	}
> }
> ```
> 
> later during a market `OnTick()` we call `m.removeExpiredOrders()` which removes an order and updates the party's position. In the _unrestored node_ this also updates the position in the settlement engine, in the _restored node_ it doesn't. We then fall out of consensus.
> 
> So we need to either (quick, dirty, and hack) reshare the pointer when we restore, or do it properly, refactor, and not share the pointers at all.

@wwestgarth see that ticket: https://github.com/vegaprotocol/vega/issues/6910. we discussed removing with the position in the settlement engine with @EVODelavega and @ze97286 offline. 

I believe if possible that's what should be done. However this PR is already pretty big. Seeing that we want to do the dedup we could:
- keep it as is, knowing it's going to break, but implement the deduping in a separate PR
- quick fix it now, and implement the proper dedup in a follow up PR.

wwestgarth:
I ran the full snapshots tests we have in the system-test suite and there is a problem loading from a snapshot after a market has been settled. We get the error:
```
2022-11-23T12:36:58.214Z	PANIC	processor	processor/abci.go:531	failed to check if snapshot has been loaded	{"error": "unknown snapshot namespace settlement.58ec60e8568915c83415beb1ef9959e1f7a0055f9a89009e8568892f33bafea1"}
```

https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/29089/tests

jeremyletang (approved):

wwestgarth (approved):

PR #6912, opened 2022-11-23
feat: simplify config and make it so only publishing nodes create segments

### LINKED ISSUE #6911: Simplify config and ensure non-publishing node does not create segments
To allow accurate assessment of impact of decentralized history creation on node performance the 'AddToStore' flag should be changed to 'Publish' and should control not just whether segments get added to IPFS but also whether they get created at all.  Also this change goes some towards supporting a 'lite' form of the datanode that will not publish decentralized history segments (to be done in a seperate PR).

ettec:
closes #6911 

To allow accurate assessment of impact of decentralized history creation on node performance the 'AddToStore' flag should be changed to 'Publish' and should control not just whether segments get added to IPFS but also whether they get created at all. Also this change goes some towards supporting a 'lite' form of the datanode that will not publish decentralized history segments (to be done in a seperate PR).

### DIFF
--- a/datanode/dehistory/config.go
+++ b/datanode/dehistory/config.go
@@ -1,8 +1,6 @@
 package dehistory
 
 import (
-	"time"
-
 	"code.vegaprotocol.io/vega/datanode/config/encoding"
 	"code.vegaprotocol.io/vega/datanode/dehistory/initialise"
 	"code.vegaprotocol.io/vega/datanode/dehistory/snapshot"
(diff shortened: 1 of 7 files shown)


jeremyletang (dismissed):

datanode/dehistory/config.go
```diff
@@ -15,8 +13,7 @@ type Config struct {
 	Enabled       encoding.Bool     `long:"enabled" description:"set to false to disable decentralized history"`
 	WipeOnStartup encoding.Bool     `long:"wipe-on-startup" description:"remove all deHistory state on startup"`
 
-	AddSnapshotsToStore  encoding.Bool     `long:"add-snapshots-to-store" description:"if true snapshot data produced by this node will be added to the decentralise history store"`
-	AddSnapshotsInterval encoding.Duration `long:"add-snapshots-interval" description:"interval between checking for and adding snapshot data to the decentralised store"`
+	Publish encoding.Bool `long:"publish" description:"if true this node will create and publish decentralized history segments"`
```
wwestgarth, line 16: How does this differ from:
```
Enabled       encoding.Bool     `long:"enabled" description:"set to false to disable decentralized history"`
```
?

ettec: That prevents the node from 'reading/recovering from' dehistory and connecting into the IPFS peer network.

wwestgarth: but in the code it also turns off actually taking snapshots?
```
if l.conf.DeHistory.Enabled {
		l.Log.Info("Initializing Decentralized History")
		err = l.initialiseDecentralizedHistory()
		if err != nil {
			return fmt.Errorf("failed to initialise decentralized history:%w", err)
		}
	}
```
`l.snapshotService` is initialised in `initialiseDecentralizedHistory()`

ettec: yep, publish is to control whether the node is a reader/writer of segments in essence.  Enabled just switches dehistory off altogether, so all IPFS stuff and including ability to use it to restore a node and includes switching off snapshot creation/publishing.

wwestgarth (dismissed):

jeremyletang (approved):

PR #6916, opened 2022-11-23
feat: support filters in ordersConnection

### LINKED ISSUE #6615: Support filters in ordersConnection
# Feature Overview

**In order to** allow traders to better track orders
**We will** update the ordersConnection query and orders subscription
**So that** client applications can retrieved data in the correct order and with filters applied

# Acceptance Criteria

- [x] I can filter by one or more order statuses
- [ ] I can filter by one or more order types
- [ ] I can filter by one or more market ids
- [ ] I can filter by one or more order TIFs
- [ ] I can filter by liquidity order

### Additional information
Suggested query
```
ordersConnection(
  dateRange: { ... }
  pagination: { ... }
  filter: {
    status: [OrderStatus!]
    type: [OrderType!]
    marketId: [ID!]
    timeInForce: [OrderTimeInForce!]
    excludeLiquidity: Boolean
  }
) {
  edges { ... } 
}
```

guoguojin:
Closes #6615


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -565,6 +565,13 @@ message GetOrderResponse {
     vega.Order order = 1;
 }
 
+message OrderFilter {
+    repeated vega.Order.Status statuses = 1;
+    repeated vega.Order.Type types = 2;
+    repeated vega.Order.TimeInForce time_in_forces = 3;
+    bool exclude_liquidity = 4;
+}
```
jeremyletang, line 573: Why not just have those at the top level of the ListOrderRequest? would make builing the query simpler + it's not reused anywhere

datanode/gateway/graphql/order_filter_resolver.go
```diff
@@ -0,0 +1,20 @@
...
+
+func (o orderFilterResolver) Status(ctx context.Context, obj *v2.OrderFilter, data []vega.Order_Status) error {
+	obj.Statuses = data
+	return nil
+}
+
+func (o orderFilterResolver) TimeInForce(ctx context.Context, obj *v2.OrderFilter, data []vega.Order_TimeInForce) error {
+	obj.TimeInForces = data
+	return nil
+}
```
pscott31, line 20: this is a bit of an unusual pattern for resolvers, normally they return the thing right, rather than setting it somewhere? But perhaps it's fine in which case ignore me :)

datanode/gateway/graphql/schema.graphql
```diff
@@ -599,7 +599,7 @@ type Query {
   ): Order!
 
   "An order in the Vega network found by referenceID"
-  orderByReference("Reference for an order" reference: String!): Order!
+  orderByReference("Reference for an order" reference: String!, filter: OrderFilter): Order!
```
pscott31, line 602: Do we actually want/need a filter on this guy?  I would have thought usually you're reference should be pretty unique and you were only expecting one result (though obv. not guaranteed)

datanode/gateway/graphql/schema.graphql
```diff
@@ -1709,6 +1709,8 @@ type Market {
     dateRange: DateRange,
     "Pagination information"
     pagination: Pagination
+    "Filter orders"
+    filter: OrderFilter
```
pscott31, line 1713: Ditto above; I would have thought this is generally for the case where you've got an order ID and you want to see it's history, I can't really think of a use-case for wanting to filter it further; though I guess it doesn't do any harm :)

datanode/gateway/graphql/order_filter_resolver.go
guoguojin: I'm not sure what you mean. `gqlgen` didn't generate any resolvers so I had to add these, but there's really no work to be done other than setting the OrderFilter property with the data provided so not sure what else there is to do. I think it's because there are already resolvers for the GraphQL enumerations to their gRPC counterparts.

pscott31: I mean, e.g. I was expecting to see here something like 
```
func (o orderFilterResolver) Status(ctx context.Context, obj *v2.OrderFilter, data []vega.Order_Status) Statuses, error {
	return obj.Statues, nil
}
```

But now I look upon it with fresh eyes maybe this is an _input_ resolver not an _output_ one and that is why it's setting not getting :)

datanode/gateway/graphql/schema.graphql
guoguojin: I'm sorry, I'm a bit confused, the ordersConnection under markets and parties lists the current orders for the market or party. It doesn't let you filter for a specific order and get the order's history. The filtering here should be valid.

There is a gRPC function that allows you to retrieve the order history for a specific order ID, which is exposed in GraphQL the orderVersionsConnection that doesn't have any filtering but lets you look up the history of an order using it's orderID.

datanode/gateway/graphql/schema.graphql
guoguojin: Yes I think you're right, a bit overzealous with the filtering here, I will remove it.

datanode/gateway/graphql/schema.graphql
guoguojin: I think I understand where the confusion is. The way GH is displaying this makes it look like the filter is being applied to orderVersionsConnection, but it's not. The filter is only applied to ordersConnection under the partiesConnection and marketsConnection query.

pscott31 (dismissed):

ValentinTrinque (approved):

PR #6936, opened 2022-11-25
feat: separate settled position from position

### LINKED ISSUE #6910: Deduplicate the positions from the settlement and positions engines
When the following issue has been merged:

- https://github.com/vegaprotocol/vega/issues/6831

we need to investigate how we can deduplicate the positions from the settlement and positions engines and implement if possible.

Further details TBC...

ze97286:
Closes #6910


wwestgarth (dismissed):

jeremyletang (dismissed):

core/settlement/engine.go
```diff
@@ -120,12 +121,8 @@ func (e *Engine) Update(positions []events.MarketPosition) {
...
 	for _, p := range positions {
 		party := p.Party()
-		if ps, ok := e.pos[party]; ok {
-			// ATM, this can't possibly return an error, hence we're ignoring it
-			_ = ps.update(p)
-		} else {
-			e.pos[party] = newPos(p)
-		}
+		e.settledPosition[party] = p.Size()
+		e.lastMarkPrice = p.Price()
```
EVODelavega, line 125: Not sure about this. The last Mark price should come from the market/snapshot IMO. As far as settlement engine is concerned, we have parties with a certain size/position, and some unsettled trades. All positions at MTM are held at the same price. When we call this function, can we really be sure that the price from the market position event is the right one?

core/settlement/engine.go
```diff
@@ -289,9 +287,13 @@ func (e *Engine) SettleMTM(ctx context.Context, markPrice *num.Uint, positions [
...
 	for _, evt := range positions {
 		party := evt.Party()
-		// get the current position, and all (if any) position changes because of trades
-		current := e.getCurrentPosition(party, evt)
-		// we don't care if this is a nil value
+		// NB:
+		// current = last mtm'ed position if there is such of the position from the position event
+		// lastSettledPrice = last mark price if there is mtm'ed position or the price from the position event
+		// In addition if there is no previous mtm'ed position, e.settledPosition is updated with the position event size.
+		// @@EVODelavega this seems wrong to me - I think in case there was no previous mtm'ed position size should be initialised to 0
```
EVODelavega, line 294: IIRC, the `getCurrentPosition` would either get the party position, or create a new one with size 0 if needed (ie initialise the position to 0). At some point, in market.go `handleConfirmation` we registered the party/position with the settlement engine even if no trades had occurred, essentially creating a 0-initialised entry. 

As far as initialising the position with the event: because we continuously marked to market, any time we would hit this point, a new party that had just traded would've traded at the new mark price, so their position didn't need to be marked to market. That's why we could just use the positions engine event. Now, though, we should initialise to 0, and MTM all trades based on the price they were made at, compared to the new mark price

core/settlement/engine.go
```diff
@@ -309,18 +311,18 @@ func (e *Engine) SettleMTM(ctx context.Context, markPrice *num.Uint, positions [
 		// and the old mark price at which the party held the position
 		// the trades slice contains all trade positions (position changes for the party)
 		// at their exact trade price, so we can MTM that volume correctly, too
-		mtmShare, mtmDShare, neg := calcMTM(markPrice, current.price, current.size, traded, e.positionFactor)
+		mtmShare, mtmDShare, neg := calcMTM(markPrice, lastSettledPrice, current, traded, e.positionFactor)
```
EVODelavega, line 314: Not sure if we can always use `lastSettledPrice` here, unless we make this call for each trade, and set the `lastSettledPrice` to the trade price. I'd have to take a closer look, but just thought I'd mention it as I go through this

core/settlement/engine.go
```diff
@@ -461,35 +462,29 @@ func (e *Engine) settleAll(assetDecimals uint32) ([]*types.Transfer, error) {
...
-}
-
-func (e *Engine) AddPosition(party string, evt events.MarketPosition) {
-	_ = e.getCurrentPosition(party, evt)
+	return p, e.lastMarkPrice
 }
 
 func (e *Engine) HasPosition(party string) bool {
-	_, okPos := e.pos[party]
+	_, okPos := e.settledPosition[party]
```
EVODelavega, line 475: The behaviour here might have changed slightly. If a party has traded, but the position was zero, and we haven't marked them to market, wouldn't this call return `false`, whereas the party is actually holding a position, it's just that it hasn't been MTM'ed yet

core/settlement/position.go
```diff
@@ -61,48 +56,13 @@ func (t settlementTrade) MarketPrice() *num.Uint {
 	return t.marketPrice.Clone()
 }
 
-func newPos(evt events.MarketPosition) *pos {
+func newPos(marketPosition events.MarketPosition, price *num.Uint) *pos {
```
EVODelavega, line 59: Do we still rely on this type?

core/settlement/engine.go
ze97286: this has nothing to do with the snapshot. it's not used by the snapshot, only by silly testing code.

core/settlement/engine.go
ze97286: I didn't change the behaviour - it's doing exactly what it did before. whether this was correct I didn't evaluate. I was trying to keep things working as they were.

core/settlement/position.go
ze97286: in its simplified form yes. Because unfortunately we need to pass a MarketPosition to create a transfer.

core/settlement/engine.go
ze97286: well, no need to rely on memory here, just look at the old code - it's creating a new position with the size from the position engine's event. 

```
newPos(evt events.MarketPosition) *pos {
	return &pos{
		MarketPosition: evt,
		party:          evt.Party(),
		size:           evt.Size(),
		price:          evt.Price().Clone(),
	}
}
```

so it would not create a new one with 0, which is exactly my point. It creates a new one with the position engine's size which doesn't make sense to me.

core/settlement/engine.go
ze97286: 1. this is not a change of behaviour introduced in this PR
2. you're reading this right but it's not clear to me from the context of the call (releaseExcessMargin) if it expects only to get parties with settled position + unsettled trades (which is what it will get now) or parties with unsettled trades (with or without settled position).

as this change was introduced by @jeremyletang I'll leave it to  him to comment.

core/settlement/position.go
EVODelavega: fair enough

core/settlement/engine.go
EVODelavega: Hmm, we pass in the trades, those should be handled there. Ignore this comment, was a bit too trigger-happy

core/settlement/engine.go
EVODelavega: OK, a new party will have traded, that's a given. We skip the MTM settlement if they hold their position at the new mark price and if the  party hasn't traded. The current position size at current mark price here would be `mtmShare := delta.Mul(delta, num.NewUint(uint64(size)))`, or `mtmShare = 0 * size`. After that we mark the individual trades to market, same as before. I agree that a size of 0 to initialise makes more logical sense, but after looking at the current implementation a bit, I don't think it makes that much of a difference when marking to market

core/settlement/engine.go
EVODelavega: I know this doens't have anything to do with the snapshot stuff, but what is the reason we set `lastMarkPrice` in the loop here? solely for testing, or is it something that is used elsewhere?

ze97286: well of course. it's artificially initialising settled position out of thin air so it must also initialise the last mark price because out side of this test scenario, it is impossible to have a settled position without having gone through mark to market which would have set lastMarkPrice.

EVODelavega (dismissed):

EVODelavega (approved):

PR #6940, opened 2022-11-25
fix: remove network name from config in store and use filename only

### LINKED ISSUE #6812: Remove the network name from network config when imported
But not when importing. We should check for name before import, and remove it when imported.

wwestgarth:
closes #6812


wallet/network/network.go
```diff
@@ -10,7 +10,7 @@ import (
 var ErrNetworkDoesNotHaveGRPCHostConfigured = errors.New("network configuration does not have any gRPC host set")
 
 type Network struct {
-	Name        string              `json:"name"`
+	Name        string              `json:"name" toml:"-"`
```
ValentinTrinque, line 13: I would prefer we use two dedicated struct instead of relying on the build tags to determine whether or not we should include the field. 

I am ok with reusing domain object for serialization when there is no differences, but ass soon as we start diverging, or introducing different version, it's better to back down and introduce dedicated structs, and convert them on the "IO" layer.

You can reuse most of the other structs, but the root struct should be different.

```go
type networkFileContent struct {
    LogLevel    vgencoding.LogLevel `json:"level"`
	TokenExpiry vgencoding.Duration `json:"tokenExpiry"`
	Port        int                 `json:"port"`
    //...
}

type Network struct {
    Name        string `json:"name"
    LogLevel    vgencoding.LogLevel `json:"level"`
	TokenExpiry vgencoding.Duration `json:"tokenExpiry"`
	Port        int                 `json:"port"`
    //...
}

**Review by ValentinTrinque — CHANGES_REQUESTED** [review]

File: wallet/network/store/v1/store.go [hunk 2]
```diff
@@ -19,6 +20,15 @@ type Store struct {
 	networksHome string
 }
 
+type networkFileContent struct {
+	Name        string              `json:"name" toml:"-"`
```
ValentinTrinque, line 24: This property can go away now, since it no longer lives in the file

ValentinTrinque (changes requested):

wallet/network/store/v1/store.go
wwestgarth: Ah sorry, I was being too quick. Its gone now.

ValentinTrinque (approved):

PR #6965, opened 2022-11-28
Mtm end block

### LINKED ISSUE #6569: Order margin double counted in auctions
When calculating margin during auctions we calculate order margin via a call to:
https://github.com/vegaprotocol/vega/blob/44e1777f32b332502595ac8639f89341cb5b696b/core/risk/margins_calculation.go#L65

(should be ` e.calculateMargins(m, markPrice, rf, **false**, true)`) 

and then add the order margin calculated again (this time with a volume-weighted average price) a few lines down:
https://github.com/vegaprotocol/vega/blob/44e1777f32b332502595ac8639f89341cb5b696b/core/risk/margins_calculation.go#L75-L80

### LINKED ISSUE #6949: MTM calculations to happen at the end of the block
Once the period specified by `markPriceUpdateMaximumFrequency`  has elapsed, we should mark positions to market at the end of that block, rather than at the start.

### LINKED ISSUE #6976: Recalculate margins on every MTM anniversary regardless of trades
On every MTM anniversary:
1. If any trades happened over the last period carry out MTM settlement
2. Regardless of trades happening or not always recalculate margins based on latest mark price and order book composition. Carry out the necessary margin search and release transfers.

### LINKED ISSUE #6977: Perform a MTM settlement prior to final settlement on market close
When parties have traded, and the market is settled before we've MTM'ed the last batch of trades on said market, we should MTM all positions before performing the final settlement in order to settle the last trades properly.

EVODelavega:
Closes #6949 
closes #6976  
Closes #6977 
Closes #6569 
Perform MTM settlement at the end of the block, rather than on time update

### DIFF
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -781,8 +781,7 @@ func (m *Market) blockEnd(ctx context.Context) {
 		// last traded price should not reflect the closeout trades
 		m.lastTradedPrice = mp.Clone()
 	}
-	// update margin requirements
-	_ = m.recheckMargin(ctx, m.position.Positions())
+	m.releaseExcessMargin(ctx, m.position.Positions()...)
 }
 
 func (m *Market) updateMarketValueProxy() {
@@ -1791,7 +1790,7 @@ func (m *Market) confirmMTM(
 	// we can safely ignore the error here
 	_ = m.recheckMargin(ctx, pos)
 	// release any excess if needed
-	m.releaseExcessMargin(ctx, pos...)
+	// m.releaseExcessMargin(ctx, pos...)
 }
 
 // updateLiquidityFee computes the current LiquidityProvision fee and updates
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -726,35 +726,11 @@ func (m *Market) OnTick(ctx context.Context, t time.Time) bool {
 	}
 
 	// check auction, if any. If we leave auction, MTM is performed in this call
-	inAuction := m.as.InAuction()
 	m.checkAuction(ctx, t)
-	hasTraded := m.settlement.HasTraded()
-	// have we left auction -> if so, hasTraded will be set to true, but should be false
-	// so we avoid triggering a second MTM call.
-	if inAuction && !m.as.InAuction() {
-		hasTraded = false
-	}
 	timer.EngineTimeCounterAdd()
 
 	m.updateMarketValueProxy()
 	m.updateLiquidityFee(ctx)
-	if hasTraded && m.closed {
-		// we have trades, and the market has been closed. Perform MTM sequence now so the final settlement
-		// works as expected.
-		mp := m.getLastTradedPrice()
-		if mp == nil {
-			m.broker.Send(events.NewMarketTick(ctx, m.mkt.ID, t))
-			return m.closed
-		}
-		m.markPrice = mp.Clone()
-		mcmp := num.UintZero().Div(mp, m.priceFactor) // create the market representation of the price
-		dummy := &types.Order{
-			ID:            m.idgen.NextID(),
-			Price:         mp,
-			OriginalPrice: mcmp,
-		}
-		m.confirmMTM(ctx, dummy, nil)
-	}
 	m.broker.Send(events.NewMarketTick(ctx, m.mkt.ID, t))
 	return m.closed
 }
@@ -886,6 +862,19 @@ func (m *Market) closeCancelledMarket(ctx context.Context) error {
 }
 
 func (m *Market) closeMarket(ctx context.Context, t time.Time) error {
+	// perform last MTM settlement if needed
+	if mp := m.getLastTradedPrice(); mp != nil && mp.IsZero() && m.settlement.HasTraded() {
+		// we have trades, and the market has been closed. Perform MTM sequence now so the final settlement
+		// works as expected.
+		m.markPrice = mp.Clone()
+		mcmp := num.UintZero().Div(mp, m.priceFactor) // create the market representation of the price
+		dummy := &types.Order{
+			ID:            m.idgen.NextID(),
+			Price:         mp,
+			OriginalPrice: mcmp,
+		}
+		m.confirmMTM(ctx, dummy, nil)
+	}
 	positions, err := m.settlement.Settle(t, m.assetDP)
 	if err != nil {
 		m.log.Error("Failed to get settle positions on market closed",
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -725,13 +725,61 @@ func (m *Market) OnTick(ctx context.Context, t time.Time) bool {
 		}
 	}
 
+	// check auction, if any. If we leave auction, MTM is performed in this call
+	inAuction := m.as.InAuction()
+	m.checkAuction(ctx, t)
+	hasTraded := m.settlement.HasTraded()
+	// have we left auction -> if so, hasTraded will be set to true, but should be false
+	// so we avoid triggering a second MTM call.
+	if inAuction && !m.as.InAuction() {
+		hasTraded = false
+	}
+	timer.EngineTimeCounterAdd()
+
+	m.updateMarketValueProxy()
+	m.updateLiquidityFee(ctx)
+	if hasTraded && m.closed {
+		// we have trades, and the market has been closed. Perform MTM sequence now so the final settlement
+		// works as expected.
+		mp := m.getLastTradedPrice()
+		if mp == nil {
+			m.broker.Send(events.NewMarketTick(ctx, m.mkt.ID, t))
+			return m.closed
+		}
+		m.markPrice = mp.Clone()
+		mcmp := num.UintZero().Div(mp, m.priceFactor) // create the market representation of the price
+		dummy := &types.Order{
+			ID:            m.idgen.NextID(),
+			Price:         mp,
+			OriginalPrice: mcmp,
+		}
+		m.confirmMTM(ctx, dummy, nil)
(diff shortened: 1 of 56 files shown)


core/integration/steps/time_is_updated.go
```diff
@@ -13,11 +13,14 @@
...
+
 	"code.vegaprotocol.io/vega/core/integration/stubs"
 )
 
-func TimeIsUpdatedTo(timeService *stubs.TimeStub, newTime string) {
+func TimeIsUpdatedTo(exec Execution, timeService *stubs.TimeStub, newTime string) {
 	t, err := Time(newTime)
 	panicW("date", err)
 	timeService.SetTime(t)
+	exec.BlockEnd(context.Background())
```
ze97286, line 25: I don't think that is correct. it may be harmless but if someone defined average block time to be 2 seconds and they move the time by 10 seconds, they'd expect blockEnd to be called 5 times. Similarly, if they set average block time to 5 seconds and they move the time by 3 seconds it shouldn't be called at all.

core/integration/steps/parties_place_the_following_orders.go
```diff
@@ -63,13 +65,17 @@ func PartiesPlaceTheFollowingOrdersWithTicks(exec Execution, time *stubs.TimeStu
 		}
 		// make it look like we start a new block
 		epochService.OnBlockEnd(context.Background())
+		// indicate block has ended, so we MTM if needed
+		exec.BlockEnd(context.Background())
```
ze97286, line 69: why every order on it's own block?

core/processor/abci.go
```diff
@@ -739,6 +739,7 @@ func (app *App) OnEndBlock(req tmtypes.RequestEndBlock) (ctx context.Context, re
 			AppVersion: AppVersion,
 		},
 	}
+	app.exec.BlockEnd(app.blockCtx)
```
jeremyletang, line 742: Not any issue I can't think of just now, but could there be any problems with calling that last, seeing that other engines are being call for the BlockEnd before it? Does order could matter here?

core/integration/steps/parties_place_the_following_orders.go
ze97286: and why is it even necessary vs the user moving ahead the time / number of blocks

core/integration/steps/blocks.go
```diff
@@ -46,22 +46,24 @@ func TheAverageBlockDurationWithVariance(block *helpers.Block, dur, variance str
...
 	}
 	now := time.GetTimeNow()
 	for i := int64(0); i < nr; i++ {
 		epochService.OnBlockEnd(context.Background())
+		exec.BlockEnd(context.Background())
 		now = now.Add(block.GetDuration())
 		// progress time
 		time.SetTime(now)
 	}
+	// exec.BlockEnd(context.Background())
```
ze97286, line 62: remove

core/integration/steps/blocks.go
```diff
@@ -73,8 +75,10 @@ func TheNetworkMovesAheadDurationWithBlocks(block *helpers.Block, ts *stubs.Time
 	now := ts.GetTimeNow()
 	target := now.Add(td)
 	for now.Before(target) {
+		exec.BlockEnd(context.Background())
 		now = now.Add(block.GetStep(bd))
 		ts.SetTime(now)
 	}
+	// exec.BlockEnd(context.Background())
```
ze97286, line 82: remove

core/integration/features/verified/verified-positions-resolution-1.feature
```diff
@@ -87,7 +87,8 @@ Feature: Position resolution case 1
       | sellSideProvider | BTC   | ETH/DEC19 | 2154700 | 999997854000 |
       | buySideProvider  | BTC   | ETH/DEC19 | 320     | 999999999680 |
       | aux              | BTC   | ETH/DEC19 | 320     | 999999999650 |
-      | aux2             | BTC   | ETH/DEC19 | 7430    | 999999992600 |
+      | aux2             | BTC   | ETH/DEC19 | 7520    | 999999992510 |
+      #| aux2             | BTC   | ETH/DEC19 | 7430    | 999999992600 |
```
ze97286, line 91: can we not keep this historical values please? (commenting here for all of the files above)

core/integration/steps/parties_place_the_following_orders.go
EVODelavega: Because the step here is `PartiesPlaceTheFollowingOrdersWithTicks`. This is was a relatively simple way to come as close as possible to the old behaviour (where MTM was performed after each trade, so by placing each order in its own block, any order that uncrosses can trigger MTM)

core/integration/features/verified/verified-positions-resolution-1.feature
EVODelavega: We sure can, I just kept them here to make reviewing/documenting things easier for @Jiajia-Cui and @witgaw

core/execution/market.go
```diff
@@ -754,28 +778,11 @@ func (m *Market) OnTick(ctx context.Context, t time.Time) bool {
...
-			ID:            m.idgen.NextID(),
-			Price:         mp.Clone(),
-			OriginalPrice: mcmp,
-		}
-		m.confirmMTM(ctx, dummy, nil)
-	}
-	m.broker.Send(events.NewMarketTick(ctx, m.mkt.ID, t))
-	return m.closed
+	// update margin requirements
+	_ = m.recheckMargin(ctx, m.position.Positions())
```
ze97286, line 785: should we also releaseExcessMargin here?

core/integration/steps/time_is_updated.go
EVODelavega: The `TimeIsUpdatedTo` is a step that predates the steps like `Then the network moves ahead "N" blocks`. These steps don't use block duration, and just move time forwards to a specific point in a single jump. They're still there because in older scenario's (e.g. price monitoring) they are still used, but gradually scenario's are being reworked to do away with this all together, and instead use blocks to progress time. For that reason, I'd say this is a non-issue here

core/processor/abci.go
EVODelavega: yeah, I'm not sure about this one either. I just added it to the bottom to not interfere with the existing calls

core/execution/market.go
jeremyletang: Wouldn't recheckMargin do that?

EVODelavega: In case we have trades, excess margin is released earlier on already. This call only release ensures that parties have enough margin to cover their current open position. I can't really think of a situation where we would end up with excess margin that isn't released here.

ze97286: @jeremyletang - no
@EVODelavega - maybe if orders get cancelled in the book? idk.

EVODelavega: We had a call where we discussed this briefly. If a party cancels all orders from the book (and has no open position), their margin is released immediately. We did conclude, however, that with this periodic MTM change, there are cases where margin will not always get released instantly, and that this was acceptable.

jeremyletang (dismissed):

core/integration/features/liquidity-provision/lp-margin-post-auction.feature
```diff
@@ -55,11 +55,11 @@ Feature: Assure LP margin is correct
       | lp1 | party0 | ETH/MAR22 | 55000             | 0.001 | buy  | BID              | 500        | 17     | amendment |
 
     Then the parties should have the following margin levels:
-      | party  | market id | maintenance | search  | initial | release |
-      | party0 | ETH/MAR22 | 1380011     | 1518012 | 1656013 | 1932015 |
+      | party  | market id | maintenance | search | initial | release |
+      | party0 | ETH/MAR22 | 695821      | 765403 | 834985  | 974149  |
```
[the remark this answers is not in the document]
TomMcL: (In general I see a combination of most margin changes are small tweaks and some are halving. Assume halving ones are expected to be auction fix?)

core/integration/features/verified/closeout-cascade.feature
```diff
@@ -108,23 +108,26 @@ Feature: Closeout-cascades
     # check that trader3 is closed-out but trader2 is not
     And the parties should have the following margin levels:
       | party   | market id | maintenance | search | initial | release |
-      | trader2 | ETH/DEC19 | 5000        | 7500   | 10000   | 15000   |
+      #| trader2 | ETH/DEC19 | 3000        | 4500   | 6000    | 9000    |
+      #| trader2 | ETH/DEC19 | 5000        | 7500   | 10000   | 15000   |
       | trader3 | ETH/DEC19 | 0           | 0      | 0       | 0       |
+      | trader2 | ETH/DEC19 | 0           | 0      | 0       | 0       |
```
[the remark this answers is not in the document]
EVODelavega: Yes, this scenario wasn't a cascade before these changes, but it sort of is now: trader2 takes over a position from trader3, and as a result gets distressed. The way we're handling this now, both get closed out at the end of the same block, so it's not really possible to step-by-step are in a cascade

core/integration/features/liquidity-provision/lp-margin-post-auction.feature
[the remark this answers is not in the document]
EVODelavega: This pr fixes a number of issues with margins, including the incorrect calculations for margin requirements during auction, which is especially visible in this feature file

jeremyletang (approved):

TomMcL (approved):

PR #6970, opened 2022-11-28
fix: specify transactions as json objects not encoded in strings

### LINKED ISSUE #6884: Investigate moving from base64 transaction to plain JSON in `client.send_transaction`
To send a transaction the dApp has to encode the transaction as base64. It is unpractical for dApp. 

Why not using a plain JSON object ?

We need to investigate what's the impact inside the backend.

pscott31:
closes #6884 although there's still a bit of an open question about how to handle the equivalent admin_xxx methods, one of which expects an encoded _proto_ message instead of _JSON_


jeremyletang (changes requested):
I don’t believe this is not gonna work. You’ll need to use the json rawmessage for this: https://pkg.go.dev/encoding/json#RawMessage instead of the interface so you can un marshall the raw bytes. Then you use the jsonpb unmarshaller to unmarshal the command

pscott31:
> I don’t believe this is not gonna work. You’ll need to use the json rawmessage for this: https://pkg.go.dev/encoding/json#RawMessage instead of the interface so you can un marshall the raw bytes. Then you use the jsonpb unmarshaller to unmarshal the command

It does work, I promise (its default deserilaization is it into a map of string -> interface{}) - but absolutely RawMessage is a more idiomatic is a more way of doing it - I was not aware of its existence. Will switch

jeremyletang:
> > I don’t believe this is not gonna work. You’ll need to use the json rawmessage for this: https://pkg.go.dev/encoding/json#RawMessage instead of the interface so you can un marshall the raw bytes. Then you use the jsonpb unmarshaller to unmarshal the command
> 
> It does work, I promise (its default deserilaization is it into a map of string -> interface{}) - but absolutely RawMessage is a more idiomatic is a more way of doing it - I was not aware of its existence. Will switch

You’ll also need the jsonpb unmarshaller not the standard one or the one offs might be wrongly unmarshaled

pscott31:
Just for the record, chatted with Jeremy on discord about RawJSON and it's not possible in the current implementation because our JsonRPC implementation already parses all the parameters into a generic `map[string]interface{}` structure when it parses the request:

```go
type Request struct {
	// Version specifies the version of the JSON-RPC protocol.
	// MUST be exactly "2.0".
	Version string `json:"jsonrpc"`

	// Method contains the name of the method to be invoked.
	Method string `json:"method"`

	// Params is a by-name Structured value that holds the parameter values to be
	// used during the invocation of the method. This member MAY be omitted.
	Params Params `json:"params,omitempty"`

	// ID is an identifier established by the Client that MUST contain a String.
	// If it is not included it is assumed to be a notification.
	// The Server MUST reply with the same value in the Response object if included.
	// This member is used to correlate the context between the two objects.
	ID string `json:"id,omitempty"`
}
```

It might be nicer if `Params` here were `json.RawMessage` and left parsing up to the individual handlers; we wouldn't have to use `mapstructure` etc.. as well, but that is a commit for another day.

jeremyletang (approved):

ValentinTrinque (approved):

Issue #6955, opened 2022-11-28
Refactor liquidity order lp orders sizing


jeremyletang:
Implement the changes introduced by the following specs changes: https://github.com/vegaprotocol/specs/pull/1438

- [ ] Remove scaling by probability when implying LP volumes.
- [ ] Only change the price if it’s outside the new “valid LP price range” - move it to the bound in that case.

PR #7014, opened 2022-12-02
Add LP price range prameter to the APIs

witgaw:
API changes to do with #6955 

LP price range specifies a percentage move up and down from the mid price which specifies the range of price levels over which automated liquidity provision orders will be deployed.

It needs to be a strictly positive number.


datanode/entities/market.go
```diff
@@ -103,6 +104,11 @@ func NewMarketFromProto(market *vega.Market, txHash TxHash, vegaTime time.Time)
 		return nil, fmt.Errorf("%d is not a valid number for position decimal places", market.PositionDecimalPlaces)
 	}
 
+	if market.LpPriceRange <= 0 {
+		return nil, fmt.Errorf("%v is not a valid number for LP price rannge", market.LpPriceRange)
```
peterbarrow, line 108:
```suggestion
		return nil, fmt.Errorf("%v is not a valid number for LP price range", market.LpPriceRange)
```

datanode/entities/market.go
```diff
@@ -43,6 +43,7 @@ type Market struct {
 	State                         MarketState
 	MarketTimestamps              MarketTimestamps
 	PositionDecimalPlaces         int
+	LpPriceRange                  float64
```
pscott31, line 46: Also I don't see any changes to the database schema, you'll need an additional column to actually store this in (see sqlstore/migrations)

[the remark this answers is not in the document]
pscott31: Edit - I read the description in protos and I see now it's a percentage*, but most of the argument still holds; given floats are an approximation only I believe it's asking for trouble to ever rely on the hash of one being the same in one place as it is in another.


* I wonder if there's a better name that would make that more obvious; I can't think of one off the top of my head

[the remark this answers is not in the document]
witgaw: Couldn't think of a better name that would be accurate and not supper long. 

I'm aware of floating-pint gotchas, but it is an externally supplied parameter (doesn't come from any calculations) and as long as things like margin scaling factors are floats (they are) so should this.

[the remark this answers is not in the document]
witgaw: `LpPriceRangePercentage` ?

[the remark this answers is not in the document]
davidsiska-vega: If it would help it can definitely be a decimal.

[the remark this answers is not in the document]
pscott31: That's better name I think. Do you mean these guys? They are stored as decimal which is different, fixed precision type.

In fact there's only one thing in the database that is a float (reward & of total) and this has gotten me thinking that that is a mistake :)

```type MarginLevels struct {
	AccountID              AccountID
	MaintenanceMargin      decimal.Decimal
	SearchLevel            decimal.Decimal
	InitialMargin          decimal.Decimal
	CollateralReleaseLevel decimal.Decimal
	Timestamp              time.Time
	TxHash                 TxHash
	VegaTime               time.Time
}
```

It is mandatory (in order for dehistory to work) on datanode that two different nodes seeing the same event stream create rows with the exact same binary representation; I'm not confident that is guaranteed to be the case if you start passing floating point numbers into it.

pscott31: yes exactly :) though as mentioned above I think we'd be better off with a NUMERIC type with has arbitrary precision e.g. 

 NUMERIC(32, 16)

Which is a bit overkill gives you 16 digits either side of the decimal place :)

protos/sources/vega/markets.proto
```diff
@@ -275,6 +275,9 @@ message Market {
   MarketTimestamps market_timestamps = 10;
   // The number of decimal places for a position
   int64 position_decimal_places = 11;
+  // Percentage move up and down from the mid price which specifies the range of 
+  // price levels over which automated liquidity provision orders will be deployed
+  double lp_price_range = 12;
```
ze97286, line 280: this should be string, we shouldn't use doubles/floats.

witgaw: Ok, thanks. So any legacy doubles (say `LiquidityMonitoringParameters.triggering_ratio`) will get changed to strings in the future?

ze97286: I think those are less important but inside the market nothing should be done with floats.

witgaw: Yes, of course, that's why I've set the `LPPriceRange` field in [core/types/market.go](https://github.com/vegaprotocol/vega/blob/630acafa1b7738c19d3eda58c3b7104cb68648fe/core/types/market.go#L633) to be a `Decimal` from the very beginning, but you're commenting on `markets.proto` which has plenty of doubles. 

Anyway, I've changed it to a string, just trying to understand what the convention should be since I just followed the same approach as was already used for other similar parameters  (e.g. `triggering_ratio`, `scaling_factor` both of which are used inside the market).

PR #7018, opened 2022-12-02
Fix issue with market update during opening auction

### LINKED ISSUE #7017: Updating market during opening auction prevents it from ever leaving it
When we update the market we do this:
https://github.com/vegaprotocol/vega/blob/5833d4bdf3ecd9bd36194e0117ee53265286f72e/core/liquidity/supplied/engine.go#L102

If probability of trading state variable calculation has already [been triggered](https://github.com/vegaprotocol/vega/blob/5833d4bdf3ecd9bd36194e0117ee53265286f72e/core/execution/auction.go#L43) by that time the `initialised` field of the price monitoring engine will be set to `true`.

However,  do not reset that field when we update the market so:
https://github.com/vegaprotocol/vega/blob/5833d4bdf3ecd9bd36194e0117ee53265286f72e/core/execution/auction.go#L35
never evaluates to `true` again and so we never trigger probability of trading state variable recalculation, so we never leave the opening auction.

witgaw:
See issue for details

Closes #7017


pscott31 (dismissed):
This looks OK to me but I don't know much about how this all works :)

pscott31 (approved):

PR #7022, opened 2022-12-03
Final settlement fix

### LINKED ISSUE #7021: Inconsistent behaviour when market is terminated/settled
Before a market is settled (final settlement), any pending trades should trigger a final MTM settlement before all positions are settled to the final price. The final MTM settlement should take place as soon as trading is terminated, but currently doesn't take place until we have a final settlement price. By that time the market is already set to state `TRADING_TERMINATED` which impacts the existing MTM flow.

EVODelavega:
Closes #7021


core/execution/market.go
```diff
@@ -3227,31 +3227,50 @@ func (m *Market) tradingTerminated(ctx context.Context, tt bool) {
...
 
 	if m.mkt.State != types.MarketStateProposed && m.mkt.State != types.MarketStatePending {
-		m.mkt.State = types.MarketStateTradingTerminated
-		m.mkt.TradingMode = types.MarketTradingModeNoTrading
-		m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
-
-		if m.settlementDataInMarket == nil {
+		if m.settlementDataInMarket != nil {
+			// because we need to be able to perform the MTM settlement, only update market state now
+			m.mkt.State = types.MarketStateTradingTerminated
```
ze97286, line 3232: this could be outside the if as it happens in both cases.

EVODelavega: The state can't be set until after the MTM stuff, but has to be set prior to the `m.settlementDataWithLock(ctx)` call, I duplicated them to get things working and ready for QA to test. Want to check, too, if by having the MTM settlement here, we can remove it from the `m.settlementDataWithLock` entirely

ze97286: does the MTM rely on the state/trading_state of the market? why does the state need to be set *after* MTM?

EVODelavega: We should be able to close out distressed traders, and check for reference moves (including submitting new orders), which fails if trading is terminated. Before the periodic MTM changes, we did this after each trade, so to preserve or at least approximate the previous behaviour, the final MTM settlement should go through as though the market is still in continuous trading. The only reason we perform this final MTM settlement is to ensure all trades since last MTM settlement have been processed before we settle all positions for the market. A cleaner solution definitely would be to handle the trades as part of the final settlement, or allow for a MTM settlement without submitting/updating LP orders which will just be cancelled afteterwards (as trading is terminated), but in order to address the immediate issue the system tests are having, this seemed like a quick fix.

ze97286 (dismissed):

ze97286 (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

PR #7025, opened 2022-12-04
chore: add csv export to ledger entries

### LINKED ISSUE #6754: Implement CSV export for ledger movement
We need to implement a CSV export for ledger movement in the datanode.

Let's keep it simple at first, the datanode should provide a grpc + rest endpoint to retrieve a CSV file containing ALL ledger entry for a party.

The endpoint should take as parameters:
- 1 pubkey (required) - we will filter ledger entry with this pubkey as a from or to account
- 1 asset (required) - we will filter ledger entry for this specific asset
- start and end date (optional) - if not specified all ledger entries since all time matching the 2 previous required filters.

Let's not over optimised for now and do it on a blocking connection with large timeout? Later if we realise these takes too long to generate on the go, we can eventually do it asynchronously, but that might be unnecessary for now (may be wrong)

AnExsomnis:
close: #6754

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -36,7 +36,9 @@ import (
 	"code.vegaprotocol.io/vega/version"
 
 	"github.com/pkg/errors"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -240,6 +242,38 @@ func (t *tradingDataServiceV2) ListLedgerEntries(ctx context.Context, req *v2.Li
 			Edges:    edges,
 			PageInfo: pageInfo.ToProto(),
 		},
+	}, nil
+}
+
+func (t *tradingDataServiceV2) ExportLedgerEntries(ctx context.Context, req *v2.ExportLedgerEntriesRequest) (*v2.ExportLedgerEntriesResponse, error) {
+	defer metrics.StartAPIRequestAndTimeGRPC("ExportLedgerEntriesV2")()
+	if t.accountService == nil {
+		return nil, apiError(codes.Internal, ErrAccountServiceSQLStoreNotAvailable, nil)
+	}
+
+	dateRange := entities.DateRangeFromProto(req.DateRange)
+	pagination, err := entities.CursorPaginationFromProto(req.Pagination)
+	if err != nil {
+		return nil, apiError(codes.InvalidArgument, fmt.Errorf("invalid cursor: %w", err))
+	}
+
+	raw, pageInfo, err := t.ledgerService.Export(ctx, req.PartyId, req.AssetId, dateRange, pagination)
+	if err != nil {
+		apiError(codes.Aborted, err)
+	}
+
+	header := metadata.New(map[string]string{
+		"Content-Type":       "text/csv",
+		"Content-diposition": fmt.Sprintf("attachment;filename=%s", "ledger_entries_export.csv"),
+	})
+
+	if err := grpc.SendHeader(ctx, header); err != nil {
+		return nil, apiError(codes.Internal, fmt.Errorf("unable to send 'x-response-id' header"))
+	}
+
+	return &v2.ExportLedgerEntriesResponse{
+		Data:     raw,
+		PageInfo: pageInfo.ToProto(),
 	}, nil
 }
 
--- a/datanode/sqlstore/ledger.go
+++ b/datanode/sqlstore/ledger.go
@@ -14,6 +14,7 @@ package sqlstore
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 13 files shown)


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -751,10 +760,22 @@ message ListLedgerEntriesRequest {
     optional DateRange date_range = 3;
 }
 
+message ExportLedgerEntriesRequest {
+    optional string path = 1;
+    optional string filename = 2;
```
wwestgarth, line 765: Similar to this https://github.com/vegaprotocol/vega/issues/7029 I don't think we can do it this way. Can we not return the file as a base64 string or is it too big?

AnExsomnis: Yes, I can do it with a base64 filename returned in the response;  without path and filename settings in the request.
Thanks!

wwestgarth: But if I don't have access to the machine that datanode is running on, how will I get to the file?

AnExsomnis: All right, so in that case by "file as a base64 string" I assume you mean the file content as a base64, is this correct?
In that case the content will be returned in the response and written to a file outside the core part, yes?

jeremyletang: we should return the file yes, no need to base64 encode it. let's just return the raw byte, and when using the REST endpoint for it. let's add the content-type:text/csv header so browser can response properly.
No need to expose on graphql I suppose?

AnExsomnis: Okay.
Yes, I havent exposed it in graphql

datanode/sqlstore/ledger.go
```diff
@@ -160,6 +164,85 @@ func (ls *Ledger) Query(
...
+	if err != nil {
+		return nil, "", err
+	}
+
+	path := fmt.Sprintf("%s-%s-%s-%s", partyID, assetID, dateRange.Start.String(), dateRange.End.String())
+	file, err := os.Create(path)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to create filename: %s", err.Error())
+	}
+	defer file.Close()
```
jeremyletang, line 197: I'm not sure you would need to write it all to a file.

datanode/sqlstore/ledger.go
```diff
@@ -160,6 +164,85 @@ func (ls *Ledger) Query(
...
+	defer w.Flush()
+
+	// Using Write
+	for _, record := range *aggregatedEntries {
+		fi, err := file.Stat()
+		if err != nil {
+			return nil, "", err
+		}
+
+		if fi.Size() >= 3670016 {
```
jeremyletang, line 215: I didn't meant about a hardcode size like this. we don't know with what max size a client is configured with. But I would have said that we could have some kind of pagination for it as we do for other endpoint. let's say that you are allowed to at most 1 month at a time of transfers, and the front end could request csv by paginating over that.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -748,10 +757,21 @@ message ListLedgerEntriesRequest {
     optional DateRange date_range = 3;
 }
 
+message ExportLedgerEntriesRequest {
+    string party_id = 3;
+    string asset_id = 4;
```
wwestgarth, line 762: Doesn't this need a `optional Pagination pagination = 6;` field to be able to use the cursor in the the response?

datanode/sqlstore/ledger.go
```diff
@@ -160,6 +163,77 @@ func (ls *Ledger) Query(
...
+				*record.ToProto().ReceiverPartyId,
+				*record.ToProto().ReceiverMarketId,
+				record.ReceiverAccountType.String(),
+				record.Quantity.String(),
+				vega.TransferType(*record.TransferType).String(),
+			}
+
+			err = w.Write(row)
+			if err != nil {
+				return nil, entities.PageInfo{}, fmt.Errorf("failed to write records to filename: %w", err)
```
pscott31, line 223: I think the error message needs updating

datanode/api/trading_data_v2.go
```diff
@@ -243,6 +245,51 @@ func (t *tradingDataServiceV2) ListLedgerEntries(ctx context.Context, req *v2.Li
...
+
+	raw, pageInfo, err := t.ledgerService.Export(ctx, req.PartyId, req.AssetId, dateRange, pagination)
+	if err != nil {
+		apiError(codes.Aborted, err)
+	}
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, apiError(codes.DataLoss, fmt.Errorf("failed to get context metadata"))
+	}
```
pscott31, line 268: I'm not sure what this part is for, xrid doesn't seem to be used?

datanode/sqlstore/ledger.go
AnExsomnis: Addresssed - it sends the data as binary payload now
Thanks!

datanode/sqlstore/ledger.go
AnExsomnis: Phil helped me figure out how to use the pagination - its done like this now.
Thanks!

datanode/api/trading_data_v2.go
AnExsomnis: This has been removed, it uses the gRPCGateway header now
Thanks!

datanode/sqlstore/ledger.go
AnExsomnis: This has been updated - Thanks!

datanode/api/trading_data_v2.go
```diff
@@ -243,6 +245,51 @@ func (t *tradingDataServiceV2) ListLedgerEntries(ctx context.Context, req *v2.Li
...
+	xrid := md["x-request-id"]
+	if len(xrid) == 0 {
+		return nil, apiError(codes.InvalidArgument, fmt.Errorf("missing 'x-request-id' header"))
+	}
+
+	if strings.Trim(xrid[0], " ") == "" {
+		return nil, apiError(codes.InvalidArgument, fmt.Errorf("empty 'x-request-id' header"))
+	}
+
+	header := metadata.New(map[string]string{
```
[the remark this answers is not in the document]
AnExsomnis: No, the response header is rewritten to ```application/grpc```
the binary data is prepared to be interpreted as csv

datanode/sqlstore/ledger.go
AnExsomnis: At the moment the solution has data encoded to be processed and it can be decoded on the other side as csv.
It can be taken from the response field and dealt with on the FE side.

Do we like this solution or we _specifically_ want to download a file?
Because in that case grpc-gateway will use a custom route for this endpoint? Not that it is a problem, just to be sure - because in the beginning the idea was not to have a file saved in the datanode system.

AnExsomnis: image (1)

candida-d:
@AnExsomnis While you're working in this area, would you mind updating the API description for `filter.closeOnAccountFilters`?

Updated: 
`close_on_account_filters` is used to filter for which AccountFrom (receiver) and AccountTo (sender) values are returned. When it’s set to `false`, the filter shows all ledger entries that match the given criteria for at least one of receiver or sender. When it’s set to `true`, the filter will only display the results if they meet the criteria for both sides (sender and receiver).

AnExsomnis:
> @AnExsomnis While you're working in this area, would you mind updating the API description for `filter.closeOnAccountFilters`?
> 
> Updated: `close_on_account_filters` is used to filter for which AccountFrom (receiver) and AccountTo (sender) values are returned. When it’s set to `false`, the filter shows all ledger entries that match the given criteria for at least one of receiver or sender. When it’s set to `true`, the filter will only display the results if they meet the criteria for both sides (sender and receiver).

As per discussion in slack between me and Candida  - the requested change already exists.

wwestgarth (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

datanode/sqlstore/ledger.go
barnabee: Sorry, I missed this thread previously but I think a direct download makes the most sense:

- it will often be a lot of data, and often the user will want to open it in a 3rd party tool, so it makes sense to me to avoid the overhead of the dapp loading it into memory and decoding it, only to serve the result up as a file for the user to save anyway;
- many users will skip the dapp altogether and just donwload the data with curl or similar.

PR #7030, opened 2022-12-05
feat: add a type to node votes

jeremyletang:
Add types to node votes transaction, this will help when debugging validator votes on events.


zale144 (dismissed):
lgtm

wwestgarth (approved):

PR #7038, opened 2022-12-05
Setup codeQL

jeremyletang:


ze97286 (approved):

gordsport:
@jeremyletang do we want this as a blocking-merge check for CI?

jeremyletang:
> @jeremyletang do we want this as a blocking-merge check for CI?

Yes, but I think it's the case, it was red before I got the lints fixed.

PR #7042, opened 2022-12-05
feat: use voting power for counting witness votes

### LINKED ISSUE #7041: change witness vote count to be based on voting power
currently witness vote count is based on the number of validators voting out of the total number of tendermint validators. This should be changed to be using the voting power out of the total voting power.

ze97286:
Closes #7041

### DIFF
--- a/core/validators/witness.go
+++ b/core/validators/witness.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"math/rand"
 	"sort"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


core/validators/witness.go
```diff
@@ -345,8 +338,6 @@ func (w *Witness) OnTick(ctx context.Context, t time.Time) {
 				}
 				w.log.Warn("resource checking was not validated by all nodes",
 					logging.String("resource-id", v.res.GetID()),
-					logging.Int("vote-count", votesLen),
```
jeremyletang, line 348: Should we actually keep some kind of logs to show how much of the total we have?

jeremyletang (dismissed):

core/validators/witness.go
ze97286: added

jeremyletang (approved):

PR #7045, opened 2022-12-05
feat: various proof of work enhancements

### LINKED ISSUE #7044: POW refactoring
1. Need to ensure that a party can submit a maximum of `SpamPoWNumberOfTxPerBlock` for a given block - regardless of how blocks are organised - this means that a party can fill a block with their transactions as long as the number of transactions generated by the party for any given height respects the difficulty and `SpamPoWNumberOfTxPerBlock` requirements. Conversely, if a party sends more than  `SpamPoWNumberOfTxPerBlock` split over many blocks, they still get banned when the number of transactions they generated with a given block height is greater than `SpamPoWNumberOfTxPerBlock` regardless how these transactions ended up in blocks. 
2. ban period is reduced to 30 minutes. 
3. If `increaseDifficulty` is enabled then for every increased difficulty a party is allowed to send `SpamPoWNumberOfTxPerBlock` transactions

ze97286:
Closes #7044


wwestgarth:
Might be worth running the system tests marked as `pow`, give QA a heads up on potentional failures. I can see one test that checks that the ban period is 4 epochs.

wwestgarth (dismissed):

ze97286:
> Might be worth running the system tests marked as `pow`, give QA a heads up on potentional failures. I can see one test that checks that the ban period is 4 epochs.

Good point. Also, as this is 30 minutes, rather than a function of anything, this will be hard to test. Maybe rather than 30 minutes, we should make it a fraction of an epoch, with some minimum, like we somewhere for the scores.

core/pow/engine.go
```diff
@@ -251,20 +243,27 @@ func (e *Engine) DeliverTx(tx abci.Tx) error {
 	e.heightToTid[tx.BlockHeight()] = append(e.heightToTid[tx.BlockHeight()], tx.GetPoWTID())
 	e.seenTid[tx.GetPoWTID()] = struct{}{}
 
-	// if we've not seen any transactions from this party this block then let it pass and save the observed difficulty
-	if _, ok := state.blockPartyToSeenCount[tx.Party()]; !ok {
-		state.blockPartyToObservedDifficulty[tx.Party()] = uint(d)
-		state.blockPartyToSeenCount[tx.Party()] = 1
+	// if it's the first transaction we're seeing from any party for this block height, initialise the state
+	if _, ok := state.blockToPartyState[txBlock]; !ok {
```
TomMcL, line 247: Is `txBlock` here the height of the block being created? As we should be using the height the PoW was generated using, but looks like this might be the new height?

ze97286: its the height of the block pow was generated with not the height of the block it is included in. txBlock = tx.BlockHeight

TomMcL: Ok cool, wasn't sure whether `tx` here was the one pow used or the new one being created 👍

core/pow/engine.go
```diff
@@ -17,20 +17,22 @@ import (
...
 )
 
 const (
-	banPeriod = 4
-	ringSize  = 500
+	ringSize       = 500
+	minBanDuration = time.Second * 30 // minimum ban duration
 )
 
+var banDurationAsEpochFaction = num.DecimalOne().Div(num.DecimalFromInt64(48)) // 1/48 of an epoch will be the default 30 minutes ban
```
TomMcL, line 34:
```suggestion
var banDurationAsEpochFraction = num.DecimalOne().Div(num.DecimalFromInt64(48)) // 1/48 of an epoch will be the default 30 minutes ban
```

core/pow/engine.go
```diff
@@ -95,35 +109,22 @@ func New(log *logging.Logger, config Config, epochEngine EpochEngine) *Engine {
...
-
-	// check if there are banned parties who can be released
-	for k, v := range e.bannedParties {
-		if epoch.Seq > v {
-			delete(e.bannedParties, k)
-			e.log.Info("released proof of work spam ban from", logging.String("party", k))
-		}
+// OnEpochDurationChanged updates the ban duration as a fraction of the epoch duration.
+func (e *Engine) OnEpochDurationChanged(_ context.Context, duration time.Duration) error {
+	epochImpliedDurationNano, _ := num.UintFromDecimal(num.DecimalFromInt64(duration.Nanoseconds()).Mul(banDurationAsEpochFaction))
```
TomMcL, line 120:
```suggestion
	epochImpliedDurationNano, _ := num.UintFromDecimal(num.DecimalFromInt64(duration.Nanoseconds()).Mul(banDurationAsEpochFraction))
```

ze97286: corrected

TomMcL (approved):

pscott31 (approved):

core/pow/engine.go
```diff
@@ -53,18 +55,29 @@ func (p *params) isActive(blockHeight uint64) bool {
...
+// of transactions generated with a given block height.
+type partyStateForBlock struct {
+	observedDifficulty uint
+	seenCount          uint
+}
+
 type state struct {
-	blockPartyToObservedDifficulty map[string]uint // party observed total difficulty in block
-	blockPartyToSeenCount          map[string]uint // party observed transactions in block
+	blockToPartyState map[uint64]map[string]*partyStateForBlock
```
ValentinTrinque, line 66: Since maps in golang don't shrink automatically in memory, I think we should find a way to clean up the maps. Like recreating a new one when appropriate, for example ?

I don't know what would be best in terms of technic.

ze97286: we clean up the state when possible, i.e. when a block becomes out of scope we remove it from blockToPartyState. Otherwise it's add only because all information is needed while the block is still in scope.

PR #7062, opened 2022-12-06
feat: catchup from dehistory

### LINKED ISSUE #7063: Dehistory - support node catchup
Initial implementation of dehistory supported recovering a node by emptying and reloading all history from dehistory,   attempted recovery of node on testnet was not feasible with just this mechanism supported.   Supporting 'catchup' behaviour that is required for this.   Catchup behaviour fetches just the history needed to append to the nodes current block span in order for it to catchup with the networks current height.

ettec:
closes #7063 

Initial implementation of dehistory supported recovering a node by emptying and reloading all history from dehistory,   attempted recovery of node on testnet was not feasible with just this mechanism supported.   Supporting 'catchup' behaviour is required for this.   Catchup behaviour fetches just the history needed to append to the nodes current block span in order for it to catchup with the networks current height.

If a node is configured with  `AutoInitialiseFromDeHistory = true` and is empty at startup it will fetch at least `MinimumBlockCount` blocks of history from the network up to the most recent history segment.   

If a node is configured with  `AutoInitialiseFromDeHistory = true`  and already has data at startup it will attempt to fetch all history between the nodes last block height and the networks current height (latest history segment) and load this into the node.  Thus, a node will self recover if it has been down for a period of time on restart.


datanode/dehistory/service_test.go
```diff
@@ -376,122 +375,241 @@ func TestMain(t *testing.M) {
...
+	broker, err := setupSQLBroker(ctx, sqlConfig, snapshotService,
+		func(ctx context.Context, service *snapshot.Service, chainId string, lastCommittedBlockHeight int64) {
+			if lastCommittedBlockHeight == 1800 {
+				cancelFn()
+			}
+		},
+		evtSource, puh)
 	if err != nil {
-		panic(err)
+		panic(fmt.Errorf("failed to setup broker:%w", err))
```
wwestgarth, line 412: Why are we panicking in tests, and not using `require.NoError(t, err)`?

datanode/dehistory/snapshot/service_load_snapshot.go
```diff
@@ -23,32 +26,58 @@ type Conn interface {
...
 
 	vegaDbConn, err := pgxpool.Connect(context.Background(), b.connConfig.GetConnectionString())
 	if err != nil {
-		return 0, fmt.Errorf("unable to connect to vega database:%w", err)
+		return LoadResult{}, fmt.Errorf("unable to connect to vega database:%w", err)
+	}
+
+	datanodeBlockSpan, err := sqlstore.GetDatanodeBlockSpan(ctx, b.connConfig)
+	if err != nil {
+		return LoadResult{}, fmt.Errorf("failed to check if datanode has data:%w", err)
```
wwestgarth, line 49: I'm sure I'm mentioned this before, can there be a space between the `:` and the `%w`.

datanode/dehistory/snapshot/service_load_snapshot.go
```diff
@@ -23,32 +26,58 @@ type Conn interface {
...
+	}
+
+	historyFromHeight := contiguousHistory[0].HeightFrom
+	historyToHeight := contiguousHistory[len(contiguousHistory)-1].HeightTo
+
+	if err = validateSpanOfHistoryToLoad(datanodeBlockSpan, historyFromHeight, historyToHeight); err != nil {
+		return LoadResult{}, fmt.Errorf("validation of history span to load failed: %w", err)
```
wwestgarth, line 56: I don't think we need to wrap the error here, we are wrapping far too much. The context given by the errors returned from `validateSpanOfHistory()` is more than enough.

datanode/dehistory/service_test.go
ettec: sorted

datanode/dehistory/snapshot/service_load_snapshot.go
ettec: ah yeah, finger muscle memory, sorted

datanode/dehistory/snapshot/service_load_snapshot.go
ettec: sorted

wwestgarth (approved):

AnExsomnis (approved):

PR #7078, opened 2022-12-07
Alias `vegawallet info` to `vegawallet describe`, before definitive renaming

### LINKED ISSUE #7068: Reword `vegawallet info`
Users have trouble with this command and don't get the intent of the command, nor its output.

We should rename it to `describe` just like the keys, network, etc.

The `wallet version` also need to be renamed bause users constantly get confused between the `wallet version` and the `software version`. A more appropriate name would be `key derivation version`.

This will be first deprecated, and then removed.

- [ ] Update the examples and description in the command lines

This changes should be reflected in the JSON-RPC API as well.

ValentinTrinque:
Close #7068


jeremyletang (dismissed):

ValentinTrinque:
```
{
   "system-tests": "extra_wallet_logging"
}
```

jeremyletang (approved):

ValentinTrinque:
```
{
   "system-tests": "develop"
}
```

jeremyletang (approved):

PR #7084, opened 2022-12-08
Check peer swarm key matches when getting latest history segment

### LINKED ISSUE #7083: Check swarm id when getting latest segment
Check swarm id when getting latest segment - suspect sys tests may be talking cross network

ettec:
closes #7083 - Datanode now checks the swarm key of sources nodes matches when filtering the latest history segment responses.


datanode/dehistory/initialise.go
```diff
@@ -55,23 +55,25 @@ func DatanodeFromDeHistory(parentCtx context.Context, cfg InitializationConfig,
...
+		if response == nil {
+			log.Infof("unable to get a most recent segment response from peers")
 			return ErrDeHistoryNotAvailable
 		}
 
-		log.Infof("got most recent history segment:%s", mostRecentHistorySegmentFromPeers)
+		mostRecentHistorySegment := response.Response.Segment
 
-		toSegmentID = mostRecentHistorySegmentFromPeers.HistorySegmentId
+		log.Infof("got most recent history segment:%s from peer:%s", mostRecentHistorySegment, response.PeerAddr)
```
jeremyletang, line 65: Can we try to just us log fields instead of fmt directives?

datanode/dehistory/initialise.go
```diff
@@ -169,41 +173,46 @@ func FetchHistoryBlocks(ctx context.Context, logInfo func(s string, args ...inte
...
-	for _, peerAddress := range activePeerAddresses {
+	for _, peerAddress := range peerAddresses {
 		for _, grpcAPIPort := range grpcAPIPorts {
-			segment, err := GetMostRecentHistorySegmentFromPeer(ctxWithTimeOut, peerAddress, grpcAPIPort)
+			resp, err := GetMostRecentHistorySegmentFromPeer(ctxWithTimeOut, peerAddress, grpcAPIPort)
 			if err == nil {
-				if segment != nil {
-					peerToSegment[peerAddress] = segment
-				}
+				peerAddress = fmt.Sprintf("%s:%d", peerAddress, grpcAPIPort)
```
jeremyletang, line 199: net.JoinHostPost?

datanode/dehistory/initialise.go
```diff
@@ -169,41 +173,46 @@ func FetchHistoryBlocks(ctx context.Context, logInfo func(s string, args ...inte
...
 			} else {
 				errorMsgs = append(errorMsgs, err.Error())
 			}
 		}
 	}
 
-	if len(peerToSegment) == 0 && len(errorMsgs) != 0 {
-		return nil, nil, fmt.Errorf(strings.Join(errorMsgs, "\n"))
+	if len(peerToResponse) == 0 {
+		return &PeerResponse{}, nil, fmt.Errorf(strings.Join(errorMsgs, "\n"))
```
jeremyletang, line 208: why not return nil?

also and error with \n looks weird, maybe want to join with `", "` instead

datanode/dehistory/initialise.go
```diff
@@ -169,41 +173,46 @@ func FetchHistoryBlocks(ctx context.Context, logInfo func(s string, args ...inte
...
 	}
 
-	rootSegment := SelectRootSegment(peerToSegment)
-	return rootSegment, peerToSegment, nil
+	rootSegment := SelectMostRecentHistorySegmentResponse(peerToResponse, swarmKey)
+	return rootSegment, peerToResponse, nil
```
jeremyletang, line 212:
```suggestion
	return rootSegment, SelectMostRecentHistorySegmentResponse(peerToResponse, swarmKey), nil
```

datanode/dehistory/initialise.go
```diff
@@ -215,33 +224,43 @@ func GetMostRecentHistorySegmentFromPeer(ctx context.Context, ip string, datanod
...
-		if segment.ToHeight == segmentsList[0].ToHeight {
+		if segment.Response.Segment.ToHeight == segmentsList[0].Response.Segment.ToHeight {
 			filteredSegments = append(filteredSegments, segment)
 		}
 	}
 
 	// Select one segment from the list at random
 	if len(filteredSegments) > 0 {
-		return filteredSegments[rand.Intn(len(filteredSegments))]
+		segment := filteredSegments[rand.Intn(len(filteredSegments))]
```
jeremyletang, line 262: Why use a random one? is it possible to have different height here? No way to have a more consistent behaviour?

datanode/dehistory/initialise.go
```diff
@@ -215,33 +224,43 @@ func GetMostRecentHistorySegmentFromPeer(ctx context.Context, ip string, datanod
...
+		}
+	}
+
 	// Filter out segments with toHeight < highest toHeight
-	var filteredSegments []*v2.HistorySegment
+	segmentsList = filteredSegments
+	filteredSegments = nil
 	for _, segment := range segmentsList {
-		if segment.ToHeight == segmentsList[0].ToHeight {
+		if segment.Response.Segment.ToHeight == segmentsList[0].Response.Segment.ToHeight {
```
jeremyletang, line 255: Comment seems invalid here, aren't we instead filtering out segment where ToHeight is < highest && ToHeight > highest?
Is that the expected behaviour?

datanode/dehistory/store/config.go
```diff
@@ -55,6 +57,17 @@ func NewDefaultConfig() Config {
 	}
 }
 
+func (c Config) GetSwarmKey(log *logging.Logger, chainID string) string {
+	swarmKey := chainID
+	if len(c.SwarmKeyOverride) > 0 {
+		swarmKey = c.SwarmKeyOverride
+		log.Infof("Using swarm key override %s as the swarm key", c.SwarmKeyOverride)
```
jeremyletang, line 64: structured logging preffered here as well.

datanode/dehistory/store/config.go
```diff
@@ -55,6 +57,17 @@ func NewDefaultConfig() Config {
...
+	} else {
+		log.Infof("Using chain id %s as the swarm key", chainID)
```
jeremyletang, line 66: same

datanode/dehistory/store/store.go
```diff
@@ -686,7 +695,10 @@ func createIpfsNodeConfiguration(log *logging.Logger, identity config.Identity,
 
 	updatedSwarmAddrs := make([]string, 0, 10)
 	for _, addr := range cfg.Addresses.Swarm {
-		updatedSwarmAddrs = append(updatedSwarmAddrs, strings.ReplaceAll(addr, "4001", strconv.Itoa(swarmPort)))
+		// Exclude ip6 addresses cause hang on lookup
+		if !strings.Contains(addr, "/ip6/") {
+			updatedSwarmAddrs = append(updatedSwarmAddrs, strings.ReplaceAll(addr, "4001", strconv.Itoa(swarmPort)))
```
jeremyletang, line 700: Why are we replacing all 4001 here? would be good to have a comment to explain here, it just seems like out of nowhere.

datanode/dehistory/initialise.go
ettec: Indeed, it probably will be random to an extent anyway depending on how and when each peer is contacted, but as per the comment on this method:

`// TODO this needs some thought as to the best strategy to select the root segment to avoid spoofing.`

might be that we could compare and pick the one where there is a majority agreement for example, and not necessarily the highest height (as currently) for example.  But that not's perfect either, so just opted to ensure it's totally random pick from the highest height set for now.  maybe we could have  some sort of consensus process whereby the validators decided amonst them what the are valid segments that way we could have a check on this, but maybe that is too much?

datanode/dehistory/initialise.go
ettec: sorted

datanode/dehistory/initialise.go
ettec: sorted

datanode/dehistory/initialise.go
ettec: sorted

datanode/dehistory/initialise.go
ettec: sorted

datanode/dehistory/initialise.go
ettec: yeah, the method was unclear, refactored it

datanode/dehistory/store/config.go
ettec: sorted - (I will try harder to remember !)

datanode/dehistory/store/config.go
ettec: sorted

datanode/dehistory/store/store.go
ettec: ah yeah, instroduced a named const to make it clearer

jeremyletang (approved):

PR #7090, opened 2022-12-08
fix: remove unsafe rpc endpoint in datanode

### LINKED ISSUE #7029: Dehistory: having an RPC endpoint that dumps data to a target path on the host is not a good idea - lets remove it
# UPDATE:
The `FetchDeHistorySegment` endpoint is still at risk being exposed, as it is possible to make any data node download arbitrary content onto their machine.

Add similar functionality as core so `FetchDeHistorySegment` is behind an admin endpoint

# Problem encountered
We seem to have this RPC endpoint to the datanode:

```
// Copy history segment data to a target file
    rpc CopyHistorySegmentToFile(CopyHistorySegmentToFileRequest) returns (CopyHistorySegmentToFileResponse);
```

which tells the datanode to copy a segment to an arbitrary location on the host machine. This is not secure for a publically exposed endpoint.

Fetching is not as bad, but still similar in that it is a activity that should be performed _only by the node operator_ but it is exposed in a public API.
```
// Fetch a history segment from another peer in the network
    rpc FetchDeHistorySegment(FetchDeHistorySegmentRequest) returns (FetchDeHistorySegmentResponse);
```

# Observed behaviour
Why does this endpoint exist?

# Expected behaviour
An endpoint like this should not exist.

# System response

# Steps to reproduce

## Manual

## Automation

# Evidence

## Logs

## Additional context

# Definition of Done

**Before Merging**

**After Merging**

jeremyletang:
close #7029


cmd/data-node/commands/dehistory/fetch.go
```diff
@@ -113,3 +114,12 @@ func verifyChainID(log *logging.Logger, connConfig sqlstore.ConnectionConfig, ch
 	}
 	return nil
 }
+
+func errorFromGrpcError(msg string, err error) error {
```
ValentinTrinque, line 118: Is it used ?

jeremyletang: yep.

ValentinTrinque (dismissed):

PR #7091, opened 2022-12-08
fix: remove unused expiry field in withdrawal

### LINKED ISSUE #7075: [api] All APIs still have expiry fields on withdrawals, which are no longer used
## Description

https://github.com/vegaprotocol/vega/pull/4403 removes the functionality of expiry, but expiry/expirationdate still exist on the protos & graphql types for withdrawals, which is confusing.

# Tasks 
- [ ] Remove redundant fields from protos
- [ ] Remove redundant fields from graphql

jeremyletang:
close #7075


edd (dismissed):
🥳

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -785,7 +785,6 @@ create table if not exists withdrawals (
     asset bytea not null,
     status withdrawal_status not null,
     ref text not null,
-    expiry timestamp with time zone not null,
```
ettec, line 788: I suppose really we should be adding this in another file as its a migration....though given were we are at I'm ok with letting this slide for now - unless we want to start testing the process?

jeremyletang: I think it’s fine for now, as we are anoit to tag the version which will go to mainnet next after this merge

ValentinTrinque (approved):

PR #7120, opened 2022-12-12
Using time.Time as a map key is tricky and should be avoided for maximum predictability

ValentinTrinque:
As stated by the golang documentation

> Note that the Go == operator compares not just the time instant but also the Location and the monotonic clock reading. Therefore, Time values should not be used as map or database keys without first guaranteeing that the identical Location has been set for all values, which can be achieved through use of the UTC or Local method, and that the monotonic clock reading has been stripped by setting t = t.Round(0). In general, prefer t.Equal(u) to t == u, since t.Equal uses the most accurate comparison available and correctly handles the case when only one of its arguments has a monotonic clock reading.


datanode/sqlstore/market_data_test.go
```diff
@@ -892,7 +892,7 @@ func setupMarketData(t *testing.T, ctx context.Context) (*sqlstore.MarketData, e
 		marketData := csvToMarketData(t, line, seqNum)
 		seqNum++
 
-		if _, alreadyAdded := addedBlocksAt[marketData.VegaTime]; !alreadyAdded {
+		if _, alreadyAdded := addedBlocksAt[marketData.VegaTime.UnixMicro()]; !alreadyAdded {
```
jeremyletang, line 895: why the use UnixMicro here? everywhere else we do Nano. Not that it matter much but eh.

jeremyletang (dismissed):

jeremyletang:
Just realised there's a couple of other places where we do use it like so, should we get rid of it then:
```
➜  grep -rn 'map\[time.Time\]' .
./core/integration/stubs/staking_account_stub.go:27:	partyToStakeForEpoch map[time.Time]map[string]*num.Uint
./core/integration/stubs/staking_account_stub.go:70:		partyToStakeForEpoch: make(map[time.Time]map[string]*num.Uint),
./core/delegation/engine_test.go:1979:	partyToStakeForEpoch map[time.Time]map[string]*num.Uint
./core/delegation/engine_test.go:1985:		partyToStakeForEpoch: make(map[time.Time]map[string]*num.Uint),
./core/banking/engine.go:139:	scheduledTransfers         map[time.Time][]scheduledTransfer
./core/banking/engine.go:193:		scheduledTransfers:         map[time.Time][]scheduledTransfer{},
./datanode/sqlstore/market_data_test.go:881:	addedBlocksAt := make(map[time.Time]struct{})
```

datanode/sqlstore/market_data_test.go
jeremyletang: alright, weird ish

ValentinTrinque:
Alright, let's address these as well ! 💪🏻

datanode/sqlstore/market_data_test.go
pscott31: yes unfortunately postgres only supports timestamps down to microseconds. there is an extension that gives you nanosecond resolution but it seemed more faff to make sure it gets properly installed everywhere. we could have stored them as integers instead of natively as timestamps but it makes querying a bit awkward

core/banking/checkpoint.go
```diff
@@ -135,15 +135,19 @@ func (e *Engine) getRecurringTransfers() *checkpoint.RecurringTransfers {
...
 		for _, v := range v {
 			transfers = append(transfers, v.ToProto())
 		}
 
-		out = append(out, &checkpoint.ScheduledTransferAtTime{DeliverOn: k.Unix(), Transfers: transfers})
+		// k is a Unix nano timestamp, and we want a Unix timestamp.
+		// So we divide it by (time.Second / time.Nanosecond).
+		deliverOnAsUnix := k / 1e9
+
+		out = append(out, &checkpoint.ScheduledTransferAtTime{DeliverOn: deliverOnAsUnix, Transfers: transfers})
```
jeremyletang, line 150: does it matter here that it's a unix seconds or nanoseconds? I'd rather keep it all in nano to be fair as it shall fit in the int64. Also wouldn't restoring the checkpoint here be broken then?

jeremyletang: hum right... because ultimately we will need to reuse the checkpoint form mainnet which have transfers
alright, makes sense. however instead of using 1e9 here why not just:
```
k / time.Nanosecond
```
or 
```
deliverOnAsUnix := time.Unix(0, k).Unix()
```

ValentinTrinque: In golang, the `time.<Nano|Micro|Milli>second` works the other way around. The smallest measure `time.Nanosecond` is equal to `1`, and it multiplies by `1000` for each upper tranche, so:

- `k / time.Nanosecond`  is basically `k / 1` that won't do it.
- `time.Unix(0, k).Unix()` would work but just more "expensive". A division should always work for us.

jeremyletang: right, typo, I mean time.Second, sorry to be picky, but that just makes it too cryptic, like anywhere else when you want a time in second your would do 4*time.Second or in Nano 41*time.Nanosecond, here just use it too.

ValentinTrinque: OK

jeremyletang (approved):

PR #7125, opened 2022-12-12
Move the authentication of wallet API v2 to the transport layer

### LINKED ISSUE #6895: Get rid of the `token` property inside the JSON-RPC API
Moving the token from the wallet API v2 to higher authentication layer.

We should use the `Authentication` HTTP header instead, as follow:
```
Authorization: VWT <TOKEN>
```

`VWT` stands for _Vega Wallet Token_. It's a custom scheme (non-standard).

ValentinTrinque:
Close #6895 

## Test the feature

1. Start the wallet service
2. On `http://localhost:1789/api/v2/requests`, send the following body:
   ```
   {
    "jsonrpc": "2.0",
    "method": "client.connect_wallet",
    "id": "wahtever-id-you-want"
   }
   ```
3. Get the token back from the header `Authorization` in the response
4. On `http://localhost:1789/api/v2/requests`, set the `Authorization` header with the previous value and send the following body:
   ```
   {
     "jsonrpc": "2.0",
     "method": "client.list_keys",
     "id": "wahtever-id-you-want"
   }
   ```

Do this for all possible methods, more detail at https://docs.vega.xyz/testnet/api/vega-wallet/v2-api


ValentinTrinque:
@wwestgarth Thanks for spotting this out. But leaving such bug in place _is unacceptable_™. Will fix it.

wwestgarth:
Really minor and not related at all to the changes here, but is a one liner so feel free to address now or ignore for later:
Looks like in `func (s *Store) DescribeToken()` we're not filling in creation data so it always comes out as `0001-01-01`.

wwestgarth (dismissed):

jeremyletang (approved):

PR #7137, opened 2022-12-13
feat: make spam protection ban 30 minutes

### LINKED ISSUE #7136: Update spam ban length from 4 epochs to 30 mins
In line with the spec to be updated we need to make sure that the length of the ban is updated:

- https://github.com/vegaprotocol/specs/pull/1415

- [ ] update the ban length as per the spec
- [ ] fix any tests this breaks

ze97286:
Closes #7136


core/spam/engine.go
```diff
@@ -118,6 +124,18 @@ func New(log *logging.Logger, config Config, epochEngine EpochEngine, accounting
 	return e
 }
 
+// OnEpochDurationChanged updates the ban duration as a fraction of the epoch duration.
+func (e *Engine) OnEpochDurationChanged(_ context.Context, duration time.Duration) error {
+	epochImpliedDurationNano, _ := num.UintFromDecimal(num.DecimalFromInt64(duration.Nanoseconds()).Mul(banDurationAsEpochFraction))
+	epochImpliedDurationDuration := time.Duration(epochImpliedDurationNano.Uint64())
```
jeremyletang, line 130: Do we need to go through a num.UintFromDecimal here? decimal.IntPart would be enough? doesn't matter much thought

jeremyletang (approved):

core/spam/engine.go
ze97286: for the frequency this happens it doesn't matter really

Issue #7028, opened 2022-12-05
Dehistory: `datanode dehistory` commands have usability issues


wwestgarth:
# Problem encountered
There are a few usability with the set of `datanode dehistory` commands:
- They do not have a `--output=json` option like all out other CLI tools, meaning [the system tests have to parse plain text](https://github.com/vegaprotocol/system-tests/blob/develop/modules/networkMod/networkMod.py#L2507)
- Overriding the config with CLI options does not work i.e `vega datanode show --api.port=DIFFERENT PORT`
- The help output is not helpful eg. for `dump-segment` we have doing `vega datanode dehistory fetch --help` gives:
```
Usage:
  datanode [OPTIONS] dehistory fetch [fetch-OPTIONS]
```

with no details of wat `fetch-options` are. You only know if you help the top-level command:
```
Available commands:
  dump-segment                       dumps the specified segment to disk
  fetch                              fetch <start from history segment id> <blocks to fetch>, fetches the given number of blocks into this node's decentralised history
  latest-history-segment-from-peers  latest-history-segment returns the id of the networks latest history segment
  list-active-peers                  list the active datanode peers
  load                               loads the most recent contiguous history from decentralised history into the datanode
  show                               shows decentralised history segments currently stored by the node
```

# Observed behaviour
Dehistory CLI is not very usable.

# Expected behaviour
Dehistory CLI is usable

# System response
Describe what the system response was, include the output from the command, automation, or else.

# Steps to reproduce

## Manual
Steps to reproduce the behaviour manually:

## Automation
Link to automation and explanation on how to run it to reproduce the problem/bug

# Evidence

## Logs
If applicable, add logs and/or screenshots to help explain your problem.

## Additional context
Add any other context about the problem here including; system version numbers, components affected.

# Definition of Done
>ℹ️ Not every issue will need every item checked, however, every item on this list should be properly considered and actioned to meet the [DoD](https://github.com/vegaprotocol/vega/blob/develop/DEFINITION_OF_DONE.md).

**Before Merging**
- [ ] Code refactored to meet SOLID and other code design principles
- [ ] Code is compilation error, warning, and hint free
- [ ] Carry out a basic happy path end-to-end check of the new code
- [ ] All APIs are documented so auto-generated documentation is created
- [ ] All bug recreation steps can be followed without presenting the original error/bug
- [ ] All Unit, Integration and BVT tests are passing
[...]

PR #7153, opened 2022-12-14
fix: empty datanode command for loading historical data into pre-populated node

### LINKED ISSUE #7152: Loading historical data requires easy way to empty the node
3 primary use cases for dehistory:

1) Get node up to latest height and processing new blocks
2) Get a node that already contains data up to the latest height (filling in the gap) and processing new blocks
3) Load all (or some) historical data 

Cases 1 and 2 are handled by the autoinitialise flag, no manual intervention required.  Case 3 however requires that the runners are archival nodes carry out the following process:

A) start datanode (with autoinitialise=true) - this will get the datanode up to the networks current height and start producing segments locally
B) Fetch the required historical asynchronously - as datanode is running it will not get behind during this process
C) Stop the datanode, empty the datanode and load all data from dehistory.

Currently there is no easy way, without connecting to DB and executing some drop database etc commands, to empty the datanode.  Automating the dropping of the data in the case where the history to load is older than that in the datanode would be possible but not desirable, better that the operator takes this decision explicitly. - The error message when load fails due to this scenario is clear e.g.

`loading history from height 1 is not possible as it is before the datanodes oldest block height 19001, to load this history first empty the datanode`

ettec:
closes #7152

### DIFF
--- a/cmd/data-node/commands/dehistory/empty-datanode.go
+++ b/cmd/data-node/commands/dehistory/empty-datanode.go
@@ -1,0 +1,55 @@
+package dehistory
+
+import (
+	"fmt"
+
+	"code.vegaprotocol.io/vega/datanode/sqlstore"
+
+	"code.vegaprotocol.io/vega/cmd/vegawallet/commands/flags"
+	"code.vegaprotocol.io/vega/logging"
+	"code.vegaprotocol.io/vega/paths"
+
+	"code.vegaprotocol.io/vega/datanode/config"
+)
+
+type emptyDatanodeCmd struct {
+	config.VegaHomeFlag
+	config.Config
+}
+
+func (cmd *emptyDatanodeCmd) Execute(_ []string) error {
+	cfg := logging.NewDefaultConfig()
+	cfg.Custom.Zap.Level = logging.WarnLevel
+	cfg.Environment = "custom"
+	log := logging.NewLoggerFromConfig(
+		cfg,
+	)
+	defer log.AtExit()
+
+	vegaPaths := paths.New(cmd.VegaHome)
+	configFilePath, err := vegaPaths.CreateConfigPathFor(paths.DataNodeDefaultConfigFile)
+	if err != nil {
+		return fmt.Errorf("couldn't get path for %s: %w", paths.DataNodeDefaultConfigFile, err)
+	}
+
+	err = paths.ReadStructuredFile(configFilePath, &cmd.Config)
+	if err != nil {
+		return fmt.Errorf("failed to read configuration:%w", err)
+	}
+
+	if datanodeLive(cmd.Config) {
+		return fmt.Errorf("datanode must be shutdown before datanode can be emptied")
+	}
+
+	yes := flags.YesOrNo("This will remove all data from datanode, are you sure?")
+
+	if yes {
+		err = sqlstore.WipeDatabaseAndMigrateSchemaToVersion(log, cmd.Config.SQLStore.ConnectionConfig, 0, sqlstore.EmbedMigrations)
+		if err != nil {
+			return fmt.Errorf("failed to wipe database and migrate to schema version 0: %w", err)
+		}
+		fmt.Println("Datanode is now empty")
+	}
+
+	return nil
+}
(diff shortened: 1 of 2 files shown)


wwestgarth:
For Scenario 3, is that not what the `MinimumBlockCount = 1` config option to auto-initialise is for?

cmd/data-node/commands/dehistory/empty-datanode.go
```diff
@@ -0,0 +1,55 @@
...
+	err = paths.ReadStructuredFile(configFilePath, &cmd.Config)
+	if err != nil {
+		return fmt.Errorf("failed to read configuration:%w", err)
+	}
+
+	if datanodeLive(cmd.Config) {
+		return fmt.Errorf("datanode must be shutdown before datanode can be emptied")
+	}
+
+	yes := flags.YesOrNo("This will remove all data from datanode, are you sure?")
```
wwestgarth, line 44: We're going to need a `--force` option or something that circumvents the `yes/no` so people can script against it.

cmd/data-node/commands/dehistory/empty-datanode.go
```diff
@@ -0,0 +1,55 @@
...
+	)
+	defer log.AtExit()
+
+	vegaPaths := paths.New(cmd.VegaHome)
+	configFilePath, err := vegaPaths.CreateConfigPathFor(paths.DataNodeDefaultConfigFile)
+	if err != nil {
+		return fmt.Errorf("couldn't get path for %s: %w", paths.DataNodeDefaultConfigFile, err)
+	}
+
+	err = paths.ReadStructuredFile(configFilePath, &cmd.Config)
```
wwestgarth, line 35: As in https://github.com/vegaprotocol/vega/issues/7028 can we make sure `config.Config` can be overriden by CLI options?

ettec:
> For Scenario 3, is that not what the `MinimumBlockCount = N` config option to auto-initialise is for? We keep fetching segments until we have it all?

That works when the history is small, as its fetched synchronously, thus the network is moving on while the fetch is happening, but thats ok if we're talking seconds/minutes.  Where the history is large (weeks/months whatever, i.e. setting up a new archive node) by the time the history is fetched the node would again be significantly behind the current network height, so for example,  there would be no suitable core snapshot available to start from.  Hence why the fetch of large sections of history has to be done async.

cmd/data-node/commands/dehistory/empty-datanode.go
ettec: phils in the middle of that, some code I will need to pickup, gonna wait till he merges

wwestgarth:
> > For Scenario 3, is that not what the `MinimumBlockCount = N` config option to auto-initialise is for? We keep fetching segments until we have it all?
> 
> That works when the history is small, as its fetched synchronously, thus the network is moving on while the fetch is happening, but thats ok if we're talking seconds/minutes. Where the history is large (weeks/months whatever, i.e. setting up a new archive node) by the time the history is fetched the node would again be significantly behind the current network height, so for example, there would be no suitable core snapshot available to start from. Hence why the fetch of large sections of history has to be done async.

Oh ok, so this would work:
```
vega datanode dehistory fetch EVERYTHING

# takes some time

vega datanode start # let autoinitiliase fill in the rest that appear while everything was fetching earlier
```

but it just means you won't have a datanode up while you're waiting for a big fetch?

I'm wondering then, whether it should just be an option to `dehistory load` instead of something separate? eg.
```
vega datanode dehistory load # get some error because you have stuff already in the db, so you do

vega datanode dehistory load --wipe-existing
```
?

ettec:
> > > For Scenario 3, is that not what the `MinimumBlockCount = N` config option to auto-initialise is for? We keep fetching segments until we have it all?
> > 
> > 
> > That works when the history is small, as its fetched synchronously, thus the network is moving on while the fetch is happening, but thats ok if we're talking seconds/minutes. Where the history is large (weeks/months whatever, i.e. setting up a new archive node) by the time the history is fetched the node would again be significantly behind the current network height, so for example, there would be no suitable core snapshot available to start from. Hence why the fetch of large sections of history has to be done async.
> 
> Oh ok, so this would work:
> 
> ```
> vega datanode dehistory fetch EVERYTHING
> 
> # takes some time
> 
> vega datanode dehistory load
> vega datanode start # let autoinitiliase fill in the rest that appeared while everything was fetching earlier
> ```
> 
> but it just means you won't have a datanode up while you're waiting for a big fetch?
> 
> I'm wondering then, whether it should just be an option to `dehistory load` instead of something separate? eg.
> 
> ```
> vega datanode dehistory load # get some error because you have stuff already in the db, so you do
> 
> vega datanode dehistory load --wipe-existing-or-something
> ```
> 
> ?

possibly that would work but with a large amount of data the autoinitialise could then take a long time to auto init and you are back to square one as by the time autoinit has pulled the data you would be a long way behind current height

ettec:
@wwestgarth closing this - I think your suggestion to add a '--wipe-existing' option to the load command is a better solution, going to close this PR and do that instead.

PR #7174, opened 2022-12-19
fix: non-deterministice order of price bounds on market data event

### LINKED ISSUE #7173: Market data events sending price bounds in non-deterministic order
Market data events sending price bounds in non-deterministic order

This code in pricemonitor.go:

sort.SliceStable(ret,
		func(i, j int) bool {
			return ret[i].Trigger.Horizon <= ret[j].Trigger.Horizon &&
				ret[i].Trigger.Probability.LessThanOrEqual(ret[j].Trigger.Probability)
		})

should be:

`sort.SliceStable(ret,
		func(i, j int) bool {
			if ret[i].Trigger.Horizon == ret[j].Trigger.Horizon {
				return ret[i].Trigger.Probability.LessThanOrEqual(ret[j].Trigger.Probability)
			}
			return ret[i].Trigger.Horizon < ret[j].Trigger.Horizon
		})`

ettec:
closes #7173

### DIFF
--- a/core/monitor/price/pricemonitoring.go
+++ b/core/monitor/price/pricemonitoring.go
@@ -245,9 +245,12 @@ func (e *Engine) GetCurrentBounds() []*types.PriceMonitoringBounds {
 
 	sort.SliceStable(ret,
 		func(i, j int) bool {
-			return ret[i].Trigger.Horizon <= ret[j].Trigger.Horizon &&
-				ret[i].Trigger.Probability.LessThanOrEqual(ret[j].Trigger.Probability)
+			if ret[i].Trigger.Horizon == ret[j].Trigger.Horizon {
+				return ret[i].Trigger.Probability.LessThan(ret[j].Trigger.Probability)
+			}
+			return ret[i].Trigger.Horizon < ret[j].Trigger.Horizon
 		})
+
 	return ret
 }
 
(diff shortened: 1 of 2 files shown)


core/monitor/price/pricemonitoring.go
```diff
@@ -245,9 +245,12 @@ func (e *Engine) GetCurrentBounds() []*types.PriceMonitoringBounds {
 
 	sort.SliceStable(ret,
 		func(i, j int) bool {
-			return ret[i].Trigger.Horizon <= ret[j].Trigger.Horizon &&
-				ret[i].Trigger.Probability.LessThanOrEqual(ret[j].Trigger.Probability)
+			if ret[i].Trigger.Horizon == ret[j].Trigger.Horizon {
+				return ret[i].Trigger.Probability.LessThanOrEqual(ret[j].Trigger.Probability)
```
ze97286, line 249: I don't know if it is valid or not but if you can have two triggers with same horizon and same probability, then you need a tie breaker with the AuctionExtension. As it is now, with the LessThanOrEqual you'll have the same problem if you have 2 with same horizon and probability

ze97286: actually this shouldn't be allowed so it would be sufficient to have `ret[i].Trigger.Probability.LessThan(ret[j].Trigger.Probability)`

but would need to add validation in new market validation, maybe as a separate ticket though.

ettec: sorted that

ze97286 (approved):

PR #7178, opened 2022-12-19
fix: 7172 market current state duplication

### LINKED ISSUE #7172: We have both markets and markets current as current state data
We have 2 tables for markets->  'markets_current' and 'markets'.    The latter should be a history table to stop snapshot bloat.  Also, the markets store is using the markets table where it should perhaps be using market_current.  This should be reviewed as well.

guoguojin:
Closes #7172


guoguojin:
Incorrect changes made due to misunderstanding of the ticket.

PR #7181, opened 2022-12-20
feat: test AC 0024-OSTA-009

Jiajia-Cui:
test AC 0024-OSTA-009

Test that market settlement cashflows only depend on parties positions and is independent of what order types there are on the book


Jiajia-Cui:
replaced with [PR #7182](https://github.com/vegaprotocol/vega/pull/7182)

PR #7197, opened 2022-12-21
chore: do not list all snapshots in memory when asked for them

### LINKED ISSUE #7198: Limit RAM usage when Listin snapshots
DO THIS.

jeremyletang:
close #7198

### DIFF
--- a/core/snapshot/engine.go
+++ b/core/snapshot/engine.go
@@ -33,6 +33,7 @@ import (
 	"code.vegaprotocol.io/vega/logging"
 	"code.vegaprotocol.io/vega/paths"
 	"github.com/cosmos/iavl"
+	tmtypes "github.com/tendermint/tendermint/abci/types"
 	db "github.com/tendermint/tm-db"
 )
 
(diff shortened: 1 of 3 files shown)


EVODelavega (approved):

core/snapshot/engine.go
```diff
@@ -218,6 +219,40 @@ func (e *Engine) ReloadConfig(cfg Config) {
...
 }
 
+// List returns all snapshots available.
+func (e *Engine) ListMeta() ([]*tmtypes.Snapshot, error) {
+	e.avlLock.Lock()
+	defer e.avlLock.Unlock()
+
+	snapshots := make([]*tmtypes.Snapshot, 0, len(e.versions))
+	// TM list of snapshots is limited to the 10 most recent ones.
+	i := len(e.versions) - 11
```
ze97286, line 229: I know it's copied from the list() but why the -11?

jeremyletang: no idea, I would have done -10, but I guess there's some black magic reason for it to be -11?

ze97286: it must be because it's the smallest two-digit prime number (◔_◔)

jeremyletang: it must definitely be the reason.

PR #7222, opened 2022-12-28
fix: fix market triggering ratio value type

### LINKED ISSUE #7171: triggeringRatio does not update correctly over GRPC and gQL
# Problem encountered
When proposing a market with a float value (e.g. **0.3**) for triggeringRatio, the proposal passes, but the outcome value is **0.0** over GRPC and **0** over gQL

# Test run here
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/36675/tests

AnExsomnis:
close: #7171


edd (dismissed):
GraphQL change looks fine to me.

Note to self: this will probably require changes in the generated proposals in docs.

wwestgarth:
I think this is a breaking change given the type has gone from `double -> string` in the market proposal? So we'll need a changelog entry and a breaking-change label.

core/types/governance_new_market.go
```diff
@@ -214,7 +218,11 @@ func NewMarketConfigurationFromProto(p *vegapb.NewMarketConfiguration) *NewMarke
 	}
 	var liquidityMonitoring *LiquidityMonitoringParameters
 	if p.LiquidityMonitoringParameters != nil {
-		liquidityMonitoring = LiquidityMonitoringParametersFromProto(p.LiquidityMonitoringParameters)
+		var err error
+		liquidityMonitoring, err = LiquidityMonitoringParametersFromProto(p.LiquidityMonitoringParameters)
+		if err != nil {
+			return nil, fmt.Errorf("error getting new market configuration from proto: %s", err)
```
wwestgarth, line 224: Minor: should probably be `%w` to wrap the error.

commands/proposal_submission.go
```diff
@@ -464,7 +464,18 @@ func checkLiquidityMonitoring(parameters *protoTypes.LiquidityMonitoringParamete
 		return errs
 	}
 
-	if parameters.TriggeringRatio < 0 || parameters.TriggeringRatio > 1 {
+	tr := float64(0)
+	var err error
+	if len(parameters.TriggeringRatio) > 0 {
+		tr, err = strconv.ParseFloat(strings.TrimSpace(parameters.TriggeringRatio), 64)
```
wwestgarth, line 470: We should be using `num.DecimalFromString()` here so that we're not dealing with floats at all.

I think also for other fields when then value is an empty string we return `ErrIsNotValidNumber` instead of treating it as `"0"`, so we should probably be consistent.

core/types/liquidity.go
```diff
@@ -412,7 +413,7 @@ func (l LiquidityMonitoringParameters) IntoProto() *proto.LiquidityMonitoringPar
 	}
 	return &proto.LiquidityMonitoringParameters{
 		TargetStakeParameters: params,
-		TriggeringRatio:       tr,
+		TriggeringRatio:       strconv.FormatFloat(tr, 'f', -1, 64),
```
wwestgarth, line 416: This should just be `l.TriggeringRatio.String()`, currently we're going from `Decimal -> float64 -> string`

core/types/governance_new_market.go
AnExsomnis: Addressed here https://github.com/vegaprotocol/vega/pull/7222/commits/935f9774a523f334d55e43742ff63338043eb2c4#diff-872de7073d1b2054247d4d9bfcdc63c49d530f5ee501083b4fc7ec970ce13051R467
Thanks!

commands/proposal_submission.go
AnExsomnis: Addressed here - https://github.com/vegaprotocol/vega/pull/7222/commits/935f9774a523f334d55e43742ff63338043eb2c4#diff-872de7073d1b2054247d4d9bfcdc63c49d530f5ee501083b4fc7ec970ce13051R467
Thanks!

core/types/liquidity.go
AnExsomnis: Corrected here as well https://github.com/vegaprotocol/vega/pull/7222/commits/935f9774a523f334d55e43742ff63338043eb2c4#diff-8bf56342620e6cec70bbb347efe846b5a22a316a7e903809ebe5707a7ffe0386R414
Thanks!

edd (dismissed):
👍 GraphQL change still looks fine to me

core/types/liquidity.go
```diff
@@ -405,14 +405,13 @@ type LiquidityMonitoringParameters struct {
...
 func (l LiquidityMonitoringParameters) IntoProto() *proto.LiquidityMonitoringParameters {
-	tr, _ := l.TriggeringRatio.Float64()
 	var params *proto.TargetStakeParameters
 	if l.TargetStakeParameters != nil {
 		params = l.TargetStakeParameters.IntoProto()
 	}
 	return &proto.LiquidityMonitoringParameters{
 		TargetStakeParameters: params,
-		TriggeringRatio:       tr,
+		TriggeringRatio:       l.TriggeringRatio.String(),
```
zale144, line 414: Can `l.TriggeringRatio` ever be `nil`?

zale144 (dismissed):
lgtm!

core/types/liquidity.go
AnExsomnis: No, it is `decimal.Decimal`, not a pointer to it.

wwestgarth (dismissed):

datanode/entities/market.go
```diff
@@ -239,9 +241,18 @@ func liquidityMonitoringParametersFromProto(lmp *vega.LiquidityMonitoringParamet
...
+		tr, err = strconv.ParseFloat(strings.TrimSpace(lmp.TriggeringRatio), 64)
+		if err != nil {
+			return LiquidityMonitoringParameters{}, err
+		}
+	}
+
 	return LiquidityMonitoringParameters{
 		TargetStakeParameters: tsp,
-		TriggeringRatio:       lmp.TriggeringRatio,
+		TriggeringRatio:       tr,
```
jeremyletang, line 255: Does that get save into postres as float? any chance it could get a different representation and fuck up the datanode snapshots accross different nodes?

jeremyletang: We may want to use decimal or string here too?

AnExsomnis: Yes, I missed that, sorry.
Here I added the change to have the `TrigerringRatio` as a string inside the datanode object as well - https://github.com/vegaprotocol/vega/pull/7222/commits/2d223ec15e292897001a794aa99bd00d630636c4
So in that way we receive it as a string from the proto, do nothing on it - insert it as a string in the datanode object and in the DB
So we will have it as a decimal.Decimal object only in the core 
https://github.com/vegaprotocol/vega/pull/7222/commits/2d223ec15e292897001a794aa99bd00d630636c4

jeremyletang (approved):

PR #7234, opened 2023-01-04
chore: add max memory usage as part of the config for datanode and core

### LINKED ISSUE #7228: Data node OOM investigations
During recent incentive the data nodes have been running out of memory.

This seems to be around the time of snapshot creation/loading

This issue is to investigate and fix this issue.

jeremyletang:
Setup max memory usage for core and datanode

close #7228

### DIFF
--- a/cmd/data-node/commands/start.go
+++ b/cmd/data-node/commands/start.go
@@ -14,6 +14,7 @@ package commands
 
 import (
 	"context"
+	"runtime/debug"
 
 	"code.vegaprotocol.io/vega/cmd/data-node/commands/start"
 	"code.vegaprotocol.io/vega/datanode/config"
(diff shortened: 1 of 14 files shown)


wwestgarth (dismissed):
Was there any reason behind the default of 33%? (I've not got a better suggestion, just out of interest really)

ze97286 (dismissed):

peterbarrow (dismissed):

jeremyletang:
> Was there any reason behind the default of 33%? (I've not got a better suggestion, just out of interest really)

No really good reason apart from the assumption that people would run the datanode the core node and postgres on the same machine, so just to keep a a third of the ram for each of them.

Although that makes me think that the validators should run the core node as a standalone, (no postgres or datanode, so we should be less conservative in this case, waht do you think @wwestgarth?)

wwestgarth (approved):

PR #7240, opened 2023-01-04
fix: when gap fill loading dehistory only attempt to load data more recent than the datanode contains

### LINKED ISSUE #7233: Dehistory considers all data in dehistory when loading data - it should look only at history at height > than current height
From this thread: https://vegaprotocol.slack.com/archives/C9E6XA8GK/p1672146115967499

When gap filling, it is considering all history in the dehistory store.  It subsequently ignores all history < current block height, however this process can take a long time until it reaches the correct height.  It should only consider history from dehistory at height > current block height.

ettec:
closes #7233 

Previously on load it was pulling all dehistory into the staging directory and then attempting to load this, only history with height > than current datanode height would be loaded, however, it would check each segment (which was quite time consuming).  This change ensures just segments that contain data the datanode doesn't already contain are put into the staging area.   Also changed the load command to better handle the various scenarios  (datanode has no data, dehistory data has older history, datanode has older data)

### DIFF
--- a/cmd/data-node/commands/dehistory/load.go
+++ b/cmd/data-node/commands/dehistory/load.go
@@ -106,30 +106,50 @@ func (cmd *loadCmd) Execute(_ []string) error {
 		return nil
 	}
 
-	fmt.Printf("Decentralized history from block height %d to %d is available to load, current datanode block span is %d to %d\n",
-		from, to, span.FromHeight, span.ToHeight)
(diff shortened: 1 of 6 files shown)


jeremyletang (approved):
Slightly related, but I suppose it would be good at some point to uniformise command line everywhere and make sure we use the proper output (e.g: human readable, json or whatever) just to make sure it works well with or without tty (thinking about the YesOrNo flag there  etc.

PR #7254, opened 2023-01-05
fix: improve weight adjustment for handling or precision errors in re…

### LINKED ISSUE #7253: Rewards payout abnormal for epoch 417
At 18:33 (GMT+1) on the 4th of January 2023 a user reported that epoch 417 appeared to have a significantly different reward payout when compared to other epoch rewards. In some cases token holders that had staked and delegated to validators received ~700 tokens for the epoch. whereas others may not have had the reward payout as expected. Some tools and front ends reported a total payout of roughly 10x the normal amount, though we believe it is more likely that some parties received nothing and the reward API reported the amount that was attempted to pay, rather than actually transferred.

Additional info:

https://vegaprotocol.slack.com/archives/CFXA4EGN7/p1672864113593079

ze97286:
…ward calculation

The issue is the result of the following: for each delegator in the validator we calculate their relative share of the reward given to the validator, so lets say if we have 3 validators with delegation to node1 being: [10,20,70], then this would give us [0.1, 0.2, 0.7]. However, with rounding errors the sum of the weights array may be > 1 which may lead to overflows. So we worked around it (before this issue happened) by checking the total weight and if it's > 1, pick at pseudorandom one delegator and adjust their weight by the precision error, so that after the adjustment we are guaranteed to have the weights sum up to 1. what happened here is that the selected party had a weight of 0.0000000000000003 (2fd6dfb33917e0438746a3d0f70175a6ba528f497649cbcd8e0d1c8607182e2d on staking facilities) on the validator and a precisionError 0.0000000000000004, so when we adjust it (to 0) we still end up with a total weight > 1 which we no longer check which leads to the overflow. This led to the core reporting a payout for the epoch that is 10x greater than the actual balance of the reward account at the time. While the core reported a total reward payout of 50k, as the reward account balance was something like 5k, only the latter has been paid by executing the transfers sequentially until there was not enough balance to pay more. The issue if fully reproduced by the unit test added in this PR. 

Closes #7253


jeremyletang (dismissed):

core/rewards/staking_reward_calculator.go
```diff
@@ -180,3 +178,24 @@ func calculateRewardsByStake(epochSeq, asset, accountID string, rewardBalance *n
...
 }
+
+func adjustWeights(delegatorWeights map[string]num.Decimal, weightSums num.Decimal, sortedParties []string, decimalOne num.Decimal, rng *rand.Rand) {
+	// NB: due to rounding errors this sum can be greater than 1
+	// to avoid overflow, we choose at random a party adjust it by the error
+	// we keep looking until we find a candidate with sufficient weight to adjust by the precision error
+	for weightSums.GreaterThan(decimalOne) {
+		precisionError := weightSums.Sub(decimalOne)
+		for precisionError.GreaterThan(num.DecimalZero()) {
+			unluckyParty := sortedParties[rng.Intn(len(delegatorWeights))]
```
davidsiska-vega, line 189: Would it be simpler / safer to take the precision error from the party with the biggest weight rather than choosing a random one?

ze97286: it will be equally safe, and slightly less fair, but I don't mind

davidsiska-vega: Sorry, I was missing the outer `for` loop in my thinking. We go and try to remove the `precisionError` repeatedly till it's gone. So the case I am describing above isn't a problem.

davidsiska-vega: Let's go with the random integer selection.

ze97286: though equally safe and slightly less fair I would actually prefer to go with adjusting the largest. For the largest it would probably be a meaningless adjustment so I don't think it matters.

jeremyletang (approved):

davidsiska-vega (approved):

PR #7275, opened 2023-01-06
Feat/7114 expose spam protection per user details

### LINKED ISSUE #7114: Expose spam protection per users details
Implement an endpoint in the core.proto server to expose per users spam information.
This needs to be implemented as forwarding the endpoint to the core node from the datanode as is done with the others endpoint in this service.

### inputs
- one required party id.

### outputs
- pow infos
- state of all spam protections for the user

guoguojin:
Closes #7114


core/spam/vote_spam_policy.go
```diff
@@ -433,3 +434,23 @@ func (vsp *VoteSpamPolicy) PreBlockAccept(tx abci.Tx) (bool, error) {
...
+
+func (vsp *VoteSpamPolicy) GetStats(partyID string) Statistic {
+	vsp.lock.RLock()
+	defer vsp.lock.RUnlock()
+
+	stats := Statistic{
+		Limit: banFactor.String(),
+	}
+
+	bStats, ok := vsp.partyBlockRejects[partyID]
```
ze97286, line 446: how is this kind of thing ever going to be useful? this starting to be meaningful once the block has been proposed and is reset at the end of the block so what's the point of it?

guoguojin: This is a misunderstanding on my part due to the ban calculation using the block stats and Gordon's comment on the original ticket regarding helping users avoiding bans. I will update in accordance to the Slack conversation.

https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1673017527936189

core/pow/engine.go
```diff
@@ -561,3 +561,25 @@ func (e *Engine) BlockData() (uint64, string) {
...
 	return e.currentBlock, e.blockHash[e.currentBlock%ringSize]
 }
+
+func (e *Engine) GetSpamStatistics(partyID string) (uint64, int64) {
+	e.lock.RLock()
+	defer e.lock.RUnlock()
+
+	totalTx := uint64(0)
+
+	for _, state := range e.activeStates {
```
ze97286, line 571: I don't think this is sufficient. I think for pow what I'd want to know as a user is for every block height that I have transactions for, how many transactions I've sent and what is the total difficulty of them. So the statistics looks like, a list of blocks, for each block, number of transactions and total difficulty. you can also throw in the required difficulty for this block height. anything else? @davidsiska-vega @TomMcL

core/spam/vote_spam_policy.go
```diff
@@ -433,3 +434,20 @@ func (vsp *VoteSpamPolicy) PreBlockAccept(tx abci.Tx) (bool, error) {
...
+	defer vsp.lock.RUnlock()
+
+	stats := Statistic{}
+
+	bStats, ok := vsp.partyBlockRejects[partyID]
+	if !ok {
+		return stats
+	}
+
+	stats.Total = strconv.FormatUint(bStats.total, formatBase)
```
ze97286, line 449: it might also be worth returning the actual min tokens required as this may be growing during a perceived attack

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,50 @@ message LastBlockHeightResponse {
...
+  // If blocked the timestamp when the party will be unblocked as RFC3339Nano
+  optional string until = 4;
+}
+
+// The vote statistics for the voting spam policies
+// which are calculated as a ratio of the total votes
+// that have been rejected.
+message VoteSpamStatistic {
+  // The current vote count received from the party for the given policy during this epoch
+  string count_for_epoch = 1;
```
ze97286, line 372: this needs to be by proposal, there's no significance to the total number of votes but the votes for a given proposal.

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,50 @@ message LastBlockHeightResponse {
...
+  // If blocked the timestamp when the party will be unblocked as RFC3339Nano
+  optional string until = 2;
```
ze97286, line 374: until is not very descriptive for what it actually is referring to.

protos/sources/vega/api/v1/core.proto
ze97286, line 372: for consistency why is max_for_epoch not included here?

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,50 @@ message LastBlockHeightResponse {
...
+}
+
+// The statistics for a given spam policy
+message SpamStatistic {
+  // The current transaction count received from the party during this epoch for this policy
+  string count_for_epoch = 1;
+  // The maximum number of transactions allowed for this policy in an epoch
+  string max_for_epoch = 2;
+  // If blocked the timestamp when the party will be unblocked as RFC3339Nano
+  optional string until = 4;
```
ze97286, line 364: could be better named

core/spam/vote_spam_policy.go
```diff
@@ -433,3 +434,20 @@ func (vsp *VoteSpamPolicy) PreBlockAccept(tx abci.Tx) (bool, error) {
...
 }
+
+func (vsp *VoteSpamPolicy) GetStats(partyID string) Statistic {
+	vsp.lock.RLock()
+	defer vsp.lock.RUnlock()
+
+	stats := Statistic{}
+
+	bStats, ok := vsp.partyBlockRejects[partyID]
+	if !ok {
```
wwestgarth, line 445: Minor: this is a little inconsistent with the other spam-stats categories in that for the others, if you pass in a party that hasn't been seen you get counts of `"0"` but for votes you get `""`:
```
{
  "statistics": {
    "proposals": {
      "countForEpoch": "0",
      "maxForEpoch": "3"
    },
    "delegations": {
      "countForEpoch": "0",
      "maxForEpoch": "390"
    },
    "transfers": {
      "countForEpoch": "0",
      "maxForEpoch": "20"
    },
    "nodeAnnouncements": {
      "countForEpoch": "0",
      "maxForEpoch": "1"
    },
    "votes": {
      "countForEpoch": "" <-- this guy
    },
    "pow": {
      "totalTxsInScope": "0"
    }
  }
}
```

core/pow/engine.go
TomMcL: :+1: . Not even sure whether you need total difficulty given you can check you aren't currently blocked, but it can't hurt and might make some wallet-side calculations easier

guoguojin: Addressed in https://github.com/vegaprotocol/vega/pull/7275/commits/9804a5d4f1f2567c57a6ee0713cb1ea7da56d315

core/spam/vote_spam_policy.go
guoguojin: Thanks Will, I believe this has been addressed now: https://github.com/vegaprotocol/vega/pull/7275/commits/9804a5d4f1f2567c57a6ee0713cb1ea7da56d315

protos/sources/vega/api/v1/core.proto
davidsiska-vega: Yes,  100%.

protos/sources/vega/api/v1/core.proto
davidsiska-vega: I am all for longer and more descriptive variable names on the interface.

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,61 @@ message LastBlockHeightResponse {
...
+  string max_for_epoch = 3;
+  // If blocked the timestamp when the party will be unblocked as RFC3339Nano
+  optional string banned_until = 4;
+}
+
+// The PoW state for a given block
+message PoWBlockState {
+  uint64 block_height = 1;
+  uint64 transactions_seen = 2;
+  uint64 difficulty = 3;
```
ze97286, line 384: I think it's worth adding here a few params for the block - the difficulty and the increaseDifficulty values in particular.

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,61 @@ message LastBlockHeightResponse {
...
+  optional string banned_until = 2;
+}
+
+// The complete spam statistics captured for a given party
+message SpamStatistics {
+  SpamStatistic proposals = 1;
+  SpamStatistic delegations = 2;
+  SpamStatistic transfers = 3;
+  SpamStatistic node_announcements = 4;
+  repeated VoteSpamStatistic votes = 5;
```
ze97286, line 401: I think this is a bit awkward that vote here is a collection rather than an object with a collection because things like banned_until and effective_Min_Tokens (which I think is worth adding) are global for party votes rather than per proposal.

core/spam/statistics.go
```diff
@@ -0,0 +1,23 @@
...
+
+type Statistics struct {
+	Proposals         Statistic
+	Delegations       Statistic
+	Transfers         Statistic
+	NodeAnnouncements Statistic
+	Votes             []VoteStatistic
+}
+
+type VoteStatistic struct {
```
ze97286, line 11: I'm not a fan of having these intermediary structs - what's the point of them? we only create them temporarily before we convert to the proto ones so why not use the proto ones directly? may not be super clean from a pure point of view but for something like that I'd prefer performance.

core/spam/statistics.go
```diff
@@ -0,0 +1,23 @@
...
+	Proposal    string
+	Total       string
```
ze97286, line 13: why are those `strings` rather than uint64?

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,61 @@ message LastBlockHeightResponse {
...
+  optional string banned_until = 4;
+}
+
+// The vote statistics for the voting spam policies
+// which are calculated as a ratio of the total votes
+// that have been rejected.
+message VoteSpamStatistic {
+  string proposal = 1;
+  // The current vote count received from the party for the given policy during this epoch
+  string count_for_epoch = 2;
```
ze97286, line 373: why are all the counts strings rather than uint64?

protos/sources/vega/api/v1/core.proto
ze97286: block hash is a good idea. 
a couple of reasons - the wallet may not need to cache historical values of difficulty so if anyone wants to generate another transaction for a particular block having the difficulty param here is useful. In addition what I thought of originally is an optimisation where given the required difficulty and the observed one, the wallet can optimise the pow calculation to not increase the difficulty if not needed (if got lucky in earlier rounds and got increased difficulty).

wwestgarth: Core's calculation of what difficulty you need to use can take into account luckily getting high difficulty too. I just think if we ever change in the future the way the difficulty is calculated its a single core change, no big deal. If we let wallet's calculate then its a breaking-change that could affects many many people's code.

I don't really follow your point about having to cache historical values? Got an example?

core/spam/statistics.go
guoguojin: It was because in a lot of other places we store numbers as strings, and in the back of my mind, I've somehow made it a rule. I will change it.

protos/sources/vega/api/v1/core.proto
wwestgarth: You've remembered it correctly (or atleast remembered it the same as me). 
What I'm saying is that a wallet (ours, or someone elses implementation) will read these values from the API and write some function `f(difficulty, increaseDifficulty, transactions_seen) -> difficulty_to_aim_for_now` 

So we might as well just have our API do `f()` for us and just return `difficulty_to_aim_for_now`

ze97286: we could, I wanted to keep it simple...

wwestgarth: Ok, fair enough lets keep it simple!

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,81 @@ message LastBlockHeightResponse {
...
+  // The current transaction count received from the party during this epoch for this policy
+  uint64 count_for_epoch = 1;
+  // The maximum number of transactions allowed for this policy in an epoch
+  uint64 max_for_epoch = 2;
+  // If blocked the timestamp when the party will be unblocked as RFC3339Nano
+  optional string banned_until = 4;
+}
+
+// The voting statistics by proposal for a given party for the current epoch
+message VoteSpamStatistics {
```
ze97286, line 369: maybe add the effective min_token_required

core/spam/vote_spam_policy.go
```diff
@@ -436,3 +436,24 @@ func (vsp *VoteSpamPolicy) PreBlockAccept(tx abci.Tx) (bool, error) {
...
+
+	stats := make([]Statistic, 0, len(partyStats))
+	bannedUntil := vsp.bannedParties[partyID]
+
+	for proposal, votes := range partyStats {
+		stats = append(stats, Statistic{
+			Name:        proposal,
+			Total:       votes,
+			Limit:       vsp.numVotes,
+			BannedUntil: bannedUntil,
```
ze97286, line 454: why is bannedUntil inside here? it should be global for the party

core/spam/engine.go
```diff
@@ -231,3 +235,60 @@ func (e *Engine) PostBlockAccept(tx abci.Tx) (bool, error) {
...
+		case txn.DelegateCommand:
+			stats.Delegations = ssProto
+		case txn.TransferFundsCommand:
+			stats.Transfers = ssProto
+		case txn.AnnounceNodeCommand:
+			stats.NodeAnnouncements = ssProto
+		case txn.VoteCommand:
+			vs := make([]*protoapi.VoteSpamStatistic, len(pStats))
+			for i, ps := range pStats {
+				vs[i] = &protoapi.VoteSpamStatistic{
```
ze97286, line 278: it's a bit redundant to generate the struct there with the redundant banned until and then copy it here. better to return already the full struct from the vote stats func.

core/spam/vote_spam_policy.go
guoguojin: It's because this is the object returned by anything implementing the spam.Policy interface. The spam.Engine will extract it out and provide only 1 in the response.

```
// The voting statistics by proposal for a given party for the current epoch
message VoteSpamStatistics {
  // List of statistics for proposals voted on by the party
  repeated VoteSpamStatistic statistics = 1;
  // If blocked the timestamp when the party will be unblocked as RFC3339Nano
  optional string banned_until = 2;
}

// The vote statistics for the voting spam policies
// which are calculated as a ratio of the total votes
// that have been rejected.
message VoteSpamStatistic {
  string proposal = 1;
  // The current vote count received from the party for the given proposal during this epoch
  uint64 count_for_epoch = 2;
  // The maximum number of votes per proposal allowed in an epoch
  uint64 max_for_epoch = 3;
}
```

core/pow/engine.go
```diff
@@ -561,3 +576,65 @@ func (e *Engine) BlockData() (uint64, string) {
...
+		bannedUntil = &untilStr
+	}
+
+	return &protoapi.PoWStatistic{
+		BlockStates: stats,
+		BannedUntil: bannedUntil,
+	}
+}
+
+func calculateSpamExpectedDifficulty(spamPoWDifficulty, spamPoWNumberOfTxPerBlock, seenTx uint, observedDifficulty uint) uint64 {
```
ze97286, line 629: can you call this func getMinDifficultyForNextTx maybe? 

func getMinDifficultyForNextTx(baseDifficulty, txPerBlock, seenTx, observedDifficulty uint) uint64

ze97286: note, that none of this is correct if the boolean increaseDifficulty is false! in which case there is no difficulty that can be used for the next transaction for the block. not sure how you want to express that.

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,84 @@ message LastBlockHeightResponse {
...
+// which are calculated as a ratio of the total votes
+// that have been rejected.
+message VoteSpamStatistic {
+  string proposal = 1;
+  // The current vote count received from the party for the given proposal during this epoch
+  uint64 count_for_epoch = 2;
+  // The maximum number of votes per proposal allowed in an epoch
+  uint64 max_for_epoch = 3;
+  // The effective minimum number of tokens required to vote on the proposal
+  uint64 min_tokens_required = 4;
```
ze97286, line 386: that can't be uint64, it has to be string.

core/spam/vote_spam_policy.go
```diff
@@ -436,3 +438,31 @@ func (vsp *VoteSpamPolicy) PreBlockAccept(tx abci.Tx) (bool, error) {
...
+
+	stats := make([]*protoapi.VoteSpamStatistic, 0, len(partyStats))
+	bannedUntil := vsp.bannedParties[partyID]
+
+	for proposal, votes := range partyStats {
+		stats = append(stats, &protoapi.VoteSpamStatistic{
+			Proposal:          proposal,
+			CountForEpoch:     votes,
+			MaxForEpoch:       vsp.numVotes,
+			MinTokensRequired: vsp.minVotingTokens.Uint64(),
```
ze97286, line 460: that would have to be vsp.minVotingTokens.String()

ze97286 (dismissed):

wwestgarth (dismissed):

core/pow/engine_test.go
```diff
@@ -149,13 +151,105 @@ func TestMempoolTidRejection(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotTotal, gotDifficulty := calculateExpectedDifficulty(tt.args.spamPowDifficulty, tt.args.spamPoWNumberOfTxPerBlock, tt.args.seenTx)
+			require.Equal(t, tt.wantTotal, gotTotal)
+			require.Equal(t, tt.wantDifficulty, gotDifficulty)
+		})
+	}
+
+	// require.Equal(t, uint(60), calculateExpectedDifficulty(20, 5, 3))   // 3 * 20
```
davidsiska-vega, line 246: Should we get rid of this commented out code? The for loop above seems to be testing against the data included.

davidsiska-vega (dismissed):

jeremyletang (dismissed):

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,84 @@ message LastBlockHeightResponse {
...
+message PoWBlockState {
+  // Block height for the current Proof of Work state statistics
+  uint64 block_height = 1;
+  // Hash of the current block
+  string block_hash = 2;
+  // Total number of transactions seen in the block
+  uint64 transactions_seen = 3;
+  // The minimum expected difficulty for the next transaction in the block
+  // This is only provided if increasing difficulty is required
+  optional uint64 expected_difficulty = 4;
```
ValentinTrinque, line 399: Why not return always return it so we don't have to compute it on the wallet side ? That would save us from implementing the algorithm inside the wallet. And consolidating the state would be much simpler.

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,84 @@ message LastBlockHeightResponse {
...
+  SpamStatistic node_announcements = 4;
+  // Statistics for proposal votes made by the party
+  VoteSpamStatistics votes = 5;
+  // Statistics for proof of work difficulty observed per block for the party
+  PoWStatistic pow = 6;
+}
+
+// A response containing all the spam statistics of a party for the current epoch
+message GetSpamStatisticsResponse {
+  SpamStatistics statistics = 1;
```
ValentinTrinque, line 430: If I understand this PR correctly, we would replace our current call to the last block information by this API. Then we need to return the ChainID in the payload.

ValentinTrinque (changes requested):

protos/sources/vega/api/v1/core.proto
```diff
@@ -343,3 +348,84 @@ message LastBlockHeightResponse {
...
+
+// The vote statistics for the voting spam policies
+// which are calculated as a ratio of the total votes
+// that have been rejected.
+message VoteSpamStatistic {
+  string proposal = 1;
+  // The current vote count received from the party for the given proposal during this epoch
+  uint64 count_for_epoch = 2;
+  // The maximum number of votes per proposal allowed in an epoch
+  uint64 max_for_epoch = 3;
```
ValentinTrinque, line 384: Either I understand this property documentation the wrong way, either it shouldn't live in this payload. `VoteSpamStatistic` seems to be related to a "single" proposal. However `max_for_epoch` seems to be a threshold set for all proposals. Wouldn't this make the information redundant ?

protos/sources/vega/api/v1/core.proto
ze97286: I don't think it is meant to be used in the way you describe - at least not explicitly.  With small changes it could, but as is, it only gives you statistics about blocks you already have transactions for, it won't tell you what the current block is nor what its hash is nor what the relevant pow params are.

protos/sources/vega/api/v1/core.proto
ze97286: it's max_per_proposal, but its definition is global and it shouldn't be here.

protos/sources/vega/api/v1/core.proto
ze97286: I'm not sure what you're saying. if it's nil it means you can't send any more transaction on this block. If it is not nil, then this is the exact (minimum) difficulty you need to use in the wallet.

protos/sources/vega/api/v1/core.proto
ze97286: I don't think you can currently do that with the current implementation and I'm not aware that this was a requirement, those as commented above, it could be but would require some changes.

protos/sources/vega/api/v1/core.proto
ValentinTrinque: Right my question was confusing, but your response is what was I was looking for.

It makes sense, and fits what I would like.

However, the comment needs to be rework, bale cause it basically says the value is present only when the difficulty is increased which is not what you explained, right ?

protos/sources/vega/api/v1/core.proto
ValentinTrinque: Alright I get it. We need both calls then.

ze97286: not necessarily, I like the idea - I think all we need for that is to always add to this the last block. But we may do it separately if decided it's desired.

protos/sources/vega/api/v1/core.proto
ze97286: I would suggest rephrasing the comment as:
This is the minimum required difficulty for the next transaction submitted on this block if it is possible to submit more transactions on this block, otherwise nil.

core/pow/engine_test.go
ze97286: we should

protos/sources/vega/api/v1/core.proto
guoguojin: As it was explained to me, when the increaseDifficulty param is false, then there is no difficulty calculation, so there's no value for the expected difficulty.

core/pow/engine_test.go
```diff
@@ -149,13 +151,105 @@ func TestMempoolTidRejection(t *testing.T) {
...
+	// require.Equal(t, uint(100), calculateExpectedDifficulty(20, 5, 5))  // 5 * 20
+	// require.Equal(t, uint(121), calculateExpectedDifficulty(20, 5, 6))  // 5 * 20 + 21
+	// require.Equal(t, uint(184), calculateExpectedDifficulty(20, 5, 9))  // 5 * 20 + 4 * 21
+	// require.Equal(t, uint(205), calculateExpectedDifficulty(20, 5, 10)) // 5 * 20 + 5 * 21
+	// require.Equal(t, uint(430), calculateExpectedDifficulty(20, 5, 20)) // 5 * 20 + 5 * 21 + 5 * 22 + 5 * 23
+	// require.Equal(t, uint(478), calculateExpectedDifficulty(20, 5, 22)) // 5 * 20 + 5 * 21 + 5 * 22 + 5 * 23 + 2 * 24
```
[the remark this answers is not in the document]
guoguojin: Yes, this should have been removed, but I forgot. I will address it along with your other comments.

core/pow/engine.go
```diff
@@ -561,3 +576,97 @@ func (e *Engine) BlockData() (uint64, string) {
...
+
+	params := activeParams[stateInd]
+	return params
+}
+
+func (e *Engine) GetSpamStatistics(partyID string) *protoapi.PoWStatistic {
+	e.lock.RLock()
+	defer e.lock.RUnlock()
+
+	stats := make([]*protoapi.PoWBlockState, 0, len(e.activeStates)) //
```
ze97286, line 597: I don't think this allocation makes sense. There's no real connection between the number of blocks in scope that the party can have transactions on and the number of active states which is the number of network param changes that are in scope.

ValentinTrinque (dismissed):

ze97286 (dismissed):

jeremyletang (approved):

PR #7277, opened 2023-01-06
fix: bug with pagination defaults

### LINKED ISSUE #7260: Datanode pagination doesn't work if 'first' or 'last' not set
# Problem encountered
@TomMcL  ran into problem whereby he was trying to fetch everything in a collection

# Observed behaviour
Going to e.g. `/api/v2/accounts`, getting the default page of 1000 items, and then using the end-of-page cursor to get the next page by going to `/api/v2/accounts?pagination.after=xxx` - you got exactly the same page back.

# Expected behaviour
You would expect the next page

# Mitigation
You can work around it by just specifying, e.g. `pagination.first=1000` in the query but one shouldn't need to

# Investigation
This is because by of the following but of code:
```go
func CursorPaginationFromProto(cp *v2.Pagination) (CursorPagination, error) {
	if cp == nil || (cp.Last == nil && cp.First == nil && cp.NewestFirst == nil) {
		return DefaultCursorPagination(true), nil
	}
```
Which says, if you don't specify first or last (e.g. a page size) you get the default pagination object, not just a default page size applied.

Should be easy to fix

pscott31:
Closes #7260

### DIFF
--- a/datanode/entities/pagination.go
+++ b/datanode/entities/pagination.go
@@ -19,6 +19,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+var defaultPageSize int32 = 1000
+
 type Pagination interface{}
 
 type OffsetPagination struct {
@@ -103,7 +105,7 @@ func NewCursorPagination(first *int32, after *string, last *int32, before *strin
 }
 
 func CursorPaginationFromProto(cp *v2.Pagination) (CursorPagination, error) {
-	if cp == nil || (cp.Last == nil && cp.First == nil && cp.NewestFirst == nil) {
+	if cp == nil {
 		return DefaultCursorPagination(true), nil
 	}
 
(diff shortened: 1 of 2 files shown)


datanode/entities/pagination.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+var defaultPageSize int32 = 1000
```
jeremyletang, line 22: should we make smaller pages? This seems massive?

pscott31: I dunno I thought it seemed rather small.. i mean ball park let's say it's 512 bytes per record on average (and I would think that's a significant overestimate) that's just a 500kb payload. The smaller you make it the more requests someone has to do to get the whole set if they want it, and there is definitely much more cost to someone doing 2 requests or 20. Really you want to minimise the number of requests/database queries; I doubt it makes a lot of difference to Postgres if you're getting 100 records or 10,000 but it's when there are millions it starts getting a bit more unmanageable

jeremyletang: Hum fair enough

jeremyletang (approved):

PR #7312, opened 2023-01-12
chore: add dependency between number of validators and min number of …

ze97286:
…multi sig signers required


wwestgarth:
I imagine there is going to be some system-tests full run issues with this change that need fixing up, because we have tests which reduce `network.validators.tendermint.number` without changing the other one. Full-run: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/15912/pipeline

gordsport:
System test issue to support resolving the test failures 

- https://github.com/vegaprotocol/system-tests/issues/1677

wwestgarth:
The issue with the failing checkpoint tests is due to an order-dependency in the rule checking in the netparam engine. We check the rules as we are loading each network-parameter and not after loading them all.

So for example if I have in my genesis file the two parameters orders:
```
"network.validators.multisig.numberOfSigners": "5",
"network.validators.tendermint.number": "5",
```
we load nSigners first and compare to the default of nTendermint: 5 <= 30, passes validation
we the load nTendermint and compare it to the loaded nSigners: 5 <= 5, passes validation

If I switch them around and instead have in my genesis file:
```
So for example if I have in my genesis file the two parameters orders:
```
"network.validators.tendermint.number": "5",
"network.validators.multisig.numberOfSigners": "5",
```
we the load nTendermint and compare it to the default nSigners: 5 <= 13, fails validation because nTendermint must be higher.

Its only showing up in the checkpoint tests because in the order of the parameters changes between when we first start the network, and the genesis we use when restarting from a checkpoint (I don't know why)

jeremyletang:
> The issue with the failing checkpoint tests is due to an order-dependency in the rule checking in the netparam engine. We check the rules as we are loading each network-parameter and not after loading them all.
> 
> So for example if I have in my genesis file the two parameters orders:
> 
> ```
> "network.validators.multisig.numberOfSigners": "5",
> "network.validators.tendermint.number": "5",
> ```
> 
> we load nSigners first and compare to the default of nTendermint: 5 <= 30, passes validation we the load nTendermint and compare it to the loaded nSigners: 5 <= 5, passes validation
> 
> If I switch them around and instead have in my genesis file:
> 
> ```
> "network.validators.tendermint.number": "5",
> "network.validators.multisig.numberOfSigners": "5",
> ```
> 
> we the load nTendermint and compare it to the default nSigners: 5 <= 13, fails validation because nTendermint must be higher.
> 
> Its only showing up in the checkpoint tests because in the order of the parameters changes between when we first start the network, and the genesis we use when restarting from a checkpoint (I don't know why)

Probably because it's  stored in a map (not sure anymore).
Can we do as you suggest to fix that, basically: 
- load all parameters
- validate them then once all loaded

wwestgarth (approved):

davidsiska-vega (approved):

PR #7324, opened 2023-01-13
chore: add a bunch of indexes following audit of queries

### LINKED ISSUE #7316: Audit our database indexes
Look at all the tables, and all the ways we query them. Make sure there are indexes where appropriate.

Here's an initial list from Jeremy:

```
ranking_scores:
 - node_id
reward_scores:
 - node_id
reward:
 - party_id
 - asset_id
 - market_id
 - epoch_id
delegations:
 - party_id
 - asset_id
 - market_id
delegations_current:
 - same as prev
deposits:
 - asset
 - party
withdrawal:
 - same as prev?
proposal:
 - party_id
vote:
 - party_id
 - proposal_id
current_margin_level:
 - account_id
positions:
 - party_id
 - market_id
positions_current:
 - party_id
 - market_id
liquidity_provision:
 - market_id
 - party_id
liquidity_provision_current:
 - market_id
 - party_id
stake_linking:
 - party_id
stake_linking_current:
 - party_id
```

pscott31:
Closes #7316 

After going through every query we do and [every table](https://docs.google.com/spreadsheets/d/15uv8BWDBuSW8nAxdfRiCRLXQ9Xbeb9-E0lKX-ORZ2Hk/edit?usp=sharing) we have, I think we now have indexes covering everything just about everything we do on any table of non-trivial size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### 🛠 Improvements
 - [7302](https://github.com/vegaprotocol/vega/issues/7302) - Add withdrawal minimal amount
 - [5487](https://github.com/vegaprotocol/vega/issues/5487) - Add `UPGRADING.md`
+- [7316](https://github.com/vegaprotocol/vega/issues/7316) - Add a bunch of database indexes following audit of queries
 
 ### 🐛 Fixes
 - [7260](https://github.com/vegaprotocol/vega/issues/7260) - Fix bug where pagination `before` or `after` cursors were ignored if `first` or `last` not set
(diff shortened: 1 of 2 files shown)


jeremyletang:
The audit CSV linked in the summary could be quite useful in the future if we need to re-assert all this, could we add it in the codebase as well somewhere?

pscott31:
It's a bit of hode-podge of a spreadsheet as I was just kind-of using it to keep track of what I'd looked at/found. I could tidy it up a bit, but I'm not sure how useful it would be going forward, as it's bound to become out of date quickly.. I think the real solution is to put some monitoring/metrics in place that reports when real queries are running that aren't using any indexes or taking a long time and letting us know about it.

Also make it a bit of dicipline when reviewing code that does new queries to check that there are relevant indexes if needed.

jeremyletang (approved):

PR #7332, opened 2023-01-14
feat: improve decrease of validator set when changing the network par…

### LINKED ISSUE #7331: Controlled decrease of the number of validators upon network parameter change
# Feature Overview

When the network parameter controlling the number of validators in the network is decreased we want to reduce it one validator an epoch as long as the remaining number of validators is sufficiently large to be able to update the multisig contract. 

This means if the number of validators is going down from n to k:

1. sort the set of n validators by decreasing rank
2. remove the lowest ranking validator from consensus validators set
3. if this validator is not a signer on the multisig and more validators need to be removed to make n==k, keep removing validators.
4. if the validator is a signer and removing it will still leave enough validators such that `num_remaining * 1000/(num_signers) > threshold ` then remove the validator.
5. repeat in the next epoch until no more validators can be removed from the tendermint validator set or the number of validators in the set is equal to the value of the network parameter

## Specs
PR - https://github.com/vegaprotocol/specs/pull/1518/files

ze97286:
…ameter

Closes #7331

### DIFF
--- a/core/validators/validator_score_test.go
+++ b/core/validators/validator_score_test.go
@@ -54,6 +54,18 @@ func (t *TestMultisigTopology) ExcessSigners(addresses []string) bool {
 	return false
 }
 
+func (t *TestMultisigTopology) GetSigners() []string {
+	signers := make([]string, 0, len(t.validators))
+	for k := range t.validators {
+		signers = append(signers, k)
+	}
+	return signers
+}
+
+func (t *TestMultisigTopology) GetThreshold() uint32 {
+	return 666
+}
+
 func TestScores(t *testing.T) {
 	t.Run("test calculation of stake score with no anti-whaling", testStakeScore)
 	t.Run("test calculation of performance score", testPerformanceScore)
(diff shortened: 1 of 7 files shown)


davidsiska-vega (dismissed):
I think it matches what's proposed here. Anyway, let's get it in!

core/validators/validator_score_test.go
```diff
@@ -54,6 +54,18 @@ func (t *TestMultisigTopology) ExcessSigners(addresses []string) bool {
 	return false
 }
 
+func (t *TestMultisigTopology) GetSigners() []string {
+	signers := make([]string, 0, len(t.validators))
+	for k := range t.validators {
+		signers = append(signers, k)
+	}
+	return signers
```
jeremyletang, line 62:
```suggestion
	return maps.Keys(t.validators)
```

jeremyletang (approved):
Approved, but you might want to wait for a review from @wwestgarth

wwestgarth (approved):
Code looks fine, but I am suprised there were no failures on the full run. Theres a test `test_pending_not_promoted` that drops the number of TM validators from 5 -> 3 without removing any signers and it is still passing.

You've added a unittest for this exact case though and it works as expected (not allowing the full drop until signers are removed), so I'm just going to assume its a system-test issue for now. I'll make a note about it when the QA ticket for the ACs pops up.

ze97286:
> Code looks fine, but I am suprised there were no failures on the full run. Theres a test `test_pending_not_promoted` that drops the number of TM validators from 5 -> 3 without removing any signers and it is still passing.
> 
> You've added a unittest for this exact case though and it works as expected (not allowing the full drop until signers are removed), so I'm just going to assume its a system-test issue for now. I'll make a note about it when the QA ticket for the ACs pops up.

let me rerun this system test because the full run precedes the latest code which includes this test you're referring to

PR #7341, opened 2023-01-16
feat: network history should pre-allocate connections to use so it is resilient to db  connection starvation

### LINKED ISSUE #7336: Segment creation should use pre-allocated connection to prevent failure due to external process consuming all connections
Segment creation should use pre-allocated connection to prevent failure due to external process consuming all connections

Currently history segment creation creates a connection for the lifetime of the segment creation, this is problematic if an external process is using the postgres instance (should we even allow this? things could go wrong in very unexpected ways perhaps have dedicated instance(s) for this sort of stuff?) and consumes all available connections, at this point the segment creation will fail as it will not be able to create a connection.

ettec:
closes #7336

### DIFF
--- a/datanode/networkhistory/snapshot/database_meta_data.go
+++ b/datanode/networkhistory/snapshot/database_meta_data.go
@@ -6,11 +6,12 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/jackc/pgx/v4"
+	"github.com/pressly/goose/v3"
+
 	"code.vegaprotocol.io/vega/datanode/sqlstore"
 	"github.com/georgysavva/scany/pgxscan"
 	"github.com/jackc/pgx/v4/pgxpool"
-	"github.com/jackc/pgx/v4/stdlib"
-	"github.com/pressly/goose/v3"
 )
 
 type DatabaseMetadata struct {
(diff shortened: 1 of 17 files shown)


datanode/networkhistory/snapshot/database_meta_data.go
```diff
@@ -170,3 +150,101 @@ func getHyperTablePartitionColumns(ctx context.Context, conn *pgxpool.Pool) (map
 	}
 	return tableNameToPartitionColumn, nil
 }
+
+// getDBVersion copied from the goose library and modified to support using a pre-allocated connection. It's worth noting
+// that this method also has the side effect of creating the goose version table if it does not exist as per the original
```
ettec, line 155: Yeah, I hate this, but could see no better way

datanode/sqlstore/connection_source.go
```diff
@@ -271,3 +263,22 @@ func (t *delegatingConnection) QueryFunc(ctx context.Context, sql string, args [
 	}
 	return t.pool.QueryFunc(ctx, sql, args, scans, f)
 }
+
+func CreateConnectionPool(conf ConnectionConfig) (*pgxpool.Pool, error) {
```
ettec, line 267: we set a minimum pool size for the network-history pool, but not, and never had, for the sqlstores pool (note: write connection is seperate and isolated) - should we?   sensible default?

jeremyletang (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

PR #7346, opened 2023-01-17
fix: accounts by id should lookup via cache

### LINKED ISSUE #7345: Account.GetByID is not cached
Account.GetByID is not cached.   This method is called frequently by server side code when constructing other protos (so for example when iterating transfers) - seeing 100+ time per seconds on grafana. We should cache it.

ettec:
closes #7345  -  accounts are looked up by id with high frequency, lookup should go via a cache


datanode/sqlstore/accounts.go
```diff
@@ -33,14 +33,16 @@ var accountOrdering = TableOrdering{
 
 type Accounts struct {
 	*ConnectionSource
-	cache     map[accountCacheKey]entities.Account
-	cacheLock sync.Mutex
+	cache            map[accountCacheKey]entities.Account
+	idToAccountCache map[entities.AccountID]entities.Account
+	cacheLock        sync.Mutex
```
jeremyletang, line 38: Should this be an RWMutex?

ettec: apparently RWMutex does not let you upgrade the read lock to a write lock (bit surprising), you have to explicity release the RLock and then obtain the Lock, and calling RUnlock when you don't have the RLock is a runtime error(?!).  So you can't then rely on defer to unlock.  Which means the code becomes a bit tangled, I don't think the perf benefit is worth it here.  Maybe I've missed something?

datanode/sqlstore/accounts.go
```diff
@@ -66,6 +68,7 @@ func (as *Accounts) checkCache(
 func (as *Accounts) updateCache(a entities.Account) {
 	key := accountCacheKey{a.PartyID, a.AssetID, a.MarketID, a.Type}
 	as.cache[key] = a
+	as.idToAccountCache[deterministicAccountID(&a)] = a
```
wwestgarth, line 71: Bit of a shame to have two caches for the same data. Is there a reason why `Obtain` can't also use one with the `deterministicAccountID()` key too?

ettec: you'd effectivley have to add a mapping index, which given the size of the account object and number of them doesn't carry any material advantage but does add a bit of complexity (IMO at least ;-) )

ettec: > Bit of a shame to have two caches for the same data. Is there a reason why `Obtain` can't also use one with the `deterministicAccountID()` key too?

actually, theres not much in it, I'll update to map the cache key

datanode/sqlstore/accounts.go
jeremyletang: Well just Lock when you write and Rlock in any other case. I could imagine the perf benefits could be there seeing that you implemented that specifically because there was lots of calls to the accounts, so the rlock would allow to read that map conccurrently.

datanode/sqlstore/accounts.go
ettec: > > Bit of a shame to have two caches for the same data. Is there a reason why `Obtain` can't also use one with the `deterministicAccountID()` key too?
> 
> actually, theres not much in it, I'll update to map the cache key

ok, bit more of a refactor than originally intending, but the code is better for it I think, the Obtain method is a bit of an odd one in that it doesn't return an account but rather updates the instance passed in, but there were (and afaik still are) good reasons for it being done that way.  @Phil might want to double check the refactor in case missed something dumb.

datanode/sqlstore/accounts.go
```diff
@@ -114,10 +100,15 @@ func (as *Accounts) GetAll(ctx context.Context) ([]entities.Account, error) {
 func (as *Accounts) Obtain(ctx context.Context, a *entities.Account) error {
 	as.cacheLock.Lock()
 	defer as.cacheLock.Unlock()
-	if account, ok := as.checkCache(a.PartyID, a.AssetID, a.MarketID, a.Type); ok {
+
+	id := deterministicIDFromAccount(a)
+	if account, ok := as.idToAccount[id]; ok {
 		a.ID = account.ID
+		a.VegaTime = account.VegaTime
```
ettec, line 107: @pscott31 I set these attrs as well, any reason this is not the right thing to do?

wwestgarth (dismissed):

datanode/sqlstore/accounts.go
ettec: yeah, because its just a map lookup so contention would be very low (writes are v.rare) I was 50/50 on it tbh given the additional logic needed,  I'll make the change

ettec: ok, thats done

jeremyletang (approved):

PR #7365, opened 2023-01-18
Add cors headers and max age

### LINKED ISSUE #7335: Expose custom headers
# Problem encountered
When using a basic JavasSript fetch request you are unable to retrieve Vega's custom headers (eg x-block-height), I believe due `'Access-Control-Expose-Headers': '*'` not being set by the backend.

# Observed behaviour
Custom response headers are not retrievable

# Expected behaviour
Should be able to access custom response headers

# System response

REST example
```
  fetch('https://api.n00.testnet.vega.xyz/statistics', {
    headers: {
      'Content-Type': 'application/json',
    },
    method: 'GET',
  })
    .then((res) => {
      const blockheight = res.headers.get('Grpc-Metadata-X-Block-Height');
      const contenttype = res.headers.get('content-type');
      console.log(contenttype); // application/json
      console.log(blockheight); // null
      return res.json();
    })
    .then(console.log);
```

GraphQL example
```
  fetch('https://api.n09.testnet.vega.xyz/graphql', {
    headers: {
      'Content-Type': 'application/json',
    },
    body: '{"query":"query {\\n  statistics {\\n    blockHeight\\n  }\\n}"}',
    method: 'POST',
  }).then((res) => {
    const blockheight = res.headers.get('x-block-height');
    const contenttype = res.headers.get('content-type');
    console.log(contenttype); // application/json
    console.log(blockheight); // null
    return res.json()
  })
  .then(console.log);
```

# Steps to reproduce

## Manual
1. Run either javascript above
2. Notice that block height headers return null

### LINKED ISSUE #7349: Set Access-Control-Max-Age for OPTIONS responses
# Feature Overview

**In order to** avoid clients having to send as many PREFLIGHT requests as they currently do
**We will** set `Access-Control-Max-Age` to `7200` for all `OPTIONS` queries (to graphql & rest endpoints)
**So that** browsers will cache OPTIONS as effectively as possible

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age - and https://developer.mozilla.org/en-US/docs/Glossary/preflight_request for more about OPTIONS/preflights

Current default: OPTIONS responses expire after 5 seconds, so will be sent on the next query after that.

# Tasks
- [ ] Make GraphQL respond with the appropriate header and value to all OPTIONS requests to the query endpoint
- [ ] Ensure REST does the same
- [ ] Ensure Block Explorer does the same

zale144:
Resolve #7335 
close #7349


libs/http/cors.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"net/http"
+	"strings"
+
+	"github.com/rs/cors"
+)
+
+// CORSConfig represents the configuration for CORS.
+type CORSConfig struct {
+	AllowedOrigins string `long:"allowed-origins" description:"Allowed origins for CORS"`
```
jeremyletang, line 12: Could/Should that be a slice?

zale144: @jeremyletang Could. Not sure if it should though. I mean, it was easier to put `string` because of `*`. I can change it if you think it'd be cleaner

jeremyletang: I don't mind, it's just that then the lib can setup a list of allowed origins, here we restrict it to 1. not that it matters much

jeremyletang (dismissed):

libs/http/cors.go
zale144: We can still have more than one, since the origin is matched with `strings.Contains` (line 37 below).
Yeah, now I realise it's not that clean. I better fix it

jeremyletang (approved):

PR #7369, opened 2023-01-18
feat: retry and panic on sync segment creation error

### LINKED ISSUE #7360: Segment Creation - Should attempt retry of segment creation if it fails in the synchronous segment creation code and panic if still failing after retries
Segment Creation - Should attempt retry of segment creation and panic if still failing after retries.

Currently the block commit handler logs the error if the synchronous part of segment creation fails.  It should instead retry X number of times, after this point if still failing it should panic.  Panic'ing is consistent with what happens if failure occurs during the asynchronous part of segment creation (an unrecoverable error as the database state will changed, due to async event processing, from the point at which it the snapshot needs to be created) .   The agreed view here is that a data-node should not continue if it is unable to produce history segments, it should be considered a critical failure.

ettec:
closes #7360 the part of the segment creation code that is synchronous with the event processing now retries to create the segment on a failure, after 6 failed attempts it will panic.

### DIFF
--- a/cmd/data-node/commands/networkhistory/fetch.go
+++ b/cmd/data-node/commands/networkhistory/fetch.go
@@ -32,7 +32,7 @@ func (cmd *fetchCmd) Execute(args []string) error {
 	defer log.AtExit()
 
 	if len(args) != 2 {
-		return errors.New("expected <start-history-segment-id> <num-blocks-to-fetch>")
+		return errors.New("expected <history-segment-id> <num-blocks-to-fetch>")
 	}
 
 	rootSegmentID := args[0]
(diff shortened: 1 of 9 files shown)


pscott31 (dismissed):

pscott31 (dismissed):

wwestgarth:
maybe kinda of related, but I was just checking through the code that does the snapshot to double check theres nothing that would slip up on a retry and noticed that this block might be missing a call to `runAllInReverseOrder(cleanup)` before returning on error? If we don't I think we'll still be holding on to the lock.
```
// To ensure reads are isolated from this point forward execute a read on last block
_, err = sqlstore.GetLastBlockUsingConnection(ctx, copyDataTx)
if err != nil {
	return MetaData{}, fmt.Errorf("failed to get last block using connection: %w", err)
}
```

ettec:
> maybe kinda of related, but I was just checking through the code that does the snapshot to double check theres nothing that would slip up on a retry and noticed that this block might be missing a call to `runAllInReverseOrder(cleanup)` before returning on error? If we don't I think we'll still be holding on to the lock.
> 
> ```
> // To ensure reads are isolated from this point forward execute a read on last block
> _, err = sqlstore.GetLastBlockUsingConnection(ctx, copyDataTx)
> if err != nil {
> 	return MetaData{}, fmt.Errorf("failed to get last block using connection: %w", err)
> }
> ```

yeah, good spot, that should also rollback....sorting...

ettec:
> 

sorted

wwestgarth (approved):

PR #7372, opened 2023-01-18
PnL updates when trade is made

### LINKED ISSUE #7169: Positions API (PnL) should get updated on trade
# Feature Overview

Now that we mark positions to market at set intervals (rather than continuously), the positions API (specifically the PnL data) doesn't get updated until the next MTM anniversary (end of block). As a result, parties won't see what their PnL and open volume is until they've been marked to market. Traders should be able to see their open volume on trade.

## Specifics

In short, the positions subscriber should update the position/PnL data on trade events (the positions engine emits a `PositionStateEvent` that contains the information we can use). This allows us to accurately provide the open volume regardless of MTM.

EVODelavega:
Add pending PnL fields to position data to track PnL per trade, rather than per settlement.

closes #7169

### DIFF
--- a/datanode/entities/position.go
+++ b/datanode/entities/position.go
@@ -32,10 +32,6 @@ type positionSettlement interface {
 	TxHash() string
 }
 
-type positionState interface {
-	Size() int64
-}
-
 type lossSocialization interface {
 	Amount() *num.Int
 	TxHash() string
--- a/datanode/entities/position.go
+++ b/datanode/entities/position.go
@@ -32,6 +32,10 @@ type positionSettlement interface {
 	TxHash() string
 }
 
+type positionState interface {
+	Size() int64
+}
+
 type lossSocialization interface {
 	Amount() *num.Int
 	TxHash() string
@@ -48,6 +52,14 @@ type settleMarket interface {
 	TxHash() string
 }
 
+type PendingPosition struct {
+	OpenVolume              int64
+	RealisedPnl             decimal.Decimal
+	UnrealisedPnl           decimal.Decimal
+	AverageEntryPrice       decimal.Decimal
+	AverageEntryMarketPrice decimal.Decimal
+}
+
 type Position struct {
 	MarketID                MarketID
 	PartyID                 PartyID
@@ -60,6 +72,9 @@ type Position struct {
 	Adjustment              decimal.Decimal // what a party was missing which triggered loss socialization
 	TxHash                  TxHash
 	VegaTime                time.Time
+	// keep track of trades that haven't been settled as separate fields
+	// these will be zeroed out once we process settlement events
+	Pending PendingPosition
 }
 
 func NewEmptyPosition(marketID MarketID, partyID PartyID) Position {
@@ -73,7 +88,16 @@ func NewEmptyPosition(marketID MarketID, partyID PartyID) Position {
 		AverageEntryMarketPrice: num.DecimalZero(),
 		Loss:                    num.DecimalZero(),
 		Adjustment:              num.DecimalZero(),
+		Pending: PendingPosition{
+			RealisedPnl:             num.DecimalZero(),
+			UnrealisedPnl:           num.DecimalZero(),
+			AverageEntryPrice:       num.DecimalZero(),
+			AverageEntryMarketPrice: num.DecimalZero(),
+		},
 	}
+}
+
+func (p *Position) UpdateWithPositionState(e positionState) {
 }
 
 func (p *Position) UpdateWithPositionSettlement(e positionSettlement) {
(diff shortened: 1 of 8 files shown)


wwestgarth:
Looks good, although there was some chat on slack yesterday about how we can no longer changing `0001_initial.sql` and should do schema upgrade instead? But I have no idea how any of that works! (cc: @pscott31 @ettec)

ettec (changes requested):

datanode/entities/position.go
```diff
@@ -32,6 +32,10 @@ type positionSettlement interface {
 	TxHash() string
 }
 
+type positionState interface {
```
pscott31, line 35: Like lint says it looks like this is unused :)

datanode/sqlstore/positions.go
```diff
@@ -49,6 +49,14 @@ func (ps *Positions) Flush(ctx context.Context) ([]entities.Position, error) {
...
 }
 
+// AddBatch just allows you to add several entities in a single call
+func (ps *Positions) AddBatch(ctx context.Context, pos ...entities.Position) error {
+	for _, p := range pos {
+		ps.batcher.Add(p)
+	}
+	return nil
+}
+
```
pscott31, line 59: Is this used anywhere?

datanode/entities/position.go
```diff
@@ -60,27 +64,61 @@ type Position struct {
 	Adjustment              decimal.Decimal // what a party was missing which triggered loss socialization
 	TxHash                  TxHash
 	VegaTime                time.Time
+	// keep track of trades that haven't been settled as separate fields
+	// these will be zeroed out once we process settlement events
+	PendingOpenVolume              int64
+	PendingRealisedPnl             decimal.Decimal
+	PendingUnrealisedPnl           decimal.Decimal
+	PendingAverageEntryPrice       decimal.Decimal
+	PendingAverageEntryMarketPrice decimal.Decimal
```
pscott31, line 73: It feels slightly inelegant to track these separately from their non-pending counterparts, is there any reason we can't just update the 'actual' fields (I guess there probably is or else you wouldn't have done it :). It's just kind of conceptually to me, when a trade is done, your realised PnL has really changed - it's not pending.. the only thing that mark-to-market is really doing is giving you a price, changing your unrealised PnL, so maybe that's the only thing that should change when we get a mark-to-market event; the rest we can figure out from just looking at trades?

* obv. it's a bit more complicated when markets settle or distressed positions get closed

EVODelavega: We no longer mark to market after trades are made, so the new/change in position isn't yet settled, that only happens when your new position gets marked to market. By the time that happens, other trades may have changed the mark price, so MTM will take your old position, at the last mark price, mark the old position to market (new mark price), and then mark the new position to the new mark price to get to the PnL values.

We can't perform the normal MTM stuff on the existing fields because we kind of need to know the old position + old mark price when the settle events come in. These fields merely exist to update the position data and provide a _"best guess"_ approximation of the realised PnL when trades are made. We know the old mark price and trade price, that gives us some profit/loss, but the old volume, from last MTM settlement is still used in the MTM settlement in the core, so the transfers performed in the core would not match up, and the position API would not accurately reflect the balance transfers

datanode/sqlstore/positions.go
EVODelavega: Probably not, I think I added this because I'm updating 2 positions when handling the trade event, but I think I ended up just calling `Add` twice (which I suppose could just as well use variadic arguments like this)

datanode/entities/position.go
EVODelavega: I thought I removed all of these, must've missed one

datanode/entities/position.go
pscott31: I get that we don't mark to market after trades any more - but if I bought one widget @ 100.. then a trade comes through that says I sold one for 101 - then I've really made some realised PnL at that point right, I don't need to wait until MtM time, I don't need any price to tell me I've made a profit of 1.. 

Or are you perhaps saying that even though my profit is 'locked in' in a sense - I won't get my margin back etc.. until the next MtM cycle and that is what makes it 'pending'?

The PositionSettlement event when it comes should have the exact same set of trades in that we got from watching the trades as they came in right? So it should be an exact match.

That said perhaps there's some wisdom in not mucking around with the existing logic as it's tried and tested. Though it feels like a bit of a kludge it's probably the best way!

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -1038,6 +1038,11 @@ CREATE TABLE positions(
   adjustment          NUMERIC NOT NULL,
```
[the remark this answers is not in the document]
EVODelavega: Done, I believe the trigger on positions update is fine, but not 100% on that

datanode/entities/position.go
EVODelavega: Oh, I see what you mean now. I suppose that, all things considered, we could probably use the trades here and continuously update (un)RealisedPnL and average entry (market) price using the trade events, and just call the `mtm` function on the position when we get the position settlement event.

For now, it might be useful to just check and see whether these pending fields match up with the values we get when we've processed the trades in the `UpdatePositionSettlement` function.
I'm just not sure what the impact of this might be when a market gets settled, or in case of loss socialisation. I'd need to look in to that a bit more. A lot of the PnL stuff (and MTM settlements for that matter) were implemented before we changed to periodic MTM settlements, so there's almost certainly room for improvement as a result

pscott31: Yes fair enough

pscott31 (approved):

ettec (approved):

PR #7411, opened 2023-01-23
Return underlying error when parsing a command failed in the wallet API version 2

### LINKED ISSUE #7410: Return underlying error when parsing a command failed in the wallet API

ValentinTrinque:
Close #7410


ze97286 (dismissed):

wwestgarth (dismissed):

wallet/api/client_sign_transaction_test.go
```diff
@@ -66,7 +67,7 @@ func testSigningTransactionWithInvalidParamsFails(t *testing.T) {
 					"type": "not vega command",
 				},
 			},
-			expectedError: api.ErrTransactionIsNotValidVegaCommand,
+			expectedError: errors.New("the transaction is not a valid Vega command: unknown field \"type\" in vega.wallet.v1.SubmitTransactionRequest"),
```
candida-d, line 70: Does this mean that someone has used a field called 'type' and it's unknown?

ValentinTrinque: Just FYI, this a test. So I created dumb payload with a field `type`, knowing there is none for that object. When parsed as a transaction, proto is unhappy and tells me `unknown field \"type\"`. This is not a error message we have control of.

candida-d: Oh! Sorry I missed that it was a test, that explains why it's so specific

jeremyletang (approved):

PR #7418, opened 2023-01-24
chore: add batch instructions to stats

ze97286:


protos/sources/vega/api/v1/core.proto
```diff
@@ -324,6 +324,10 @@ message Statistics {
   string epoch_start_time = 37;
   // Epoch expected end time
   string epoch_expiry_time = 38;
+  // Number of events in the last block
+  uint64 event_count = 39;
+  // Rate of events per second in total
```
candida-d, line 329: Don't quite understand this description, could you perhaps reword, or explain in a comment and I can provide a suggestion?

jeremyletang (dismissed):

protos/sources/vega/api/v1/core.proto
ze97286: `number_of_events_emitted_in_last_block / duration_in_seconds_last_block`

candida-d: Thank you!

candida-d (approved):

EVODelavega (approved):

edd:
I've added these to the stats page with the current logic:
- events/second & eventCount shown as ✅  if the event count is less than (`txsPerSecond` * `59`), red if above

Magic number 59 was got from two numbers @ze97286 gave me as an example, so they might not be good. It can be changed here: https://github.com/vegaprotocol/stats/blob/f8eca96c4ffab3a97a44e0c702ccaaaec382a529/src/index.html#L346-L347

PR #7421, opened 2023-01-24
feat: 7289 postitionConnection filtering

### LINKED ISSUE #7289: positionsConnection filters
# Feature Overview

**In order to** show all positions across multiple keys
**We will** update the positionsConnection query
**So that** we can pass multiple partyId variables to the query

# Acceptance Criteria
- It is possible to filter by one or more partyIds
- It is possible to filter by one or more marketIds

# Additional Details

Ideally this would end up looking something like the AccountFilter

Its likely that we will also need to:
- Un-nest the positionsConnection field from under the party field, making it a top level query
- Deprecate `party.positionsConnection`

# Examples
```
input PositionFilter {
  partyIds: [ID!]
  marketIds: [ID!]
}

Query {
  positionsConnection(
    filter: PositionFilter,
    pagination: 
  ): PositionConnection
}
```

guoguojin:
Closes #7289 

To avoid introducing breaking changes to the gRPC API to allow for filtering by multiple parties and markets, I've deprecated the old gRPC function `ListPositions` and introduced a new `ListAllPositions` that allows the filtering based on multiple parties and/or markets.


datanode/gateway/graphql/schema.graphql
```diff
@@ -4038,6 +4043,11 @@ type PositionEdge {
   cursor: String
 }
 
+input PositionsFilter {
```
candida-d, line 4046: Could you add a description here for documentation?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -63,7 +63,15 @@ service TradingDataService {
     // Positions
```
candida-d, line 63:
```suggestion
    // Positions (deprecated)
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -63,7 +63,15 @@ service TradingDataService {
...
+    rpc ListPositions(ListPositionsRequest) returns (ListPositionsResponse) {
+        option deprecated = true;
+    };
+
+    // Positions
+    //
+    // Get a list of positions by party (public key) using cursor based pagination
+    rpc ListAllPositions(ListAllPositionsRequest) returns (ListAllPositionsResponse);
 
     // Positions subscription
```
candida-d, line 76: Is everything under "Positions subscription" deprecated? If so - could you add (deprecated) to this bit of text, and can you add a title - as was done here - to the replacement?

guoguojin: No, positions subscription is not deprecated, it has not been affected and should continue to work as it does.

candida-d: Thanks for clarifying! Hard to tell in the protos formatting sometimes

mattrussell36 (dismissed):

candida-d (dismissed):

dexturr (dismissed):

datanode/gateway/graphql/resolvers.go
```diff
@@ -481,6 +481,18 @@ func (r *myDepositResolver) CreditedTimestamp(ctx context.Context, obj *types.De
 
 type myQueryResolver VegaResolverRoot
 
+func (r *myQueryResolver) PositionsConnection(ctx context.Context, filter *v2.PositionsFilter, pagination *v2.Pagination) (*v2.PositionConnection, error) {
```
pscott31, line 484: We've got a bit of a mix of these but maybe just call this `Positions` ? The 'connection' part is a bit redundant and as I remember we only did it to disambiguate from the non-paged versions which don't exist (would be nice to go through and rename all the XXXConnection to XXX at some point too.

candida-d: From a usability/docs perspective, I second this!

guoguojin: That's a good point, we can do it for this one, but the other XXXConnections will need us to go through the same deprecation/rename process as it's going to be quite a major breaking change if we don't deprecate.

pscott31: yeah.. we could have XXXConnection and XXX for a bit with XXXConnection deprecated

pscott31 (approved):

macqbat (approved):

PR #7431, opened 2023-01-25
Do not mark wallet and network as incompatible when the patch version doesn't match

### LINKED ISSUE #7429: Do not mark wallet and network as incompatible when the patch version doesn't match
Today, we mark the wallet as not compatible with the network when the wallet core dependency then embed are not strictly equal. This is frightening users.

We decided to exclude the comparison on the patch version. 

The should offer the best balance between warning the user about incompatibilities and tolerance.

ValentinTrinque:
Close #7429


wwestgarth (approved):

jeremyletang (commented):
Shouldn't we use something like that: https://pkg.go.dev/golang.org/x/mod/semver

Just in case we need to cover some weird version with suffixes or something?

ValentinTrinque:
@jeremyletang This should be handled by I have done.

jeremyletang (approved):

PR #7437, opened 2023-01-26
fix: network history snapshots causing intermittent fails

guoguojin:
During the CI data-node integration tests, we can encounter issues due to network history snapshots.

I've disabled the network history when running integration tests as they aren't needed at this time and I've run over 100 consecutive iterations of the integration tests without any issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - [](https://github.com/vegaprotocol/vega/issues/xxxx) -
 
 ### 🛠 Improvements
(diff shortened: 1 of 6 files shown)


pscott31 (dismissed):

guoguojin:
Looks like there's a race condition in the clean up code when data-node terminates that wasn't caught. Looking through my logs it only occurred once in the 100+ iterations I did to test my changes.

pscott31 (approved):

PR #7448, opened 2023-01-27
feat: AC 0034-PROB-006, check impact from tau.scaling parameter

Jiajia-Cui:
create an example in feature test as a guidance for system test, for AC 0034-PROB-006

close vegaprotocol/system-tests#1094 .


core/integration/features/verified/0034-PROB-liquidity_TauScaling.feature
```diff
@@ -0,0 +1,299 @@
+Feature: Tests impact from change of tau.scaling parameter on probability of trading, 0038-OLIQ-006
+
+  Scenario: 001: set tau.scaling to 1
+  Background:
+    Given the following network parameters are set:
```
davidsiska-vega, line 5: let's explicitly state the LP range and also somewhere print what the price monitoring bounds are or show where on the book the LP volume is actually landing

core/integration/features/verified/0034-PROB-liquidity_TauScaling.feature
```diff
@@ -0,0 +1,604 @@
...
+      | name                                                | value |
+      | market.value.windowLength                           | 1h    |
+      | market.stake.target.timeWindow                      | 24h   |
+      | market.stake.target.scalingFactor                   | 1     |
+      | market.liquidity.targetstake.triggering.ratio       | 0     |
+      | market.liquidity.providers.fee.distributionTimeStep | 10m   |
+      | network.markPriceUpdateMaximumFrequency             | 0s    |
+      | market.liquidity.probabilityOfTrading.tau.scaling   | 10    |
```
davidsiska-vega, line 13: On line 3 we say tau scaling is 1, is 10 what we want here? Because then the next scenario also looks like the scaling is 10.

davidsiska-vega (changes requested):

davidsiska-vega (approved):

jeremyletang (approved):

PR #7449, opened 2023-01-27
Add ability to filter rewards between epochs

### LINKED ISSUE #6650: Historic reward summary per epoch
# Feature Overview

**In order to** provide an overview of the rewards earned 
**We will** provide a way to query reward types over a specified window of epochs
**So that** there is a way to accurately present aggregated reward information to the user

Example for an API that would facilitate this, not set in stone if this doesn't make sense data node side, but filtering by epoch would enable this really easily, as suggested by Phil.

```
query PartyRewards {
    party (id: 'foo') {
        rewards ( fromEpoch: 1, toEpoch: 10) {
           pagination // as normal
           edges {
                 asset
                 amount
                 rewardType
                 marketId
                 // ....ect/
              }
           }
        }
    }
}
```

# Background

Currently we cannot query reward types per epoch, only the last N rewards and therefore have no way to accurately present to the user the rewards they have earned per epoch.

Screenshot 2022-10-31 at 16 18 29

Implementing this solution with existing API would mean we take a risk on presenting an epoch with an incomplete data set - the totals could be wrong, and it's possible we miss out whole reward types as well as individual rewards - and we'd have no way of knowing it.

# Tasks
- [ ]
- [ ]

# Acceptance Criteria
A list of criteria (aim for 3!) that have to be met for this feature to be accepted as 'done' by the product owner.

- It is possible to

Acceptance Criteria and Feature Test Scenarios can, in some cases, be closely related. If acceptance criteria become fully covered by feature test scenarios they can be removed leaving on acceptance criteria that can't be directly or fully proven with tests.

# Test Scenarios
Detailed scenarios (1-3!) that can be executed as feature tests to verify that the feature has been implemented as expected. We use the follow format:

GIVEN (setup/context) 
WHEN (action) 
See [here](https://github.com/vegaprotocol/vega/tree/develop/integration/) for more format information and examples.

# Additional Details (optional)

# Examples (optional)
Code snippets from the spec for reference

# Definition of Done

**Before Merging**

**After Merging**

pscott31:
Closes #6650


dexturr (dismissed):

datanode/gateway/graphql/schema.graphql
```diff
@@ -2029,6 +2029,10 @@ type Party {
     assetId: ID
     "Optional pagination information"
     pagination: Pagination
+    "Optionally only return rewards >= this epoch"
+    fromEpoch: Int
+    "Optionally only return rewards <= this epoch"
+    toEpoch: Int
```
edd, line 2035: **Nitpick (non-blocking)**: Ultra nitpick, but this is a uint64 in the GRPC, so I guess it’s technically possible that this *will eventually* overflow, if the epoch is set unreasonably short and the network lives for a long time. Hopefully by that point all browsers properly support big ints..

Regardless, I’m happy seeing an Int and not another string 😄

edd (dismissed):

datanode/gateway/graphql/schema.graphql
pscott31: yes fair enough, we use Int elsewhere for epoch though so at least it's consistent.. I'm going to just pretend it's never going to happen for now :)

edd (approved):

dexturr (approved):

datanode/gateway/graphql/resolvers.go
```diff
@@ -1181,11 +1181,30 @@ func (r *myPartyResolver) TransfersConnection(
 	return r.r.transfersConnection(ctx, &party.Id, direction, pagination)
 }
 
-func (r *myPartyResolver) RewardsConnection(ctx context.Context, party *types.Party, assetID *string, pagination *v2.Pagination) (*v2.RewardsConnection, error) {
+func (r *myPartyResolver) RewardsConnection(ctx context.Context, party *types.Party, assetID *string, pagination *v2.Pagination, fromEpoch *int, toEpoch *int) (*v2.RewardsConnection, error) {
```
zale144, line 1184: Maybe if pagination would be the last parameter would be a bit cleaner, although I'd bundled these (party, assetID, fromEpoch, toEpoch) into a single filter struct. No biggie

datanode/sqlstore/rewards_test.go
```diff
@@ -702,7 +671,7 @@ func testRewardsCursorPaginationLastPageBeforeNewestFirst(t *testing.T) {
 	before := entities.NewCursor(entities.RewardCursor{PartyID: partyID, AssetID: assetID, EpochID: 643}.String()).Encode()
 	pagination, err := entities.NewCursorPagination(nil, nil, &last, &before, true)
 	require.NoError(t, err)
-	got, pageInfo, err := rs.GetByCursor(ctx, &partyID, &assetID, pagination)
+	got, pageInfo, err := rs.GetByCursor(ctx, &partyID, &assetID, nil, nil, pagination)
```
zale144, line 674: Yeah, with a filter struct we wouldn't have these `nil`s everywhere :)

zale144 (approved):
Some minor comments on aesthetics, but lgtm

PR #7463, opened 2023-01-30
Network history load and current order set tracking

### LINKED ISSUE #7361: Can we improve startup time when auto-initialising from network history
As part of the process of loading network history into the datanode we run the query below after all data is loaded to ensure that the latest version of an order only is marked with the 'infinity' vegatime.  The reason for this is that is would be possible to load two consecutive history segments (for example) both of which have a given order marked as infinity, so the loading code needs to take account of this after loading segments and update the orders such that only the latest version is market as vegatime='infinity'  (which will bring it into agreement with with other nodes state that are processing events normally).

From testing we see that it takes roughly 1 minute to process an orders table with 3 million rows.   Potentially problematic if a node is behind for whatever reason and set to auto-initialise from network history as it will incur this time in order to get back up to latest height at startup.  

Can we do anything to improve this?  Could we bound the query?  the risk here is that an old order that has only recently changed would not correctly have its vegatime set to infinity.....is there a better way?

WITH vegatimetomapping as (
		SELECT id,
		vega_time,
		COALESCE((LEAD(vega_time) OVER w), 'infinity') as vega_time_to
	FROM orders
	WINDOW w AS (PARTITION BY id order by vega_time))
	UPDATE orders SET vega_time_to=vegatimetomapping.vega_time_to
	FROM vegatimetomapping
	WHERE orders.id=vegatimetomapping.id AND orders.vega_time=vegatimetomapping.vega_time

### LINKED ISSUE #7363: Network history load command no longer indicates progress
Network history load command no longer indicates progress

When running  `data-node network-history load`
Console outputs
`Loading...`

But no indication of how far along it is - previously (IIRC) it would output a line after each segment was loaded - it should do this again

ettec:
closes #7361   closes #7363  - As discussed decision was made not to put the required database changes as a migration, databases will need dropping and recreating.

Fixes the following issues:

1. Datanode's restored from network history can have different state in the order tables due to a bug discovered when load testing (next point) the query that is used to ensure only one order version, the most recent version, is marked as 'current' after loading network history segments.

2. During load testing the query that updates the effective current order version after network history load was found to scale worse then linearly (roughly its look to scale quadratically).  So for example, loading 2 months data would take 4 times as long as loading 1 months data.   To fix this, the way the data-node tracks the current order set has been updated to no longer maintain a chain between an order versions vega_to_time  (information that was is in fact not being used anywhere and if ever needed could be recreated from the vega_time  attribute), and instead use just a current boolean flag.   This simplification means that the query used to update the current order set after network history load can be change to a query that exhibits scaling behaviour significantly better than linear.   

3. When appending new history segments onto exisiting data, the data node was using the same procesure as that used to populated an empty node.  This does not perform well as in the case of an empty indexes and triggers are dropped prior to the load which dramatically speeds up the load time (~10x faster).   However, this comes with a cost in that indexes have to be restored and the current order set has to be recreated after load, compared to the savings on the load itself this makes sense when loading an empty datanode.   However, when a datanode contains a significant amount of data, doing to to append a few new segments does not make sense, in this case, the indexes and triggers should be left in place, this does slow down the segment load but overall is much quicker when adding a small amount of new data to a large amount of existing data.

In order to verify and test the above changes, 2 additional changes were made.  Load logging was fixed (issue 7363) and support was added for loading a subset of network history to make it easy to test the append behaviour (these are the changes related to contiguous history), this functionality is potentially also beneficial for ops.

Rough guides from testing:

When loading data into an empty datanode, it takes just under 30 mins to load 369 segments (~ 3 and 3/4 days)  -> https://docs.google.com/document/d/1OhunF57g9Qm17af7QX7JLMLKe48LSBGGkCnmICVUxts/edit?usp=sharing

When appending network history to a datanode that is already populated, testing shows it taking around ~20s per segment, this will depend of course on the content of the segment. (https://docs.google.com/document/d/1TyuI1eKjet38UoGn-4nlLnduxMM9S3b495jJdXBOQ4M/edit?usp=sharing)

The previous method of updating the orders current set was scaling like so (testnet data), i.e very non-linear, pretty much quadratic: 

3956280 orders takes 9 seconds
33320905 orders takes 1m 36 seconds
66641810  orders takes 10m 54 seconds
133000000 orders takes 37m

The new method scales better than linear by taking advantage of the fact that most of the order data is different versions of the same order.


datanode/gateway/graphql/schema.graphql
```diff
@@ -3927,8 +3927,6 @@ type HistorySegment {
     fromHeight:       Int!
     "To block height of the history segment"
     toHeight: Int!
-    "Chain ID of the history segment"
-    chainID: String!
```
pscott31, line 3931: What's the motivation for removing chainID - not an issue, just curious!

pscott31 (approved):
I can't say I fully grok everything here as there's quite a bit going on but it all seems very sensible

edd (approved):
👍 Can't see any queries depending on this

PR #7425, opened 2023-01-24
Revert periodic resetting of market cache, add settled markets filter

pscott31:
Due to my misunderstanding of the problem this [previous PR](https://github.com/vegaprotocol/vega/pull/7405) did not actually address [the issue](https://github.com/vegaprotocol/vega/issues/7399).

The actual issue here is that the graphql query used to check data-node consistency is something like 

```gql
{markets{data{..}}}
```

However market data is considered a 'history' table, and is part of historical data snapshots created by datanode, not the _current state_ snapshot. So, if you initialise a datanode using only a 'current state' snapshot, and no history, it will have no market data.

This is not usually a problem because core emits a market data event every block, so datanode very quickly is able to answer graphql queries on market data; but for markets that are _settled_, we (very reasonably) do not get any further events from core.

After some discussion it was decided that the best thing to do was to make it possible to only query non-settled markets over graphql to side-step the problem. e.g. now the query vaguer and friends can use to check for up-to-date-ness can be something like

```gql
{markets(includeSettled=false){data{..}}}
```

Really closes #7399 
closes #5729

PR #7478, opened 2023-02-01
feat: fetch only active markets

### LINKED ISSUE #5729: API: Fetch only active markets
Add a property to a market that would allow me to fetch a list of markets, filtered by markets that are currently active.

Active is defined as:

- The market exists
- The market has not expired
- I can place orders on the market (of any type - auction mode is fine, continuous trading is fine, FBA is fine. It might be worth allowing the user to filter by those also.)

# Initial proposal

```graphql
{
   markets (active: true) {
     id
  }
}
```

```
/markets?active=true
```

Note: this would currently be possible by selecting all markets, and then manually filtering by the maturity field on a product. However this will not continue to be true if a market can close for reasons other than maturity (for example liquidity or price finding auctions failed repeatedly to return a market to a tradable state, it will eventually be closed). It’s also just a lot of work when we have more markets being proposed, and longer lived networks.

guoguojin:
Closes #5729


guoguojin:
Closing as @pscott31 has it covered with https://github.com/vegaprotocol/vega/pull/7425

PR #7479, opened 2023-02-01
Verify LP fee transfer to general account

Jiajia-Cui:
according to spec 0042: A network parameter `market.liquidity.providers.fee.distributionTimeStep` will control how often fees are distributed from the LP fee account. Starting with the end of the opening auction the clock starts ticking and then rings every time `market.liquidity.providers.fee.distributionTimeStep` has passed. Every time this happens the balance in this account is transferred to the liquidity provider's margin account for the market. If `market.liquidity.providers.fee.distributionTimeStep` is set to `0` then the balance is distributed either immediately upon collection or at then end of a block.

LP fee is transfers to margin account. 

According to the feature test, LP fee is transfers to general account.


Jiajia-Cui:
This PR was created to clarify where the LP fees are transferred to

PR #7484, opened 2023-02-02
feat: network history information endpoints

### LINKED ISSUE #7473: Add network history peers information to the datanodes info endpoint
When setting up a core node an operator can obtain the information for persistent-peers config via the tendermint 'status' end point.  An equivalent end point is needed in the data-node to source network history peer config.  This should be added to the exisiting 'info' endpoint.

ettec:
closes #7473

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -16,6 +16,8 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math"
+	"math/rand"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -205,7 +207,7 @@ func (t *tradingDataServiceV2) sendAccountsSnapshot(ctx context.Context, req *v2
 	return nil
 }
 
-func (t *tradingDataServiceV2) Info(ctx context.Context, _ *v2.InfoRequest) (*v2.InfoResponse, error) {
+func (t *tradingDataServiceV2) Info(_ context.Context, _ *v2.InfoRequest) (*v2.InfoResponse, error) {
 	return &v2.InfoResponse{
 		Version:    version.Get(),
 		CommitHash: version.GetCommitHash(),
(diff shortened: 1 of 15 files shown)


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2314,6 +2327,31 @@ message GetActiveNetworkHistoryPeerAddressesResponse {
...
+// Response containing the status of network history
+message NetworkHistoryStatusResponse {
+    // The ipfs address of this node
+    string ipfs_address = 1;
+    // The swarm key
+    string swarm_key = 2;
+    // The swarm key seed
+    string swarm_key_seed = 3;
+    // The connected peers
+    repeated string connected_peers = 5;
```
wwestgarth, line 2342: Any reason why we can't just return the bootstrap peers in the status response, instead of it having a dedicated endpoint?

datanode/api/trading_data_v2.go
```diff
@@ -3194,13 +3196,46 @@ func (t *tradingDataServiceV2) GetActiveNetworkHistoryPeerAddresses(_ context.Co
...
 
+func (t *tradingDataServiceV2) NetworkHistoryStatus(_ context.Context, _ *v2.NetworkHistoryStatusRequest) (*v2.NetworkHistoryStatusResponse, error) {
+	connectedPeerAddresses, err := t.networkHistoryService.GetConnectedPeerAddresses()
+	if err != nil {
+		return nil, t.formatE(fmt.Errorf("failed to get connected peer addresses: %w", err))
+	}
+
+	// A subset of the connected peer addresses are likely to be copied to form another nodes peer set, randomise the list
+	// to minimise the chance that the same sub set are copied each time.
+	rand.Shuffle(len(connectedPeerAddresses), func(i, j int) {
```
wwestgarth, line 3213: Not sure I get this. Are people not just going to ping the endpoint and copy the entire list into there new datanode's config?

protos/sources/data-node/api/v2/trading_data.proto
ettec: yeah, I wanted to encourage people not to use the same bootstrap peers set, but have it available if needed, so my thinking was to split this into a seperate call, and have the status display a randomised list of  the current active peers so that there is a better chance that the new nodes use this and don't all end up with the same bootstrap peers -  but at the same time it could be useful information to have available - I did toy with not exposing bootstrap peers at all.

datanode/api/trading_data_v2.go
ettec: true, that may happen, I was optimistically thinking forward to the day when we have lots of peers, and a subset would be selected, just tyring to do what we can to ensure we don't have a dependency  on a central set of nodes.  Perhaps unnecessary.

wwestgarth (approved):

PR #7485, opened 2023-02-02
7450 position details

### LINKED ISSUE #7450: Position API to return close-out reason and loss socialisation status
The Position API should be able to indicate whether a party was involved in loss socialisation, and whether or not the position (e.g. open volume) was updated due to trades with the network, or, if the open volume is closed, we should have a flag indicating whether or not this happened due to the party being liquidated.

EVODelavega:
Closes #7450 

Add close-out to position data, track loss socialisation amounts (although it seems we were already doing so?)


datanode/sqlsubscribers/position.go
```diff
@@ -112,11 +119,29 @@ func (p *Position) Push(ctx context.Context, evt events.Event) error {
...
+	defer p.mutex.Unlock()
+	// @TODO implement this
+	positions, err := p.store.GetByMarketAndParties(ctx, event.MarketID, event.Parties())
+	if err != nil {
+		return errors.Wrap(err, "error getting positions")
+	}
+	for _, pos := range positions {
+		pos.UpdateOrdersClosed()
+		// an update of an existing position here can't really fail
+		_ = p.updatePosition(ctx, pos)
```
ettec, line 140: I get what you're saying but still be better to handle the error and return though, may that a future change results in error? method already returns error

datanode/sqlsubscribers/position.go
```diff
@@ -112,11 +119,29 @@ func (p *Position) Push(ctx context.Context, evt events.Event) error {
...
 	}
 }
 
+func (p *Position) handleOrdersClosedEvent(ctx context.Context, event ordersClosed) error {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	// @TODO implement this
+	positions, err := p.store.GetByMarketAndParties(ctx, event.MarketID, event.Parties())
+	if err != nil {
+		return errors.Wrap(err, "error getting positions")
```
ettec, line 135: we are a bit inconsistent on this so its not always obvious from looking at the code the correct approach is, but the agreement is that we should not use errors.Wrap any more (we should really remove all locations...still), so in this case:
fmt.Errorf("failed to get by market and parties: %w", err) or similar  (and yes its a maintenance PITA - but overall was agreed the best way)

core/execution/market.go
```diff
@@ -1911,15 +1911,22 @@ func (m *Market) resolveClosedOutParties(ctx context.Context, distressedMarginEv
 		// which have acceptable margins
 		okPos, closed = m.risk.ExpectMargins(distressedMarginEvts, m.lastTradedPrice.Clone())
 
+		parties := make([]string, 0, len(okPos))
 		if m.log.GetLevel() == logging.DebugLevel {
```
ettec, line 1915: we only send the event if debug logging is on?  can't be right can it?

ettec (changes requested):

core/execution/market.go
EVODelavega: The send bit is outside of the `if`

datanode/sqlsubscribers/position.go
EVODelavega: Method signature returns an error, but IIRC, it's just adding the entity to a batch, which actually doesn't return an error. I can add a check and return (or ignore?) the error. We should update as many positions as possible, but when that fails, I'm not sure whether we should continue trying

datanode/sqlstore/positions.go
```diff
@@ -70,6 +71,27 @@ func (ps *Positions) GetByMarketAndParty(ctx context.Context,
...
+	bindNum := 2
+	for _, p := range partyIDsRaw {
+		partyIDs = append(partyIDs, entities.PartyID(p))
+		in = append(in, fmt.Sprintf("$%d", bindNum))
+		bindNum++
+	}
+	bind := make([]interface{}, 0, len(in)+1)
+	// set all bind vars
+	bind = append(bind, marketID)
+	bind = append(bind, partyIDs...)
```
pscott31, line 87: We have a nice little helper called `nextBindVar` for this sort of thing

```go
	args := []interface{}{}

	conditions := []string{}
	if marketID.String() != "" {
		conditions = append(conditions, fmt.Sprintf("market_id=%s", nextBindVar(&args, marketID)))
	}
```

it sticks your argument onto an arg list and returns the right `$n` guy

datanode/sqlstore/positions.go
```diff
@@ -70,6 +71,27 @@ func (ps *Positions) GetByMarketAndParty(ctx context.Context,
 		marketID, partyID))
 }
 
+func (ps *Positions) GetByMarketAndParties(ctx context.Context, marketIDRaw string, partyIDsRaw []string) ([]entities.Position, error) {
```
pscott31, line 74: If this is going to be called a lot it might be worth figuring out if we can cache somehow in the service, we want as few SQL calls as possible per block as each one adds a few 10s of ms of latency (hence all this batching and caching and flushing)

datanode/sqlstore/positions.go
EVODelavega: Yeah, I know about that helper, I just figured that having everything here, because I know the `len()` of the party ID's, this would be a bit more efficient, and because I'm using `IN()`, I just decided to do it all in-situ

datanode/sqlstore/positions.go
EVODelavega: I've added this to the service, using it to first retrieve parties from cache, and add missing ones if needed. This isn't going to be called often, though. The only time this can happen is during MTM, and we will, at most, receive this event once per market, every N seconds (when we MTM)

core/events/bus.go
```diff
@@ -202,6 +203,7 @@ var (
 		eventspb.BusEventType_BUS_EVENT_TYPE_TRANSACTION_RESULT:               TransactionResultEvent,
 		eventspb.BusEventType_BUS_EVENT_TYPE_SNAPSHOT_TAKEN:                   CoreSnapshotEvent,
 		eventspb.BusEventType_BUS_EVENT_TYPE_PROTOCOL_UPGRADE_DATA_NODE_READY: ProtocolUpgradeDataNodeReadyEvent,
+		eventspb.BusEventType_BUS_EVENT_TYPE_DISTRESSED_ORDERS_CLOSED:         DistressedOrdersClosedEvent,
```
ze97286, line 206: see line 208 :)

core/execution/market.go
```diff
@@ -1911,15 +1911,23 @@ func (m *Market) resolveClosedOutParties(ctx context.Context, distressedMarginEv
 		// which have acceptable margins
 		okPos, closed = m.risk.ExpectMargins(distressedMarginEvts, m.lastTradedPrice.Clone())
 
+		parties := make([]string, 0, len(okPos))
+		for _, v := range okPos {
+			parties = append(parties, v.Party())
+		}
 		if m.log.GetLevel() == logging.DebugLevel {
```
ze97286, line 1918: IsDebug() would be nicer

core/events/bus.go
EVODelavega: Damn... Can't believe I missed that, will add now

core/execution/market.go
```diff
@@ -1911,15 +1911,23 @@ func (m *Market) resolveClosedOutParties(ctx context.Context, distressedMarginEv
...
-			for _, v := range okPos {
+			for _, pID := range parties {
 				if m.log.GetLevel() == logging.DebugLevel {
```
ze97286, line 1920: this isn't needed

EVODelavega: That's what was already there. I only changed this because I added the loop outside the `if` creating a slice of party ID's, and used that slice here to not repeatedly call `v.Party()`

datanode/sqlsubscribers/position.go
```diff
@@ -112,11 +120,32 @@ func (p *Position) Push(ctx context.Context, evt events.Event) error {
...
+		return p.handleOrdersClosedEvent(ctx, event)
 	default:
 		return errors.Errorf("unknown event type %s", evt.Type().String())
 	}
 }
 
+func (p *Position) handleOrdersClosedEvent(ctx context.Context, event ordersClosed) error {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	// @TODO implement this
```
ze97286, line 133: implement what?

protos/sources/vega/events/v1/events.proto
```diff
@@ -420,6 +420,15 @@ message SettleDistressed {
   string price = 4;
 }
 
+// A Distressed order contains the party ID's for all parties that were distressed, but had their orders closed,
+// but did not need to be closed out after cancelling their orders
```
candida-d, line 424:
```suggestion
// A distressed order contains the party IDs for all parties that were distressed and had orders closed,
// but did not need to be closed out after cancelling their orders
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -2157,6 +2167,12 @@ type Position {
   "Average entry price for this position"
   averageEntryPrice: String!
 
+  "The total amount of profit and Loss that was not transferred due to loss socialization"
```
candida-d, line 2170:
```suggestion
  "The total amount of profit and loss that was not transferred due to loss socialisation"
```

datanode/sqlsubscribers/position.go
EVODelavega: removed

core/execution/market.go
EVODelavega: Removed the inner log-level check, and replaced the outer one with `IsDebug()`

core/events/bus.go
EVODelavega: done

ze97286 (dismissed):

datanode/sqlsubscribers/position.go
```diff
@@ -112,11 +120,31 @@ func (p *Position) Push(ctx context.Context, evt events.Event) error {
...
+func (p *Position) handleOrdersClosedEvent(ctx context.Context, event ordersClosed) error {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	positions, err := p.store.GetByMarketAndParties(ctx, event.MarketID(), event.Parties())
+	if err != nil {
+		return fmt.Errorf("failed to get positions: %w", err)
+	}
+	for _, pos := range positions {
+		pos.UpdateOrdersClosed()
+		// this update can't really fail, the entity is just added to the batch, but add the check
```
ettec, line 139: remove the comment?  doesn't add anything

core/execution/market.go
ettec: right but the parties was only added to if debug logging was on, but I think you've fixed this now?

ettec (approved):

datanode/gateway/graphql/schema.graphql
```diff
@@ -2157,6 +2167,12 @@ type Position {
   "Average entry price for this position"
   averageEntryPrice: String!
 
+  "The total amount of profit and loss that was not transferred due to loss socialisation"
+  lossSocializationAmount: String!
+
+  "Enum set if the position was closed out or orders were removed because party was distressed"
+  positionStatus: PositionStatus!
```
mattrussell36, line 2174: suggestion (non blocking): would prefer this as just `status` the code will read nicer as a result: `position.status` rather than `position.positionStatus`

mattrussell36 (approved):

PR #7486, opened 2023-02-02
fix: 7469 sanitise prometheus request type label

### LINKED ISSUE #7469: Sanitize Prometheus requestType label
https://github.com/vegaprotocol/vega/blob/develop/core/metrics/prometheus.go#L534

`apiRequestCallCounter` and `apiRequestTimeCounter` are used with unsanitized request string from any http call used on the service. This leads to potentially extremely high number of time-series created by bad actors (one per label, ie one per path called - even non-existent) and can slow-down, break or incur very high costs on the Prometheus host.

Solution is to only instrument existing routes, optionally instrument non-existing routes into a single label.

guoguojin:
Closes #7469


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - [6963](https://github.com/vegaprotocol/vega/issues/6963) - Remove the legacy fields from network API
 - [7361](https://github.com/vegaprotocol/vega/issues/7361) - Network history loading and current order set tracking - database requires database to be dropped
 - [6963](https://github.com/vegaprotocol/vega/issues/7382) - `IssueSignatures` is no longer a validator command and is now protected by the spam engine
+- [7361](https://github.com/vegaprotocol/vega/issues/7361) - Network history loading and current order set tracking - database requires database to be dropped
+
```
pscott31, line 11: I think this is from another commit :)

pscott31 (dismissed):

jeremyletang (dismissed):

wwestgarth (approved):

PR #7507, opened 2023-02-06
Verify balances vs transfers

### LINKED ISSUE #7508: Transfer events not always sent after margin recheck
A loop returning early may result in transfer events not being sent if the first few transfers succeed, but the `n`-th one fails:
https://github.com/vegaprotocol/vega/blob/8d76c5e6865fad12cdc497189285121e9eb2655b/core/execution/collateral.go#LL93C5-L93C5

### LINKED ISSUE #7509: Automatically reconcile account balance changes with transfer events after each integration test step

witgaw:
After each integration step: reconcile account balance changes over the step with the emitted transfer events, 
tidy up step naming (`Given the initial insurance pool balance is "xxxx" for all the markets`),
remove redundant step (`deposit_to_reward_account.go`),
bugfix

Closes #7509 
close #7508 .


witgaw (commented):
Ready for review

core/execution/collateral.go
```diff
@@ -86,11 +86,14 @@ func (m *Market) transferRecheckMargins(ctx context.Context, risk []events.Risk)
...
+	var err error
+	var closed events.Margin
 	for _, r := range risk {
+		var tr *types.LedgerMovement
 		responses := make([]*types.LedgerMovement, 0, 1)
-		tr, closed, err := m.collateral.MarginUpdateOnOrder(ctx, mID, r)
+		tr, closed, err = m.collateral.MarginUpdateOnOrder(ctx, mID, r)
 		if err != nil {
-			return err
+			break
```
witgaw, line 96: @vegaprotocol/core probably worth continuing here instead and logging the error somehow. Any thoughts? Should I create something like:
```
errors.New("one or more margin updates failed during margin recheck" 
```
and hide away the details coming in the error returned by `.MarginUpdateOnOrder` or somehow combine the errors? (or leave as is?)

EVODelavega: Hmm, I need to go back through everything I was doing when adding this function, but I believe that the point at which this is called, it technically shouldn't be possible for errors occurring here (ie balances should be fine). That would mean that, if the collateral engine returns an error, that means that we're dealing with margin transfer requests for parties who don't exist in the collateral engine, and things are going quite dramatically wrong.

witgaw: OK, that may well be the case given there's a distressed party there. I can take a look and try to get to the bottom of it.

witgaw: @EVODelavega I'd be keen to get this merged as is (it's same functionality, just reports the event correctly now) and deal with the rest as part of this, what do you think: https://github.com/vegaprotocol/vega/issues/7543 ?

EVODelavega (dismissed):

core/execution/collateral.go
```diff
@@ -80,17 +80,21 @@ func (m *Market) transferMarginsAuction(ctx context.Context, risk []events.Risk,
...
 	for _, r := range risk {
+		var tr *types.LedgerMovement
 		responses := make([]*types.LedgerMovement, 0, 1)
 		tr, closed, err := m.collateral.MarginUpdateOnOrder(ctx, mID, r)
 		if err != nil {
-			return err
+			m.log.Warn("margin recheck failed",
+				logging.MarketID(m.GetID()),
+				logging.PartyID(r.Party()),
+				logging.Error(err))
```
witgaw, line 97: @EVODelavega please confirm that you're happy with this. This seemed like a best choice to me as it still logs a warning (possibly more than one now, before it was an umbrella warning here https://github.com/vegaprotocol/vega/blob/develop/core/execution/market.go#L478C10-L484.

Could also just `log.Debug` if `closed!=nil` (it just means somone's marked for closeout but it cannot be done yet as there's not enough volume, so nothing sinister) and leave `Warn` for other cases (there shouldn't be any).

EVODelavega: yeah, that's fair enough

EVODelavega (approved):

davidsiska-vega (approved):

PR #7520, opened 2023-02-07
Memory leak and CPU spikes on data streams

### LINKED ISSUE #7519: Datanode uses resources to handle closed graphql subscriptions
If we repeatedly create graphQL subscriptions to the datanode and then close them, over time we generate an increased idle CPU usage in the datanode.

Steps:
* Spin us a new network using Vega Capsule with at least one data node.
* Create a new market.
* Record the normal idle CPU usage of the datanode.
* In a script, use the consoleLoadTest application to pretend to be 500 consoles and then kill the tool after 5 seconds.
* Repeat for around 10 minutes.
* Check that the CPU usage of the datanode is close to the original idle.

EVODelavega:
Fixes #7519 

Tickers need a `Stop` call or they leak memory, added some checks on the request and stream contexts on graphQL subscriptions in case the connection was closed between receiving and sending the data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - [7468](https://github.com/vegaprotocol/vega/issues/7468) - `Datanode` network history load command only prompts when run from a terminal
 - [7164](https://github.com/vegaprotocol/vega/issues/7164) - The command `vega wallet transaction send` now returns verbose errors
 - [7514](https://github.com/vegaprotocol/vega/issues/7514) - Network names cannot contain `/`, `\` or start with a `.`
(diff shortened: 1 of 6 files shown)


pscott31 (commented):
Nice find.. though is this liable to fill our logs up with ('...stream closed')

peterbarrow:
I have run the tests for around 3 hours (which created around 30M subscriptions) and once the client app had stopped, the idle CPU for the datanode returned to ~1% which is the same value as I recorded when I first started the network. This bug is no longer here.

peterbarrow (dismissed):

peterbarrow (approved):

PR #7524, opened 2023-02-07
fix: data-node tests fail intermittently

guoguojin:
This fixes intermittent database issues when running the integration tests. From my investigations, it looks like the retention policies may be causing database chunks to be cleared out while data is loading into the database. After changing the data-node mode to archive for integration tests, thus removing all retention policies, I managed to complete 150 iterations of the integration tests without any failures.

I think the issue may be related to the version of TimescaleDB that is currently deployed to the embedded postgres that is used for the integration and unit tests. We will need to update the TimescaleDB libraries to a later version and investigate further, but for now, this should stabilise the integration tests and stop them from intermittently failing.


wwestgarth (approved):

ettec:
As discussed, issue here is that the data used is old, therefore retention policies will clean up the data, but it will be a bit 'random' as retention policies are run in an unpredicatable  way.  So don't think an issue with timescale.

PR #7526, opened 2023-02-07
fix: 7494 memory leak in event bus stream subscriber when consumer is…

### LINKED ISSUE #7494: Memory Leak in event bus stream subscriber
FYI related to the main issue:
core/subscribers/stream_subscriber.go is a duplicate of datanode/subscribers/stream_subscriber.go
and
datanode/subscribers/service.go  and  core/core/subscribers/service.go
are duplicates of each, so suffer from the same bug, we may want to consolidate these as part of this fix,unless there is good reason to have the duplicates?  but can't see any.

the service.go has a memory leak that manifest itself if a client subscribes but is a slow consumer, see the printf in the code below, the datalength2 (and datalength1) has been observer to grow continuously is this scenario, we should set an allowed upper limit, beyond which the connection is closed.  Based on observeration, 10000 is probably a reasonable default.   Note, do consider if this limit makes sense for the cores version of the service as well (I think it does), if so consolidate the code.

`for {
			select {
			case <-ctx.Done():
				return
			case bs := <-in:
				// batch size changed: drain buffer and send data
				data = append(data, sub.UpdateBatchSize(ctx, bs)...)
				datalength1 := len(data)

				if datalength1 > lastLoggedDataLength+10000 {
					lastLoggedDataLength = datalength1
					fmt.Printf("SUBSCRIBER %s, DATALENGTH 1: %d\n", sub.StreamID, lastLoggedDataLength)
				}

				if datalength1 > 0 {
					trySend()
				}
			default:
				// wait for actual changes
				data = append(data, sub.GetData(ctx)...)

				datalength2 := len(data)
				if datalength2 > lastLoggedDataLength+10000 {
					lastLoggedDataLength = datalength2
					fmt.Printf("SUBSCRIBER %s, DATALENGTH 2: %d\n", sub.StreamID, lastLoggedDataLength)
				}

				// this is a very rare thing, but it can happen
				if len(data) > 0 {
					trySend()
				}
			}
		}`

ettec:
closes #7494 fixes the memory leak cause by slow consumer of the event bus and deduplicates the stream and stream subscription code into a 'shared' area (better place/name?)


shared/subscribers/service.go
```diff
@@ -17,32 +17,47 @@ import (
 	"time"
```
ettec, line 18: better place than the shared package for this stuff?

core/broker/config.go
```diff
@@ -23,9 +23,10 @@ const namedLogger = "broker"
 
 // Config represents the configuration of the broker.
 type Config struct {
-	Level  encoding.LogLevel `long:"log-level"`
-	Socket SocketConfig      `group:"Socket" namespace:"socket"`
-	File   FileConfig        `group:"File" namespace:"file"`
+	Level                    encoding.LogLevel `long:"log-level"`
+	Socket                   SocketConfig      `group:"Socket" namespace:"socket"`
+	File                     FileConfig        `group:"File" namespace:"file"`
+	EventBusClientBufferSize int               `long:"event-bus-client-buffer-size"`
```
ettec, line 29: Seemed like the most appropriate place for this, but maybe not?

core/broker/config.go
```diff
@@ -47,6 +48,7 @@ func NewDefaultConfig() Config {
 		File: FileConfig{
 			Enabled: false,
 		},
+		EventBusClientBufferSize: 100000,
```
ettec, line 51: I've made the default for core bigger than that for the datanode,  10,000 events per sec perfectly reasonable, so even with buffer etc we shoujld allow some headroom (GC pause whatever) -  maybe the default should be bigger?

core/broker/broker.go
```diff
@@ -18,25 +18,12 @@ import (
...
+
 	"code.vegaprotocol.io/vega/core/events"
 	"code.vegaprotocol.io/vega/logging"
 )
 
-// Subscriber interface allows pushing values to subscribers, can be set to
-// a Skip state (temporarily not receiving any events), or closed. Otherwise events are pushed
-//
-//go:generate go run github.com/golang/mock/mockgen -destination mocks/subscriber_mock.go -package mocks code.vegaprotocol.io/vega/core/broker Subscriber
-type Subscriber interface {
```
EVODelavega, line 29: Interfaces should be declared in the package that uses/relies on them. By moving this to a shared package, the broker in datanode and core become tightly coupled in a way that they shouldn't be. Let's say that we want to change snapshots in core to work asynchronously, using subscribers, and want to change the interface, that should not impact the interface on the data node side of things.
The broker interface so far was the only exception, and we only made that change after having discussed the matter rather extensively.

IMO, because the broker in data node and core are separate entities, and the role of their subscribers may diverge in the future, I don't think this shared package is a good idea.

shared/subscribers/service.go
EVODelavega: as mentioned earlier: not having the shared package

shared/subscribers/service_test.go
```diff
@@ -0,0 +1,148 @@
...
+		}).StreamMessage(),
+		events.NewAccountEvent(ctx, dtypes.Account{
+			ID: "acc-6",
+		}).StreamMessage(),
+	}
+
+loop:
+	for {
+		select {
+		case <-serviceContext.Done():
```
EVODelavega, line 72: This looks like a really elaborate way to write the equivalent of:

```
<-ctx.Done()
// or
for _ := range <-ctx.Done() {
    break
}
```

don't really see a point to the `loop` label either

shared/subscribers/stream_subscriber.go
```diff
@@ -32,6 +32,20 @@ type MarketStreamEvent interface {
 	StreamMarketMessage() *eventspb.BusEvent
 }
 
+type StreamSubscription interface {
```
EVODelavega, line 35: Why have this interface here?

shared/subscribers/stream_subscriber.go
```diff
@@ -80,7 +94,7 @@ func getBufSize(batch int, types []events.Type) int {
 	return base
 }
 
-func NewStreamSub(ctx context.Context, types []events.Type, batchSize int, filters ...EventFilter) *StreamSub {
+func NewStreamSub(ctx context.Context, types []events.Type, batchSize int, filters ...EventFilter) StreamSubscription {
```
EVODelavega, line 97: This is considered bad practice in golang. A function should be as liberal as possible in terms of its returns, but as restrictive as possible in terms of what it accepts as arguments. Returning the underlying type as in interface defined in this package itself is doing the opposite

shared/subscribers/stream_subscriber_test.go
```diff
@@ -20,19 +20,20 @@ import (
...
 	eventspb "code.vegaprotocol.io/vega/protos/vega/events/v1"
 
 	"github.com/stretchr/testify/assert"
 )
 
 type tstStreamSub struct {
-	*subscribers.StreamSub
-	ctx   context.Context
-	cfunc context.CancelFunc
+	subscription subscribers.StreamSubscription
```
EVODelavega, line 34: Why remove the embedding?

EVODelavega (changes requested):

shared/subscribers/service.go
```diff
@@ -82,16 +97,30 @@ func (s *Service) ObserveEvents(ctx context.Context, retries int, eTypes []event
...
 				// batch size changed: drain buffer and send data
 				data = append(data, sub.UpdateBatchSize(ctx, bs)...)
-				if len(data) > 0 {
+				dataLength := len(data)
+
+				if dataLength > 0 {
 					trySend()
 				}
+
+				if dataLength > s.maxBufferSize {
```
EVODelavega, line 106: What if `trySend` actually managed to send the data? It makes more sense to me to check `dataLength` _before_ trying to send

shared/subscribers/service.go
```diff
@@ -82,16 +97,30 @@ func (s *Service) ObserveEvents(ctx context.Context, retries int, eTypes []event
...
 				data = append(data, sub.GetData(ctx)...)
+				dataLength := len(data)
+
 				// this is a very rare thing, but it can happen
-				if len(data) > 0 {
+				if dataLength > 0 {
 					trySend()
 				}
+
+				if dataLength > s.maxBufferSize {
```
EVODelavega, line 120: Same here

EVODelavega (changes requested):

core/broker/broker.go
ettec: can you take that up with @jeremyletang, questioned was asked and  agreed they should be consolidated (see original issue)

shared/subscribers/service_test.go
ettec: ah yeah, sorted

shared/subscribers/service.go
ettec: if it can send the data then by definition its not a slow consumer, so checking before doesn't make sense

shared/subscribers/service.go
ettec: if it can send the data then by definition its not a slow consumer, so checking before doesn't make sense

shared/subscribers/service_test.go
```diff
@@ -0,0 +1,148 @@
...
+		[]*eventspb.BusEvent{
+			events.NewAccountEvent(ctx, dtypes.Account{ID: "acc-3"}).StreamMessage(),
+			events.NewAccountEvent(ctx, dtypes.Account{ID: "acc-4"}).StreamMessage(),
+			events.NewAccountEvent(ctx, dtypes.Account{ID: "acc-5"}).StreamMessage(),
+			events.NewAccountEvent(ctx, dtypes.Account{ID: "acc-6"}).StreamMessage(),
+		},
+		events1)
+
+	// We expect this channel to close
+	for range out {
```
wwestgarth, line 82: If the channel doesn't get closed, this is going to hang.

shared/subscribers/service.go
ettec: see previous comment

shared/subscribers/service.go
ettec: fair point, switched them around, though if we get that close to the buffer limit pretty much nailed on that the consume is slow

shared/subscribers/service.go
ettec: switched them

shared/subscribers/stream_subscriber.go
ettec: oops, should be service.go, moved it there

shared/subscribers/stream_subscriber.go
ettec: it was due needing the internal context of the event_observer for the test (need to bend the black box testing rules cause of the threading) , which ultimately required this change  however just realised the context is not needed can just rely on the out channel closing, this has been moved back to *StreamSub

shared/subscribers/stream_subscriber_test.go
ettec: see previous comment

shared/subscribers/service_test.go
ettec: true, it should always close, if not we have bug and we'll know about it cause the test times out, does it need more?  it should never happen

wwestgarth: The timeout is 30 whole minutes, and then when it does timeout it won't tell you why

Anyway I think the test could be better if we `EXPECT()` on the connection being unsubscribed. Its also a bit confusing and not accurate that the subscriber _is_ reading quick enough but we just artificially send too much into the channel. Something like this would be better:
```
// send in 4 things, one sitting waiting in the out channel and 3 in the data so we're at buffer max
	for i := 0; i < 4; i++ {
		testStreamSubscription.events <- []*eventspb.BusEvent{
			events.NewAccountEvent(ctx, dtypes.Account{
				ID: "acc-1",
			}).StreamMessage(),
		}
	}

	// this one will fill up the buffer too much causes unsubscription from the broker
	broker.EXPECT().Unsubscribe(gomock.Any()).Times(1)
	testStreamSubscription.events <- []*eventspb.BusEvent{
		events.NewAccountEvent(ctx, dtypes.Account{
			ID: "acc-3",
		}).StreamMessage(),
	}
	<-serviceContext.Done()
```

ettec: btw you do get the stack on timeout:


goroutine 18 [sleep]:
time.Sleep(0x22ecb25c000)
	/usr/local/go/src/runtime/time.go:195 +0x135
code.vegaprotocol.io/vega/shared/subscribers.TestSlowConsumerIsDisconnected(0xc000291860?)
	/home/matthewpendrey/projects/vegadevelop/vega/shared/subscribers/service_test.go:66 +0xf17
testing.tRunner(0xc000291860, 0xde4310)
	/usr/local/go/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1493 +0x35f
FAIL	code.vegaprotocol.io/vega/shared/subscribers	60.070s
?   	code.vegaprotocol.io/vega/shared/subscribers/mocks	[no test files]
FAIL

wwestgarth: Then do this:
```
wait, cancel := context.WithCancel(context.Background())
	broker.EXPECT().Unsubscribe(gomock.Any()).Times(1).Do(
		func(_ interface{}) {
			cancel()
		},
	)
	testStreamSubscription.events <- []*eventspb.BusEvent{
		events.NewAccountEvent(ctx, dtypes.Account{
			ID: "acc-3",
		}).StreamMessage(),
	}

	select {
	case <-wait.Done():
	case <-time.After(2 * time.Second):
		assert.Fail(t, "broker never unsubscribed")
	}
```

ettec: think that's at risk of intermittent timing failure under race test isn't ?  the original seems more robust

shared/subscribers/service.go
ettec: as discussed  'We should follow "the go proverbs", specifically: a little bit of copying is better than a a little dependency.'    is fine for trivial bits of code, but this is clearly not that, and is actually a perfect example of why we should not duplicate code willy nilly, it was pure chance that I stumbled across the duplicate code, had I not we would have been in a situation where the mem leak would have been fixed in datanode but not in core.

core/broker/broker.go
ettec: as discussed the reason given against this change, i.e. -> 'We should follow "the go proverbs", specifically: a little bit of copying is better than a a little dependency.' is fine for trivial bits of code, but this is clearly not that, and is actually a perfect example of why we should not duplicate code willy nilly, it was pure chance that I stumbled across the duplicate code, had I not we would have been in a situation where the mem leak would have been fixed in datanode but not in core.

jeremyletang (approved):

shared/subscribers/service_test.go
wwestgarth: How?

PR #7545, opened 2023-02-08
feat: add API rate limiting

### LINKED ISSUE #7445: Datanode API rate limiting
As of today, the datanode API do rate limite only the number of subscribtion open per IPs. Unary ueries (graphql, rest or GRPC) are not rate limited.

We want to implement rate limiting on these 3 API per IP address. Because the graphl API uses the grpc API, and because it is sligltly more difficult to control the amount of request done by the graphql API to the grpc API, the rate limiting needs to be accounted separately for graphql, and grpc endpoint (to be set in the configuration. e.g: 10 graphql request per second , 20 grpc request per second).

In order for a request to not be rate limited twice, all request to the grpc API coming from the graphql layer will need to embed a metadata saying so, all these request are then excluded from the grpc rate limiting.

A user, which break the maximum number of request per seconds will be rate limited. A breach is valid as soon as in a second:
`(number graphql request for ip / max graphql request per second + number of grpc request fro ip / max grpc request per second) > 1`

A user breaching the rate limit, is banned exponentially applicable immediately. All following request are rejected with an appropriate error message. Once the ban period is lifted, for the duration of the previous ban period any breach will of the rate limite will provoke a ban period of an increase duration.
Example:
- Rate limit breached at 1PM: initial ban period of 10 minutes
- User can start using the API again at 1:10PM.
-  until 1:20PM, any new breach will incure a ban of an extended time of 1h instead of 10 minutes, etc.
-  after 1:20PM, any breach wil incure a ban of 10 minutes again.

The APIs should also return in some ways (headers, or metada for graphql) the amount of request allowed by the client in the current rate limited scope.

The rate limiting should be optional on the datanode, and configured by the operator to eventually be totally disabled.

pscott31:
This is possibly (probably) breaking for frontend & system tests - though it's easily disabled in the config.

I wrote a little doc on the implementation [here](https://github.com/vegaprotocol/vega/blob/develop/datanode/ratelimit/README.md). The executive summary is that graphql, grpc, and REST requests will by default,

- allow only 20 API calls/second on average 
- with a burst of up to 100 api calls in a short period of time

If you go over that, you'll get an appropriate error response. There are HTTP headers and gRPC metadata showing you after each request how close to the limit you are.

If the client continues making requests such that they are getting > 10 rate limited _error responses_ per second they will eventually get banned for a period of time (default 10 mins).

Pulled the defaults out of my bottom somewhat, very open to discussion on changing them.

Closes #7445

### DIFF
--- a/datanode/ratelimit/naughtystep.go
+++ b/datanode/ratelimit/naughtystep.go
@@ -1,0 +1,92 @@
+package ratelimit
+
+import (
+	"sync"
+	"time"
+
+	"code.vegaprotocol.io/vega/logging"
+	"github.com/didip/tollbooth/v7"
+	"github.com/didip/tollbooth/v7/limiter"
+	"go.uber.org/zap"
+)
+
+// naughtyStep is a struct for keeping track of bad behavior and bans.
+//
+// You get put on the naughty step if you make requests despite having run out of tokens.
+// The naughty step has it's own rate limiter, and its tokens are spent every time a failed
+// (due to rate limiting) API call is made. If you run out of naughty tokens, then you get
+// banned for a period of time.
+
+type naughtyStep struct {
+	log    *logging.Logger
+	lmt    *limiter.Limiter
+	bans   map[string]time.Time
+	mu     sync.RWMutex
+	banFor time.Duration
+}
+
+func newNaughtyStep(log *logging.Logger, rate float64, burst int, banFor, pruneEvery time.Duration) *naughtyStep {
+	limitOpts := limiter.ExpirableOptions{DefaultExpirationTTL: pruneEvery}
+	lmt := tollbooth.NewLimiter(rate, &limitOpts)
+	lmt.SetBurst(burst)
+
+	ns := naughtyStep{
+		log:    log,
+		lmt:    lmt,
+		bans:   make(map[string]time.Time),
+		banFor: banFor,
+	}
+
+	go func() {
+		for range time.Tick(pruneEvery) {
+			ns.prune()
+		}
+	}()
+
+	return &ns
+}
+
+func (n *naughtyStep) enabled() bool {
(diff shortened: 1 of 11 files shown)


pscott31:
{  
"system-tests": "disable_ratelimit",  
}

datanode/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,196 @@
...
+	ip = libstring.CanonicalizeIP(ip)
+
+	// Check the naughty step
+	if r.naughtyStep.isBanned(ip) {
+		return nil, status.Error(codes.Unavailable, banMsg)
+	}
+
+	setRateLimitXResponseHeaders(ctx, r.log, r.lmt, ip)
+	if r.lmt.LimitReached(ip) {
+		r.naughtyStep.smackBottom(ip)
```
ettec, line 160: isn't smacking banned these days?

datanode/ratelimit/naughtystep.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	if n.lmt.LimitReached(ip) {
+		n.ban(ip)
+		n.log.Info("banned for requesting past rate limit", zap.String("ip", ip))
+	}
+}
+
+func (n *naughtyStep) ban(ip string) {
+	n.mu.Lock()
```
ettec, line 65: takeAwayIPad would be a better name ;-)

pscott31: ha yes it would I am actually tempted to change it :)

datanode/ratelimit/ratelimit.go
pscott31: I guess datanode could ban itself if it smacks too often :)

ValentinTrinque (approved):

PR #7546, opened 2023-02-08
feat: add support for vega and data node sync restart

### LINKED ISSUE #7503: Use correct snapshot when restarting with data-node
# Feature Overview

Can we change the behaviour of the default value of `-1` of `--snapshot.load-from-block-height`, to use `latest snapshot compatible with data-node`.
`data-node` can be behind/at a lower block than the latest vega core snapshot. Trying to restart with default value `-1` (i.e. use the latest snapshot) doesn't make sense, cos it will 100% fail.
This is problematic for:
- automatic restart by `vegavisor`,
- manual restart, cos it requires extra dummy work: figure out block by running CLI commands manually, modify startup arguments, and after starting the manual cleanup of startup arguments,

# Details
It can happen that `data-node` gets behind `vega` core. I.e. latest block of vega core is higher than latest block of data-node.
It can also happen that `snapshot` is created in between the blocks. i.e. `vega` core block is higher than the latest snapshot, and `data-node` latest block is below/behind the latest snapshot.
If we stop the node in such conditions, restarting it will require manual intervention.

# Problem encountered
It happens that `data-node` can struggle because of external conditions and get behind vega core. Further, it can be forced to restart. But the automatic restart can fail because of the above problem.

**One of the main reasons for having a visor is to auto-restart the node, which stopped.**

# Observed behaviour
It happens from time to time on Fairground and requires manual intervention from Ops team. The last time it happened was on Friday/Sunday night.

# Expected behaviour
`visor` should be able to restart the node successfully without manual config changes.

# Definition of Done

**Before Merging**

**After Merging**

karlem:
close #7503 .


cmd/data-node/commands/networkhistory/latest_history_segment.go
```diff
@@ -56,47 +45,36 @@ func (cmd *latestHistorySegment) Execute(_ []string) error {
...
+	networkHistoryStore, err := store.New(ctx, log, cmd.Config.ChainID, cmd.Config.NetworkHistory.Store, vegaPaths.StatePathFor(paths.DataNodeNetworkHistoryHome), false)
 	if err != nil {
-		handleErr(log, cmd.Output.IsJSON(), "failed to get datanode client", err)
+		handleErr(log, cmd.Output.IsJSON(), "failed to create network history store", err)
 		os.Exit(1)
 	}
-	defer func() { _ = conn.Close() }()
 
-	resp, err := client.GetActiveNetworkHistoryPeerAddresses(context.Background(), &v2.GetActiveNetworkHistoryPeerAddressesRequest{})
+	segments, err := networkHistoryStore.ListAllIndexEntriesOldestFirst()
```
karlem, line 57: @ettec Hey, I think there is a slight problem with using the store directly when the node is running. It seems like the file is being used by node and it's locked somehow.

jeremyletang (approved):

PR #7548, opened 2023-02-09
feat: cap slippage component in margin calculation

### LINKED ISSUE #7542: Cap slippage using market max slippage fraction
# Feature Overview

To address the issue where LP has been liquidated when the order book was empty save a far from mid offer order which made the slippage component of the margin calculation massive we want to cap the slippage component. 

## Specs
- [https://github.com/vegaprotocol/specs/pull/1583](xyz) spec PR.

ze97286:
Closes #7542


core/integration/steps/the_markets.go
```diff
@@ -516,9 +534,39 @@ func (r marketRow) lpPriceRange() float64 {
 	return r.row.MustF64("lp price range")
 }
 
+func (r marketRow) linearSlippageFactor() float64 {
+	if !r.row.HasColumn("linear slippage factor") {
+		// set to 1e6 by default
+		return 1000000
```
davidsiska-vega, line 540: I wouldn't set it this way to 10^6 if the "real" default is 0.1 I would rather add a line to every feature test this impacts to set it "large" there. 

Otherwise, it may be confusing for ppl who don't look here and just look at the default value and then can't figure out maintenance margin numbers.

davidsiska-vega (changes requested):
One comment above, you may disagree...

core/integration/steps/the_markets.go
ze97286: I don't think this should hold off this PR or even be done as part of it even if I agree it makes sense - in case we want to tag and push it to testnet. This change though mechanic will include 10s of FT files.

ze97286: @davidsiska-vega this would add 152 FTs to this PR - I don't mind, but I don't think it's a good idea.

davidsiska-vega: Ok I agree, let's do that separately.

davidsiska-vega (approved):

core/risk/engine.go
```diff
@@ -188,7 +194,7 @@ func (e *Engine) UpdateMarginAuction(ctx context.Context, evts []events.Margin,
 	// for now, we can assume a single asset for all events
 	rFactors := *e.factors
 	for _, evt := range evts {
-		levels := e.calculateAuctionMargins(evt, price, rFactors)
+		levels := e.calculateAuctionMargins(evt, price, rFactors, e.linearSlippageFactor, e.quadraticSlippageFactor)
```
zale144, line 197: I could be wrong about this, but do we have to pass `e.linearSlippageFactor, e.quadraticSlippageFactor` here, if they are accessible through the receiver? Can these parameters have different values?

jeremyletang (approved):

core/types/market.go
```diff
@@ -637,6 +639,8 @@ type Market struct {
 func MarketFromProto(mkt *proto.Market) (*Market, error) {
 	asset, _ := mkt.GetAsset()
 	lppr, _ := num.DecimalFromString(mkt.LpPriceRange)
+	linearSlippageFactor, _ := num.DecimalFromString(mkt.LinearSlippageFactor)
+	quadraticSlippageFactor, _ := num.DecimalFromString(mkt.QuadraticSlippageFactor)
```
zale144, line 643: Shouldn't we handle these errors?

zale144 (approved):
Nice work! Just a couple of minor comments

core/types/market.go
ze97286: no. it's not possible for this error to occur here.

core/risk/engine.go
ze97286: you're right but if you don't mind I'll fix it in a followup pr

protos/sources/vega/governance.proto
```diff
@@ -59,6 +59,10 @@ message NewMarketConfiguration {
   // Percentage move up and down from the mid price which specifies the range of 
   // price levels over which automated liquidity provision orders will be deployed
   string lp_price_range = 8;
+  // Linear slippage factor is used to cap the slippage component of maintainence margin - it is applied to the slippage volume
```
edd, line 62: **thought (non-blocking)**: Why do the protos get better comments than the GraphQL fields? 😢

edd (approved):

protos/sources/vega/governance.proto
ze97286: I'll copy them over on the followup PR. thanks.

PR #7552, opened 2023-02-09
chore: update feature tests with linear and quadratic slippage factor…

### LINKED ISSUE #7550: Feature test runner to use core defaults for slippage cap
In https://github.com/vegaprotocol/vega/pull/7548 we are adding cap on the slippage comp. of maintenance margin. 

In that PR the default for feature tests is being set to `10^6` to avoid having to modify all the feature test files impacted. 

To close this ticket:
- [ ] add feature test runner functionality to specify custom slippage parameters per market (if it doesn't already exist)
- [ ] add an override of the default to every feature test; thus making the tests pass. The override should be as small as possible to make the test pass so ideally less than `10^6` 
- [ ] finally remove the override of the default value set in `core/integration/steps/the_markets.go` around line `537`.

ze97286:
…s and change default to 0.1

Closes #7550


jeremyletang (approved):

edd (approved):

Jiajia-Cui (approved):
LGTM

PR #7554, opened 2023-02-09
Remove redundant margin update for all parties during auction

witgaw:
Updating margins for all parties after every order during auction had no material impact on the margin figures:
  - during opening auction nobody has a position, so only the bit that calculates margin based on vwap of orders placed by a party is relevant (and that's not affect by other orders in any way)
  - during other auction types the vwap bit from above was the same, and for the position part we always used the `lastTradedPrice`, which does not change during auction. 

but it was generating a lot of events.

Also: removed the hack preventing margin release during auction and replaced it with a conditional statement based on the auction state.

### DIFF
--- a/core/integration/features/auctions/opening-auction-price.feature
+++ b/core/integration/features/auctions/opening-auction-price.feature
@@ -65,8 +65,8 @@ Feature: Set up a market, create indiciative price different to actual opening a
       | party7 | ETH/DEC19 | sell | 1      | 11000 | 0                | TYPE_LIMIT | TIF_GFA | t7-s-1    |
     And the parties should have the following margin levels:
       | party  | market id | maintenance | search | initial | release |
-      | party1 | ETH/DEC19 | 11200       | 12320  | 13440   | 29120   |
-      | party2 | ETH/DEC19 | 10900       | 11990  | 13080   | 28340   |
+      | party1 | ETH/DEC19 | 11200       | 12320  | 13440   | 15680   |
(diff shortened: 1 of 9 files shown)


core/integration/features/auctions/opening-auction-price.feature
```diff
@@ -65,8 +65,8 @@ Feature: Set up a market, create indiciative price different to actual opening a
       | party7 | ETH/DEC19 | sell | 1      | 11000 | 0                | TYPE_LIMIT | TIF_GFA | t7-s-1    |
     And the parties should have the following margin levels:
       | party  | market id | maintenance | search | initial | release |
-      | party1 | ETH/DEC19 | 11200       | 12320  | 13440   | 29120   |
-      | party2 | ETH/DEC19 | 10900       | 11990  | 13080   | 28340   |
+      | party1 | ETH/DEC19 | 11200       | 12320  | 13440   | 15680   |
```
davidsiska-vega, line 68: I see maintenance margin staying the same but release is changing - why? (probably an easy answer, but I am slow)

witgaw: If you look at the last file in the diff (`core/risk/margins_calculation.go`) you'll see that there used to be a hack in place to prevent a margin release during auction. An auction state object has been added since, so I'm now just doing properly using that object.

EVODelavega (approved):

jeremyletang (approved):

davidsiska-vega (approved):

PR #7557, opened 2023-02-09
refactor: mtm after all orders back on book

### LINKED ISSUE #7543: Initiate post-auction closeout once parked orders get redeployed
Currently, when leaving an auction we attempt a closeout before the parked (pegged and LP) orders get redeployed:

https://github.com/vegaprotocol/vega/blob/9fd529c7f8554fb8d19b2fbae068319bf5a7203a/core/execution/market.go#LL1119C3-L1119C3

It should be done after these orders are placed on the book.

witgaw:
- Moved after-auction MTM check (and hence closeouts) to a point after all the parked orders have been redeployed.
- Integration test changes are due to a different slippage component:
    - margin requirements usually grow as a result (as before the slippage was calculated on a sparse order book),
    - however they can also diminish if LPs use orders pegged to MID (then the slippage can be less than it used to).
- Simplified `confirmMTM` signature to avoid passing in dummy objects.
- Improved account balance cucumber step so that checks for parties with accounts denominated in multiple assets look sensibly.

closes #7543

### DIFF
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -1141,11 +1141,12 @@ func (m *Market) leaveAuction(ctx context.Context, now time.Time) {
 		cmp := m.getLastTradedPrice()
 		m.markPrice = cmp.Clone()
 		mcmp := num.UintZero().Div(cmp, m.priceFactor) // create the market representation of the price
+		// checkForReferenceMoves has already been called with updatedOrders so passing an empty array in now
 		m.confirmMTM(ctx, &types.Order{
 			ID:            m.idgen.NextID(),
 			Price:         cmp,
 			OriginalPrice: mcmp,
-		}, updatedOrders, false)
+		}, []*types.Order{}, false)
 		// set next MTM
 		m.nextMTM = m.timeService.GetTimeNow().Add(m.mtmDelta)
 	}
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -2115,7 +2115,7 @@ func (m *Market) zeroOutNetwork(ctx context.Context, parties []events.MarketPosi
 	}
 
 	tradeEvts := make([]events.Event, 0, len(parties))
-	for i, party := range parties {
+	for _, party := range parties {
 		tSide, nSide := types.SideSell, types.SideSell // one of them will have to sell
 		if party.Size() < 0 {
 			tSide = types.SideBuy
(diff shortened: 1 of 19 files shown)


core/execution/market.go
```diff
@@ -1964,19 +1947,14 @@ func (m *Market) resolveClosedOutParties(ctx context.Context, distressedMarginEv
 		size = uint64(-networkPos)
 	}
 
-	ref := "LS"
-	if orderID != nil {
-		ref = fmt.Sprintf("LS-%s", *orderID)
```
witgaw, line 1969: This was just the `m.idgen.NextID()` pulled for a dummy order object passed into the old confirmMTM signature, so hopefully not too controversial...

witgaw (commented):
Ready for review

core/execution/market.go
```diff
@@ -1964,19 +1948,14 @@ func (m *Market) resolveClosedOutParties(ctx context.Context, distressedMarginEv
...
-
 	no := types.Order{
 		MarketID:    m.GetID(),
 		Remaining:   size,
 		Status:      types.OrderStatusActive,
 		Party:       types.NetworkParty, // network is not a party as such
 		Side:        types.SideSell,     // assume sell, price is zero in that case anyway
 		CreatedAt:   now.UnixNano(),
-		Reference:   ref,                       // liquidity sourcing, reference the order which caused the problem
+		Reference:   "LS",                      // liquidity sourcing, reference the order which caused the problem
```
ze97286, line 1958: [not important] why not keep the convention of ref = fmt.Sprintf("LS-%s", *orderID) just do it yourself here with m.idgen.NextID()?

core/execution/market.go
```diff
@@ -2182,7 +2146,7 @@ func (m *Market) zeroOutNetwork(ctx context.Context, parties []events.MarketPosi
 			Price:         settleOrder.Price.Clone(), // average price
 			OriginalPrice: settleOrder.OriginalPrice.Clone(),
 			CreatedAt:     now,
-			Reference:     ref,
+			Reference:     fmt.Sprintf("distressed-%d", i),
```
ze97286, line 2149: maybe party is better than i for tracking, idk.

ze97286 (dismissed):

core/execution/market.go
witgaw: Can do, just didn't think it was adding any value (that order will have its ID like any other one), some random reference didn't seem like it'd help with anything, but I can put it back in.

witgaw: I think it didn't have sense one we were MTMing after each trade, then it'd would get the ID of the order that resulted in the price move that brought about the closeout, but after moving to time-based MTM it lost it's meaning.

core/execution/market.go
witgaw: Sounds good, happy to put it in place once there's a consensus we're ok to remove the random ID in the reference.

ze97286 (approved):

TomMcL (approved):

PR #7561, opened 2023-02-10
feat: archive and roll event buffer files

### LINKED ISSUE #7517: Core File Client - Roll files
The cores file client currently persists all events to a single file.  This means there is no way to compress and cleanup storage as the file cannot be cleanly split (without a process to read it in and do this).  If we make the file client split the file after X events (such that we get reasonable event file size say ~ 100MB uncompressed which should be about 10MB compressed), then it will easy to add a job to cleanup and compress these as we go meaning that, if we cannot always have all event data as it may be too large to store from block 0, we can at least have the most recent few days data which will make it easier to diagnose issues.   Related to this, the data-node file reader would need to change to sequentially read in a time sorted collection of event files.

ettec:
closes #7517 

This change adds functionality to the datanode to optinally archive the event buffer files that are already created.  This is part 1 of 2 part change, part 1 is the archiving of the files, part 2 will be the change to allow the datanode to read archive buffer files (buffer file format is different from that of the core's file event dumper).

Originally, was going to just move the files into an archive directory and rely on ops to setup the compression and rolling of the files, however on reflection it will  be exceptionally useful to always be able to get a datanodes recent events in the event of a bug/issue.  Combined with load a recent network history segment this will allow us to replay any scenario.  For this reason I opted to make archiving on by default with a max archive dir size of 10GB.

  @vegaprotocol/ops @fkondej the maximum size of the archive can be configure and turned on and off using the following properties.   Ideally the maximum size would be set as a percentage of the available disk space in out env so we can hold onto as much event history as possible without it impacting anything.  As a guide a reasonable number would be say 100GB although more is always better.

`Archive = true 
 ArchiveMaximumSizeBytes = 10000000 
`
Which is part of the buffered event source config, e.g:

`  [Broker.BufferedEventSourceConfig]
    EventsPerFile = 100000
    SendChannelBufferSize = 10000
    MaxBufferedEvents = 100000000
    Archive = true 
    ArchiveMaximumSizeBytes = 10000000 
`

### DIFF
--- a/datanode/broker/buffered_event_source.go
+++ b/datanode/broker/buffered_event_source.go
@@ -5,10 +5,14 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
-	"io/ioutil"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
+	"strings"
 	"time"
+
+	"code.vegaprotocol.io/vega/datanode/utils"
 
 	"code.vegaprotocol.io/vega/core/events"
 	"code.vegaprotocol.io/vega/datanode/metrics"
@@ -24,6 +28,7 @@ type FileBufferedEventSource struct {
 	sendChannelBufferSize int
 	source                EventReceiver
 	bufferFilePath        string
+	archiveFilesPath      string
 	config                BufferedEventSourceConfig
 }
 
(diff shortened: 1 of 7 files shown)


pscott31 (dismissed):

datanode/broker/buffered_event_source.go
```diff
@@ -270,14 +301,41 @@ func (m *FileBufferedEventSource) removeBufferFile(bufferFile *os.File) error {
...
+
+// moveBufferFileToArchive encodes the creation time into the archive file name to ensure that the correct order
+// of files can always be determined even if the archive files are copied etc.
+func (m *FileBufferedEventSource) moveBufferFileToArchive(bufferFilePath string) error {
+	bufferFileName := filepath.Base(bufferFilePath)
+	bufferSeqSpan := strings.ReplaceAll(bufferFileName, bufferFileNamePrepend, "")
+	timeNowUtc := time.Now().UTC()
+
+	archiveFileName := fmt.Sprintf("%s-%s-%d-seqnumspan%s", bufferFileNamePrepend,
+		timeNowUtc.Format("2006-01-02-15:04:05"), timeNowUtc.UnixNano(), bufferSeqSpan)
```
guoguojin, line 327: I don't think having colons in file names are a good idea. It's a bit messy to escape the characters.

datanode/broker/buffered_event_source.go
```diff
@@ -322,8 +380,10 @@ func readEvent(eventFile *os.File, offset int64) (event *eventspb.BusEvent, seqN
 	return event, seqNum, totalBytesRead, nil
 }
 
+const bufferFileNamePrepend = "datanode-buffer"
+
 func (m *FileBufferedEventSource) getBufferFileName(fromSeqNum uint64, toSeqNum uint64) string {
-	return fmt.Sprintf("%s/datanode-buffer-%d-%d", m.bufferFilePath, fromSeqNum, toSeqNum)
+	return fmt.Sprintf("%s/%s-%d-%d.bevt", m.bufferFilePath, bufferFileNamePrepend, fromSeqNum, toSeqNum)
```
guoguojin, line 386: I think you need to specify the number of digits for the integer values here, otherwise the sorting of the filenames can be corrupted: e.g.

prefix-1.ext
prefix-10.ext
prefix-2.ext

etc.

ettec: I prepended the time to ensure the sort works, the seq num is really just for info, it serves no purpose

ettec: that is to say, the time is preppended on the archive files, you don't need it on the non archived files as the buffer reader knows what file nam,e to expect

datanode/broker/buffered_event_source.go
ettec: good point - sorted

guoguojin (approved):

PR #7572, opened 2023-02-13
fix: send market depth event at end of block

### LINKED ISSUE #7492: Force market depth updates to be sent at the end of every block
Currently the market depth service generates a message with updates in every 100ms (if updates are available). This had the side effect of allowing a batch of orders to have their updates split across multiple update messages. To prevent his we want to force the update message to be sent at the end of every block to guarantee that the information in the update message does not split apart a batch of orders. The 100ms timer is no longer required.

(This ticket is a quick fix to correct the current issue, a better fix will be proposed in a new ticket which will describe the better way to handle this)

AnExsomnis:
close #7492


jeremyletang (approved):
Looks good to me, the only difference is that previously we would publish to the client asynchronously, now it's being done synchronously (haven't gone deep, so might be wrong and their maybe some mutex anyway around in the publishChanges method). If it's not a concern, then that looks sane to me.

guoguojin (approved):

peterbarrow (approved):

PR #7573, opened 2023-02-13
chore: extend protos to add rest docs

jeremyletang:


edd:
List accounts | Vega Protocol Documentation

* There's a new '500' error, which is good, I like that
* There's a response example. It's good, but doesn't actually match the schema, which kinda sucks and sounds like manual work to keep in sync
* The 'Submit request' thing doesn't actually restrict data by the regex you give us, which is a shame
* No `required` annotations (which was the original impetus for looking in to more annotations)

PR #7578, opened 2023-02-13
fix: Use correct trade size when calculating pending open volume

### LINKED ISSUE #7577: Pending open volume not calculated correctly
# Problem encountered
Pending open volume grows disproportionately to the trades done.

# Observed behaviour
In stagnet1, parties got to trillion $ open volume as returned by the API. 

# Expected behaviour
Pending Open Volume should track open position and trades. 

Pending Open Volume is calculated in datanode and tracks trades and position changes. When a trade occurs it calculates the newly opened or closed volume and updates the pending open volume. There is a bug in the implementation that updates (in certain cases) the open volume not by the size of the new trade but by the size of the new trade + the pending volume preceding the trade. For example:

Say we have a pending volume of -100k, and in a certain block there are 5 trades with size 50, rather than getting a pending volume of -100250k, we'll get ~100k * 2^5.

ze97286:
Closes #7577


jeremyletang (dismissed):
looks good. we might want @EVODelavega to look into it as well before merging.

EVODelavega (approved):

Issue #7588, opened 2023-02-14
Slippage_per_unit is calculated in the old way by using exiting orders on the book rather than `slippage_per_unit = +Infinity` 


Jiajia-Cui:
# Problem encountered

When there is insufficient order book volume on the relevant side of the order book to calculate the `exit_price`, then slippage_per_unit is calculated in the old way by using exiting orders on the book rather than `slippage_per_unit = +Infinity`  ,which is in the newly updated spec: link to [PR #1583](https://github.com/vegaprotocol/specs/pull/1583/files)

# Expected behaviour
When there is insufficient order book volume on the relevant side of the order book to calculate the `exit_price`, then take `slippage_per_unit = +Infinity` which means that `min(slippage_volume * slippage_per_unit, slippage_volume * mark_price * market.maxSlippageFraction) = slippage_volume * mark_price * market.maxSlippageFraction` above.

PR #7589, opened 2023-02-14
feat: create test for bug 7588

Jiajia-Cui:
link to #7588


Jiajia-Cui:
This test was created for work 7588, it can be closed now since the 7588 is almost done

PR #7602, opened 2023-02-15
feat: use ExpiredOrders event

### LINKED ISSUE #7601: Create a single event for order expiry
When orders are expired, we currently emit full-blown order events for each expired order event. We can easily replace this with a single event containing just the market and order IDs of those orders we wish to expire. This will cut down on the amount of noise, and volume of data we need to handle.

EVODelavega:
Closes #7601


vega-paul:
```{"system-tests":"order_expiry"}```

**Review by jeremyletang — DISMISSED** [review]

File: datanode/sqlsubscribers/order.go [hunk 1]
```diff
@@ -44,17 +51,43 @@ func NewOrder(store OrderStore) *Order {
...
 func (os *Order) Flush(ctx context.Context) error {
 	return os.store.Flush(ctx)
 }
 
+func (os *Order) expired(ctx context.Context, eo ExpiredOrdersEvent, seqNum uint64) error {
+	orders, err := os.store.GetByMarketAndID(ctx, eo.MarketID(), eo.OrderIDs())
+	if err != nil {
+		return err
+	}
+	for _, o := range orders {
```
jeremyletang, line 79: Won't tyou have the wrong TxHash here? You'll need to set it sa well, see the consume function.

EVODelavega: Wrong is such a strong word, but it would indeed be, erm, less correct. Updated

datanode/networkhistory/service_test.go
```diff
@@ -1264,17 +1264,23 @@ func getDatabaseDataSummary(ctx context.Context, connConfig sqlstore.ConnectionC
 	}
 
 	for table, meta := range dbMetaData.TableNameToMetaData {
+		if len(meta.SortOrder) == 0 {
```
ettec, line 1267: why is this needed?

datanode/networkhistory/snapshot/database_meta_data.go
```diff
@@ -112,7 +112,11 @@ func getTableSortOrders(ctx context.Context, conn *pgxpool.Pool) (map[string]str
 		if len(split) != 2 {
 			return nil, fmt.Errorf("unexpected primary key index definition:%s", pkIdx.Indexdef)
 		}
-		tableNameToSortOrder[pkIdx.Tablename] = strings.Replace(split[1], ")", "", 1)
+		so := strings.Replace(split[1], ")", "", 1)
```
ettec, line 115: Whats the reason for this change?

datanode/networkhistory/snapshot/history.go
```diff
@@ -42,6 +42,9 @@ func (h History) GetCopySQL(dbMetaData DatabaseMetadata, databaseSnapshotsPath s
 	var copySQL []string
 	for tableName, meta := range dbMetaData.TableNameToMetaData {
 		if dbMetaData.TableNameToMetaData[tableName].Hypertable {
+			if len(meta.SortOrder) == 0 {
```
ettec, line 45: why would we we have an empty sort order?

datanode/sqlstore/migrations/0005_orders_pk.sql
```diff
@@ -0,0 +1,12 @@
...
+-- CREATE TABLE orders_new AS orders;
+ALTER TABLE orders DROP CONSTRAINT IF EXISTS orders_pkey;
+
+-- ALTER TABLE orders ADD CONSTRAINT orders_pkey PRIMARY KEY (vega_time, seq_num, id);
+
+-- +goose Down
+
+-- ALTER TABLE orders DROP CONSTRAINT IF EXISTS orders_pkey;
+
+ALTER TABLE orders ADD CONSTRAINT orders_pkey PRIMARY KEY (vega_time, seq_num); -- not sure if we really can go back
```
ettec, line 12: the PK already is already `PRIMARY KEY(vega_time, seq_num)`   why drop and recreate the same PK?

datanode/networkhistory/service_test.go
```diff
@@ -1264,17 +1264,23 @@ func getDatabaseDataSummary(ctx context.Context, connConfig sqlstore.ConnectionC
...
 		if err != nil {
 			panic(err)
 		}
 
 		if summary.rowCount > 0 {
-			err = conn.QueryRow(ctx, fmt.Sprintf("SELECT md5(CAST((array_agg(f.* order by %s))AS text)) FROM %s f; ",
-				meta.SortOrder, table)).Scan(&summary.dataHash)
+			q := fmt.Sprintf("SELECT md5(CAST((array_agg(f.* order by %s))AS text)) FROM %s f; ", meta.SortOrder, table)
+			err = conn.QueryRow(ctx, q).Scan(&summary.dataHash)
+			// err = conn.QueryRow(ctx, fmt.Sprintf("SELECT md5(CAST((array_agg(f.* order by %s))AS text)) FROM %s f; ",
```
ettec, line 1279: looks like the same thing as original?  could we restore the commented out code?

ettec (changes requested):

datanode/sqlstore/orders.go
```diff
@@ -98,6 +99,39 @@ func (os *Orders) GetOrder(ctx context.Context, orderIDStr string, version *int3
...
+	bindNum := 2
+	for _, o := range orderIDs {
+		IDs = append(IDs, entities.OrderID(o))
+		in = append(in, fmt.Sprintf("$%d", bindNum))
+		bindNum++
+	}
+	bind := make([]interface{}, 0, len(in)+1)
+	// set all bind vars
+	bind = append(bind, marketID)
+	bind = append(bind, IDs...)
```
pscott31, line 124: Have you seen the `nextBindVar` helper function, it makes keeping track of these bind numbers much simpler

```go
func (s *Store) DoQuery(thing, wotsit string) {
	args := []interface{}{}
	query := fmt.Sprintf("select * from foo where thing=%s and whotsit=%s",
		nextBindVar(&args, thing),
		nextBindVar(&args, wotsit)
	)
}
```

datanode/networkhistory/snapshot/database_meta_data.go
```diff
@@ -112,7 +112,11 @@ func getTableSortOrders(ctx context.Context, conn *pgxpool.Pool) (map[string]str
...
+		if noSpace := strings.ReplaceAll(so, " ", ""); len(noSpace) == 0 {
```
ettec, line 116: Whats the reason for this change? we should not need to default the sort order, and vega_time is not the correct sort order in some cases

datanode/networkhistory/service_test.go
```diff
@@ -1264,17 +1264,23 @@ func getDatabaseDataSummary(ctx context.Context, connConfig sqlstore.ConnectionC
 	}
 
 	for table, meta := range dbMetaData.TableNameToMetaData {
+		if len(meta.SortOrder) == 0 {
+			meta.SortOrder = "vega_time, seq_num"
```
ettec, line 1268: why is this needed? It should never happen, and defaulting the sort order to vega_time,seq_num is not the right thing to do as some tables have different sort order

datanode/networkhistory/service_test.go
```diff
@@ -1264,17 +1264,23 @@ func getDatabaseDataSummary(ctx context.Context, connConfig sqlstore.ConnectionC
...
+		}
```
pscott31, line 1269: I'm not sure why this magic is needed?

datanode/sqlstore/migrations/0005_orders_pk.sql
ettec, line 12: the PK already is already PRIMARY KEY(vega_time, seq_num) why drop and recreate the same PK?

datanode/networkhistory/snapshot/history.go
ettec, line 45: why would we we have an empty sort order?

ettec (changes requested):

datanode/networkhistory/service_test.go
```diff
@@ -1264,17 +1264,23 @@ func getDatabaseDataSummary(ctx context.Context, connConfig sqlstore.ConnectionC
...
-			err = conn.QueryRow(ctx, fmt.Sprintf("SELECT md5(CAST((array_agg(f.* order by %s))AS text)) FROM %s f; ",
-				meta.SortOrder, table)).Scan(&summary.dataHash)
+			q := fmt.Sprintf("SELECT md5(CAST((array_agg(f.* order by %s))AS text)) FROM %s f; ", meta.SortOrder, table)
+			err = conn.QueryRow(ctx, q).Scan(&summary.dataHash)
+			// err = conn.QueryRow(ctx, fmt.Sprintf("SELECT md5(CAST((array_agg(f.* order by %s))AS text)) FROM %s f; ",
+			// meta.SortOrder, table)).Scan(&summary.dataHash)
 			if err != nil {
-				panic(err)
+				panic(fmt.Sprintf("Query: %s\nWhere: '%s'\nError: %v\n", q, meta.SortOrder, err))
+				// panic(q + " " + err.Error())
```
pscott31, line 1283: Slight tidy up maybe?

datanode/networkhistory/snapshot/database_meta_data.go
```diff
@@ -112,7 +112,11 @@ func getTableSortOrders(ctx context.Context, conn *pgxpool.Pool) (map[string]str
...
+			so = "vega_time" // default to time sort
+		}
+		tableNameToSortOrder[pkIdx.Tablename] = so
```
pscott31, line 119: Why does vega_time need special treatment?

datanode/sqlstore/orders.go
```diff
@@ -98,6 +99,39 @@ func (os *Orders) GetOrder(ctx context.Context, orderIDStr string, version *int3
...
+	// select directly from orders_live table, the current view searches in orders
+	// this is used to expire orders, which have to be, by definition, live. This table uses ID as its PK
+	// so this is a more optimal way of querying the data.
```
pscott31, line 127: nitpick: This is slightly at-odds with the name of the function, which doesn't mention anything about expiring or whatnot.. I guess if the function were called GetLiveOrdersByMarketAndID it would be a bit more obvious but it is a bit of a mouthful.. it's perhaps slightly surprising to a user that two functions:
`GetByMarket` and `GetByMarketAndID` are actually querying different things

datanode/sqlsubscribers/order.go
```diff
@@ -44,17 +51,45 @@ func NewOrder(store OrderStore) *Order {
...
+	}
+	return nil
 }
 
 func (os *Order) Flush(ctx context.Context) error {
 	return os.store.Flush(ctx)
 }
 
+func (os *Order) expired(ctx context.Context, eo ExpiredOrdersEvent, seqNum uint64) error {
+	orders, err := os.store.GetByMarketAndID(ctx, eo.MarketID(), eo.OrderIDs())
```
ettec, line 75: I'm not sure I see the point of this change, I see you get reduced event numbers, but from the POV of the datanode that's not the bottleneck  (is it?  can we see the pprof), the bottle neck is the number of order updates we have to persist into the database, and having to first lookup the order before repersisting will not reduce the number of persisted rows and may even slow down order persistence processing which IS the bottle neck.   This change seems more harmful than beneficial, we should have a group discussion on it I think, may be I am missing something but I have significant concern about it.

ettec (changes requested):

pscott31 (commented):
So if I'm reading this right this should reduce the number of order events sent when orders are expired - which is good... however now we have to do a query (or potentially bunch of queries; I'm not sure how tightly the expired event volumes are batched) to get those orders instead.

It's not obvious to me that that is an improvement, but perhaps I'm missing something? Is this about making datanode go faster or reducing the volume of traffic on the event bus? Would be delighted to be proved wrong

EVODelavega:
Superseded by #7626

PR #7603, opened 2023-02-15
feat: send 1 margin levels events per transaction processing.

jeremyletang:
With a sample of 1.65k event in a block, we send ~300 events for margin levels. This PR try to reduce this amount by sending the events only once per transaction per party which had their margin levels re-calculated.

### DIFF
--- a/core/risk/engine.go
+++ b/core/risk/engine.go
@@ -15,6 +15,7 @@ package risk
 import (
 	"context"
 	"errors"
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


EVODelavega (dismissed):

karlem:
@jeremyletang Hey, quick question about this. This will still produces the same amount of events right? You just hope that they will not happen to be all around the same time?

vega-paul:
ST full https://jenkins.ops.vega.xyz/job/common/job/system-tests-nightly/785/

core/risk/engine.go
```diff
@@ -142,9 +151,50 @@ func NewEngine(log *logging.Logger,
...
+	e.updateMarginLevels = e.bufferMarginLevels
+	if e.StreamMarginLevelsVerbose {
+		e.updateMarginLevels = e.sendMarginLevels
+	}
+
 	return e
 }
 
+func (e *Engine) FlushMarginLevelsEvents() {
+	if e.StreamMarginLevelsVerbose || len(e.marginLevelsUpdates) <= 0 {
```
karlem, line 163: This is a mega nit. The length can't be less then zero. So I think `len(e.marginLevelsUpdates) < 1` would do.

karlem (dismissed):

ze97286 (approved):

PR #7610, opened 2023-02-15
fix: reduce amount of liquidity orders events emitted

jeremyletang:
Looking at a run emitting 1.65k events per blocks, ~500 where order events, out of theose ~120 where just updating liqudity orders, out of these 60 where cancelling the order, to then re-submit it straight away. This ensure that no cancel order event is send while repricing LPs if not required.

### DIFF
--- a/core/execution/2965_refresh_lp_orders_size_test.go
+++ b/core/execution/2965_refresh_lp_orders_size_test.go
@@ -244,12 +244,12 @@ func TestRefreshLiquidityProvisionOrdersSizes(t *testing.T) {
 				types.OrderStatusActive,
 				size - 20,
 			},
-			{
-				// this is the replacement order created
-				// by engine.
-				types.OrderStatusCancelled,
-				size - 20,
-			},
+			// {
+			// 	// this is the replacement order created
+			// 	// by engine.
+			// 	types.OrderStatusCancelled,
+			// 	size - 20,
+			// },
 			{
 				// this is the re-deployment
 				types.OrderStatusActive,
--- a/core/execution/2965_refresh_lp_orders_size_test.go
+++ b/core/execution/2965_refresh_lp_orders_size_test.go
@@ -244,12 +244,6 @@ func TestRefreshLiquidityProvisionOrdersSizes(t *testing.T) {
 				types.OrderStatusActive,
 				size - 20,
 			},
-			// {
-			// 	// this is the replacement order created
-			// 	// by engine.
-			// 	types.OrderStatusCancelled,
-			// 	size - 20,
-			// },
 			{
 				// this is the re-deployment
 				types.OrderStatusActive,
--- a/core/execution/special_orders.go
+++ b/core/execution/special_orders.go
@@ -233,9 +233,10 @@ func (m *Market) updateLPOrders(
 		now       = m.timeService.GetTimeNow().UnixNano()
 	)
 
-	// now we gonna map all the all order which
-	// where to be cancelled, and just do nothing in
-	// those case.
+	// now we gonna map all the order which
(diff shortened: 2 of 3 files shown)


zale144 (dismissed):

EVODelavega (dismissed):

core/execution/2965_refresh_lp_orders_size_test.go
```diff
@@ -244,12 +244,12 @@ func TestRefreshLiquidityProvisionOrdersSizes(t *testing.T) {
 				types.OrderStatusActive,
 				size - 20,
 			},
-			{
-				// this is the replacement order created
-				// by engine.
-				types.OrderStatusCancelled,
-				size - 20,
-			},
+			// {
```
ze97286, line 247: might as well remove

core/execution/special_orders.go
```diff
@@ -241,6 +242,13 @@ func (m *Market) updateLPOrders(
 		}
 	}
 
+	// now we gonna map all the all order which
```
ze97286, line 245: you may want to translate this to English :)

core/execution/special_orders.go
```diff
@@ -261,10 +269,14 @@ func (m *Market) updateLPOrders(
...
 		order.UpdatedAt = now
 
+		_, toCancel := cancelIDs[order.ID]
+		_, toSubmit := submitIDs[order.ID]
 		// these order were actually cancelled, just send the event
-		if _, ok := cancelIDs[order.ID]; ok {
-			order.Status = types.OrderStatusCancelled
-			orderEvts = append(orderEvts, events.NewOrderEvent(ctx, order))
+		if toCancel {
+			if !toSubmit {
```
ze97286, line 276: something is not clear to me though is not a regression from the previous behaviour. Suppose an order id is in the submit and is not in the cancel - wouldn't it be submitted here and then again submitted in line 286?

jeremyletang: Nope, I asked myself the same thing, and basically here the allOrders is all that were in the book? The cancels are all that needs to cancel, and the submit are all that are new order instructed to be created (as in they were not in the book before but they are now needed)

core/execution/special_orders.go
jeremyletang: Lol. Nope I like it like this!

core/execution/special_orders.go
ze97286: so the order of things as I understand it is the following:

1. we remove all LP orders from the book
`m.removeLPOrdersFromBook(ctx, lpOrders)`

2. then we resubmit all pegged orders 
	
```
partiesPos := map[string]events.MarketPosition{}
if needsPeggedUpdates {
_, partiesPos = m.reSubmitPeggedOrders(ctx, toSubmit)
}
```	

3. then we're calling update to reprice LP order
`newOrders, cancels := m.liquidity.Update(ctx, minLpPrice, maxLpPrice, m.repriceLiquidityOrder)`

and this newOrders and cancels is what goes into the function you changed above. So I can see that submits are not in the book as they've been removed. I'm not sure how there can be cancels if all orders have been removed in step 1, but anyways, this isn't a regression.

core/execution/2965_refresh_lp_orders_size_test.go
jeremyletang: done

jeremyletang:
> ST Full https://jenkins.ops.vega.xyz/job/common/job/system-tests-nightly/786/

https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-nightly/detail/system-tests-nightly/787/pipeline

ze97286 (approved):

PR #7616, opened 2023-02-16
fix: save m.parties in snapshot so that we don't have to worry the lo…

### LINKED ISSUE #7614: Snapshot: `m.parties` not restore properly in `PostRestore`
# Problem encountered
This snapshot restore failure appeared in the overnight soak tests:
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsnapshot-soak-tests/detail/snapshot-soak-tests/2165/pipeline

It is related to the restoration of a market after a snapshot restore, namely that `m.parties` has length 4 in a non-restored node but length 5 in a restored node.

# Observed behaviour
Snapshot gone wrong.

# Expected behaviour
Snapshot not gone wrong.

# System response

# Steps to reproduce

## Manual

## Automation

# Evidence

## Logs

## Additional context

# Definition of Done

**Before Merging**

**After Merging**

wwestgarth:
closes #7614

The issue here is that after a snapshot restore the map of parties with positions in a market, `m.parties` becomes out of sync.

The reason is because we assume in `PostRestore` that if a party has a position then it is in `m.parties` but this is not true in this edge case. At the end of each block we all `releaseExcessMargin()` which will remove a party with no positions from `m.parties` but it remains in the positions engine.

We often have situations where an engine has state that is derived from other state, and where it is trivial we can get away with re-deriving it on snapshot restore. I think here this is not a case where we should do that.

The code in `PostRestore()` at the moment is sensitive to changes across multiple engines. Previously the assumption it made workedd, but a code change made 3 months ago broke that assumption and it went unnoticed until now.

So I think we should try not to be clever and serialise `m.parties` along with the market so we can always restore it exactly how it was before.

The PR is marked as breaking because it will cause problems over a protocol upgrade -- `m.parties` will remain empty since the old snapshot will not contain them.


core/execution/market_snapshot.go
```diff
@@ -190,6 +195,12 @@ func (m *Market) getState() *types.ExecMarket {
 		sp = m.settlementDataInMarket.Clone()
 	}
 
+	parties := make([]string, 0, len(m.parties))
+	for p := range m.parties {
+		parties = append(parties, p)
+	}
```
jeremyletang, line 201: Just style so be free to ignore, but I've started to like doing these: `parties := maps.Keys(m.parties)`, it ultimately does exactly the same as your code.

jeremyletang (approved):

core/execution/market_snapshot.go
wwestgarth: Thats nice. Changed.

PR #7626, opened 2023-02-17
Expired orders sequence fix

### LINKED ISSUE #7601: Create a single event for order expiry

### LINKED ISSUE #7621: ExpiredOrders - duplicate key value violates unique constraint "9_9_orders_pkey"
Running expired-orders-event branch which delivers ExpiredOrder events

Full run has postgres has errors and datanode connection issues

https://jenkins.ops.vega.xyz/job/common/job/system-tests-wrapper/51633/

https://jenkins.ops.vega.xyz/job/common/job/system-tests-wrapper/51633/artifact/testnet/logs/postgres-1/postgres-1.stderr-2023-02-16T20%3A01%3A31Z.log#:~:text=2%0A2023%2D02-,%2D,-16%2021%3A15

```
023-02-16 21:14:59.247 UTC [7438] LOG:  refreshing continuous aggregate "trades_candle_1_minute" in window [ 2023-02-15 21:15:00+00, 2023-02-16 21:13:00+00 ]
2023-02-16 21:15:23.670 UTC [192] ERROR:  duplicate key value violates unique constraint "9_9_orders_pkey"
2023-02-16 21:15:23.670 UTC [192] DETAIL:  Key (vega_time, seq_num)=(2023-02-16 21:15:22.906619+00, 70) already exists.
2023-02-16 21:15:23.670 UTC [192] CONTEXT:  COPY orders, line 2
2023-02-16 21:15:23.670 UTC [192] STATEMENT:  copy "orders" ( "id", "market_id", "party_id", "side", "price", "size", "remaining", "time_in_force", "type", "status", "reference", "reason", "version", "pegged_offset", "batch_id", "pegged_reference", "lp_id", "created_at", "updated_at", "expires_at", "tx_hash", "vega_time", "seq_num" ) from stdin binary;
20
```

EVODelavega:
Closes #7621 
and 
close #7601 

Using a different approach (no migrations)


EVODelavega:
```{"system-tests":"order_expiry"}```

vega-paul:
full run https://jenkins.ops.vega.xyz/job/common/job/system-tests-nightly/791/

pscott31 (commented):
I got a bit confused here and left a bunch of comments on [7062](https://github.com/vegaprotocol/vega/pull/7602) which I think is an earlier version of this PR. I can go through and copy-paste relevant suggestions if it would be helpful - but this was my overall comment which I think still applies:

> So if I'm reading this right this should reduce the number of order events sent when orders are expired - which is good... however now we have to do a query (or potentially bunch of queries; I'm not sure how tightly the expired event volumes are batched) to get those orders instead.

> It's not obvious to me that that is an improvement, but perhaps I'm missing something? Is this about making datanode go faster or reducing the volume of traffic on the event bus? Would be delighted to be proved wrong

EVODelavega:
> So if I'm reading this right this should reduce the number of order events sent when orders are expired - which is good... however now we have to do a query (or potentially bunch of queries; I'm not sure how tightly the expired event volumes are batched) to get those orders instead.
 
That's the same as before: each expired order resulted in an order event being sent, which we handled individually, though we didn't perform the select, we spent a lot of time unmarshalling the event, converting it to an entity, and upserting it. True we are performing a single SELECT that wasn't there before, in exchange, we are spending a lot less time passing through and mapping data from events that we are currently doing. The benchmarks show that this is a sizeable improvement in terms of performance

> It's not obvious to me that that is an improvement, but perhaps I'm missing something? Is this about making datanode go faster or reducing the volume of traffic on the event bus? Would be delighted to be proved wrong

Here's the benchmarking results: https://docs.google.com/spreadsheets/d/1fzafG5Wz8II_SWRlqNb_C7Sd4aUbwxzKwec7n6_olY4/edit?pli=1#gid=0

The most notable improvement:
<google-sheets-html-origin><style type="text/css"></style>

75 | 1 | 10 | 10 | 50 |   |   | 49411 | 47.00% | 248.00% | 14.00% |   | 17252 | 44.00% | 67.00%
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

ettec:
> > So if I'm reading this right this should reduce the number of order events sent when orders are expired - which is good... however now we have to do a query (or potentially bunch of queries; I'm not sure how tightly the expired event volumes are batched) to get those orders instead.
> 
> That's the same as before: each expired order resulted in an order event being sent, which we handled individually, though we didn't perform the select, we spent a lot of time unmarshalling the event, converting it to an entity, and upserting it. True we are performing a single SELECT that wasn't there before, in exchange, we are spending a lot less time passing through and mapping data from events that we are currently doing. The benchmarks show that this is a sizeable improvement in terms of performance
> 
> > It's not obvious to me that that is an improvement, but perhaps I'm missing something? Is this about making datanode go faster or reducing the volume of traffic on the event bus? Would be delighted to be proved wrong
> 
> Here's the benchmarking results: https://docs.google.com/spreadsheets/d/1fzafG5Wz8II_SWRlqNb_C7Sd4aUbwxzKwec7n6_olY4/edit?pli=1#gid=0
> 
> The most notable improvement: <style type="text/css"></style>
> 
> 75	1	10	10	50	 	 	49411	47.00%	248.00%	14.00%	 	17252	44.00%	67.00%

Yeah, I also commented on the wrong PR, I won't bring over the other comments, but this one I think is worth repeating:
If this change reduces the number of inserts into the order table, then I am all for it, as this is the bottleneck, not event marshalling, or anything else as can quite clearly be seen by looking the block processing stats on any env, e.g->

https://metrics.vega.xyz/d/datanode-stagnet1/data-node?orgId=1&var-instance=n05.stagnet3.vega.xyz

As you can see the VAST majority of real world time processing a block is inserting the orders into the DB, therefore anything that can reduce this number is invaluable and is where our optimising effort should be spent, so if this change does that, great, but from what I see it does not, am I wrong?   If it does not I don't see the point of the optimisation, it would seem that we are optimising things that will have no real world benefit but come with risk as it introduces a select on the orders table as part of the block processing which was not happening before, which will probably be ok most of the time, but maybe if it references an order not in the latest hyper table chunk, wherein there is a risk it slows down processing, this is speculation, but if there is no real world benefit EVEN IF it saves CPU cycles, but risks impacting the key known bottleneck,  it is not a risk a risk worth taking IMO.

jeremyletang:
> > > So if I'm reading this right this should reduce the number of order events sent when orders are expired - which is good... however now we have to do a query (or potentially bunch of queries; I'm not sure how tightly the expired event volumes are batched) to get those orders instead.
> > 
> > 
> > That's the same as before: each expired order resulted in an order event being sent, which we handled individually, though we didn't perform the select, we spent a lot of time unmarshalling the event, converting it to an entity, and upserting it. True we are performing a single SELECT that wasn't there before, in exchange, we are spending a lot less time passing through and mapping data from events that we are currently doing. The benchmarks show that this is a sizeable improvement in terms of performance
> > > It's not obvious to me that that is an improvement, but perhaps I'm missing something? Is this about making datanode go faster or reducing the volume of traffic on the event bus? Would be delighted to be proved wrong
> > 
> > 
> > Here's the benchmarking results: https://docs.google.com/spreadsheets/d/1fzafG5Wz8II_SWRlqNb_C7Sd4aUbwxzKwec7n6_olY4/edit?pli=1#gid=0
> > The most notable improvement: <style type="text/css"></style>
> > 75	1	10	10	50	 	 	49411	47.00%	248.00%	14.00%	 	17252	44.00%	67.00%
> 
> Yeah, I also commented on the wrong PR, I won't bring over the other comments, but this one I think is worth repeating: If this change reduces the number of inserts into the order table, then I am all for it, as this is the bottleneck, not event marshalling, or anything else as can quite clearly be seen by looking the block processing stats on any env, e.g->
> 
> https://metrics.vega.xyz/d/datanode-stagnet1/data-node?orgId=1&var-instance=n05.stagnet3.vega.xyz
> 
> As you can see the VAST majority of real world time processing a block is inserting the orders into the DB, therefore anything that can reduce this number is invaluable and is where our optimising effort should be spent, so if this change does that, great, but from what I see it does not, am I wrong? If it does not I don't see the point of the optimisation, it would seem that we are optimising things that will have no real world benefit but come with risk as it introduces a select on the orders table as part of the block processing which was not happening before, which will probably be ok most of the time, but maybe if it references an order not in the latest hyper table chunk, wherein there is a risk it slows down processing, this is speculation, but if there is no real world benefit EVEN IF it saves CPU cycles, but risks impacting the key known bottleneck, it is not a risk a risk worth taking IMO.

Points taken. However as I mentioned it to you what you are saying is coming without proof compared to the benchmarks we ran earlier. I’ll more than happy to roll it back if when we have tested it with a real network we can confirm this brings issues. In the meantime these debates are just wasting time that we could use just using the code and see.

EVODelavega:
> Yeah, I also commented on the wrong PR, I won't bring over the other comments, but this one I think is worth repeating: If this change reduces the number of inserts into the order table, then I am all for it, as this is the bottleneck, not event marshalling, or anything else as can quite clearly be seen by looking the block processing stats on any env, e.g->
> 
> https://metrics.vega.xyz/d/datanode-stagnet1/data-node?orgId=1&var-instance=n05.stagnet3.vega.xyz
> 
> As you can see the VAST majority of real world time processing a block is inserting the orders into the DB, therefore anything that can reduce this number is invaluable and is where our optimising effort should be spent, so if this change does that, great, but from what I see it does not, am I wrong? If it does not I don't see the point of the optimisation, it would seem that we are optimising things that will have no real world benefit but come with risk as it introduces a select on the orders table as part of the block processing which was not happening before, which will probably be ok most of the time, but maybe if it references an order not in the latest hyper table chunk, wherein there is a risk it slows down processing, this is speculation, but if there is no real world benefit EVEN IF it saves CPU cycles, but risks impacting the key known bottleneck, it is not a risk a risk worth taking IMO.

Every change in the order status has to result in record being appended to the `orders` table. Nothing can change that. The number of inserts is what it is. The results in the spreadsheet linked above, though, are quite revealing. The concern about the select on the orders table is unfounded. We're not querying the `orders` table. We are looking for _active_ orders that have just expired, so we know that the records we're interested in can only be found in the `orders_current` table, and we  select using order ID, which is that tables PK. That's not the sort of query to show up in slow query logs.

ze97286 (approved):

guoguojin:
Small correction to @EVODelavega last comment. The active orders are in the `orders_live` table and the code correctly searches there when looking for orders to expire.

`orders_current` is a view of the historic orders table showing the current state of all orders.

PR #7636, opened 2023-02-20
feat: add test for "loss socialisation amount" AC 0007-POSN-014/15/16

Jiajia-Cui:
link to [PR #1594](https://github.com/vegaprotocol/specs/pull/1594/files)

add feature test to cover Acs:

- 0007-POSN-015
- 0007-POSN-016
- 0007-POSN-017


core/integration/main_test.go
```diff
@@ -462,6 +462,14 @@ func InitializeScenario(s *godog.ScenarioContext) {
 		execsetup.assetsEngine.SetPermissive()
 		return nil
 	})
+
```
EVODelavega, line 465: Is this the bit I added a while back? Because if it is, through some other PR a while back I added the position status bit to the existing positions table, so this probably isn't needed anymore.

Jiajia-Cui: I am not sure which other PRs you meant, could you remove this part, and then pull develop, and then rerun the feature tests?

core/integration/features/verified/closeout-lognormal.feature
```diff
@@ -209,13 +207,104 @@ Feature: Closeout scenarios
...
+      | trader2 | ETH/DEC20 | 0           | 0      | 0       | 0       |
 
-    And the parties should have the following profit and loss:
-      | party   | volume | unrealised pnl | realised pnl | status                     |
-      | trader2 | 0      | 0              | -1026        | POSITION_STATUS_CLOSED_OUT |
+    And the parties should have the following position changes for market "ETH/DEC20":
+      | party   | status                     |
+      | trader2 | POSITION_STATUS_CLOSED_OUT |
+
+Scenario: 003, Position becomes distressed when market is in consitous mode
```
cdummett, line 224:
```suggestion
Scenario: 003, Position becomes distressed when market is in continuous mode 
```

jeremyletang (dismissed):

core/integration/features/verified/closeout-lognormal.feature
```diff
@@ -209,13 +207,104 @@ Feature: Closeout scenarios
...
+      | party      | market id | side | volume | price | resulting trades | type       | tif     | reference  |
+      | auxiliary2 | ETH/DEC20 | buy  | 5      | 5     | 0                | TYPE_LIMIT | TIF_GTC | aux-b-5    |
+      | auxiliary1 | ETH/DEC20 | sell | 10     | 1000  | 0                | TYPE_LIMIT | TIF_GTC | aux-s-1000 |
+      | auxiliary2 | ETH/DEC20 | buy  | 1      | 10    | 0                | TYPE_LIMIT | TIF_GTC | aux-b-1    |
+      | auxiliary1 | ETH/DEC20 | sell | 1      | 10    | 0                | TYPE_LIMIT | TIF_GTC | aux-s-1    |
+    When the opening auction period ends for market "ETH/DEC20"
+    Then the market data for the market "ETH/DEC20" should be:
+      | mark price | trading mode            | auction trigger             | horizon | min bound | max bound | target stake | supplied stake | open interest |
+      | 10         | TRADING_MODE_CONTINUOUS | AUCTION_TRIGGER_UNSPECIFIED | 5       | 10        | 10        | 355          | 400            | 1             |
+      | 10         | TRADING_MODE_CONTINUOUS | AUCTION_TRIGGER_UNSPECIFIED | 10      | 10        | 10        | 355          | 400            | 1             |
```
cdummett, line 250: The bounds and stake values are somewhat irrelevant to testing the closeout feature, maybe this instead:
```suggestion
    Then the market data for the market "ETH/DEC20" should be:
      | mark price | trading mode            | open interest |
      | 10         | TRADING_MODE_CONTINUOUS | 1             |
      | 10         | TRADING_MODE_CONTINUOUS | 1             |
```

core/integration/features/verified/closeout-lognormal.feature
```diff
@@ -209,13 +207,104 @@ Feature: Closeout scenarios
...
+      | buy  | 1     | 400    |
+
+    When the parties place the following orders with ticks:
+      | party      | market id | side | volume | price | resulting trades | type       | tif     |
+      | auxiliary2 | ETH/DEC20 | buy  | 12      | 10    | 0                | TYPE_LIMIT | TIF_GTC |
+      | trader2    | ETH/DEC20 | sell | 12      | 10    | 1                | TYPE_LIMIT | TIF_GTC |
+
+    Then the market data for the market "ETH/DEC20" should be:
+      | mark price | trading mode            | auction trigger             | target stake | supplied stake | open interest |
+      | 10         | TRADING_MODE_CONTINUOUS | AUCTION_TRIGGER_UNSPECIFIED | 4623         | 400            | 13            |
```
cdummett, line 278: Similarly here:
```suggestion
    Then the market data for the market "ETH/DEC20" should be:
      | mark price | trading mode            | open interest |
      | 10         | TRADING_MODE_CONTINUOUS | 13            |
```

core/integration/features/verified/closeout-lognormal.feature
```diff
@@ -209,13 +207,104 @@ Feature: Closeout scenarios
...
+    #   | party   | status                        |
+    #   | trader2 | POSITION_STATUS_ORDERS_CLOSED |
+
+    Then the parties should have the following profit and loss:
+      | party   | volume | unrealised pnl | realised pnl | 
+      | trader2 | -12    | 0              | 0            | 
+
+    Then the market data for the market "ETH/DEC20" should be:
+      | mark price | trading mode            | auction trigger             | target stake | supplied stake | open interest |
+      | 10         | TRADING_MODE_CONTINUOUS | AUCTION_TRIGGER_UNSPECIFIED | 4623         | 400            | 13            |
```
cdummett, line 309: Not sure this step adds anything to test the feature?

core/integration/features/verified/closeout-lognormal.feature
```diff
@@ -209,13 +207,104 @@ Feature: Closeout scenarios
...
+      | sell | 1002  | 1      |
+      | sell | 1000  | 10     |
+      | buy  | 5     | 5      |
+      | buy  | 1     | 400    |
+
+    Then the network moves ahead "5" blocks
+
+    # And the parties should have the following position changes for market "ETH/DEC20":
+    #   | party   | status                        |
+    #   | trader2 | POSITION_STATUS_ORDERS_CLOSED |
```
cdummett, line 301: Should this be uncommented now?

core/integration/features/verified/closeout-lognormal.feature
```diff
@@ -209,13 +207,104 @@ Feature: Closeout scenarios
...
+    Then the parties should have the following account balances:
+      | party   | asset | market id | margin | general |  
+      | trader2 | USD   | ETH/DEC20 | 999    | 0       | 
+
+    # trader2's order (price 1003) has been canceled
+    Then the order book should have the following volumes for market "ETH/DEC20":
+      | side | price | volume |
+      | sell | 1003  | 0      |
+      | sell | 1002  | 1      |
+      | sell | 1000  | 10     |
```
cdummett, line 293: I think the trader is still distressed, there just isn't enough volume on the book to close the position.. Does that conflict with the purpose of the test or is that okay?

cdummett (changes requested):
Hey, a few comments re minimising the number of steps in a feature test:

cdummett (dismissed):

core/integration/main_test.go
```diff
@@ -462,6 +462,14 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
+	s.Step(`^the parties should have the following position changes for market "([^)]+)":$`, func(mkt string, table *godog.Table) error {
+		return steps.PartiesShouldHaveTheFollowingPositionStatus(execsetup.broker, mkt, table)
+	})
+
+	s.Step(`^the parties should have the following aggregated position changes for market "([^)]+)":$`, func(mkt string, table *godog.Table) error {
```
witgaw, line 470: Do we really need this if it's not been used even in this PR?

core/integration/main_test.go
```diff
@@ -462,6 +462,14 @@ func InitializeScenario(s *godog.ScenarioContext) {
 		execsetup.assetsEngine.SetPermissive()
 		return nil
 	})
+
+	s.Step(`^the parties should have the following position changes for market "([^)]+)":$`, func(mkt string, table *godog.Table) error {
```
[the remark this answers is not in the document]
Jiajia-Cui: @EVODelavega I have changed the test to only using the "the parties should have the following profit and loss", so we can remove the "the parties should have the following position changes for market" :)

core/integration/main_test.go
EVODelavega: We don't use it in this PR, indeed. I added these steps before the position status was added to the PnL table. I don't think we need to include the position status as a separate step per se, but the aggregated position change step might be useful in future. It allows us to check all states a party passes through in a more complex scenario (e.g. -> party is distressed, but can't be closed out, price changes increase their balance so they are no longer distressed, price changes again, and the party gets closed out). With this step we can ensure the party passes through all the expected states in a single step/check. For that reason, I'd quite like to keep this one.

It's also worth keeping in mind that core, currently, still has the positions plugin in the code, solely for use in the integration tests. Having these steps checking position status specifically might be worth keeping, because the positions plugin is increasingly something we can classify as legacy code.

witgaw: Fair enough, thanks for the explanation.

EVODelavega (approved):

EVODelavega:
Moved to new PR: https://github.com/vegaprotocol/vega/pull/8172

PR #7640, opened 2023-02-21
Change margin requirement to initial (from maintenance) on new order

### LINKED ISSUE #7655: Require initial margin level to be met on new orders
Require initial margin level to be met on new orders unless they'd reduce party's exposure in which case margin check is skipped and order is allowed.

witgaw:
Closes https://github.com/vegaprotocol/vega/issues/7655

### DIFF
--- a/core/positions/market_position.go
+++ b/core/positions/market_position.go
@@ -201,3 +201,32 @@ func (p MarketPosition) VWSell() *num.Uint {
 	}
 	return num.UintZero()
 }
+
+func (p MarketPosition) OrderReducesExposure(ord *types.Order) bool {
+	if ord == nil || p.Size() == 0 || ord.PeggedOrder != nil {
+		return false
+	}
+	// long position and short order
+	if p.Size() > 0 && ord.Side == types.SideSell {
+		// market order reduces exposure and doesn't flip position to the other side
+		if p.Size()-int64(ord.Remaining) >= 0 && ord.Type == types.OrderTypeMarket {
+			return true
+		}
+		// sum of all short limit orders wouldn't flip the position if filled (ord already included in pos)
+		if p.Size()-p.Sell() >= 0 && ord.Type == types.OrderTypeLimit {
+			return true
+		}
+	}
+	// short position and long order
+	if p.Size() < 0 && ord.Side == types.SideBuy {
+		// market order reduces exposure and doesn't flip position to the other side
+		if p.Size()+int64(ord.Remaining) <= 0 && ord.Type == types.OrderTypeMarket {
+			return true
+		}
+		// sum of all long limit orders wouldn't flip the position if filled (ord already included in pos)
+		if p.Size()+p.Buy() <= 0 && ord.Type == types.OrderTypeLimit {
+			return true
+		}
+	}
+	return false
+}
--- a/core/risk/engine.go
+++ b/core/risk/engine.go
@@ -568,7 +568,7 @@ func orderReducesExposure(ord *types.Order, pos events.Margin) bool {
 			return true
 		}
 		// sum of all short limit orders wouldn't flip the position if filled (ord already included in pos)
-		if pos.Size()-pos.Sell() >= 0 {
+		if pos.Size()-pos.Sell() >= 0 && ord.Type == types.OrderTypeLimit {
 			return true
 		}
 	}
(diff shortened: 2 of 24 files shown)


EVODelavega (dismissed):

witgaw (commented):
Ready for another review

davidsiska-vega (dismissed):

EVODelavega:
IIRC, we changed this to maintenance level because previously, parties that had dipped below maintenance/search level were unable to submit orders that closed out their open position manually

core/risk/engine.go
```diff
@@ -554,3 +556,32 @@ func (m marginChange) Transfer() *types.Transfer {
 func (m marginChange) MarginLevels() *types.MarginLevels {
 	return m.margins
 }
+
+func orderReducesExposure(ord *types.Order, pos events.Margin) bool {
```
EVODelavega, line 560: We have this code (or something very similar to it) in the execution engine. If the order reduces exposure, we outright skip the margin check, so I'm not sure if we'd ever hit a case where this returns `true` (in execution/market.go, the `submitValidatedOrder` function):

```go
	pos := m.position.RegisterOrder(ctx, order)
	checkMargin := true
	if !isPegged && ok {
		oldVol, newVol := pos.Size()+pos.Buy()-pos.Sell(), oldPos.Size()+pos.Buy()-pos.Sell()
		if oldVol < 0 {
			oldVol = -oldVol
		}
		if newVol < 0 {
			newVol = -newVol
		}
		// check margin if the new volume is greater, or the same (implying long to short, or short to long)
		checkMargin = oldVol <= newVol
	}

	// Perform check and allocate margin unless the order is (partially) closing the party position
	if checkMargin {
		if err := m.checkMarginForOrder(ctx, pos, order); err != nil {
			if m.log.GetLevel() <= logging.DebugLevel {
				m.log.Debug("Unable to check/add margin for party",
					logging.Order(*order), logging.Error(err))
			}
			_ = m.unregisterAndReject(
				ctx, order, types.OrderErrorMarginCheckFailed)
			return nil, nil, ErrMarginCheckFailed
		}
```

witgaw: oh so it's still here, I thought it disappeared because I wasn't getting the behaviour I expected with the test.

witgaw: Ok so I moved the new method up and using it instead of the above now.

core/positions/market_position.go
```diff
@@ -201,3 +201,32 @@ func (p MarketPosition) VWSell() *num.Uint {
...
+		return false
+	}
+	// long position and short order
+	if p.Size() > 0 && ord.Side == types.SideSell {
+		// market order reduces exposure and doesn't flip position to the other side
+		if p.Size()-int64(ord.Remaining) >= 0 && ord.Type == types.OrderTypeMarket {
+			return true
+		}
+		// sum of all short limit orders wouldn't flip the position if filled (ord already included in pos)
+		if p.Size()-p.Sell() >= 0 && ord.Type == types.OrderTypeLimit {
```
EVODelavega, line 216: This looks a bit off to me: A position of 10 long, with a sell volume on the book of 10, but submitting a sell order for a total volume of 20 would increase the exposure (potential short becomes 20). Shouldn't this check be something like `if p.Size() - (p.Sell() + int64(ord.Remaining)) >= 0`?

core/positions/market_position.go
```diff
@@ -201,3 +201,32 @@ func (p MarketPosition) VWSell() *num.Uint {
...
+		}
+	}
+	// short position and long order
+	if p.Size() < 0 && ord.Side == types.SideBuy {
+		// market order reduces exposure and doesn't flip position to the other side
+		if p.Size()+int64(ord.Remaining) <= 0 && ord.Type == types.OrderTypeMarket {
+			return true
+		}
+		// sum of all long limit orders wouldn't flip the position if filled (ord already included in pos)
+		if p.Size()+p.Buy() <= 0 && ord.Type == types.OrderTypeLimit {
```
EVODelavega, line 227: Same here, we're not including the `order.Remaining` to work out what the new exposure could be. If this is (roughly) a copy-paste of the check we previously had in the execution package, then I suspect this is the reason why the observed behaviour wasn't correct

EVODelavega: Alright, `pos` already includes the order, should've paid more attention. All good

EVODelavega (approved):

davidsiska-vega (approved):

PR #7645, opened 2023-02-21
Fix issue with LP order crossing on entry

### LINKED ISSUE #7647: Panic when LP orders gets capped/floored
It is possible (with a small spread, small LP price range and LP order pegged to MID) to create a situation when LP order crosses on entry triggering a panic.

witgaw:
LP price range was disregarding best bid/ask (only considered MID), hence it was possible to create bounds that would result in LP orders crossing on entry, this PR fixes that issue.

Closes https://github.com/vegaprotocol/vega/issues/7647


witgaw (commented):
Ready for review

davidsiska-vega (approved):

jeremyletang (approved):

PR #7648, opened 2023-02-21
Add special 'block' interval for candles by block

### LINKED ISSUE #7288: Add block interval type
# Feature Overview

**In order to** View hloc data for a single block
**We will** add block height as a special interval type
**So that** candles can be grouped by block height

# Tasks
A checklist of the tasks that are needed to develop the feature and meet the acceptance criteria and feature test scenarios. Ideally, tasks would reflect the issues/pull requests likely to be created when developing the feature. 
- [ ] Add INTERVAL_BLOCK
- [ ] Allow INTERVAL_BLOCK to be used for candlesConnection
- [ ] Allow INTERVAL_BLOCK to be used for the candles subscription 

# Acceptance Criteria
- I can query for candles by block interval
- I can subscribe to candles by block interval

pscott31:
Closes #7288


datanode/gateway/graphql/schema.graphql
```diff
@@ -3061,6 +3061,9 @@ enum Side {
 
 "The interval for trade candles when subscribing via Vega GraphQL, default is I15M"
 enum Interval {
+  "block interval"
```
candida-d, line 3064: How can someone define this, is it by block height, number of blocks, or something else? It would be good to specify in the docs?

protos/sources/vega/vega.proto
```diff
@@ -24,6 +24,8 @@ enum Interval {
   // Default value, always invalid
   INTERVAL_UNSPECIFIED = 0;
   // 1 minute.
+  INTERVAL_BLOCK = -1;
+  // 1 minute.
```
candida-d, line 28: Is this description correct?

datanode/gateway/graphql/schema.graphql
pscott31: It's the interval which corresponds to a single block in the chain, e.g. for trades candles it is all the trades in a block.. so it's not precisely a fixed interval, it depends on how frequently they are generated... I'll improve the docs to make that clear

protos/sources/vega/vega.proto
pscott31: err nope, will fix

ettec (approved):

dexturr (approved):

PR #7649, opened 2023-02-21
fix: network history load performance issues

### LINKED ISSUE #7535: Performance problems when loading data from Network History
The process becomes very slow when loading data from Network History over time.
I have tested it with the situation when there was an empty database, and I was loading data for the past 15-30 hours.
It starts as expected: `2sec/segment`, but after a time (30-60min) it slows down to `15-20sec/segment` and then further to `30sec/segment`.
The segment processing time does not depend on the size of a segment: I run a test where the first half of segments was big `100k rows a segment`, then another half was `2k rows a segment`, and a couple of `100k rows a segment` at the end. At the begging `100k segments` were processed in `1-2sec`, but over time they slowed down to `20sec`. When segments switched to `2k`, the processing time was still `~20sec` and over time slowing to `30sec`, at the end a couple of `100k` segments got processed in the same time `~30sec` as the `2k` segments.

I have been trying to correlate when the first slowness happens (when times jump from `2sec` to `15sec`) with the Order table size, but in one run it was 3.5GB, and in another 2.25 GB.

https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1675846017348689

guoguojin:
Closes #7535 

The problem was caused by a query that was run before the copy command on every insert to ensure that partial loads do not overwrite existing history. The network history loads temporarily removes all indexes to make the bulk load quicker, but without those indexes, the query that ran before the copy would need to scan the table to get the information it needs, this is why the insert time increased as the amount of data inserted increased.

Attached a partial log of the import towards the end of the files showing the time is still consistent even towards the end.

[network-history-load.except.log](https://github.com/vegaprotocol/vega/files/10796252/network-history-load.except.log)

Running the imports with the data I downloaded from stagnet1 last week, I was able to copy all the data from the snapshot history into the database in about 24 minutes where before it was taking several hours.


datanode/networkhistory/snapshot/service_load_snapshot.go
```diff
@@ -272,15 +278,15 @@ func (b *Service) loadSnapshot(ctx context.Context, loadLog LoadLog, snapshotDat
...
 	rowsCopied, err := b.copyDataIntoDatabase(ctx, vegaDbConn, decompressedFilesDestination,
 		filepath.Join(b.snapshotsCopyFromPath, snapshotData.UncompressedDataDir()), dbMetaData,
-		withIndexesAndOrderTriggers)
+		withIndexesAndOrderTriggers, historyTableLastTimestamps)
 	if err != nil {
 		return 0, fmt.Errorf("failed to copy uncompressed data into the database %s : %w", snapshotData.UncompressedDataDir(), err)
 	}
-
-	loadLog.Infof("copied %d rows from %s into database", rowsCopied, snapshotData.UncompressedDataDir())
+	elapsed := time.Since(startTime)
```
EVODelavega, line 288: Do we still need these time logs here, or was that part of the debugging/detective work?

datanode/networkhistory/snapshot/service_load_snapshot.go
```diff
@@ -314,6 +320,28 @@ func getLastPartitionColumnEntryForHistoryTable(ctx context.Context, vegaDbConn
 	return time.Time{}, nil
 }
 
+func getLastHistoryTimestampMap(ctx context.Context, vegaDbConn *pgxpool.Pool, dbMetadata DatabaseMetadata) (map[string]time.Time, error) {
+	lastHistoryTimestampMap := make(map[string]time.Time)
```
EVODelavega, line 324: `map[string]time.Time{}` is shorter, and IMO easier to read

datanode/networkhistory/snapshot/service_load_snapshot.go
```diff
@@ -79,6 +79,12 @@ func (b *Service) LoadSnapshotData(ctx context.Context, log LoadLog, currentStat
 	var totalRowsCopied int64
 	var rowsCopied int64
 	var removedConstraintsAndIndexes *constraintsAndIndexes
+
+	historyTableLastTimestampMap, err := getLastHistoryTimestampMap(ctx, b.connPool, dbMetaData)
```
zale144, line 83: Why do we pass the `b.connPool` and not have `getLastHistoryTimestampMap` to be a method of `*Service` and simply use it directly?

EVODelavega (dismissed):

datanode/networkhistory/snapshot/service_load_snapshot.go
guoguojin: The elapsed time was added for the debugging, but I think it is nicer for the end user to see and if there are problems can be helpful so I kept it in.

datanode/networkhistory/snapshot/service_load_snapshot.go
```diff
@@ -256,7 +262,7 @@ type compressedFileMapping interface {
 }
 
 func (b *Service) loadSnapshot(ctx context.Context, loadLog LoadLog, snapshotData compressedFileMapping, copyFromDirectory string,
-	dbMetaData DatabaseMetadata, vegaDbConn *pgxpool.Pool, withIndexesAndOrderTriggers bool,
+	dbMetaData DatabaseMetadata, vegaDbConn *pgxpool.Pool, withIndexesAndOrderTriggers bool, historyTableLastTimestamps map[string]time.Time,
```
zale144, line 265: I don’t quite get why we have `vegaDbConn *pgxpool.Pool` as a parameter here (and in other methods of `*Service`), and not use `b.connPool`?

zale144 (dismissed):
Apart from those comments I left, looks good.

datanode/networkhistory/snapshot/service_load_snapshot.go
guoguojin: Good point, it was something inherited from the original implementation and I have now fixed.

ettec (approved):

PR #7674, opened 2023-02-23
feat: remove default network parameters

### LINKED ISSUE #7378: Remove unused netparm market.liquidity.targetstake.triggering.ratio
Following https://github.com/vegaprotocol/specs/pull/1534 remove the never used netparam `market.liquidity.targetstake.triggering.ratio`.

Jiajia-Cui:
close #7378 .


davidsiska-vega:
I thought you just meant the price monitoring network parameter - this is a bit drastic. The spec issue is for discussion - I don't think we've necessarily agreed. 

I think this will break stuff (on nodes) - so it would need to be mentioned in changelog. 

Not convinced now is the time to do this - speak with @jeremyletang / @ze97286 / @gordsport .

jeremyletang:
It’s not the time. Also, all network parameters must have a default.

The ticket you linked is different to what was done here as well. Here you remove their default value, this ticket is about removing the parameters themselves if not mistaken.

davidsiska-vega:
> this ticket is about removing the parameters themselves if not mistaken.

Yes, the idea would be to start removing network parameters that have no other purpose than to provide a default for a market parameter. 

I think for now the sane thing now is EITHER:
1. just *add* validation on the market proposal to match the validation on the network param OR 
2. remove the network parameter for default price monitoring config

I don't have an opinion as to what is best to do now, but to not disrupt launch it's probably 1.

Jiajia-Cui:
close with this [PR](https://github.com/vegaprotocol/vega/pull/7672)

PR #7679, opened 2023-02-23
Add CLIs to work with the wallet service configuration

### LINKED ISSUE #7656: Add a command to describe the wallet service config
There is no way to access the service config. It feels like it is hidden. 

1. We need a `vega wallet service config describe` CLI that returns the content of the config 
2. We need a `vega wallet service config reset` CLI that reset the file to its default values.
3. We need a `vega wallet service config locate` CLI that returns the location of the file so the user knows where to look at to edit it.

ValentinTrinque:
Close #7656


cmd/vegawallet/commands/service_config_describe.go
```diff
@@ -0,0 +1,87 @@
...
+	"code.vegaprotocol.io/vega/wallet/service"
+	svcStoreV1 "code.vegaprotocol.io/vega/wallet/service/store/v1"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	describeServiceConfigLong = cli.LongDesc(`
+	    Describe the service configuration.
+	`)
```
EVODelavega, line 20: nit-pick, but why the trailing whitespace in the strings here?

pscott31 (dismissed):

cmd/vegawallet/commands/service_config_describe.go
ValentinTrinque: They stripped by `cli.LongDesc()`. It's easier to get the text right.

jeremyletang (approved):

PR #7706, opened 2023-02-27
Cache ListMarket sql store queries

### LINKED ISSUE #7696: Datanode spends a lot of CPU in ListMarkets
# Problem encountered

Certain bots repeatedly poll the ListMarkets API endpoint. We had a case the other day with ~100 bots running where were were seeing ~200 calls/sec to this API which is high (and the bots are going to be modified not to do that), but we should be able to cope with it.

Turns out parsing JSON that we get back from postgres into structs is quite expensive (~1ms to desterilise the `TradableInstrument`) struct, there were ~20 markets * 200 request/seconds much CPU usage.

We currently cache individual markets, when queried by name but not  the full list of markets. So lets do that so we don't have to reparse JSON each time.

pscott31:
Closes #7696


jeremyletang (approved):

datanode/sqlstore/markets.go
```diff
@@ -21,12 +21,58 @@ import (
...
+func newCacheKey(marketID string, pagination entities.CursorPagination, includeSettled bool) cacheKey {
+	k := cacheKey{
+		marketID:       marketID,
+		newestFirst:    pagination.NewestFirst,
+		includeSettled: includeSettled,
+	}
+
+	if pagination.Forward != nil {
+		if pagination.Forward.Limit != nil {
+			k.forwardOffset = *pagination.Forward.Limit
```
wwestgarth, line 51: Question: Given the limits are part of the cache key, could I as a naughty person fill up the cache and the data-node's memory just by sending in the same valid query over and over again, but just with +1 on the forward limit from `0 -> infinity`?

wwestgarth (dismissed):

jeremyletang (dismissed):

jeremyletang (dismissed):

jeremyletang (approved):

PR #7712, opened 2023-02-28
feat: potential bug, all traders closed out when market is in continuous trading mode

Jiajia-Cui:


Jiajia-Cui:
research team has agreed that its not a bug

PR #7719, opened 2023-03-01
fix: replace the slow query

### LINKED ISSUE #7535: Performance problems when loading data from Network History

EVODelavega:
Closes #7535 (slow query marking the current orders)


AnExsomnis (dismissed):

datanode/networkhistory/snapshot/orders/restore_current_orders_set.go
```diff
@@ -13,21 +13,27 @@ import (
 func RestoreCurrentOrdersSet(ctx context.Context, conn sqlstore.Connection) error {
 	// Note, testing shows that dropping the index on the current flag does not net improve performance of the following update queries
 
+	query := `WITH current_orders AS (
```
ettec, line 16: How does this perform compared to the current query and has the state of the orders table after restore been compared to the source table state to check it's the same?

datanode/networkhistory/snapshot/orders/restore_current_orders_set.go
```diff
@@ -13,21 +13,27 @@ import (
...
+  SELECT id, MAX(vega_time) AS max_vega_time
```
ettec, line 17: How does this perform compared to the current query and has the state of the orders table after restore been compared to the source table state to check it's the same?

ettec (dismissed):

datanode/networkhistory/snapshot/orders/restore_current_orders_set.go
EVODelavega: @guoguojin is testing this ATM. This query completed in 9m37, versus the old version, which is still running (23 minutes and counting). The first of the 2 old queries (setting all `current` fields to false) alone took in excess of 15 minutes, so longer than this single query version.

This query is pretty straightforward in terms of what it does, so I don't expect the results to be any different, but the plan was always to check prior to merging

guoguojin: I can confirm the query is a big improvement and the results are the same.

Test done:

Loaded the snapshot data I took from a couple of weeks ago to fix the bug with the copy time into the using network history load so that we have real data in the orders table. This resulted in 60,624,142 rows in the orders table.

The new query completed much more quickly in both iterations I tried:

```
vega.public> WITH current_orders AS (
                 SELECT id, MAX(vega_time) AS max_vega_time
                 FROM orders
                 WHERE current = true
                 GROUP BY id
                 HAVING COUNT(id) > 1
             )
             UPDATE orders o
             SET current = CASE
                               WHEN o.vega_time = co.max_vega_time AND o.seq_num = (
                                   SELECT MAX(seq_num)
                                   FROM orders
                                   WHERE id = o.id AND vega_time = co.max_vega_time
                               ) THEN true ELSE false
                 END
             FROM current_orders co
             WHERE o.id = co.id AND o.current = true
[2023-03-01 15:53:04] 2,859,040 rows affected in 9 m 37 s 701 ms

vega.public> WITH current_orders AS (
                 SELECT id, MAX(vega_time) AS max_vega_time
                 FROM orders
                 WHERE current = true
                 GROUP BY id
                 HAVING COUNT(id) > 1
             )
             UPDATE orders o
             SET current = CASE
                               WHEN o.vega_time = co.max_vega_time AND o.seq_num = (
                                   SELECT MAX(seq_num)
                                   FROM orders
                                   WHERE id = o.id AND vega_time = co.max_vega_time
                               ) THEN true ELSE false
                 END
             FROM current_orders co
             WHERE o.id = co.id AND o.current = true
[2023-03-02 08:37:59] 1,372,607 rows affected in 18 m 35 s 453 ms
```

The existing query:

```
vega.public> update orders set current = false where current = true
[2023-03-01 16:29:37] 12,368,053 rows affected in 15 m 11 s 581 ms
vega.public> with current_orders as (select r.id, max(r.vega_time) as vega_time from orders r group by id),
             	current_seq_num as (select o.vega_time, max(o.seq_num) as seq_num from orders o join current_orders co on o.id = co.id and o.vega_time = co.vega_time group by co.id, o.vega_time)
             	update orders set current = true from current_seq_num where orders.vega_time = current_seq_num.vega_time and orders.seq_num =  current_seq_num.seq_num
[2023-03-01 17:49:30] 10,879,905 rows affected in 1 h 19 m 53 s 447 ms


vega.public> update orders set current = false where current = true
[2023-03-02 09:40:40] 12,261,330 rows affected in 21 m 48 s 91 ms
vega.public> with current_orders as (select r.id, max(r.vega_time) as vega_time from orders r group by id),
             	current_seq_num as (select o.vega_time, max(o.seq_num) as seq_num from orders o join current_orders co on o.id = co.id and o.vega_time = co.vega_time group by co.id, o.vega_time)
             	update orders set current = true from current_seq_num where orders.vega_time = current_seq_num.vega_time and orders.seq_num =  current_seq_num.seq_num
[2023-03-02 10:35:25] 10,879,905 rows affected in 54 m 28 s 510 ms
```

After exporting the current orders from after each one, there were no diffs in the 2 files.

guoguojin (approved):

datanode/networkhistory/snapshot/orders/restore_current_orders_set.go
ettec: Its a bit late, and my fault for not flagging earlier, but will the new query scale linearly with order table size?  This was a critical aspect of the old slower query   (the query before that did not).

PR #7736, opened 2023-03-01
feat: upgrade ipfs libs to latest releaes

ettec:


jeremyletang:
fixed in: https://github.com/vegaprotocol/vega/pull/7754

PR #7737, opened 2023-03-01
feat: upgrade ipfs to latest version

### LINKED ISSUE #7731: Upgrade IPFS libs
We see intermittent IPFS failures when pulling segments, the cause of which we have yet to pin down.   The IPFS libs we are using are over 0.5 year old and there are a lot of changes/fixes in the underlying p2p lib.  We should upgrade the libs and see if this reduces/eliminates the intermittent failures.

ettec:
closes #7731   Will require that either the existing network history store is erased, or failing that the node will fail with instructions on how to migrate the node to the new version.


guoguojin (dismissed):

guoguojin (approved):

jeremyletang:
fixed in: https://github.com/vegaprotocol/vega/pull/7754

PR #7754, opened 2023-03-03
Feature/ipfs version stuff

### LINKED ISSUE #7731: Upgrade IPFS libs

jeremyletang:
close #7731


guoguojin (dismissed):

guoguojin (approved):

guoguojin:
Just carrying forward information regarding the breaking change. The updated IPFS version will require a migration of the IPFS database. Instructions are provided by the IPFS library. Alternatively, the IPFS database needs to be wiped and rebuilt.

PR #7767, opened 2023-03-06
Assure new parties get included in nearest MTM

### LINKED ISSUE #7766: New orders not included in first MTM calculation
When party posts their first order that does't result in a trade their position won't be included in the first mark-to-market that occurs since posting their order.

witgaw:
Currently if a new party posts an order they won't be included in the nearest MTM - we assume that position has already been marked to market at the current mark price, which need not be true (in fact it definitely won't be true if the mark price changes since previous). As shown in the impacted feature test it could result in that party becoming a closeout counter-party to the network, despite being distressed itself.

Closes https://github.com/vegaprotocol/vega/issues/7766


ze97286 (dismissed):

ze97286 (dismissed):

Jiajia-Cui (approved):
The feature test looks good to me

ze97286 (approved):

PR #7772, opened 2023-03-06
Fix/7686 ipfs retry when cannot connect

### LINKED ISSUE #7686: Can we solve 5% failure rate when pulling history segment at datanode startup?
As per this thread-> https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1677235413898689

About 5% of the time during snapshot pipeline run, IPFS will fail to fetch the latest history segment at startup.  Previous effort was expended to see if altering config or identification of a bug could resolve this without success (though a fresh pair eyes....).  However, given the number of moving parts with IPFS its probably a good idea to build in some resilience.  Right now we are effectively doing that in sys tests by restarting the node when this startup failure occurs.   Could we effectively move that restart behaviour into the node by tearing down and recreating the IPFS store (/datanode/networkhistory/store/Store.go) created by datanode to achieve the same thing?

guoguojin:
Closes #7686


wwestgarth:
Just wondering: Have you been able to catch the timeouts locally while initializing against testnet?

wwestgarth (dismissed):

jeremyletang (approved):

PR #7774, opened 2023-03-06
fix: save and restore collateral bridge starting block from snapshot

### LINKED ISSUE #7751: Possible lost deposits during a whole network snapshot-restore
# Problem encountered
As mentioned here: https://vegaprotocol.slack.com/archives/CEE07A084/p1677836123037479?thread_ts=1677825165.254859&cid=CEE07A084

there were reports of deposits getting lost around the time of the testnet outage. 

Theres not much to go on but my _hunch_ is that when the entire network was started from a snapshot the EEF didn't restore correctly. So any deposits made during the downtime were not getting picked up upon restore.

If true, this bug would also present during a protocol upgrade.

The snapshot and some block data from the point at which the network restarted is here: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/private%2FSnapshots%2FFairground/detail/Fairground/15665/pipeline/111/

I've also attached the artefacts to the ticket incase jenkins cleans them up.

Two deposits that got lost:
https://sepolia.etherscan.io/tx/0x5f7fe077a21db8bc758505e1d8d3c1c00f74efe5f87c4bc6fee444c9e5678d15
10c36272e050adafe1b274a6194939e22d356a0ed8e3bacb7e02510abb9818ee

https://sepolia.etherscan.io/tx/0x9597a9e9af0f002020b01a1d9b3f69ae94dc9541e178a4447bdfbcb3ad3bb16a
58aaf8e18fcc89ad099cbdeaba658830330499e7e6f4ca7b1bfa0820f1daa670

# Observed behaviour
When testnet went down, any deposits made during/near were not seen by the vega chain after it was restored via snapshots.

# Expected behaviour
All deposits are always picked up.

# System response

# Steps to reproduce

## Manual

## Automation

# Evidence

## Logs

## Additional context

# Definition of Done

**Before Merging**

**After Merging**

wwestgarth:
closes #7751 

During network downtime any deposit made on the collateral bridge would get lost when the network was started from a snapshot. This was because we were not restoring a starting block height in the EEF and it would default to the current-block-height.

Snapshotting the EEF is a bit awkward because it is only run by validator nodes so we cannot save the state directly We do what we do with the other contracts and infer a block height from the last seen ERC20 chain-event. We save that height in the banking engines snapshot, then push it into the EEF on restore.

The following test now works:
- stop the network
- wait 10 seconds
- do a deposit
- start the network
- watch the deposit appear on the vega chain

https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/56627/pipeline

I've also fixed some off-bby-one errors in the other engines where we were restoring the block height to the one where we saw an event, meaning we would instantly re-pick up that event. It was harmless because the `fowarder` would block it as an event already seen, but whatever.


jeremyletang (dismissed):

jeremyletang:
looks good. buf lint not passing though.

ze97286 (commented):
about the off-by-one - wouldn't we potentially be ignoring events from the last block height which would have been received when vega was already down?

wwestgarth:
> about the off-by-one - wouldn't we potentially be ignoring events from the last block height which would have been received when vega was already down?

I think you're right, it does leave the tiniest opportunity for one to get lost, reverted the +1s.

ValentinTrinque (approved):

PR #7780, opened 2023-03-06
Improve bond penalty mechanics

### LINKED ISSUE #7745: Bug: LP did not get closed out when they should
see feature test below (currently in branch 7588-exit-price, will be soon merged into develop)

core/integration/features/liquidity-provision/lp-distressed-closeout.feature

on line 142, with additional LP commitment from party1, market get out of auction, and a new trade, party0 is still not closed out

witgaw:
- We were not recalculating margin events following the application of a bond penalty, which meant LP could be falsely perceived as having sufficient collateral during margin check.
- We never deleted a bond account, but logic within collateral engine uses the existence of it to assert that someone is an LP. That means parties that used to be LPs would get a bond penalty applied whenever they had a margin shortfall.

Closes https://github.com/vegaprotocol/vega/issues/7745

### DIFF
--- a/core/execution/liquidity_provision.go
+++ b/core/execution/liquidity_provision.go
@@ -658,6 +658,7 @@ func (m *Market) cancelLiquidityProvision(
 			return err
 		}
 		m.broker.Send(events.NewLedgerMovements(ctx, []*types.LedgerMovement{tresp}))
+		m.collateral.RemoveBondAccount(party, m.GetID(), asset)
 	}
 
 	// now let's update the fee selection
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -146,6 +146,7 @@ type MarketCollateral interface {
 	GetPartyGeneralAccount(party, asset string) (*types.Account, error)
 	GetPartyBondAccount(market, partyID, asset string) (*types.Account, error)
 	BondUpdate(ctx context.Context, market string, transfer *types.Transfer) (*types.LedgerMovement, error)
+	RemoveBondAccount(partyID, marketID, asset string) error
 	MarginUpdateOnOrder(ctx context.Context, marketID string, update events.Risk) (*types.LedgerMovement, events.Margin, error)
 	GetPartyMargin(pos events.MarketPosition, asset, marketID string) (events.Margin, error)
 	GetPartyMarginAccount(market, party, asset string) (*types.Account, error)
--- a/core/execution/market.go
+++ b/core/execution/market.go
@@ -1731,7 +1731,27 @@ func (m *Market) confirmMTM(
 				m.log.Error("Failed to perform bond slashing",
 					logging.Error(err))
 			}
-			m.broker.Send(events.NewLedgerMovements(ctx, transfers))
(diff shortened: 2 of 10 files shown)


witgaw (commented):
Ready for review

witgaw:
Btw. were we not meant to not rely on events in the internal logic?

witgaw:
> > Btw. were we not meant to not rely on events in the internal logic?
> 
> we shall never rely on the broker for internal logic indeed. did you see any offenses to that rules?

No, not intercepting anything from broker or anything like that. Just the events passed around collateral engine.

core/execution/liquidity_provision.go
```diff
@@ -658,6 +658,7 @@ func (m *Market) cancelLiquidityProvision(
 			return err
 		}
 		m.broker.Send(events.NewLedgerMovements(ctx, []*types.LedgerMovement{tresp}))
+		m.collateral.RemoveBondAccount(party, m.GetID(), asset)
```
jeremyletang, line 661: you don't even check the error for it...

core/execution/market.go
```diff
@@ -1731,7 +1732,27 @@ func (m *Market) confirmMTM(
 				m.log.Error("Failed to perform bond slashing",
 					logging.Error(err))
 			}
-			m.broker.Send(events.NewLedgerMovements(ctx, transfers))
+			// if bond slashing occurred then amounts in "closed" will not be accurate
+			if len(transfers) > 0 {
+				m.broker.Send(events.NewLedgerMovements(ctx, transfers))
+				asset, _ := m.mkt.GetAsset()
+				closedRecalculated := make([]events.Margin, 0, len(closed))
```
jeremyletang, line 1739: That might be what you mean about the events being used for internal stuff it's fine here, it's not very pretty, and it should probably defined somewhere else but these are just interface of data being carried around the core.

core/execution/market.go
```diff
@@ -1731,7 +1732,27 @@ func (m *Market) confirmMTM(
...
+					recalculated := c
+					if pos, ok := m.position.GetPositionByPartyID(c.Party()); ok {
+						margin, err := m.collateral.GetPartyMargin(pos, asset, m.mkt.ID)
+						if err != nil {
+							m.log.Error("couldn't get party margin",
+								logging.PartyID(c.Party()),
+								logging.Error(err))
+							continue
+						}
+						recalculated = margin
```
jeremyletang, line 1750: why the unecessary copy here?
just happen margin in closedRecalculated instead?

core/execution/market.go
witgaw: Cool, thanks for confirming

core/execution/market.go
witgaw: Wanted to keep the old value in case we weren't able to recalculate - so we don't make the "closed" list shorter. Replaced with two separate appends now.

core/execution/liquidity_provision.go
witgaw: Changed to panic now as you recommended

jeremyletang (approved):

PR #7793, opened 2023-03-07
feat: park lp orders when not on book

jeremyletang:
This PR does 2 things:
- set the status of LP order to parked while they are undeployed
- set the createdAt field of lp orders to the creation date of the LiquidityProvision


witgaw (dismissed):
LGTM

wwestgarth:
Really minor: the proto doc string for PARKED needs updating not that its not just pegged orders:
```
// Order has been removed from the order book and has been parked, this applies to pegged orders only
    STATUS_PARKED = 8;
```

wwestgarth (dismissed):

witgaw (dismissed):
LGTM

ze97286 (approved):

PR #7808, opened 2023-03-09
7588 exit price

### LINKED ISSUE #7588: Slippage_per_unit is calculated in the old way by using exiting orders on the book rather than `slippage_per_unit = +Infinity`

EVODelavega:
Closes #7588


Jiajia-Cui:
a few feature tests are still failing, and in test: 0026-AUCT-auction_interaction.feature there are 4 scenarios still show the bug (liquidity auction does not end even with enough supplied stake)

Jiajia-Cui (approved):
feature tests look good to me

jeremyletang (approved):

PR #7831, opened 2023-03-13
Add ListTransactionEntities endpoint

### LINKED ISSUE #7480: Filter by transaction ID
We need to be able filter API requests buy transaction ID

details TBC...

karlem:
close #7480 .


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2282,6 +2284,28 @@ message ObserveDelegationsResponse {
 
 // -- Nodes --
 
+// Represents a basic node without any epoch specific details like delegations, staking, rewards etc.
+message NodeBasic {
```
wwestgarth, line 2288: Is there a reason we can't use the normal `message Node` and just not fill in the epoch specific bits?

karlem: Well I guess what epoch? The latest one?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -494,6 +494,8 @@ service TradingDataService {
     // Response contains the bootstrap peers
     rpc NetworkHistoryBootstrapPeers(NetworkHistoryBootstrapPeersRequest) returns (NetworkHistoryBootstrapPeersResponse);
 
+    rpc ListTransactionEntities(ListTransactionEntitiesRequest) returns (ListTransactionEntitiesResponse);
```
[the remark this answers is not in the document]
wwestgarth: I also wonder to stay consistent with all the other API's whether this should just be `ListEntities()` with the required `transaction_hash` saying it returns them by-transaction.

protos/sources/data-node/api/v2/trading_data.proto
karlem: I am honestly not sure whether we need all the other stuff. Especially because this endpoint should return the entities changed by a tx hash.

karlem: Hm, I though this might be more clear tbh. It might look like a bug when people never see the epoch specific bit.

candida-d (commented):
Fixes to error messages

datanode/entities/utils.go
```diff
@@ -172,3 +172,7 @@ func (h *TxHash) DecodeBinary(ci *pgtype.ConnInfo, src []byte) error {
 func NanosToPostgresTimestamp(nanos int64) time.Time {
 	return time.Unix(0, nanos).Truncate(time.Microsecond)
 }
+
+func toPointer[T any](v T) *T {
+	return &v
+}
```
pscott31, line 178: We have this in libs/ptr: 
```go
package ptr

func From[T any](t T) *T {
	return &t
}
```

datanode/sqlstore/accounts.go
```diff
@@ -64,13 +64,6 @@ func (as *Accounts) Add(ctx context.Context, a *entities.Account) error {
 }
 
 func (as *Accounts) GetByID(ctx context.Context, accountID entities.AccountID) (entities.Account, error) {
-	if account, ok := as.getAccountFromCache(accountID); ok {
-		return account, nil
-	}
-
-	as.cacheLock.Lock()
-	defer as.cacheLock.Unlock()
-
```
pscott31, line 73: This cache is pretty critical - without it we are constantly hammering the DB for account IDs; though perhaps there's some more cunningness in the PR I haven't read yet that makes us not need it

karlem: As in the code still tries to read from the cache later no.

datanode/gateway/graphql/schema.graphql
```diff
@@ -3467,6 +3490,35 @@ type Proposal {
...
+  rationale: ProposalRationale!
+  "Why the proposal was rejected by the core"
+  rejectionReason: ProposalRejectionReason
+  "Error details of the rejectionReason"
+  errorDetails: String
+  "Required majority for this proposal to succeed"
+  requiredMajority: String!
+  "Required participation for this proposal to succeed"
+  requiredParticipation: String!
+  "Required liquidity provider equity like share majority for this proposal to succeed"
```
candida-d, line 3516:
```suggestion
  "The equity-like share required for a market amendment proposal to be enacted (if not met, the proposal will not be enacted), represented as a fraction of the total LPs for the market"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3465,9 +3488,38 @@ type Proposal {
...
+  "Error details of the rejectionReason"
+  errorDetails: String
+  "Required majority for this proposal to succeed"
+  requiredMajority: String!
+  "Required participation for this proposal to succeed"
+  requiredParticipation: String!
+  "Equity-like share required for a market amendment proposal to be enacted (if not met, the proposal will not be enacted), represented as a fraction that can be converted to a percentage"
   requiredLpMajority: String
-  "Required liquidity provider equity like share participation for this proposal to succeed"
+  "The market share of LPs' equity-like share that must take part in the market amendment vote for the proposal to pass. This means the votes of LPs that have submitted more liquidity to that market, or have been LPs from the start carry more weight. If it requires 50% of a market's equity-like share for a majority, and the proposal receives only YES votes but only LPs with 49% of the equity-like share voted, the proposal will not pass"
```
candida-d, line 3522:
```suggestion
  "Market share of LPs' equity-like share that must take part in the market amendment vote for the proposal to pass. This means the votes of LPs that have submitted more liquidity to that market, or have been LPs from the start carry more weight. Represented as a fraction that can be converted to a percentage"
```

candida-d (commented):
Ok - last batch hopefully. Trimmed down the very long annotation on ELS majority

pscott31 (dismissed):

jeremyletang (approved):

PR #7832, opened 2023-03-14
fix: add live liquidity provisions table

### LINKED ISSUE #7326: current_liquidity_provisions should take LP state into account
@Phil spotted that current_liquidity_provisions should take LP state into account.

If this is not done, over time the current liquidity provisions table will continue to grow as LPs are created/deleted, this in turn will cause the network history segments to grow unnecessarily.  Stats indicate that this is not an immediate concern as the growth is low, however diffferent app usage pattern could change this.

AnExsomnis:
close: #7326


pscott31 (dismissed):

datanode/sqlstore/migrations/0010_live_liquidity_provisions.sql
```diff
@@ -0,0 +1,57 @@
...
+CREATE OR REPLACE FUNCTION update_live_liquidity_provisions()
+   RETURNS TRIGGER
+   LANGUAGE PLPGSQL AS
+$$
+BEGIN
+
+DELETE FROM live_liquidity_provisions
+WHERE id = NEW.id;
+
+-- We take into consideration Liquidity provisioons with statuses:
```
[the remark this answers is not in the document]
AnExsomnis: This has been fixed, thanks!

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -1241,7 +1241,7 @@ create table current_liquidity_provisions
     reference text,
     tx_hash bytea not null,
     vega_time timestamp with time zone not null,
-    primary key (id)
+    primary key (id, vega_time)
```
ettec, line 1244: This change should be made as part of the 0010_live_liquidity_provisions.sql  migration  (unless I have missed something?)

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -1282,7 +1282,7 @@ INSERT INTO current_liquidity_provisions(id,
                                       NEW.reference,
                                       NEW.tx_hash,
                                       NEW.vega_time)
-    ON CONFLICT(id) DO UPDATE SET
+    ON CONFLICT(id, vega_time) DO UPDATE SET
```
ettec, line 1285: ditto - should be added to the 0010 migration

datanode/sqlstore/migrations/0010_live_liquidity_provisions.sql
```diff
@@ -0,0 +1,57 @@
...
+                NEW.market_id, NEW.commitment_amount, NEW.fee, NEW.sells,
+                NEW.buys, NEW.version, NEW.status, NEW.reference, NEW.tx_hash, NEW.vega_time);
+END IF;
+
+RETURN NEW;
+END;
+$$;
+-- +goose StatementEnd
+
+select create_hypertable('current_liquidity_provisions', 'vega_time', chunk_time_interval => INTERVAL '1 day');
```
ettec, line 48: convention in datanode is that any table that is 'current'  does not contain contain historical data, i.e. is not a hypertable.  Ideally this table would be called just 'liquidity_provisions', could could we add an altertable command to the migration to rename the table?

ettec (changes requested):

datanode/sqlstore/migrations/0010_live_liquidity_provisions.sql
```diff
@@ -0,0 +1,57 @@
+-- +goose Up
+
+create table live_liquidity_provisions
```
ettec, line 3: probably my misunderstanding, but I ee this table has been added, but does not seem it is used anywhere?  I would have expected to see a change to the corresponding sqlstore for liquidity provisions?

datanode/sqlstore/migrations/0001_initial.sql
AnExsomnis: This has been corrected, thanks!

ettec (approved):

PR #7838, opened 2023-03-14
7816 require slippage factors

### LINKED ISSUE #7816: Set sensible core defaults for market parameters `linear` and `quadratic`
Following https://github.com/vegaprotocol/vega/pull/7808 we're now calculating slippage and using the cap as per spec. 

When the cap factors were first introduced core included deliberately high default values to not change necessitate changing every feature test (cc @ze97286).

There is one bit of housekeeping to do: set default value of `linear = 0.001` and `quadratic = 0.0`. Then modify all feature tests to include custom values which make the test happy.
```
And the markets:
      | id        | quote name | asset | risk model                | margin calculator   | auction duration | fees         | price monitoring | data source config     | linear slippage factor | quadratic slippage factor |
      | ETH/DEC19 | ETH        | USD   | lognormal-risk-model-fish | margin-calculator-1 | 1                | default-none | default-none     | default-eth-for-future | 1e6                    | 1e6                       |

```

EVODelavega:
Closes #7816 

Makes slippage factors required in system & integration tests

This probably will require some work on the system-test side

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [7731](https://github.com/vegaprotocol/vega/issues/7731) - Upgrade the interplanetary file system library to latest release
 - [7802](https://github.com/vegaprotocol/vega/issues/7802) - Split liquidity auction trigger into two cases
 - [7728](https://github.com/vegaprotocol/vega/issues/7728) - Remove current order flag from table - adds restrictions to how orders can be paged
+- [7816](https://github.com/vegaprotocol/vega/issues/7816) - Require slippage factors to alays be set
 
 ### 🗑️ Deprecation
 - [](https://github.com/vegaprotocol/vega/issues/xxxx) -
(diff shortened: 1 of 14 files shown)


wwestgarth:
I think this might cause problems restoring from a checkpoint which has a market proposal in it which doesn't have the slippage factors?

EVODelavega:
> I think this might cause problems restoring from a checkpoint which has a market proposal in it which doesn't have the slippage factors? I'm not sure how we've handled breaking-checkpoint changes in the past, but whatever it was we should remember to to that.

I don't really see how that could cause an issue here. The checkpoint can only contain accepted proposals (ie proposals that passed validation), which by definition will have the slippage factors set. Even if they are set to zero, the proto in the checkpoint will have a non-zero length string (`"0"`).

wwestgarth:
> > I think this might cause problems restoring from a checkpoint which has a market proposal in it which doesn't have the slippage factors? I'm not sure how we've handled breaking-checkpoint changes in the past, but whatever it was we should remember to to that.
> 
> I don't really see how that could cause an issue here. The checkpoint can only contain accepted proposals (ie proposals that passed validation), which by definition will have the slippage factors set. Even if they are set to zero, the proto in the checkpoint will have a non-zero length string (`"0"`).

I was more thinking when we restore from a checkpoint/snapshot-via-protocol-upgrade when stagnet gets restarted with this new version of core.

But anyway, I missed the `len()` bit and thought the validation was checking if the slippage value was 0. So ignore me.

vega-paul:
```{"system-tests":"supply_slippage"}```

witgaw (approved):

jeremyletang (approved):

PR #7849, opened 2023-03-15
WIP: position estimate

### LINKED ISSUE #7847: Refactor `EstimateOrder` API method
* Rename to `EstimatePosition`
* Return maintenance margin level per specified orders and open volume,
   * for open volume part carry out calculation with 0 slippage and slippage cap, hence the result will be a range - which will effectively be a single point if open volume==0).
* Return liquidation price
    * it is based on open volume open and ignores any open orders (as these get cancelled when party gets distressed and party's margin then gets re-evaluated to check if party is still distressed),
    * calculation is carried out assuming both 0 slippage and the slippage cap, hence the result is a range.

witgaw:
Closes https://github.com/vegaprotocol/vega/issues/7847


core/risk/margins_calculation.go
```diff
@@ -121,25 +115,80 @@ func (e *Engine) calculateMargins(m events.Margin, markPrice *num.Uint, rf types
...
 					}
 				}
 			}
-			slippagePerUnit, negSlippage = num.UintZero().Delta(markPrice, exitPrice)
+			longSlippagePerUnit, _ = num.UintZero().Delta(markPrice, exitPrice)
+		}
+	}
+	// calculate margin maintenance short only if riskiest is < 0
+	// marginMaintenanceSht will be 0 by default
+	if riskiestSht.IsNegative() {
```
jeremyletang, line 130: I can't stop myself to read that as `riskiest shit`. Maybe not shortening the name here would be nicer.

davidsiska-vega: True that but I thought our aim is to make changes minimal 🤣 .

witgaw: I'll just rename the variable to `riskiestShit` to avoid confusion.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -659,10 +659,14 @@ service TradingDataService {
...
-      };
-    }
+        option deprecated = true;
+    };
+
+    // Estimate position
+    //
+    // Estimate the margin that would be required for maintaining the specified position
+    // if the optional collataral available argument is supplied the reposnse also contains the estimate of the liquidation price
+    rpc EstimatePosition(EstimatePositionRequest) returns (EstimatePositionResponse);
```
jeremyletang, line 669: Please add back the tags here.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3316,3 +3322,31 @@ message PingRequest {
...
+message LiquidationEstimate {
+    LiquidationPrice worst_case = 1;
+    LiquidationPrice best_case = 2;
+}
+
+message LiquidationPrice {
+    string position_only = 1;
+    string including_buy_orders = 2;
+    string including_sell_orders = 3;
+}
```
jeremyletang, line 3352: All fields needs to be documented with a doc commented (ended with a `.` so the documentation is generated properly.)

jeremyletang (changes requested):
Lots of changes in the risk package which make me not really at ease with the PR.
Also lots of variable have been renamed and introduce lots of changes in the diff, while I appreciate the changes in names, it makes it very difficult to review what were the actual changes introduced here. Any chance you could rename things back as they were? or propose the renames in a separate PR, that we'd merge first?

witgaw:
> Lots of changes in the risk package which make me not really at ease with the PR.
> Also lots of variable have been renamed and introduce lots of changes in the diff, while I appreciate the changes in names, it makes it very difficult to review what were the actual changes introduced here. Any chance you could rename things back as they were? or propose the renames in a separate PR, that we'd merge first?

Sure, I'll isolate risk package changes from the API ones.

witgaw:
> > > Lots of changes in the risk package which make me not really at ease with the PR.
> > > Also lots of variable have been renamed and introduce lots of changes in the diff, while I appreciate the changes in names, it makes it very difficult to review what were the actual changes introduced here. Any chance you could rename things back as they were? or propose the renames in a separate PR, that we'd merge first?
> > 
> > Sure, I'll isolate risk package changes from the API ones.
> 
> Also aside from that, we've been trying to "freeze" any code change on the core and limit them to only bug fixes pretty much now while getting to the mainnet release. I know we have quite a bunch of coverage for this part of the code, but I still have to say that changing code which "works" makes me uneasy at this point in time, even more so just to use it from an API.

So what's the decision? Should I use duplicated code for API? Add regression tests on develop first to give you more confidence?

witgaw:
Broken down into:
https://github.com/vegaprotocol/vega/pull/8108
https://github.com/vegaprotocol/vega/pull/8110

PR #7852, opened 2023-03-15
feat: implement post only and reduce only orders

### LINKED ISSUE #7851: PostOnly and ReduceOnly orders
Do this: https://github.com/vegaprotocol/specs/pull/1562

jeremyletang:
close #7851


core/positions/market_position.go
```diff
@@ -230,3 +230,20 @@ func (p MarketPosition) OrderReducesExposure(ord *types.Order) bool {
 	}
 	return false
 }
+
+// OrderReducesOnlyExposure returns true if the order reduce the position and the extra size if it was to flip the position side.
+func (p MarketPosition) OrderReducesOnlyExposure(ord *types.Order) (reduce bool, extraSize uint64) {
+	// if already closed, or increasing position, we shortcut
+	if p.Size() == 0 || (p.Size() < 0 && ord.Side == types.SideSell) || (p.Size() > 0 && ord.Side == types.SideBuy) {
```
AnExsomnis, line 237: Comparison operators and logical AND are with higher precedence than logical OR, but thats okay as well.

AnExsomnis (dismissed):

core/positions/market_position.go
AnExsomnis: No, in case for better readability makes sense :+1:

core/positions/market_position.go
```diff
@@ -230,3 +230,20 @@ func (p MarketPosition) OrderReducesExposure(ord *types.Order) bool {
 	}
 	return false
 }
+
+// OrderReducesOnlyExposure returns true if the order reduce the position and the extra size if it was to flip the position side.
+func (p MarketPosition) OrderReducesOnlyExposure(ord *types.Order) (reduce bool, extraSize uint64) {
```
ze97286, line 235: I think this would make more sense as:

```
func (p MarketPosition) GetOrderExposureReducingSize(ord *types.Order) uint64 {
     if p.Size() == 0 || (p.Size() < 0 && ord.Side == types.SideSell) || (p.Size() > 0 && ord.Side == types.SideBuy) {
             return 0
     }
     size := p.Size()
     if size < 0 {
        size = -size
     }
    extra := int64(ord.Remaining)-size
    if extra < 0 {
        return 0
    }
    return extra
}
     
```

jeremyletang: hum correct. my brain is broken.

EVODelavega: OK, we seem to implement this logic in different places. It used to sit in execution package, I believe @witgaw moved it to risk (not sure exactly where, though). Maybe we should move it to a receiver on the order type, and pass it a position as an argument

datanode/sqlstore/migrations/0010_post_reduce_only.sql
```diff
@@ -0,0 +1,85 @@
...
+CREATE OR REPLACE VIEW orders_current_desc_by_reference
+AS
+SELECT DISTINCT ON (orders.created_at, orders.reference, orders.id) *
+        FROM orders
+        ORDER BY orders.created_at DESC, orders.reference, orders.id, orders.vega_time DESC, orders.seq_num DESC;
+
+
+-- +goose Down
+
+drop view orders_current_versions;
```
EVODelavega, line 43: nit-pick: if we drop views explicitly, I would not use `CREATE OR REPLACE`, we drop the views, there should be nothing to replace. If there is, then something is broken

core/positions/market_position.go
jeremyletang: hum. they are doing slighlty different things to be fair.

datanode/sqlstore/migrations/0010_post_reduce_only.sql
jeremyletang: right, can change this.

EVODelavega (approved):

PR #7853, opened 2023-03-15
fix: network history load error handling

ettec:
Fix to stop the defer masking the underlying error

### DIFF
--- a/datanode/networkhistory/snapshot/service_load_snapshot.go
+++ b/datanode/networkhistory/snapshot/service_load_snapshot.go
@@ -497,8 +497,8 @@ func copyDataIntoDatabase(ctx context.Context, vegaDbConn sqlstore.Connection, c
 		return 0, fmt.Errorf("failed to disable triggers, setting session replication role to replica failed: %w", err)
 	}
 	defer func() {
-		_, err = vegaDbConn.Exec(ctx, "SET session_replication_role = DEFAULT;")
-		if err != nil {
+		_, triggersErr := vegaDbConn.Exec(ctx, "SET session_replication_role = DEFAULT;")
+		if err == nil && triggersErr != nil {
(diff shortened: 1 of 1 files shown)


datanode/networkhistory/snapshot/service_load_snapshot.go
```diff
@@ -497,10 +497,12 @@ func copyDataIntoDatabase(ctx context.Context, vegaDbConn sqlstore.Connection, c
 		return 0, fmt.Errorf("failed to disable triggers, setting session replication role to replica failed: %w", err)
 	}
 	defer func() {
-		_, err = vegaDbConn.Exec(ctx, "SET session_replication_role = DEFAULT;")
-		if err != nil {
-			err = fmt.Errorf("failed to re-enable triggers, setting session replication role to default failed: %w", err)
-			rowsCopied = 0
+		_, triggersErr := vegaDbConn.Exec(ctx, "SET session_replication_role = DEFAULT;")
+		if err == nil {
+			if triggersErr != nil {
```
guoguojin, line 502: Wouldn't it be easier to have an && here?

if err == nil && triggersErr != nil

datanode/networkhistory/snapshot/service_load_snapshot.go
```diff
@@ -530,7 +532,7 @@ func copyDataIntoDatabase(ctx context.Context, vegaDbConn sqlstore.Connection, c
 		}
 	}
 
-	return totalRowsCopied, nil
+	return totalRowsCopied, fmt.Errorf("it broke son")
```
guoguojin, line 535: This doesn't look right, you want to always return an error?

ettec: bollux, that definitely aint right ;-)

guoguojin (approved):

PR #7891, opened 2023-03-20
fix: fix precision of ref price in GetMarketData

### LINKED ISSUE #7829: Reference price from price monitoring bounds propagated as an event with incorrect precision
Min/max valid price for each bound gets modified so that the correct number of decimal places is returned, but not there reference price. It should get the same treatment as the values mentioned above.

https://github.com/vegaprotocol/vega/blob/eeb4a4d9cc396d0e5621eb756e3e755ecf9e55a1/core/execution/market.go#L549-L555

AnExsomnis:
close: #7829


witgaw (approved):
LGTM

jeremyletang (approved):

PR #7896, opened 2023-03-20
Fix auction margins

### LINKED ISSUE #7894: Use slippage cap when market is in auction mode

### LINKED ISSUE #7895: Fix margin calculation during auction
Orders are included twice (once with mark price, once with vwap price) in margin calculation

witgaw:
* Apply slippage cap in auction (no orders on the book)
* Don't double count orders part of margin during auction (used to calculated with mark price once and then again with vwap, should only be vwap)

Closes https://github.com/vegaprotocol/vega/issues/7894 
Closes https://github.com/vegaprotocol/vega/issues/7895

### DIFF
--- a/core/risk/margins_calculation.go
+++ b/core/risk/margins_calculation.go
@@ -78,27 +78,22 @@ func (e *Engine) calculateMargins(m events.Margin, markPrice *num.Uint, rf types
 		var (
 			slippageVolume  = num.MaxD(openVolume, num.DecimalZero())
 			slippagePerUnit = num.UintZero()
+			noExit          = true
 			negSlippage     bool
-			noExit          bool
 		)
 		if slippageVolume.IsPositive() {
-			var (
-				exitPrice *num.Uint
-				err       error
-			)
-			if auction {
-				noExit = true
-			} else {
+			if !auction {
 				svol, _ := num.UintFromDecimal(slippageVolume.Abs().Mul(e.positionFactor))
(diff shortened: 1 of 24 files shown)


core/risk/margins_calculation.go
```diff
@@ -116,7 +116,7 @@ func (e *Engine) calculateMargins(m events.Margin, markPrice *num.Uint, rf types
 				err       error
 			)
 			if auction {
-				exitPrice = e.ob.GetIndicativePrice()
+				noExit = true
```
EVODelavega, line 119: I'm not sure about this. If we can get an accurate exit price during an auction, why would we not use it?

EVODelavega: Ah, that makes sense. Although, doing things this way, there might be an incongruity WRT opening auctions. IIRC, the margin requirements for orders during opening auctions were supposed to be based on the order price (if we don't have an indicative price - so until we have uncrossing orders on the book), and after that, the margin will be based on the indicative price.

I'll check the spec if that's mentioned anywhere, but if so, we might need to update that.

witgaw: It's always meant to be vwap for orders now in auctions, though that might be on the `cosmicelevator` spec branch.

EVODelavega (dismissed):

davidsiska-vega (dismissed):
Feature tests LGTM.

davidsiska-vega (approved):

jeremyletang (approved):

PR #7905, opened 2023-03-21
Add integration test template that matches system tests

### LINKED ISSUE #7904: Create an easy way to replicate system tests as an integration test
To be able to quickly and accurately determine what is causing a system test to fail after a breaking change, we need to be able to verify what is causing a previously working test to fail. For this reason, having a system-test base template for the integration tests is a must.

This template should consist of:

* A like-for-like match of the relevant network parameters
* Create the default market that the system tests run against (same asset,  decimal places, initial orders, LP, ...)
* The  template should create the market, place the orders, and leave opening auction.

EVODelavega:
Closes #7904


Jiajia-Cui (dismissed):
LGTM

Jiajia-Cui (approved):
LGTM

jeremyletang (approved):

PR #7909, opened 2023-03-21
fix: add missing graphql details for liq prov list

AnExsomnis:


ettec (approved):

datanode/gateway/graphql/schema.graphql
```diff
@@ -1792,6 +1792,8 @@ type Market {
    liquidityProvisionsConnection(
      "An optional party ID"
      partyId: ID
+     "An optional live flag to determine whether to list only live LPs or not"
```
mattrussell36, line 1795: suggestion (non blocking): would be good to perhaps include the 'live' status here?

```suggestion
     "An optional live flag to determine whether to list only live LPs (Active, Undeployed, Pending) or not"
```

mattrussell36 (approved):

PR #7914, opened 2023-03-21
Offer signatures snaphost

### LINKED ISSUE #7758: List asset signatures go missing if an asset proposal is enacted on the same block as an upgrade
# Problem encountered
We have a test which (accidentally, as it happens) sometimes enacts an ERC20 asset on an upgrade block. When it does, the list asset bundle is never found after the upgrade has happened.

# Observed behaviour
`    def wait_for_signature_bundle(self, update=False):
        """
        Retrieves the signature bundle to list or update the asset on the smart contract.
        """
        if not update:
            # Create/list asset
            sig_call = networkMod.get_list_asset_bundle
        else:
            # Update asset
            sig_call = networkMod.get_set_asset_limit_bundle
        # Signature bundle may be empty until data node catches up with core
        timeout = 5
        now = time.time()
        result = None
        num_nodes = len(networkMod.get_node_details())
        expected_sig_length = 130 * num_nodes + 2
        while time.time() <= now + timeout:
            result = sig_call(self.current_proposal.id)
            assert "signatures" in result
            if len(result["signatures"]) == 0:
                sigs_returned = 0
            else:
                sigs_returned = (len(result["signatures"]) - 2) / 130
            assert int(sigs_returned) == sigs_returned, f"Weird data returned for signature: {result}"
            if len(result["signatures"]) <= 2:
                logging.info(f"Waiting for valid signature bundle in data node.")
                time.sleep(0.2)
            elif sigs_returned < num_nodes:
                logging.info(f"Waiting for signatures from all nodes. Currently have: {sigs_returned} / {num_nodes}")
                time.sleep(0.2)
            elif sigs_returned > num_nodes:
                raise Exception(f"Too many signatures returned: {result}")
            else:
                break
        bundle = result["signatures"]
        nonce = result["nonce"]
>       assert len(bundle) > 2, f"Bad list asset bundle: {result}"
E       AssertionError: Bad list asset bundle: {'assetSource': '0xad221c9cbAF618890fC12136865Aad9F8dc888E2', 'vegaAssetId': '17838bbfd518edc7c44e47857eb860ad0239d3edf43fe2a63f0a6bd757b90b54', 'nonce': '10635616997434885844920268286264680989497494871490412493598980090653440543572', 'signatures': ''}
E       assert 0 > 2
E        +  where 0 = len('')

modules/govMod/governMod.py:1569: AssertionError`

# Expected behaviour
List asset bundle is found.

# Steps to reproduce

## Automation
Current develop version of test_proposals_can_span_upgrade can reproduce this. This is because of this line:
`erc20_asset.propose(proposer, state="Open", closeTimeOffSec=30, enactTimeOffSec=31)`
I'm about to fix it to be 90 & 91, so if you want to repro, you may need to change it back, and then run the following:
`pytest tests/protocol_upgrade -l test_proposals_can_span_upgrade -x --count 20`
This will cause it to attempt the test 20 times each, and fail the first time it goes wrong.

# Evidence

## Logs
```
{"id":"314-7","block":"695663EC5ADE10B1224ACADA04AA43108EEC352C2B0CFC51E5AF77785B729881","type":"BUS_EVENT_TYPE_PROPOSAL","proposal":{"id":"75fb1b2b62b1b643635501b563e10aa3d4e022a3eefc33f909a93a73841a72f1","reference":"A70F84023D","partyId":"95433207e7d0ee71e26c5e1bfcd379c3803ea22105530641a28507398262e1fd","state":"STATE_ENACTED","timestamp":"1677810932680274010","terms":{"closingTimestamp":"1677810962","enactmentTimestamp":"1677810963","validationTimestamp":"1677810943","newAsset":{"changes":{"name":"Test Asset","symbol":"TEST","decimals":"18","quantum":"1000000000000000000","erc20":{"contractAddress":"0xad221c9cbAF618890fC12136865Aad9F8dc888E2","lifetimeLimit":"100000000000000000000000000","withdrawThreshold":"1000000000000000000000"}}}},"rationale":{"description":"New asset","title":"Asset Proposal: Test Asset"},"requiredParticipation":"0.00000015","requiredMajority":"0.66"},"version":1,"chainId":"testnet-001","txHash":"695663EC5ADE10B1224ACADA04AA43108EEC352C2B0CFC51E5AF77785B729881"}
```
^ Asset enacted
```
{"id":"314-8","block":"695663EC5ADE10B1224ACADA04AA43108EEC352C2B0CFC51E5AF77785B729881","type":"BUS_EVENT_TYPE_PROTOCOL_UPGRADE_PROPOSAL","protocolUpgradeEvent":{"upgradeBlockHeight":"314","vegaReleaseTag":"v99.99.0+dev.3601bb6b.modified","approvers":["343d1bedf0c7cfc65d87c6d9dd87f3d0199145a439e0e8d371336e0a548630b6","478d2917d60f60e2124a19bca62077d9c3d59f3a3fd87c58f2207a0efc404ce3","54edab78feef1d7facd487f5ab734484ab4df13d11718b8371af220805a38f58","83c9eb2cbafde0ee072dcd48bca45a4fc161a2ad06572913523b16f9cb59d94b","b2339ffbf32b72067e7c9cb984cf6cfb3a6123f913e34971df422788e562c7ff"],"status":"PROTOCOL_UPGRADE_PROPOSAL_STATUS_APPROVED"},"version":1,"chainId":"testnet-001","txHash":"695663EC5ADE10B1224ACADA04AA43108EEC352C2B0CFC51E5AF77785B729881"}
```
^ Protocol upgrade happens

## Additional context
Tried waiting quite a while after upgrade, bundle never appears.

# Definition of Done

**Before Merging**

**After Merging**

karlem:
Close #7758


core/validators/signatures.go
```diff
@@ -106,6 +108,49 @@ type NodeIDAddress struct {
...
+	s.notary.OfferSignatures(types.NodeSignatureKindERC20MultiSigSignerAdded, s.offerValidatorAddedSignatures)
+	s.notary.OfferSignatures(types.NodeSignatureKindERC20MultiSigSignerRemoved, s.offerValidatorRemovedSignatures)
+}
+
+func (s *ERC20Signatures) offerValidatorAddedSignatures(id string) []byte {
+	if s.isValidatorSetup {
+		return nil
+	}
+
+	sd, ok := s.pendingSignatures[id]
```
wwestgarth, line 121: I don't think this is right, the keys for `s.pendingSignatures` are ethereum addresses. It's the `issuedSignatures` map that has the notary resourceID as its keys, but then it maps to nothing which is unhelpful.

So I think we might need to do something where `s.issuedSignatures -> {eth-address, submitter-address}` and then you can get do `s.pendingSignatures[s.issuedSignatures[id].EthAddress]` to get the other bits, or something like that.

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -634,6 +634,7 @@ message PendingERC20MultisigControlSignature {
   string nonce = 3;
   uint64 epoch_seq = 4;
   bool added = 5;
+  string submitter_address = 6;
```
wwestgarth, line 637: Related to my other comment, the pending signatures aren't related to a submitter and you can have multiple submitters per-pending signatures. So I think we need this the `issue_signatures` part instead.

core/validators/signatures.go
```diff
@@ -106,6 +108,49 @@ type NodeIDAddress struct {
 	SubmitterAddress string
 }
 
+func (s *ERC20Signatures) OnTick(ctx context.Context, _ time.Time) {
```
wwestgarth, line 111: I don't think we need an `OnTick` we can just throw these calls in `func (t *Topology) BeginBlock()`. 

Actually looking at `OfferSignatures` is doesn't look like it checks that `v.signature == nil` before calling the callback, so I think we're always re-signed everything every block??

karlem: I think you have a good point - it should check for nil. Although it doesn't regenerates forever - when the current signature is delivered from other nodes it will remove this ID from the list of retries.

karlem: https://github.com/vegaprotocol/vega/blob/9409f972c7be326ad2eba695c2d47b1fddfb21c6/core/notary/notary.go#L185

candida-d (dismissed):

core/validators/signatures_snapshot.go
```diff
@@ -37,11 +37,17 @@ func (s *ERC20Signatures) SerialisePendingSignatures() *snapshot.ToplogySignatur
...
+	for resID, data := range s.issuedSignatures {
+		issued = append(issued, &snapshot.IssuedERC20MultisigControlSignature{
+			ResourceId:       resID,
+			EthereumAddress:  data.EthAddress,
+			SubmitterAddress: data.SubmitterAddress,
+		})
 	}
-	sort.Strings(issued)
+	sort.SliceStable(issued, func(i, j int) bool {
+		return issued[i].EthereumAddress < issued[j].EthereumAddress
```
wwestgarth, line 49: I don't think `EthereumAddress` is unique for the issued signature, so it would be best to sort by `resID` instead.

candida-d (dismissed):

EVODelavega (dismissed):

wwestgarth (dismissed):

gordsport (dismissed):
Approving on behalf of docs team

jeremyletang (dismissed):

wwestgarth:
Somethings not quite right here, I ran Jake's test on jenkins and occassionaly the asset list signatures are never coming through after the upgrade:
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/66921/tests

jeremyletang (approved):

PR #7925, opened 2023-03-22
feat: implement specific marshaller for the websockets subscribtion to improve payload sizes

jeremyletang:
Add specific marshaller for streams:
- marshal enums as ints instead of strings (TIME_IN_FORCE_GTT will be 1)
- do not write default values ("peggedOrder": null, will be completely omitted from the payload).


JonRay15 (approved):
Looks like some excellent coding here, I approve.

ze97286 (approved):

PR #7942, opened 2023-03-23
chore: enable subscription rate limiter on grpc and rest subscriptions

### LINKED ISSUE #7923: `NewSubscriptionRateLimiter` seems only to be used in the graphql server
When testing the data node rate limiting it was observed that `NewSubscriptionRateLimiter` seems only to be used in the graphql server.

We should ideally move the subscription rate limiting to be at the GRPC level so this is not just enabled for GraphQL

guoguojin:
Closes #7923


jeremyletang:
@guoguojin seeing that rest is actually calling grpc internally, shouldn’t this be implemented on grpc only?

guoguojin:
It is implemented on only the gRPC layer, but the REST inherits this hence the comment in the change log.

jeremyletang (approved):

PR #7953, opened 2023-03-23
feat: datanode batcher statistics

### LINKED ISSUE #7505: Metrics flush count and conflation ratio on stores
from the chart below: you see the orders flush time towards the end is pretty constant at ~1s, despiste the fact number of orders events drops, I guess this means that the size of persisted conflated events is similar (ie we are getting say 5 events per order in a  block instead of 10) - probably......can't be sure with the info we have, another way to interpret would be the DB is slowing down with data szie, which would be obviously bad.    To disitinguish we should metric the flush count and conflation ratio:
[11:09](https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1675681771884139)
https://metrics.vega.xyz/d/datanode-stagnet1/data-node?orgId=1&var-instance=n09.testnet.vega.xyz&from=now-7d&to=now

ettec:
closes #7505 added stats for the datanode batchers (possible useful metric for future compression analysis)

### DIFF
--- a/datanode/metrics/prometheus.go
+++ b/datanode/metrics/prometheus.go
@@ -84,6 +84,9 @@ var (
 	// Per table segment creation time.
 	networkHistoryCopiedRowsCounter *prometheus.CounterVec
 	networkHistoryCopyTimeCounter   *prometheus.CounterVec
+
+	batcherAddedEntities   *prometheus.CounterVec
+	batcherFlushedEntities *prometheus.CounterVec
 )
 
 // abstract prometheus types.
@@ -486,6 +489,36 @@ func setupMetrics() error {
 		return err
 	}
 	networkHistoryCopyTimeCounter = ct
+
+	h, err = addInstrument(
+		Counter,
+		"batcher_added_entities",
+		Namespace("datanode"),
+		Vectors("table"),
+	)
+	if err != nil {
+		return err
+	}
+	baet, err := h.CounterVec()
+	if err != nil {
+		return err
+	}
+	batcherAddedEntities = baet
+
+	h, err = addInstrument(
+		Counter,
+		"batcher_flushed_entities",
+		Namespace("datanode"),
+		Vectors("table"),
+	)
+	if err != nil {
+		return err
+	}
+	bfe, err := h.CounterVec()
+	if err != nil {
+		return err
+	}
+	batcherFlushedEntities = bfe
 
 	// sqlQueryTime
 	h, err = addInstrument(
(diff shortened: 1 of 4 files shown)


guoguojin (dismissed):

pscott31 (dismissed):

guoguojin (dismissed):

datanode/metrics/prometheus.go
```diff
@@ -84,6 +84,9 @@ var (
 	// Per table segment creation time.
 	networkHistoryCopiedRowsCounter *prometheus.CounterVec
 	networkHistoryCopyTimeCounter   *prometheus.CounterVec
+
+	batcherAddedEntities   *prometheus.CounterVec
```
EVODelavega, line 88: Just wondering whether a gauge here might be more informative (WRT to monitoring/alerts)

ettec: Its possible to set an alert on rate (count/time), though the key reason for this metric is to be able to monitor overtime the ratio of events received to flushed, so its got potentially a dual purpose.

EVODelavega: Yes, but for ratios, wouldn't a summary make the most sense? Rather than a direct division of sums over time, using quantiles over time windows. We'd get a baseline using, say, the 95th percentile over a 15m or even 1h time window, and compare that to the 99th percentile over 5 minutes. WRT alerts, I've found this approach much less cumbersome. Sometimes a lot of events fire, or a lot of updates happen due to a single event (e.g. a market terminates, positions close, expiring orders get cancelled, those are just 2 events, but with lots of entities being flushed). That's a one-off however, and shouldn't raise an alert. For things like that, summaries and histograms are much more useful metric types

ettec: oh, right sorry I thought you originally suggested to use a gauge? I was replying to that, did you update the original comment or am I going mad?

EVODelavega: yeah, I originally commented gauge, wanted to edit with strike-through but couldn't remember the markdown for that

ettec: the reason for the two metrics, added and flushed, is to give some insight on the 'shape' of the data (alerting wasn't the purpose here).  Receiving 1000 updates for the same order in a block results in a very different load on data-node to receiving one update for 1000 different orders (as an extreme comparison) - right now we have no insight into this and so we could see big differences in load which would be essentially be unexplainable with the current stats datanode publishes.

pscott31 (approved):

PR #7963, opened 2023-03-23
fix: initialise isValidator when loading from checkpoint

### LINKED ISSUE #7962: Validators that are not in genesis can't send node vote
# Problem encountered
Observed in mainnet but still correct for current develop (post 0.70.0). Validators that are not configured in genesis are not able to send node votes for chain events. The reason for this is that when loaded from checkpoint we're not setting the topology isValidator flag which then along with the isValidatorSetup determine whether the node should act as a validator and run witness checks and vote.

ze97286:
Closes #7962 

plus bonus fix of the witness warning log which is completely wrong.


jeremyletang (dismissed):

wwestgarth (dismissed):
We should probably do `if t.SelfNodeId(k){t.isValidator = false}` when a node is kicked out in `validator_set.go`, but I guess any validator txs sent in by a kicked out node would get stopped at `checkTx` anyway so its not a big deal.

EVODelavega (approved):

PR #8002, opened 2023-03-27
fix: order subscriptions filter handling

### LINKED ISSUE #8001: Problem with order subscription
If party and market are not both specified, subscription doesn't return anything apart from the initial snapshot.

Also initial snapshot includes non-live orders, which is potentially a huge amount.

pscott31:
closes #8001


zale144 (approved):
Lgtm

jeremyletang (approved):

PR #8016, opened 2023-03-28
feat: remove all checkpoints

### LINKED ISSUE #8015: Add ability to automatically remove checkpoint files
Currently checkpoint files are created but are never removed. This leads to excess storage usage on the file system. Add the option to remove older checkpoint files either by age or size to cap the maximum amount of file storage used up by the checkpoints.

jeremyletang:
close #8015 .

### DIFF
--- a/core/processor/abci.go
+++ b/core/processor/abci.go
@@ -18,8 +18,11 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


core/processor/abci.go
```diff
@@ -942,6 +945,65 @@ func (app *App) handleCheckpoint(cpt *types.CheckpointState) error {
...
+
+	files, err := ioutil.ReadDir(cpDirPath)
+	if err != nil {
+		return fmt.Errorf("could not open open the checkpoint directory: %w", err)
+	}
+
+	// we assume that the files in this directory are only
+	// from the checkpoints
+	// and always keep the last 20, so return if we have less than that
+	if len(files) <= 20 {
```
wwestgarth, line 965: Is there any value in making this an option in the config? Then the system-tests and the snapshot pipeline can set it really low and save reduce the size of the Jenkins artefacts.

core/processor/abci.go
```diff
@@ -942,6 +945,65 @@ func (app *App) handleCheckpoint(cpt *types.CheckpointState) error {
...
+
+func (app *App) removeOldCheckpoints() error {
+	cpDirPath, err := app.vegaPaths.CreateStatePathFor(paths.StatePath(paths.CheckpointStateHome.String()))
+	if err != nil {
+		return fmt.Errorf("couldn't get checkpoints directory: %w", err)
+	}
+
+	files, err := ioutil.ReadDir(cpDirPath)
+	if err != nil {
+		return fmt.Errorf("could not open open the checkpoint directory: %w", err)
```
wwestgarth, line 959:
```suggestion
		return fmt.Errorf("could not open the checkpoint directory: %w", err)
```

core/processor/abci.go
jeremyletang: hum yes I suppose so. I'll do it.

core/processor/abci.go
jeremyletang: addressed.

core/processor/abci.go
jeremyletang: addressed

pscott31 (approved):

PR #8028, opened 2023-03-28
Reject old commands on check Tx

### LINKED ISSUE #7969: Timebased pruning of the mempool
## Issue

In order to best manage the mempool (before the below mentioned further optimisations) we need to investigate and implement time based pruning of the mempool

- [ ] Investigate how we can prune the mempool with Tendermint 0.34
- [ ] Update the specification(s) where applicable (https://github.com/vegaprotocol/specs/issues/1639)
- [ ] Implement the changes 

Further optimisations to the mempool are to be designed along with the upcoming tendermint releases and ABCI++
- https://github.com/vegaprotocol/specs/issues/1439

EVODelavega:
Closes #7969 

Add TTL/BTL (blocks to live) to user commands


protos/sources/vega/commands/v1/transaction.proto
```diff
@@ -33,6 +33,12 @@ message InputData {
   // `block_height` prevents replay attacks in conjunction with `nonce` (see above).
   uint64 block_height = 2;
 
+  // A number of blocks this transaction is valid for starting from block_height.
+  // If block_height is set to 100, and this field is set to 10, this transaction
+  // will be rejected if it is received after block 110.
```
ze97286, line 38: I think it could be rephrased more accurately because it's not so much about when it is *received* but more that if it's not included in the blocks preceding 111, it will be rejected. In this case though it will just disappear into the void. The other case is that it's already expired when being added to the mempool which is I guess what you're describing, but just wanted to highlight that this is not the only case.

core/txn/command.go
```diff
@@ -106,6 +106,15 @@ func (cmd Command) IsValidatorCommand() bool {
 	}
 }
 
+func (cmd Command) IsTTLCommand() bool {
+	switch cmd {
+	case SubmitOrderCommand, CancelOrderCommand, AmendOrderCommand, WithdrawCommand, LiquidityProvisionCommand, CancelLiquidityProvisionCommand, AmendLiquidityProvisionCommand, DelegateCommand, UndelegateCommand, StateVariableProposal, ProposeCommand, VoteCommand, TransferFundsCommand, CancelTransferFundsCommand:
```
ze97286, line 111: just curious why is it not true for all except for validator commands?

core/processor/tx.go
```diff
@@ -42,13 +43,19 @@ func DecodeTxNoValidation(payload []byte) (*Tx, error) {
 	if err := err.ErrorOrNil(); err != nil {
 		return nil, err
 	}
+	ttl := uint64(0)
```
ze97286, line 46: why is this only here (in null blockchain context) rather than (also) below in the `DecodeTx` function?

protos/sources/vega/commands/v1/transaction.proto
```diff
@@ -33,6 +33,12 @@ message InputData {
...
+  // A maximum of 50 blocks applies, so setting this to any value > 50 is considered invalid.
```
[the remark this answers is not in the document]
EVODelavega: Not validated yet, will be added shortly

core/processor/tx.go
EVODelavega: You're right, should be added there, too. I suspect a `git checkout . && git checkout develop` without committing changes is to blame here

core/txn/command.go
EVODelavega: Things like the `AnnounceNodeCommand` aren't validator commands, and I'm not sure if we should, essentially `&>/dev/null` those commands after a number of blocks

core/processor/tx.go
```diff
@@ -42,13 +43,19 @@ func DecodeTxNoValidation(payload []byte) (*Tx, error) {
...
+		ttl = inputData.BlockHeight + inputData.GoodForBlocks
+	}
 
 	return &Tx{
 		originalTx: payload,
 		tx:         tx,
 		inputData:  inputData,
 		pow:        tx.Pow,
 		version:    tx.Version,
+		ttl:        ttl,
```
wwestgarth, line 58: Why do we need to save the tll separately and not just do:
```
func (t Tx) TTL() uint64 {
	return t.inputData.BlockHeight + t.inputData.GoodForBlocks
}
```

then in `checkTx` we just simply have:
```
if tx.GetVersion() > 3 && tx.Command().IsTTLCommand() {
    // do thing
}
```
because we also avoid having the equality here `tx.Version == commandspb.TxVersion_TX_VERSION_V4` which someone will no doubt forget to update when we roll into tx-version 5.

wwestgarth:
The ticket mentions updating a spec, but I can't find any spec for this?

Also how is this going to be wired into the `vegawallet`? Its currently still just using transaction version 3.

core/processor/tx.go
EVODelavega: So we're saving TTL separately because the TTL depends not just on the `GoodForBlocks` part of the TX, but a network parameter. The max TTL also doubles as the default for TxV3, and it overrides any value set here if the `GoodForBlocks` >  `MaxTTL`. So when decoding the TX, we check the netparam, and set the TTL accordingly. That being said, I suspect the `TX_VERSION_V4` check now can probably be replaced with a `if tx.inputData.GoodForBlocks > 0 && tx.inputData.GoodForBlocks < maxTTL`.

core/processor/abci.go
```diff
@@ -981,11 +981,20 @@ func (app *App) OnCheckTx(ctx context.Context, _ tmtypes.RequestCheckTx, tx abci
...
-		resp.Code = blockchain.AbciTxnValidationFailure
+		resp.Code = blockchain.AbciExpiredCommandError
 		resp.Data = []byte(err.Error())
 		return ctx, resp
 	}
 
+	if ttl := tx.TTL(); ttl > 0 && tx.Command().IsTTLCommand() {
+		// if current block == TTL, the tx won't be valid for the next block, and should be rejected.
+		if ttl <= app.stats.Height() {
+			resp.Code = blockchain.AbciTxnValidationFailure
```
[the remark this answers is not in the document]
EVODelavega: Yes, I think I added this before adding the error code for expired commands

core/processor/tx.go
wwestgarth: Makes sense now that its a network-parameter.

wallet/api/client_check_transaction.go
```diff
@@ -131,7 +131,12 @@ func (h *ClientCheckTransaction) Handle(ctx context.Context, rawParams jsonrpc.P
 	h.interactor.Log(ctx, traceID, SuccessLog, "The transaction passes the anti-spam rules.")
 
 	// Sign the payload.
-	rawInputData := wcommands.ToInputData(request, stats.LastBlockHeight)
+	rawInputData := wcommands.ToInputData(request, stats.LastBlockHeight, uint64(ttl))
+	// always set TTL if it wasn't previously set, and override the specified value if it's > the max set by network params
+	// alternatively leave zero and set TxV3?
```
ValentinTrinque, line 136: >// alternatively leave zero and set TxV3?

Can be removed. Also, we are missing the tests that assert that behavior. The tests will document this behavior.

wallet/api/api.go
ValentinTrinque, line 1: I think we are missing the `SignTransaction`.

wallet/api/client_send_transaction.go
```diff
@@ -135,7 +135,7 @@ func (h *ClientSendTransaction) Handle(ctx context.Context, rawParams jsonrpc.Pa
 	h.interactor.Log(ctx, traceID, SuccessLog, "The transaction passes the anti-spam rules.")
 
 	// Sign the payload.
-	rawInputData := wcommands.ToInputData(request, stats.LastBlockHeight)
+	rawInputData := wcommands.ToInputData(request, stats.LastBlockHeight, uint64(ttl))
 	inputData, err := commands.MarshalInputData(rawInputData)
 	if err != nil {
 		h.interactor.NotifyError(ctx, traceID, InternalError, fmt.Errorf("could not marshal input data: %w", err))
```
ValentinTrinque, line 141: Why is the code from `CheckTransaction` not added here ?

```go
if rawInputData.GoodForBlocks == 0 || rawInputData.GoodForBlocks > stats.MaxTTL {
    rawInputData.GoodForBlocks = stats.MaxTTL
}
```

wallet/api/client_send_transaction_test.go
ValentinTrinque, line 1: There is not test proving the TTL is properly set ?

wallet/service/starter.go
```diff
@@ -220,7 +220,7 @@ func (s *Starter) buildAPIV1(ctx context.Context, logger *zap.Logger, networkCfg
 	return servicev1.NewAPI(apiV1Logger, handler, auth, forwarder, policy, networkCfg, spam), nil
 }
 
-func (s *Starter) buildAPIV2(ctx context.Context, logger *zap.Logger, cfg *network.Network, pow api.SpamHandler, closer *vgclose.Closer) (*servicev2.API, error) {
+func (s *Starter) buildAPIV2(ctx context.Context, logger *zap.Logger, cfg *network.Network, pow api.SpamHandler, closer *vgclose.Closer, ttl int64) (*servicev2.API, error) {
```
ValentinTrinque, line 223: Is there a reason to not have a `uint` ? Can we have negative TTL ?

ValentinTrinque (changes requested):

wallet/api/client_check_transaction.go
```diff
@@ -152,6 +157,11 @@ func (h *ClientCheckTransaction) Handle(ctx context.Context, rawParams jsonrpc.P
 		Algo:    signature.Algo,
 		Version: signature.Version,
 	})
+	// so if we don't always default to TTL != 0, we should do something like this:
+	// but for now, because we set TTL for all transactions, we should just return V4 from NewTransaction
+	// if rawInputData.GoodForBlocks != 0 {
+	// tx.Version = commandspb.TxVersion_TX_VERSION_V4
+	// }
```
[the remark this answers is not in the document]
EVODelavega: the comment was to get feedback in this review, whether we should always set it or not

wallet/service/starter.go
EVODelavega: Could (and probably should) cast it earlier on. Flags (I don't think) have a `Uint64Var` option, so the value from there is `int64`

wallet/api/client_send_transaction.go
EVODelavega: knew I was likely to forget something, will add it

core/txn/command.go
EVODelavega: After talking it through some more, we're just going to use `!IsValidatorCommand()` instead

core/processor/abci.go
```diff
@@ -66,6 +67,7 @@ var (
 	ErrAssetProposalDisabled                         = errors.New("asset proposal disabled")
 	ErrAwaitingCheckpointRestore                     = errors.New("transactions not allowed while waiting for checkpoint restore")
 	ErrOracleNoSubscribers                           = errors.New("there are no subscribes to the oracle data")
+	ErrTransactionExpired                            = errors.New("transaction expired")
```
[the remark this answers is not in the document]
EVODelavega: I think a static error makes more sense here than `fmt.Errorf`. This error is used in one specific case, so when seeing this error, it's clear what happened.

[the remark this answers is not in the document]
candida-d: How about a compromise. "The transaction expired because it was not included(?) in the defined(?) block"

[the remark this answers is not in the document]
ValentinTrinque: I think Candida's suggestion is the right one indeed. What matters to me is the clarity of the error message. We need to make user facing error message very clear, otherwise it tends to create a lot of support ticket for explanations.

wallet/service/starter.go
ValentinTrinque: Alright, I think we should leverage the type system to make things easier to reason about and avoid silly mistakes.

wallet/api/client_send_transaction_test.go
ValentinTrinque, line 1: Looks like we are missing the test for the TTL there.

wallet/api/client_sign_transaction_test.go
ValentinTrinque, line 1: Looks like we are missing the test for the TTL there, _as well_.

wallet/api/client_sign_transaction.go
ValentinTrinque, line 1: This method is missing the TTL bits:
```go
rawInputData := wcommands.ToInputData(request, stats.LastBlockHeight, uint64(ttl))
// always set TTL if it wasn't previously set, and override the specified value if it's > the max set by network params
if rawInputData.GoodForBlocks == 0 || rawInputData.GoodForBlocks > stats.MaxTTL {
  rawInputData.GoodForBlocks = stats.MaxTTL
}
```

Also I think we might have a tiny inconsistency compare to the check and send transaction. They are using `commands.MarshalInputData(rawInputData)` and not `wcommands.ToMarshaledInputData(...)` which is probably why it slept through.

ValentinTrinque (changes requested):

wallet/api/client_sign_transaction.go
EVODelavega: We perform the check before the call `MarshalInputData`, which is fine. In this case (`ToMarshaledInputData`), we just take the transaction as-is, and sign it. If no TTL is provided, it will be set to 0, which is fine. `ToMarshaledInPutData` calls `ToInputData`, setting TTL to 0.

cmd/vegawallet/commands/service_run.go
```diff
@@ -129,6 +129,11 @@ func BuildCmdRunService(w io.Writer, handler RunServiceHandler, rf *RootFlags) *
 		"",
 		"Path to the file containing the tokens database passphrase",
 	)
+	cmd.PersistentFlags().Int64Var(&f.TTL,
```
wwestgarth, line 132: `.Uint64Var()` would be better here.

ValentinTrinque (dismissed):

cmd/vegawallet/commands/service_run.go
EVODelavega: It is, for some reason it didn't show up in autocomplete, but looks like it's possible. Will change

wwestgarth (dismissed):

gordsport (dismissed):
Approving on behalf of the docs team

jeremyletang (approved):

PR #8042, opened 2023-03-30
API for getting csv data from network history

### LINKED ISSUE #8030: Add an API for fetching data from network history
# Feature Overview

**In order to** retrieve archived data that has been removed from the database due to data retention policies
**We will** Provide an API point which retrieves network history segments from the local IPFS (tbc: should it fetch files not locally present?), extracts relevant data and returns it
**So that** We have a method for retrieving archived data

## Specs
- [Link](xyz) tbc; @ettec volunteered to write one

Decisions:
 - What should API look like, e.g. query by block height range | time range | segment id at a time
 - Should we get data for one table or multiple at once
 - Return format? CSV? Currently use postgres' text format so requires a bit of parsing to do  that, we could switch to CSV for segments.

pscott31:
Closes #8030 

Sort-of breaking for ops in that it changes the format in which network history segments are stored. After this, datanode will not be able to read history segments written by previous versions (and previous versions will not be able to read history segments produced by datanodes with this commit)

If this is a dealbreaker I could add some logic to handle it


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -499,6 +500,15 @@ service TradingDataService {
     // List all entities created by transaction hash
     rpc ListEntities(ListEntitiesRequest) returns (ListEntitiesResponse);
 
+    // Export Network History
+    //
+    // Extract CSV table data from network history between two block heights.
+    // Note: The block heights must fall on network history segment boundaries
+    // which can be discovered with ListAllNetworkHistorySegments
```
wwestgarth, line 507: Because of the way the proto docstrings end up also being the REST docs we can't really reference API by their GRPC name like `ListAllNetworkHistorySegments`.

We're going to have to just be vague and say something like "which can be discovered by listing all network history segments". The plan is to have similar functionality grouped in the docs, so _hopefully_ if you're looking at the doc page for exporting history segments, in the sidebar the docs for the API to list them will be staring you in the face.

pscott31: ah ok I'll change that then :)

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,78 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+
+		segmentData, err := fsutil.ReadNetworkHistorySegmentData(outerTarFileReader, req.TableName)
+		if err != nil {
+			return formatE(ErrNetworkHistoryExtractingSegment, err)
+		}
+
+		scanner := bufio.NewScanner(segmentData)
+		for ; scanner.Scan(); rowNum++ {
+			row := scanner.Text()
+			row = strings.Replace(row, "\t", ",", -1) // postgres dump file separated by tabs
```
ettec, line 3097: Would it be better to switch the postgres format to csv?,  it may be better to do this in case there is any formatting going on that we are not aware of, that way an exported csv will be guaranteed to be importable to postgres for a given schema version, this could be very useful,  just need to do a before and after test on segment size and load time to check no meaningful impact.

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,78 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+	if err != nil {
+		return formatE(ErrListAllNetworkHistorySegment, err)
+	}
+
+	ch, err := networkhistory.GetContiguousHistory(segments, req.FromBlock, req.ToBlock)
+	if err != nil {
+		return formatE(ErrNetworkHistoryGetContiguousSegments, err)
+	}
+
+	header := metadata.Pairs("Content-Disposition", fmt.Sprintf("attachment;filename=vega-%s-%06d-%06d.csv", req.TableName, ch.HeightFrom, ch.HeightTo))
```
ettec, line 3075: If we are not normalising across schema, then I think the files should be split up across schema version and the filename should encode the schema version

ettec (dismissed):

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,78 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+			return formatE(ErrNetworkHistoryExtractingSegment, err)
+		}
+
+		scanner := bufio.NewScanner(segmentData)
+		for ; scanner.Scan(); rowNum++ {
+			row := scanner.Text()
+			row = strings.Replace(row, "\t", ",", -1) // postgres dump file separated by tabs
+			row = strings.Replace(row, `\\x`, "", -1) // get rid of the \x prefix from BYTEA columns
+			chunk.WriteString(row)
+			chunk.WriteByte('\n')
```
ettec, line 3100: Also, on the csv column headers, probably easiest way would be to add this information into each segment when it is created, as a sort of meta file or something as will be relatively very small, not an issue.  And then this tool can optionally return with or without headers in the csv.

datanode/api/trading_data_v2.go
pscott31: Yes probably - it addresses a number of problems actually;
- you can include headers e.g. `COPY assets TO stdout WITH (FORMAT csv, HEADER);`
- I think what I have at the moment will break if, for example, there's a comma in a text field - it needs escaping, then you need to escape the escape character if that's in a field too. 

I'm more than happy to change it to do that but it means then that network history files won't be compatible over that change unless we add some check or something when importing

jeremyletang (dismissed):

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,78 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+	if req.TableName == "" {
+		return formatE(ErrNetworkHistoryNoTableName, errors.New("empty table name"))
+	}
+
+	segments, err := t.networkHistoryService.ListAllHistorySegments()
+	if err != nil {
+		return formatE(ErrListAllNetworkHistorySegment, err)
+	}
+
+	ch, err := networkhistory.GetContiguousHistory(segments, req.FromBlock, req.ToBlock)
```
ettec, line 3070: This will fail if the passed heights do not sit on segment boundary, which is less that ideal, its solvable, but perhaps too much, but if so it needs to made clear that the heights passed must sit on segment boundary.  Else do something like-> 
  Ideally a user should be able to get any block which means some sort of interpolation on the network history segment index would be required  i.e for block at 1200, it would get the one immediately before  (span 1000->2000), then when the rows are being read out any with vega_time before that would be ignored.  Every history table has a vega_time column, but it would mean parsing on the fly and using the blocks file to get the relevent vega time, maybe this is too much?

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,78 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
 
 // -- NetworkHistory --.
 
+func (t *tradingDataServiceV2) ExportNetworkHistory(req *v2.ExportNetworkHistoryRequest, stream v2.TradingDataService_ExportNetworkHistoryServer) error {
```
ettec, line 3058: It would be super useful if a user could give a time range, this would mean an additional index in the store by vega time and a bit of extra meta data in the segments to record the block time span.

datanode/api/trading_data_v2.go
pscott31: Yes it says in the docstring for the request
```
message ExportNetworkHistoryRequest {
    // Block to begin exporting from. Must be the first block a history segment,
    // which by default are 1000 blocks each - so 1, 1001 etc. are probably valid values.
    // you can check by first calling the API to list all network history segments.
    int64 from_block = 1;
    // Block to end exporting to (inclusive). Must be the last block of a history segment
    // which by default are 1000 blocks each - so 2000, 2000 etc. are probably valid values
    // you can check by first calling the API to list all network history segments.
    int64 to_block = 2;
```

But it is a little surprising. Before I noticed the function `GetContiguousHistory` I had actually written my own which just included the extra blocks before and after the requested window up to the initial/last segment boundary; but that's kind of surprising behaviour in it's own way, I can imagine users getting confused about duplicate rows. At least his way it's explicit.

I considered doing something like you suggested, but it makes things quite a bit more complicated. If there's a demand for it I'll do it (it sort of goes in hand with the time range comment above) - but I think for an initial cut what we have is a sufficient MVP.

datanode/api/trading_data_v2.go
ettec: ok, well if postgres handles that on import, thats the easiest for sure.   WRT schema version, this is already in the segment, I think just putting it in the file name would be sufficient (and then splitting data over schema boundary as per above comment)

datanode/api/trading_data_v2.go
ettec: Yeah, as agreed on slack, if we do it now, no need to worry about backwards compatibility.

datanode/networkhistory/fsutil/fsutil.go
```diff
@@ -212,3 +212,51 @@ func UntarFile(source io.Reader, decompressedFilesDestination string) error {
...
+	// Find the table dump file in the inner .tar file
+	for {
+		header, err := innerTarFile.Next()
+		if err == io.EOF {
+			return nil, fmt.Errorf("table file '%s' not found in segment", historyFileName)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error reading inner tar file: %w", err)
+		}
+		if filepath.Base(header.Name) == historyFileName {
```
wwestgarth, line 258: I'm thinking about that issue we saw on the snapshot-pipeline where arbitrary non-vega-schema table was being picked up and added to the network history (was a ticket ever raised for that?)

Knowing that, is us not validating `historyFileName` against a list of known table names potentially a security issue (albeit unlikely) in that we are exposing an API that _could_ let someone read arbitrary non-vega tables from an unsuspecting data-node?

I feel like we should either validate against a known set of tables first in this API, or do something so that we're 100% sure the network history won't accidently contain non-vega tables

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2834,6 +2844,23 @@ message NetworkHistoryBootstrapPeersResponse {
     repeated string bootstrap_peers = 1;
 }
 
+// Request to export network history data in CSV format
+message ExportNetworkHistoryRequest {
+    // Block to begin exporting from. Must be the first block a history segment,
+    // which by default are 1000 blocks each - so 1, 1001 etc. are probably valid values.
```
wwestgarth, line 2850: I don't think we can say `are probably valid values` in our docs :p

"Block to begin exporting from. Must be the first block a history segment" is probably all we need. We're already saying how to get a list of history segment in the docstring for `ExportNetworkHistory` we don't need it here aswell.

datanode/networkhistory/fsutil/fsutil.go
pscott31: That's a good point.

Though, I think we're OK in some sense because this is looking at network history files that are semi-public on IPFS at this point anyway; it can't hurt to do a bit of validation; I'll make a whitelist.

datanode/api/trading_data_v2.go
pscott31: So I've semi-resolved this by allowing multiple schema versions in the same file, effectively concatenating two CSVs together. I'd rather not expose users to our schema versions and what have you, it's a bit of an implementation detail - here's the comment I wrote:

```
    // The first row of the CSV file is a header which allows you to identify the columns
    // of subsequent rows.
    //
    // It is possible that your (from_block, to_block) request spans over a range of blocks
    // in which the underlying schema of the tables changes. For example, a column may have
    // been added, removed, or renamed.
    //
    // If this happens, at the point of the schema migration the CSV file will contain a 
    // single blank line, and then a new set of headers, followed by the data in the new schema.
    //
    // An example of a CSV file containing such a schema change would look like:
    //
    // timestamp, value
    // 1, foo
    // 2, bar
    // 
    // timestamp, value, extra_value
    // 3, baz, apple
    //
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -499,6 +500,45 @@ service TradingDataService {
...
+    // be discovered calling the API to list all network history segments. By default
+    // segments contain 1000 blocks, if so ranges such as (1, 1000), (1001, 2000), (1, 3000)
+    // would all fall on segment boundaries and be valid.
+    //
+    // In gRPC, results are returned in a chunked stream of base64 encoded CSV data
+    //
+    // Through the REST gateway, the base64 data chunks are decoded and streamed as a
+    // `content-type: text/csv` HTTP response.
+    //
+    // CSV data is escaped & quoted in the as described in the postgresql documentation at
```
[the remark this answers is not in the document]
pscott31: well sort-of; CSV isn't really a standard - there are quite a few different conventions for specifying how to escape and quote special characters, newlines, NULL values etc.. 

We're using the postgres standard because, well, that's how it comes to us and its a reasonable one. I could copy-paste what's in the postgres docs instead of linking to it if you prefer?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -499,6 +500,45 @@ service TradingDataService {
...
+    //
+    // In gRPC, results are returned in a chunked stream of base64 encoded CSV data
+    //
+    // Through the REST gateway, the base64 data chunks are decoded and streamed as a
+    // `content-type: text/csv` HTTP response.
+    //
+    // CSV data is escaped & quoted in the as described in the postgresql documentation at 
+    // https://www.postgresql.org/docs/14/sql-copy.html#id-1.9.3.55.9.3
+    //
+    // The first row of the CSV file is a header which allows you to identify the columns
```
candida-d, line 520:
```suggestion
    // The first row of the CSV file is a header that allows you to identify the columns
```

protos/sources/data-node/api/v2/trading_data.proto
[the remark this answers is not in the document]
candida-d: I think the origin of the standard doesn't matter as much as what the standard is - so copying out what it is seems more helpful to me :)

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2834,6 +2874,40 @@ message NetworkHistoryBootstrapPeersResponse {
...
+    REWARDS = 15;
+}
+
+// Request to export network history data in CSV format
+message ExportNetworkHistoryRequest {
+    // Block to begin exporting from. Must be the first block of a history segment,
+    // which by default are 1000 blocks each; in that case - 1, 1001, 2001 etc. are valid values
+    // you can check by first calling the API to list all network history segments.
+    int64 from_block = 1;
+    // Block to end exporting to (inclusive). Must be the last block of a history segment
```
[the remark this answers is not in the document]
pscott31: yep fixing

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,102 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+			}
+
+			// If we're in a subsequent segment and headers have changed, write a blank line before the new headers
+			if firstRowInSegment && rowNum != 1 && csvHeaders != row {
+				chunk.WriteString("\n")
+				csvHeaders = row
+			}
+
+			firstRowInSegment = false
+			row = strings.Replace(row, `\x`, "", -1) // get rid of the \x prefix from BYTEA columns
```
ettec, line 3120: Do we want to do this?  wouldn't be better to just leave it in the format postgres understands for now (which may be beneficial) until we learn otherwise?  any end user can handle this easily enough

pscott31: Yeah I dunno, I was kind of torn - took it out, then put it back again.. I was just imagining having to repeatedly answer 'what is this \x for' - but this solution is kind of crappy, e.g. it would remove \x from any text data as well. 

I'll remove it and if we get complains we can always revisit. To do it right it would know the column types and only do it for bytea columns but then we have to get into parsing the CSV data and making sure we do all the escaping right etc.. which I was tyring to avoid.

datanode/api/trading_data_v2.go
```diff
@@ -3048,6 +3055,102 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+			}
+
+			// Don't repeat headers if they're the same in subsequent segments
+			if firstRowInSegment && rowNum != 1 && csvHeaders == row {
+				firstRowInSegment = false
+				continue
+			}
+
+			// If we're in a subsequent segment and headers have changed, write a blank line before the new headers
+			if firstRowInSegment && rowNum != 1 && csvHeaders != row {
```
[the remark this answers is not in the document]
pscott31: I made a thread on #project-datanode about this; it's kind of unpleasant because there's no way to do one HTTP request that triggers multiple downloads as far as I know. We'd have to make a couple of files and zip them up or something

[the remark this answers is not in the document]
ettec: (the history segment reader I think should be able to return the schema version)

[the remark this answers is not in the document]
pscott31: Also most of the time when the overall schema version changes - the table being looked at won't have had any columns added or removed so it's just making life more complicated for the user.

datanode/api/trading_data_v2.go
```diff
@@ -3047,6 +3055,139 @@ func (t *tradingDataServiceV2) GetVegaTime(ctx context.Context, _ *v2.GetVegaTim
...
+func (t *tradingDataServiceV2) ExportNetworkHistory(req *v2.ExportNetworkHistoryRequest, stream v2.TradingDataService_ExportNetworkHistoryServer) error {
+	defer metrics.StartAPIRequestAndTimeGRPC("ExportNetworkHistory")()
+
+	if req.Table == v2.Table_TABLE_UNSPECIFIED {
+		return formatE(ErrNetworkHistoryNoTableName, errors.New("empty table name"))
+	}
+
+	tableName := strings.TrimPrefix(strings.ToLower(req.Table.String()), "table_")
+
+	allSegments, err := t.networkHistoryService.ListAllHistorySegments()
```
ettec, line 3067: you are going to hate me......but....could you use the setup in the network history tests which will give you a fully setup service with segment data and use this in here so you could write an end to end test, confirming the file structure and contents?  or maybe this is overkill, the code does look fairly involved though so may be benefit

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -499,6 +500,68 @@ service TradingDataService {
...
+    // while an empty string data value is written with double quotes ("").
+    // 
+    // Note that this scheme can produce CSV files with quoted values containing embedded carriage returns and line feeds. 
+    // Thus the files are not strictly one line per table row like text-format files.
+    //
+    // The first row of the CSV file is a header that allows you to identify the columns
+    // of subsequent rows.
+    //
+    // Usually the ZIP file will contain only a single CSV file. However it is possible that
+    // your (from_block, to_block) request spans over a range of blocks in which the underlying
```
candida-d, line 539:
```suggestion
    // the (from_block, to_block) request spans over a range of blocks in which the underlying
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -499,6 +500,68 @@ service TradingDataService {
...
+    // schema of the database changes. For example, a column may have been added, removed, or renamed.
+    //
+    // If this happens, the CSV file will be split at the point of the schema change and the zip
+    // file will contain multiple CSV files, with a potentially different set of headers. The
+    // 'version' number of the database schema is part of the in the CSV filename:
+    //
+    //  <chain id>-<table name>-<schema version>-<start block>-<end block>.zip
+    //
+    // For example of a zip file might be called mainnet-sometable-000001-003000.zip
```
candida-d, line 548:
```suggestion
    // For example, a zip file might be called mainnet-sometable-000001-003000.zip
```

datanode/api/trading_data_v2.go
pscott31: OK I started doing that and it turned into quite a faff so I've made a unit test instead which does some basic sanity checking of the `ExportNetworkHistory` API function, mocking out the network history store.

I've run it end to end myself and have checked that it all works.

ettec (dismissed):

AnExsomnis (approved):

ettec (approved):

candida-d (approved):

PR #8059, opened 2023-03-31
feat: notary to keep track of pending signature and send events once in OnTick if needed

jeremyletang:


wwestgarth (dismissed):

candida-d (dismissed):
Approving but as there are no docs it's just a formality

candida-d (approved):

ze97286 (approved):

PR #8070, opened 2023-04-03
fix: annotate required fields in protobuf API request definitions

### LINKED ISSUE #7108: [swagger] Required fields not marked as required
This REST api request will fail:
```shell
curl -L -X GET 'https://api.n06.testnet.vega.xyz/api/v2/candle' -H 'Accept: application/json'
```

It'll fail with this error:
```json
{
  "code": 13,
  "message": "Internal error",
  "details": [
    {
      "code": 60001,
      "message": "failed to get candle data",
      "inner": "getting candle data for time span:invalid candle id:"
    }
  ]
}
```

So the ID is required. However in the swagger file (generated by https://github.com/vegaprotocol/vega/blob/develop/.github/workflows/release-docs.yml) we get:

```javascript
// .... snip 
{
   "/api/v2/candle": {
      "get": {
        "summary": "Candles list",
        "description": "Get candle data for a given candle id",
        "operationId": "TradingDataService_ListCandleData",
        "responses": {
          "200": {
            "description": "A successful response.",
            "schema": {
              "$ref": "#/definitions/v2ListCandleDataResponse"
            }
          },
          "default": {
            "description": "An unexpected error response.",
            "schema": {
              "$ref": "#/definitions/googlerpcStatus"
            }
          }
        },
        "parameters": [
          {
            "name": "candleId",
            "description": "Candle identifier, required field.",
            "in": "query",
            "required": false,   // <---------
            "type": "string"
          },
          {
            "name": "fromTimestamp",
            "description": "Timestamp to retrieve candles since, in nanoseconds since the epoch,\nrequired field - See [`VegaTimeResponse`](#api.VegaTimeResponse).`timestamp`.",
            "in": "query",
            "required": false,
            "type": "string",
            "format": "int64"
          },
// .....

```

Which is untrue, and causes the docs page (e.g. https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-list-candle-data to render the parameter as optional

# Tasks
- [ ] For all REST services (`core`, `corestate`, `trading_data_v2`, `block_explorer`)
  - [ ] For all REST api requests
    - [ ] Find the required parameters
    - [ ] Ensure the protos are marked with a comment that generates them as `required` in swagger
    - [ ] Ensure that the fields that are not required (like candle id above) are not marked as required

## Notes
- The fields are generally marked  correctly in graphql

guoguojin:
Closes #7108 

ST https://jenkins.ops.vega.xyz/job/common/job/system-tests/20501/


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1189,7 +1177,7 @@ message ListTransfersRequest {
     // Public key for which to request transfers
     optional string pubkey = 1;
     // Transfer directions (sender, receiver) to include in the response data
-    TransferDirection direction = 2;
+    TransferDirection direction = 2 [(google.api.field_behavior) = REQUIRED];
```
wwestgarth, line 1180: Not sure if this one is quite true at the moment. If I don't supply a public key I can omit the transfer direction:
```
https://api.n06.testnet.vega.xyz/api/v2/transfers
```

I found this API a bit weird actually in that if the above list all transfers it feels like:
```
https://api.n06.testnet.vega.xyz/api/v2/transfers?partyId=BLAH
```
would list all transfers involving `BLAH` but it doesn't because then we care for the direction and `UNSPECIFIED` makes it an error. But then also having `UNSPECIFIED => TRANSFER_TO_OR_FROM` looks a bit weird from the proto's point of view. Maybe we just need better doc-strings.

I don't know, but anyway for now I don't think it should be marked as required?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1283,16 +1271,18 @@ message ObserveCandleDataResponse {
 // Request that is used when listing candles for a market at an interval
 message ListCandleDataRequest {
     // Candle ID, required field.
-    string candle_id = 1;
+    string candle_id = 1 [(google.api.field_behavior) = REQUIRED];
 
     // Timestamp to retrieve candles since, in nanoseconds since the epoch, required field
-    int64 from_timestamp = 2;
+    int64 from_timestamp = 2 [(google.api.field_behavior) = REQUIRED];
```
wwestgarth, line 1277: Even though in the doc-string it says from/to timestamps are required the truth is they are not:
```
https://api.n06.testnet.vega.xyz/api/v2/candle?candleId=trades_candle_5_minutes_10c7d40afd910eeac0c2cad186d79cb194090d5d5f13bd31e14c49fd1bded7e2
```
maybe they should be required true? I don't know.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1813,9 +1803,9 @@ message OrderEdge {
 // Request for listing margin levels
 message ListMarginLevelsRequest {
     // Party ID for which to list the margin levels
-    string party_id = 1;
+    string party_id = 1 [(google.api.field_behavior) = REQUIRED];
```
wwestgarth, line 1806: Again I can list all margin levels across all markets and parties:
```
https://api.n06.testnet.vega.xyz/api/v2/margin/levels
```
so these aren't required.

datanode/service/risk.go
```diff
@@ -81,7 +81,7 @@ func (r *Risk) ObserveMarginLevels(
 				return false
 			}
 			return (len(marketID) == 0 || marketID == acc.MarketID.String()) &&
-				(len(partyID) == 0 || partyID == acc.PartyID.String())
+				partyID == acc.PartyID.String()
```
wwestgarth, line 84: Is this a breaking change? before I could stream margin levels across all parties:
```
https://api.n06.testnet.vega.xyz/api/v2/stream/margin/levels
```
but now I can't? I suspect we also don't return a `InvalidArguement is required` error and just silently not match on anything in the stream.

wwestgarth:
I think this needs another pass.

We don't want any functional changes we just need to align the documentation against what the data node actually does for REST. Anything not marked as `optional` should not automatically be annotated as required, we need to first work out that field accepts `""` "do nothing" because those cases should not be marked as required.

jeremyletang:
> I think this needs another pass.
> 
> We don't want any functional changes we just need to align the documentation against what the data node actually does for REST. Anything not marked as `optional` should not automatically be annotated as required, we need to first work out if that field accepts `""` as being optional , because for those cases we should not mark them as required.

Should we actually have a look at the messages and mark things as optional as well as add the annotation when they are really optional?

wwestgarth:
> > I think this needs another pass.
> > We don't want any functional changes we just need to align the documentation against what the data node actually does for REST. Anything not marked as `optional` should not automatically be annotated as required, we need to first work out if that field accepts `""` as being optional , because for those cases we should not mark them as required.
> 
> Should we actually have a look at the messages and mark things as optional as well as add the annotation when they are really optional?

Definitely if we're feeling brave we could go and make every field where we do `"" => ignore field` be truly protos-optional. That would be the best thing. It'll just be a breaking change and I have no idea how many there are!

datanode/service/risk.go
guoguojin: So the API documentation for this was:

```
// Request to subscribe to a stream of MarginLevels data matching the given party identifier
// Optionally, the list can be additionally filtered by market
message ObserveMarginLevelsRequest {
    // Party identifier, required field
    string party_id = 1 [(google.api.field_behavior) = REQUIRED];
    // Market identifier
    optional string market_id = 2;
}
```

I did check with @pscott31 and we agreed it would be more correct to make the implementation work as has been documented rather than the other way around. Happy to change it back and update the documentation on the API though.

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: I don't think they should be required, it looks like the implementation allows the user to have an open ended result set, e.g. candlesticks since T, or candlesticks until T, as well as candlesticks between T and T' and the doc-string is incorrect. I'll remove the required status and update the doc string.

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: You're right, I saw the check for the direction, but missed the fact it is only checked if a party ID is provided. I will update the doc-string to provide the information accordingly.

guoguojin: Addressed https://github.com/vegaprotocol/vega/pull/8070/commits/16a86a18b2d74fe341c6e3d4e19eb7563618cc95

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: Addressed https://github.com/vegaprotocol/vega/pull/8070/commits/16a86a18b2d74fe341c6e3d4e19eb7563618cc95

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: Address https://github.com/vegaprotocol/vega/pull/8070/commits/16a86a18b2d74fe341c6e3d4e19eb7563618cc95

datanode/service/risk.go
guoguojin: Addressed https://github.com/vegaprotocol/vega/pull/8070/commits/16a86a18b2d74fe341c6e3d4e19eb7563618cc95

guoguojin:
I've reviewed and removed the required field on the ones I mistakenly marked as required. I've also added some additional validation where they are required, but aren't checked.
I have not created any breaking changes, but have updated the documentation on the APIs to reflect the actual behaviour of the API as written.

datanode/api/trading_data_v2.go
```diff
@@ -1325,6 +1325,10 @@ func (t *TradingDataServiceV2) sendPositionsSnapshot(ctx context.Context, req *v
 func (t *TradingDataServiceV2) GetParty(ctx context.Context, req *v2.GetPartyRequest) (*v2.GetPartyResponse, error) {
 	defer metrics.StartAPIRequestAndTimeGRPC("GetParty")()
 
+	if len(req.PartyId) == 0 {
+		return nil, formatE(ErrMissingPartyID)
+	}
```
jeremyletang, line 1330: For all these check for the IDs being here or not, would you also check if they are well formed?
Using crypto.IsValidVegaPubkey or crypto.IsValidVegaID?

Also for parties, "network" is a valid ID as an exception.

guoguojin: Resolved https://github.com/vegaprotocol/vega/pull/8070/commits/c16b06e4dcbd9087c4ee1fc4790d58f4f5c2eb71 and https://github.com/vegaprotocol/vega/pull/8070/commits/7c776bc1d6960656dfbad71ef5f7a9eece986e97

jeremyletang (dismissed):

jeremyletang:
{
"system-tests": "feature/remove-VOTE-token-id"
}

jeremyletang (approved):

PR #8071, opened 2023-04-03
fix: handle zero return value for total memory

### LINKED ISSUE #8069: Call to set IPFS resource uses an API that returns 0 instead of error, set a default to handle this
Call to set IPFS resource uses an API that returns 0 instead of error, set a default to handle this

ettec:
closes #8069

### DIFF
--- a/datanode/networkhistory/store/store.go
+++ b/datanode/networkhistory/store/store.go
@@ -14,9 +14,10 @@ import (
 	"strconv"
 	"strings"
 
+	"code.vegaprotocol.io/vega/libs/memory"
+
 	"github.com/dustin/go-humanize"
 	"github.com/ipfs/kubo/core/node/libp2p/fd"
-	"github.com/pbnjay/memory"
 
 	"github.com/libp2p/go-libp2p/core/peer"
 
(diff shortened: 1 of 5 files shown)


pscott31 (dismissed):

datanode/networkhistory/store/store.go
```diff
@@ -877,14 +877,22 @@ func createNode(ctx context.Context, log *logging.Logger, repoPath string, maxMe
...
+	totalMem := memory.TotalMemory()
+	// The above call can return 0 if the underlying system call errors which can lead to further panics,
+	// check for 0 and set to default to protect against this.
+	if totalMem == 0 {
+		totalMem = 100_000_000
+		log.Warningf("failed to determine total memory from system call, setting to %d bytes", totalMem)
+	}
+
 	// Set the maximum to a quarter of the data-nodes max memory
-	maxMemoryString := humanize.Bytes(uint64(float64(memory.TotalMemory()) * (float64(maxMemoryPercent) / (4 * 100))))
```
wwestgarth, line 887: Theres other places we call `memory.TotalMemory()`. We should handle `0` there too:
```
// only set max memory if user didn't require 100%
if memFactor != 1 {
	totalMem := memory.TotalMemory()
	debug.SetMemoryLimit(int64(float64(totalMem) * memFactor))
}
```

wwestgarth:
I think an error is better because it indicates a problem with the environment running the nodes, and earlier feedback on that to the operator is probably better than trying to muddle along.

wwestgarth (dismissed):

wwestgarth (approved):

PR #8077, opened 2023-04-04
fix: add missing enum variant to OrderRejectionReason in gql

### LINKED ISSUE #8074: Add missing rejection reasons for post only and reduce only to graphql schema
# Task Overview

The following order rejection reasons are missing from the graphql schema

```
REDUCE_ONLY_ORDER_WOULD_NOT_REDUCE

ORDER_ERROR_POST_ONLY_ORDER_WOULD_TRADE

ORDER_ERROR_UNSPECIFIED
```

jeremyletang:
close #8074

### DIFF
--- a/datanode/gateway/graphql/schema.graphql
+++ b/datanode/gateway/graphql/schema.graphql
@@ -3021,7 +3021,7 @@ enum OrderRejectionReason {
   "Unable to submit pegged order, temporarily too many pegged orders across all markets"
   ORDER_ERROR_TOO_MANY_PEGGED_ORDERS
 
-  "A reduce ony order would not reduce the party's position"
+  "A reduce-ony order would not reduce the party's position and thus it has been rejected"
   ORDER_ERROR_REDUCE_ONLY_ORDER_WOULD_NOT_REDUCE
 
   "A post only order would produce an aggressive trade"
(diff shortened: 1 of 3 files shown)


datanode/gateway/graphql/schema.graphql
```diff
@@ -3020,6 +3020,12 @@ enum OrderRejectionReason {
 
   "Unable to submit pegged order, temporarily too many pegged orders across all markets"
   ORDER_ERROR_TOO_MANY_PEGGED_ORDERS
+
+  "A reduce ony order would not reduce the party's position"
```
candida-d, line 3024:
```suggestion
  "A reduce-ony order would not reduce the party's position and thus it has been rejected"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3020,6 +3020,12 @@ enum OrderRejectionReason {
...
+  ORDER_ERROR_REDUCE_ONLY_ORDER_WOULD_NOT_REDUCE
+
+  "A post only order would produce an aggressive trade"
```
candida-d, line 3027:
```suggestion
  "A post-only order would produce an aggressive trade and thus it has been rejected"
```

candida-d (dismissed):

AnExsomnis (dismissed):

ze97286 (approved):

PR #8091, opened 2023-04-05
chore: improve estimateFees documentations

### LINKED ISSUE #7991: The `EstimateFee` API docs needs a full review
### Issue

The `EstimateFee` API docs needs a full review 

see:
- https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-estimate-fee
- https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-estimate-margin

### Tasks:

- [ ] There is usage of “this order” when the query parameters are not that of a Vega order object. This should be reviewed and corrected
- [ ] There ate parentheses everywhere - do we really need all these? These should be reviewed and corrected
- [ ] There is a description that states "when submitting", this is incorrect and should be "when trades"... Needs an overhaul and tidy up.

jeremyletang:
close #7991

### DIFF
--- a/protos/sources/data-node/api/v2/trading_data.proto
+++ b/protos/sources/data-node/api/v2/trading_data.proto
@@ -2821,7 +2821,7 @@ message GetEpochResponse {
 
 // Request to fetch the estimated fee if an order were to trade immediately
 message EstimateFeeRequest {
-  // MarketID used to specify the fee factors.
+  // Market ID, used to specify the fee factors.
   string market_id = 1;
   // Price at which the potential order is expected to trade.
   string price = 2;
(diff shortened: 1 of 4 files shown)


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2820,11 +2821,11 @@ message GetEpochResponse {
 
 // Request to fetch the estimated fee if an order were to trade immediately
 message EstimateFeeRequest {
-  // MarketID (used to specify the fee factors).
+  // MarketID used to specify the fee factors.
```
candida-d, line 2824:
```suggestion
  // Market ID, used to specify the fee factors.
```

candida-d (dismissed):

candida-d (approved):

PR #8108, opened 2023-04-07
Expose margin and liquidation price calculations for new API endpoint

witgaw:
Expose margin and liquidation price calculations.

Margin calculations are duplicated for now to avoid risk of introducing unintentional changes prior to launch, will add a refactoring ticket once this gets merged.

### DIFF
--- a/core/risk/liquidation_calculation.go
+++ b/core/risk/liquidation_calculation.go
@@ -1,0 +1,106 @@
+package risk
+
+import (
+	"fmt"
+	"sort"
+
+	"code.vegaprotocol.io/vega/libs/num"
+)
+
+type OrderInfo struct {
+	Size          uint64
+	Price         num.Decimal
+	IsMarketOrder bool
+}
+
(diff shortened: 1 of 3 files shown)


core/risk/liquidation_calculation.go
```diff
@@ -0,0 +1,106 @@
...
+	"code.vegaprotocol.io/vega/libs/num"
+)
+
+type OrderInfo struct {
+	Size          uint64
+	Price         num.Decimal
+	IsMarketOrder bool
+}
+
+func CalculateLiquidationPriceWithSlippageFactors(sizePosition int64, buyOrders, sellOrders []*OrderInfo, currentPrice, collateralAvailable num.Decimal, positionFactor, linearSlippageFactor, quadraticSlippageFactor, riskFactorLong, riskFactorShort num.Decimal) (liquidationPriceForOpenVolume, liquidationPriceWithSellOrders, liquidationPriceWithBuyOrders num.Decimal, err error) {
```
EVODelavega, line 16: The combination of long function and variable names, coupled with the sheer number of arguments and return values make this really rather difficult to read. Why are we passing things like position factor, linear and quadratic factors, and the risk factors as individual values, when having these functions as receiver functions on the risk engine would eliminate the need to pass these values all together?

witgaw: I wanted to have a static function that the data node can pass all the inputs into when queried for a liquidation price for a particular market.

EVODelavega (approved):

PR #8110, opened 2023-04-07
Estimate position API

### LINKED ISSUE #7847: Refactor `EstimateOrder` API method

witgaw:
Closes https://github.com/vegaprotocol/vega/issues/7847


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+ // Side for the order, e.g. SIDE_BUY or SIDE_SELL.
+  vega.Side side = 1;
+  // Price for the order, the price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places
+  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  string price = 2;
+  // Size remaining, when this reaches 0 then the order is fully filled and status becomes STATUS_FILLED
+  // This field is an unsigned integer passed as a string and needs to be scaled using the market's position decimal places.
+  uint64 remaining = 3;
+  // Boolean indicative a market order.
```
candida-d, line 3329:
```suggestion
  // Boolean that indicates if it is a market order.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
 // Ping response from the data node
 message PingResponse {
 }
+
+// Basic description of an order
+message OrderInfo { 
+ // Side for the order, e.g. SIDE_BUY or SIDE_SELL.
```
candida-d, line 3320:
```suggestion
 // Side for the order, e.g. buy or sell.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+  vega.Side side = 1;
+  // Price for the order, the price is an integer, for example `123456` is a correctly
```
candida-d, line 3322:
```suggestion
  // Price for the order. The price is an integer, for example `123456` is a correctly
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+  string price = 2;
+  // Size remaining, when this reaches 0 then the order is fully filled and status becomes STATUS_FILLED
+  // This field is an unsigned integer passed as a string and needs to be scaled using the market's position decimal places.
+  uint64 remaining = 3;
+  // Boolean indicative a market order.
+  bool is_market_order = 4;
+}
+
+
+// Request for the estimate of the margin level and (if available collateral is provided) liqudation price for the specified position
```
candida-d, line 3334:
```suggestion
// Request for the estimate of the margin level and, if available collateral is provided, liqudation price for the specified position.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+message EstimatePositionRequest {
+     // Market identifier.
+    string market_id = 1;
+    // Open volume (negative for short position).
```
candida-d, line 3338:
```suggestion
    // Open volume - negative for short position.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+    int64 open_volume = 2;
+    // Open and/or hypothetical orders.
+    repeated OrderInfo orders = 3;
+    // Optional argument specifying collateral available for the position, if provided then response will contain the liquidation price estimate.
+    string collateral_available = 4; 
+}
+
+// Response for the estimate of the margin level and (if available collateral was provided in the request) liqudation price for the specified position
```
candida-d, line 3346:
```suggestion
// Response for the estimate of the margin level and, if available collateral was provided in the request, liqudation price for the specified position.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+message EstimatePositionResponse {
+    // Margin level range estimate for the specified position.
+    MarginEstimate margin = 1;
+    // Liquidation price range estimate for the specified position (only populated if available collateral was specified in the request).
```
candida-d, line 3350:
```suggestion
    // Liquidation price range estimate for the specified position. Only populated if available collateral was specified in the request.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+    LiquidationEstimate liquidation = 2;
+}
+
+// Margin level estimate for both worst and best case possible
```
candida-d, line 3354:
```suggestion
// Margin level estimate for both worst and best case possible.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2829,7 +2837,8 @@ message EstimateMarginRequest {
 
 // Response to an estimate margin request, containing the estimated margin levels for a given order
 message EstimateMarginResponse {
-    // Summary of the estimated margins for this order if it were to trade now.
+    option deprecated = true;
+    // Summary of the estimated margins for this order if it were to trade now
```
candida-d, line 2841:
```suggestion
    // Summary of the estimated margins for this order if it were to trade now.
```
This needs a full stop to generate the documentation correctly.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2813,9 +2820,10 @@ message EstimateFeeResponse {
 
 // Request to fetch the estimated MarginLevels if an order were to trade immediately
 message EstimateMarginRequest {
+    option deprecated = true;
     // Market ID for the order.
     string market_id = 1 [(google.api.field_behavior) = REQUIRED];
-    // Party ID of the order.
+    // Market ID for the order
```
candida-d, line 2826: Why was this changed? Looks like it's referencing the party ID string, not Market ID (also needs to keep the full stop)

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+// Margin level estimate for both worst and best case possible
+message MarginEstimate {
+    // Margin level estimate assuming slippage cap is applied.
+    vega.MarginLevels worst_case = 1;
+    // Margin level estimate assuming no slippage.
+    vega.MarginLevels best_case = 2;
+}
+
+
+// Liquidation estimate for both worst and best case possible
```
candida-d, line 3363:
```suggestion
// Liquidation estimate for both worst and best case possible.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+message LiquidationEstimate {
+    // Liquidation price estimate assuming slippage cap is applied.
+    LiquidationPrice worst_case = 1;
+    // Liquidation price estimate assuming no slippage.
+    LiquidationPrice best_case = 2;
+}
+
+// Liquidation price estimate for: only the current open volume, position given some or all buy orders get filled, position given some or all sell orders get filled
```
candida-d, line 3371:
```suggestion
// Liquidation price estimate for either only the current open volume and position given some or all buy orders get filled, or position given some or all sell orders get filled.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+
+// Basic description of an order
+message OrderInfo { 
+ // Side for the order, e.g. SIDE_BUY or SIDE_SELL.
+  vega.Side side = 1;
+  // Price for the order, the price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places
+  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  string price = 2;
+  // Size remaining, when this reaches 0 then the order is fully filled and status becomes STATUS_FILLED
```
candida-d, line 3326:
```suggestion
  // Size remaining, when this reaches 0 then the order is fully filled and status becomes filled.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
 // Ping response from the data node
 message PingResponse {
 }
+
+// Basic description of an order
+message OrderInfo { 
+ // Side for the order, e.g. SIDE_BUY or SIDE_SELL.
+  vega.Side side = 1;
+  // Price for the order, the price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places
```
candida-d, line 3323:
```suggestion
  // formatted price of `1.23456` assuming market configured to 5 decimal places.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
 // Ping response from the data node
 message PingResponse {
 }
+
+// Basic description of an order
```
candida-d, line 3318:
```suggestion
// Basic description of an order.
```

candida-d (commented):
Documentation suggestions

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Some accidental change, reverted now.

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
 	}, nil
 }
 
+func (r *myQueryResolver) EstimateFees(
```
jeremyletang, line 858: Didn't we already have an EstimateFees API?

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
...
+	}
+
+	order.PartyId = party
+	order.TimeInForce = timeInForce
+	order.Side = side
+	order.Type = ty
+
+	// GTT must have an expiration value
+	if order.TimeInForce == types.Order_TIME_IN_FORCE_GTT && expiration != nil {
+		order.ExpiresAt = vegatime.UnixNano(*expiration).UnixNano()
```
jeremyletang, line 896: Not sur this makes sense?

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
...
+	ctx context.Context,
+	market, party string,
+	price *string,
+	size string,
+	side vega.Side,
+	timeInForce vega.Order_TimeInForce,
+	expiration *int64,
+	ty vega.Order_Type,
+) (*FeeEstimate, error) {
+	order := &types.Order{}
```
jeremyletang, line 868: Why do we have an order here?

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
...
+	openVolume string,
+	orders []*OrderInfo,
+	collateralAvailable *string,
+) (*PositionEstimate, error) {
+	collat := ""
+	if collateralAvailable != nil {
+		collat = *collateralAvailable
+	}
+
+	ov, err := safeStringInt64(openVolume)
```
jeremyletang, line 949: Why not use an int64 directly here?

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
...
+
+	ord := make([]*v2.OrderInfo, 0, len(orders))
+	for _, o := range orders {
+		mktOrder := false
+		if o.IsMarketOrder != nil {
+			mktOrder = *o.IsMarketOrder
+		}
+
+		r, err := safeStringUint64(o.Remaining)
+		if err != nil {
```
jeremyletang, line 962: Same here?

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
...
+	}
+
+	resp, err := r.tradingDataClientV2.EstimatePosition(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	var liquidation *LiquidationEstimate
+	if resp.Liquidation != nil {
+		worst := &LiquidationPrice{
```
jeremyletang, line 988: Could you set the bindings to the proto types in the graphql config so we can avoid doing those mappings here?

datanode/gateway/graphql/resolvers.go
witgaw: No, it was part of `EstimateOrder`, but that uses `EstimateMargin` which I've marked as deprecated (would like us to delete it once frontend switches to the new method), so I've extracted the fees bit to a separate method so they can use that one.

datanode/gateway/graphql/resolvers.go
```diff
@@ -855,6 +855,161 @@ func (r *myQueryResolver) EstimateOrder(
...
+) (*FeeEstimate, error) {
+	order := &types.Order{}
+
+	var err error
+
+	// We need to convert strings to uint64 (JS doesn't yet support uint64)
+	if price != nil {
+		order.Price = *price
+	}
+	s, err := safeStringUint64(size)
```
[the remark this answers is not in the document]
witgaw: I just copied the fees part from the method above (`EstimateOrder`)

datanode/gateway/graphql/resolvers.go
witgaw: It seems to be the convention in `schema.graphl` to model these things as strings.

datanode/gateway/graphql/resolvers.go
witgaw: Done now.

datanode/gateway/graphql/schema.graphql
```diff
@@ -2206,6 +2238,18 @@ type PositionUpdate {
...
+"Basic description of an order"
+input OrderInfo {
+  "Whether the order is to buy or sell"
+  side: Side!
+  "Price for the order"
+  price: String!
+  "Number of units remaining of the total that have not yet been bought or sold (uint64)"
+  remaining: String!
+  "Boolean indicating a market order"
+  isMarketOrder: Boolean
```
mattrussell36, line 2250: is this truly Boolean or undefined? Update to `Boolean!` if not

witgaw: I thought an optional boolean (assume false if not specified) will be more user friendly, but perhaps it might just lead to confusion.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+  // This field is an unsigned integer passed as a string and needs to be scaled using the market's position decimal places.
+  uint64 remaining = 3;
+  // Boolean that indicates if it is a market order.
+  bool is_market_order = 4;
+}
+
+
+// Request for the estimate of the margin level and, if available collateral is provided, liqudation price for the specified position.
+message EstimatePositionRequest {
+     // Market identifier.
```
gordsport, line 3336:
```suggestion
     // Market ID.
```

gordsport (changes requested):

protos/sources/data-node/api/v2/trading_data.proto
witgaw: This field is labeled as "Market identifier" in most cases in .proto.

gordsport: The API clean ups in: https://github.com/vegaprotocol/vega/pull/8126

Try to make this consistent across all comments that become API docs.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -660,10 +660,17 @@ service TradingDataService {
...
 	tags: "Orders";
       };
-    }
+    };
+
+    // Estimate position
+    //
+    // Estimate the margin that would be required for maintaining the specified position.
+    // If the optional collateral available argument is supplied, the response also contains the estimate of the liquidation price.
+    rpc EstimatePosition(EstimatePositionRequest) returns (EstimatePositionResponse);
```
wwestgarth, line 673: Can we add a tag to this API so that it gets put in the correct category in the auto-generated docs:
```
option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
	tags: "???";
      };
```
`EstimateMargin` was under `Orders` but we also have a category for `Positions` which might be better?

datanode/api/trading_data_v2.go
```diff
@@ -2632,16 +2644,198 @@ func (t *TradingDataServiceV2) estimateMargin(
...
+		CollateralReleaseLevel: maintenanceMargin.Mul(num.DecimalFromFloat(scalingFactors.CollateralRelease)).Round(0).String(),
+	}
+}
+
+func (t *TradingDataServiceV2) EstimatePosition(ctx context.Context, req *v2.EstimatePositionRequest) (*v2.EstimatePositionResponse, error) {
+	defer metrics.StartAPIRequestAndTimeGRPC("EstimatePosition")()
+
+	rf, err := t.riskFactorService.GetMarketRiskFactors(ctx, req.MarketId)
+	if err != nil {
+		return nil, errors.Wrapf(err, "getting risk factors: %s", req.MarketId)
```
wwestgarth, line 2669: Can we use `formatE()` for every return to make sure the right error code is used? For example if `req.MarketId` doesn't exist, or is `""` it should be an `InvalidArgument` error.

datanode/api/trading_data_v2.go
```diff
@@ -2632,16 +2644,198 @@ func (t *TradingDataServiceV2) estimateMargin(
...
+	}
+	quadraticSlippageFactor, err := num.DecimalFromString(mktProto.QuadraticSlippageFactor)
+	if err != nil {
+		return nil, errors.Wrapf(err, "parsing quadratic slippage factor: %s", mktProto.QuadraticSlippageFactor)
+	}
+
+	buyOrders := make([]*risk.OrderInfo, 0, len(req.Orders))
+	sellOrders := make([]*risk.OrderInfo, 0, len(req.Orders))
+
+	for _, o := range req.Orders {
```
wwestgarth, line 2733: Should we check that `len(req.Orders) != 0` right at the top of the function call? I'm imagining this would be an `InvalidArgument` type of error.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -18,7 +18,7 @@ option go_package = "code.vegaprotocol.io/vega/protos/data-node/api/v2";
 option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
   info: {
     title: "Vega data node APIs";
-    version: "v0.71.0+dev";
+    version: "v0.70.0";
```
wwestgarth, line 21: Merge error?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+
+// Basic description of an order.
+message OrderInfo { 
+ // Side for the order, e.g. buy or sell.
+  vega.Side side = 1;
+  // Price for the order. The price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places.
+  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  string price = 2;
+  // Size remaining, when this reaches 0 then the order is fully filled and status becomes filled.
```
wwestgarth, line 3326: Was this comment copied from elsewhere? I'm not sure if this bit makes sense in this context:
```
when this reaches 0 then the order is fully filled and status becomes filled.
```

Also the field is a `uint64` but the doc says:
"// This field is an unsigned integer **passed as a string**"

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+// Request for the estimate of the margin level and, if available collateral is provided, liqudation price for the specified position.
+message EstimatePositionRequest {
+     // Market identifier.
+    string market_id = 1;
+    // Open volume - negative for short position.
+    int64 open_volume = 2;
+    // Open and/or hypothetical orders.
+    repeated OrderInfo orders = 3;
+    // Optional argument specifying collateral available for the position, if provided then response will contain the liquidation price estimate.
+    string collateral_available = 4;
```
wwestgarth, line 3343: If this is optional, can we have it as `optional string collateral_available = 4;`? 

Admittedly we are inconsistent with this in other API and use `""` as "not set" and its too late for those, but going forward in new API I think we should use `optional` when its optional.

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Ah ok, didn't know - thanks! Committed your suggestions now.

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Changed now.

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Yeah, probably - corrected now.

datanode/api/trading_data_v2.go
witgaw: `len(req.Orders)` == 0 is fine, party may not have any open. Added a check for nil order.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3305,3 +3314,66 @@ message PingRequest {
...
+  uint64 remaining = 3;
+  // Boolean that indicates if it is a market order.
+  bool is_market_order = 4;
+}
+
+
+// Request for the estimate of the margin level and, if available collateral is provided, liqudation price for the specified position.
+message EstimatePositionRequest {
+     // Market identifier.
+    string market_id = 1;
```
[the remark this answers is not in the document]
witgaw: Added now

datanode/api/trading_data_v2.go
witgaw: Done now (_I think,_ please check).

datanode/gateway/graphql/resolvers.go
jeremyletang: it's not for sizes, look at the Order type.

datanode/gateway/graphql/resolvers.go
jeremyletang: well, it's unused, lets just remove it?

datanode/gateway/graphql/resolvers.go
jeremyletang: hum alright.

datanode/gateway/graphql/resolvers.go
```diff
@@ -870,6 +870,135 @@ func (r *myQueryResolver) EstimateOrder(
...
+	expiration *int64,
+	ty vega.Order_Type,
+) (*FeeEstimate, error) {
+	order := &types.Order{}
+
+	var err error
+
+	// We need to convert strings to uint64 (JS doesn't yet support uint64)
+	if price != nil {
+		order.Price = *price
```
EVODelavega, line 889: Just wondering, because I noticed the price factor function a bit earlier, what decimal places will this price be in? Seeing as this is part of the user-facing API, are we using asset decimals, and later converting to market precision? 

TL;DR: are we taking decimals in to account here?

gordsport (dismissed):
Approving on behalf of docs team

datanode/api/trading_data_v2.go
```diff
@@ -2632,16 +2644,201 @@ func (t *TradingDataServiceV2) estimateMargin(
...
+		rf,
+		auction,
+		asset,
+		mkt.TradableInstrument.MarginCalculator.ScalingFactors)
+
+	var liquidationEstimate *v2.LiquidationEstimate
+	if req.CollateralAvailable != nil && len(*req.CollateralAvailable) > 0 {
+		collateralAvailable, err := num.DecimalFromString(*req.CollateralAvailable)
+		if err != nil {
+			return nil, formatE(fmt.Errorf("could not parse collateral available ('%s')", *req.CollateralAvailable), err)
```
jeremyletang, line 2770: Could you make the first error  here a newInvalidArgumentError (look at the errors.go) it will map to the approriate error in the http response.
Also I'd add this validation at the beginning as it unecessary to do all the previous stuff if there's an user input.

datanode/api/trading_data_v2.go
```diff
@@ -2632,16 +2644,201 @@ func (t *TradingDataServiceV2) estimateMargin(
...
-		SearchLevel:            maintenanceMargin.Mul(num.DecimalFromFloat(mkt.TradableInstrument.MarginCalculator.ScalingFactors.SearchLevel)).Round(0).String(),
-		InitialMargin:          maintenanceMargin.Mul(num.DecimalFromFloat(mkt.TradableInstrument.MarginCalculator.ScalingFactors.InitialMargin)).Round(0).String(),
-		CollateralReleaseLevel: maintenanceMargin.Mul(num.DecimalFromFloat(mkt.TradableInstrument.MarginCalculator.ScalingFactors.CollateralRelease)).Round(0).String(),
+		SearchLevel:            maintenanceMargin.Mul(num.DecimalFromFloat(scalingFactors.SearchLevel)).Round(0).String(),
+		InitialMargin:          maintenanceMargin.Mul(num.DecimalFromFloat(scalingFactors.InitialMargin)).Round(0).String(),
+		CollateralReleaseLevel: maintenanceMargin.Mul(num.DecimalFromFloat(scalingFactors.CollateralRelease)).Round(0).String(),
+	}
+}
+
+func (t *TradingDataServiceV2) EstimatePosition(ctx context.Context, req *v2.EstimatePositionRequest) (*v2.EstimatePositionResponse, error) {
```
jeremyletang, line 2664: In general could we  move all user input validation straight at the beginning? Moreover all the one which requires no dependent calls to the services?

datanode/gateway/graphql/resolvers.go
witgaw: Do you mean the `order` or `err`? Order is clearly being used, I've removed the unused error from the methods.

datanode/gateway/graphql/resolvers.go
witgaw: This one?
https://github.com/vegaprotocol/vega/blob/3ef1e3b381c988cfeab3050c5e21ba1f18a34569/datanode/gateway/graphql/schema.graphql#L2270

datanode/gateway/graphql/resolvers.go
witgaw: It does the conversion of price and volume / order size, but would be good if you could have a closer look at the method and check that it's doing it correctly.

datanode/api/trading_data_v2.go
witgaw: Done now

datanode/gateway/graphql/resolvers.go
EVODelavega: My concern is that the price the datanode ingests is using market precision (e.g. asset being ETH - 18 decimals, market using 5 decimals => if the price on the event received by the datanode is 1, that equates to 0.00001 ETH, not 1 wei). I'll take a look though

datanode/api/trading_data_v2.go
```diff
@@ -2647,16 +2659,205 @@ func (t *TradingDataServiceV2) estimateMargin(
...
+	if err != nil {
+		return nil, formatE(err)
+	}
+
+	auction := mktData.AuctionEnd > 0
+	if auction && mktData.MarketTradingMode == types.MarketTradingModeOpeningAuction.String() {
+		marketObservable = mktData.IndicativePrice
+	}
+
+	uMarketObservable, overflowed := num.UintFromDecimal(marketObservable)
```
EVODelavega, line 2752: This is weird... a few lines higher up, we use `num.DecimalFromString` to set `marketObservable`, only to convert it to `num.Uint` here, and by the looks of it, we pass the uint in to `scaleFromMarketToAssetPrice`. Why go through the decimal type here? It makes much more sense to use `num.Uint` as much as possible, and only convert it to decimal if we absolutely have to.

datanode/api/trading_data_v2.go
```diff
@@ -2647,16 +2659,205 @@ func (t *TradingDataServiceV2) estimateMargin(
...
+	linearSlippageFactor, err := num.DecimalFromString(mktProto.LinearSlippageFactor)
+	if err != nil {
+		return nil, formatE(fmt.Errorf("can't parse linear slippage factor: %s", mktProto.LinearSlippageFactor), err)
+	}
+	quadraticSlippageFactor, err := num.DecimalFromString(mktProto.QuadraticSlippageFactor)
+	if err != nil {
+		return nil, formatE(fmt.Errorf("can't parse quadratic slippage factor: %s", mktProto.QuadraticSlippageFactor), err)
+	}
+
+	marginEstimate := t.computeMarginRange(
```
EVODelavega, line 2776: A function with the name `computeMarginRange`, and in general this much processing/logic really doesn't shouldn't be sitting in a handler like this. Convert the input required, and pass it to a component that contains the meat of the logic/work that needs to be done. It makes it easier to review, maintain, and work on

witgaw: Did you look inside the function? It just calls an external component twice, so not sure what you mean?

datanode/gateway/graphql/schema.graphql
mattrussell36: I'd prefer a stricter type, but consider this non blocking

mattrussell36 (dismissed):

datanode/gateway/graphql/schema.graphql
witgaw: Fair enough, changed now.

gordsport (approved):

jeremyletang (approved):

PR #8115, opened 2023-04-11
Fix protocol upgrade LevelDB error

zale144:
Closes #7938


datanode/networkhistory/store/store.go
```diff
@@ -226,12 +228,16 @@ func (p *Store) Stop() {
 	p.log.Info("Cleaning up network history store")
 	if p.ipfsNode != nil {
 		p.log.Info("Closing IPFS node")
-		_ = p.ipfsNode.Close()
+		if err := p.ipfsNode.Close(); err != nil {
+			p.log.Errorf("Failed to close IPFS node:%s", err)
```
jeremyletang, line 232: We usually used structured logs instead of formtated ones like so:
```suggestion
			p.log.Error("Failed to close IPFS node", logging.Error(err))
```

datanode/networkhistory/store/store.go
```diff
@@ -313,7 +319,8 @@ func (p *Store) ResetIndex() error {
 		return fmt.Errorf("failed to create index path:%w", err)
 	}
 
-	p.index, err = NewIndex(p.indexPath)
+	idxLog := p.log.With(zap.String("component", "index"))
```
jeremyletang, line 322: Same about the usage of zap directly here.

guoguojin (approved):

pscott31 (approved):

PR #8127, opened 2023-04-12
refactor: modify price monitoring behaviour

### LINKED ISSUE #8128: Price monitoring auction should trigger at most one bound at a time

witgaw:
Allow breaching at most one bound at a time

Closes https://github.com/vegaprotocol/vega/issues/8128


davidsiska-vega (approved):

ze97286 (approved):

gordsport:
When merging please also merged the approved specs PR:

- https://github.com/vegaprotocol/specs/pull/1660

core/monitor/price/pricemonitoring.go
```diff
@@ -446,6 +446,8 @@ func (e *Engine) checkBounds(trades []*types.Trade) []*types.PriceMonitoringTrig
 				ret = append(ret, b.Trigger)
 				// deactivate the bound that just got violated so it doesn't prevent auction from terminating
 				b.Active = false
+				// only allow breaking one bound at a time
```
EVODelavega, line 449: This is fine IMO, but I believe back when this was first implemented, the auction duration was supposed to depend on by how much the price had moved out of the bounds, hence we checked all price bounds, and to me, it looks like the spec still reflects this: https://github.com/vegaprotocol/specs/blob/master/protocol/0032-PRIM-price_monitoring.md#view-from-the-price-monitoring-engine-side specifically the bit that reads: _"to give an example, with 3 triggers the price protection auction can be calculated as follows"_, where 3 examples are listed, with an auction of 5m, 15m, and 1h respectively depending on how much the price bound was exceeded

jeremyletang (approved):

PR #8138, opened 2023-04-13
feat: add reset all command

### LINKED ISSUE #8131: Wipe On Startup Flags should default to false
The flags to wipe the database and network-history at startup are defaulted to true, which was convenient for development, at this stage we should default these to false.

ettec:
closes #8131 Remove wipe on startup flags and add unsafe reset all command

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - [8030](https://github.com/vegaprotocol/vega/issues/8030) - Add `API` for fetching `CSV` data from network history.
 - [7943](https://github.com/vegaprotocol/vega/issues/7943) - Add version to network file to be future-proof.
 - [7759](https://github.com/vegaprotocol/vega/issues/7759) - Support for rolling back data node to a previous network history segment
+- [8131](https://github.com/vegaprotocol/vega/issues/8131) - Add reset all command to data node and remove wipe on start up flags 
 - [7505](https://github.com/vegaprotocol/vega/issues/7505) - `Datanode` batcher statistics
 - [8045](https://github.com/vegaprotocol/vega/issues/8045) - Fix bug in handling internal sources data.
 - [7843](https://github.com/vegaprotocol/vega/issues/7843) - Report partial batch market instruction processing failure
--- a/cmd/data-node/commands/unsafe_reset_all.go
+++ b/cmd/data-node/commands/unsafe_reset_all.go
@@ -13,12 +13,14 @@
 package commands
 
 import (
+	"context"
+	"fmt"
+
 	"code.vegaprotocol.io/vega/cmd/data-node/commands/start"
 	"code.vegaprotocol.io/vega/datanode/config"
 	"code.vegaprotocol.io/vega/logging"
 	"code.vegaprotocol.io/vega/paths"
-	"context"
-	"fmt"
+
 	"github.com/jessevdk/go-flags"
 )
 
--- a/cmd/data-node/commands/unsafe_reset_all.go
+++ b/cmd/data-node/commands/unsafe_reset_all.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2022 Gobalsky Labs Limited
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at https://www.mariadb.com/bsl11.
+//
+// Change Date: 18 months from the later of the date of the first publicly
+// available Distribution of this version of the repository, and 25 June 2022.
+//
+// On the date above, in accordance with the Business Source License, use
+// of this software will be governed by version 3 or later of the GNU General
+// Public License.
+
+package commands
+
+import (
+	"code.vegaprotocol.io/vega/cmd/data-node/commands/start"
+	"code.vegaprotocol.io/vega/datanode/config"
+	"code.vegaprotocol.io/vega/logging"
+	"code.vegaprotocol.io/vega/paths"
+	"context"
+	"fmt"
+	"github.com/jessevdk/go-flags"
+)
+
+type UnsafeResetAllCmd struct {
+	config.VegaHomeFlag
+	*config.Config
+}
+
+//nolint:unparam
+func (cmd *UnsafeResetAllCmd) Execute(_ []string) error {
+	ctx, cfunc := context.WithCancel(context.Background())
+	defer cfunc()
+	log := logging.NewLoggerFromConfig(
+		logging.NewDefaultConfig(),
+	)
+	defer log.AtExit()
+
+	vegaPaths := paths.New(cmd.VegaHome)
+
+	cfgLoader, err := config.InitialiseLoader(vegaPaths)
+	if err != nil {
+		return fmt.Errorf("couldn't initialise configuration loader: %w", err)
+	}
+
+	cmd.Config, err = cfgLoader.Get()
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 - [8030](https://github.com/vegaprotocol/vega/issues/8030) - Add `API` for fetching `CSV` data from network history.
 - [7943](https://github.com/vegaprotocol/vega/issues/7943) - Add version to network file to be future-proof.
 - [7759](https://github.com/vegaprotocol/vega/issues/7759) - Support for rolling back data node to a previous network history segment
+- [7759](https://github.com/vegaprotocol/vega/issues/7759) - Support for rolling back data node to a previous network history segment
```
guoguojin, line 19: This has been duplicated from above.

cmd/data-node/commands/unsafe_reset_all.go
```diff
@@ -0,0 +1,75 @@
...
+	*config.Config
+}
+
+//nolint:unparam
+func (cmd *UnsafeResetAllCmd) Execute(_ []string) error {
+	log := logging.NewLoggerFromConfig(
+		logging.NewDefaultConfig(),
+	)
+	defer log.AtExit()
+
```
guoguojin, line 39: I'm not sure if you want to add a warning and prompt for confirmation that they really want to do this, while they have to specifically want to run the command, usually when you are about to wipe all your data, you are warned about it as a precaution. Just a suggestion though.

cmd/data-node/commands/unsafe_reset_all.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return fmt.Errorf("couldn't initialise configuration loader: %w", err)
+	}
+
+	cmd.Config, err = cfgLoader.Get()
+	if err != nil {
+		return fmt.Errorf("failed to load configuration: %w", err)
+	}
+
+	if err := sqlstore.WipeDatabaseAndMigrateSchemaToLatestVersion(log, cmd.Config.SQLStore.ConnectionConfig, sqlstore.EmbedMigrations,
```
wwestgarth, line 52: I don't think we want the `AndMigrateSchemaToLatestVersion` bit? If I have a fresh postgres instance and I run unsafe_reset_all I would expect it to leave the it alone and not add tables and things. 

For comparison the `unsafe_reset_all` on the blockexplorer just calls `DropAllTablesAndViews()`.

cmd/data-node/commands/unsafe_reset_all.go
ettec: Yeah, I based this on what the action does for core, it doesn't prompt for confirmation, I believe the reasoning here is that the name of the command itself is sufficient warning, which is reasonable

cmd/data-node/commands/unsafe_reset_all.go
ettec: good point, I now just drop the DB and recreate an empty DB using some existing code

cmd/data-node/commands/unsafe_reset_all.go
```diff
@@ -0,0 +1,75 @@
...
+		bool(cmd.Config.SQLStore.VerboseMigration)); err != nil {
+		return fmt.Errorf("failed to wipe database:%w", err)
+	}
+	log.Info("Wiped all existing data from the database")
+
+	err = os.RemoveAll(vegaPaths.StatePathFor(paths.DataNodeNetworkHistoryHome))
```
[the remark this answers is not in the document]
ettec: sorted

ettec:
> > There are a handful of places in system tests where they set the `WipeOnStartup` config to false.
> > I don't think this change will cause any of the tests to fail because the default is now false, but can you open a system-test ticket with a some sort of label like `refactor` so that they QA can go through an remove all their uses of the config option?
> > Edit: actually looking at the smoke test run this change does break things. They might now need to call `datanode unsafe-reset-all` before a checkpoint restore.
> 
> Yeah, going to run a full nightly to figure out what I need to fix.

So, unpicking the sys tests reliance on the wipe flag is proving a little trickier than expected, mainly because my knowledge of this stuff is low (though better now!), and to verify requires a 3 hour full run.  So, I've updated this branch and added back in the wipe-on flags and marked them deprecated, also defaulted to FALSE instead of TRUE.   The unsafe_reset_all command exists on this branch still and the deprecated wipe flag uses this command in place of the old wipe logic.  The idea here is we can merge this branch without breaking the system tests  (not quite true at the moment as changing the flag to default false has broken a few tests as well - working on fixing this),  and then, as we have time over the next few days the system tests can be updated to remove the dependency on the wipe flag and instead use the new command.   Once this is done, we can remove the deprecated flag, so the change to sort out the system tests dependency need not hold up this change.

MuthuVega:
```
{"system-tests": "fix_wipe_on_startup_flag"}
```

ettec:
> > > There are a handful of places in system tests where they set the `WipeOnStartup` config to false.
> > > I don't think this change will cause any of the tests to fail because the default is now false, but can you open a system-test ticket with a some sort of label like `refactor` so that they QA can go through an remove all their uses of the config option?
> > > Edit: actually looking at the smoke test run this change does break things. They might now need to call `datanode unsafe-reset-all` before a checkpoint restore.
> > 
> > 
> > Yeah, going to run a full nightly to figure out what I need to fix.
> 
> So, unpicking the sys tests reliance on the wipe flag is proving a little trickier than expected, mainly because my knowledge of this stuff is low (though better now!), and to verify requires a 3 hour full run. So, I've updated this branch and added back in the wipe-on flags and marked them deprecated, also defaulted to FALSE instead of TRUE. The unsafe_reset_all command exists on this branch still and the deprecated wipe flag uses this command in place of the old wipe logic. The idea here is we can merge this branch without breaking the system tests (not quite true at the moment as changing the flag to default false has broken a few tests as well - working on fixing this), and then, as we have time over the next few days the system tests can be updated to remove the dependency on the wipe flag and instead use the new command. Once this is done, we can remove the deprecated flag, so the change to sort out the system tests dependency need not hold up this change.

@MuthuVega has fixed the smoke tests, pending merge of sys test branch.  Started running 2 nightly tests (~3hrs):

this is base line running against develop

https://jenkins.ops.vega.xyz/job/common/job/system-tests/21054/

This is against the branch:
https://jenkins.ops.vega.xyz/job/common/job/system-tests/21055/

If these have the same test results, then we are good to merge this branch.

pscott31:
So in the full run matt kicked off these three extra tests failed with this change; I think we should just merge this and then fix up the system tests.. any objections?

Call tests / full a-f / tests.datanode.datanode_test.test_datanode_history_segments_are_deterministic
Call tests / full a-f / tests.datanode.datanode_test.test_datanode_restore_from_dehistory
Call tests / network_infra validators a-p / tests.validators.multisig_test.test_multisig_scores_swap_with_sigs

cmd/data-node/commands/unsafe_reset_all.go
pscott31: It would be nice if we could just either delete all tables + views, or maybe just the schema (e.g. 'public') then this command wouldn't need DB admin rights but we can add that later

jeremyletang (approved):

PR #8151, opened 2023-04-17
Network History Fetch command should retry

### LINKED ISSUE #8137: datanode hangs on fetching history for segment
# Problem encountered
One of the validators run into the issue that its datanode wrote `fetching history for segment` into the logs, and nothing more. Visor and core kept waiting for datanode forever. They were using `v0.70.2`.

We also see this one on our Snapshot pipeline:
- Fairground running `v0.70.0`: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/private%2FSnapshots%2FFairground/detail/Fairground/20773/pipeline/111
- Validators Testnet running `v0.70.2`: https://jenkins.ops.vega.xyz/blue/organizations/jenkins/private%2FSnapshots%2FValidators-Testnet/detail/Validators-Testnet/4030/pipeline/112
-

### LINKED ISSUE #8145: Restored node will not act a source for old history segments.
Segments that are fetched by a node during restoration of node from network history are not pinned by default, this means the node will only act as a source of segments created after it is restored (segments created by the node are pinned).  Fetched segments should be pinned.

### LINKED ISSUE #8146: Improve network history fetch command behaviour
We have observed the network-history fetch command failing to retrieve segments (@fkondej on validator test net).  Testing (using the ipfs daemon) shows that when this failure occurs, issuing a refetch resolves the problem.  We have this re-fetch behaviour already when auto-initialising a node from network history, similar behaviour should be added to the fetch logic (possibly replacing the equivalent logic in auto-initialise)

ettec:
closes #8146 closes #8137 closes #8145

Network History Fetch command should retry and consolidate the retry behaviour in node initialisation into one place.

### DIFF
--- a/datanode/networkhistory/config.go
+++ b/datanode/networkhistory/config.go
@@ -19,6 +19,9 @@ type Config struct {
 	Store    store.Config    `group:"Store" namespace:"store"`
 	Snapshot snapshot.Config `group:"Snapshot" namespace:"snapshot"`
 
+	FetchRetryMax int               `long:"fetch-retry-max" description:"maximum number of times to retry fetching segments - default 10"`
+	RetryTimeout  encoding.Duration `long:"retry-timeout" description:"time to wait between retries, increases with each retry - default 5s"`
+
 	Initialise InitializationConfig `group:"Initialise" namespace:"initialise"`
 }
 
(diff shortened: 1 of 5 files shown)


datanode/networkhistory/config.go
```diff
@@ -32,6 +35,8 @@ func NewDefaultConfig() Config {
 		Publish:       true,
 		Store:         store.NewDefaultConfig(),
 		Snapshot:      snapshot.NewDefaultConfig(),
+		FetchRetryMax: 10,
```
guoguojin, line 38: The default values don't match the values specified in the CLI command.

guoguojin (dismissed):

jeremyletang (approved):

Issue #7291, opened 2023-01-10
Investigate what we can do to make GraphQL more robust


pscott31:
# Spike Overview

It is currently not very hard to craft GraphQL queries which spawn hundreds of thousands of child gRPC requests (often simultaneously). These sort of requests can cause data-nodes to become unresponsive.

We discussed [at some length](https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1670930980721929) on slack the idea of removing GraphQL altogether, and I don't think it's really feasible; TFE are very long GraphQL and as pointed out it just punts some of the n+1 problems we have down the stack to them.

Some ideas include:

### Flattening the GraphQL schema
- A good principle might be to not allow nesting of things that are big enough to need paging. It doesn't really make sense to have
```graphql
fooConnection{ edges { node {  barConnection { edges {  node { } } } } } } 
``` 
Anyway as we don't have a very sensible way of paging through the 'bars'. So we could make a rule that all the `connections` should live at the top level of the query, and if you want to do something like
```graphql
{ parties { trades } }
```
It would just have to become two graphql queries
```graphql
parties { edges { nodes { party_id } } }
trades(party_id=xxx) { edges { nodes { ... } } }
```

To prevent having to make very many queries we could do something like
```graphql
parties { edges { nodes { party_id } } }
trades(party_ids=[xxx, yyy]) { edges { nodes { ... } } }
```

**Pros:**
  - Quite simple to do in backend
  - Should be reasonably effective at preventing datanode DDOS

**Cons:**
  - Some front-end refactoring required
  - Some of the graphQL magic powers is reduced
     - If we want to keep some of those superpowers but not have nested pagination we could keep but remove pagination for nested lists things of a reasonable size, e.g.
```graphql
parties { edges { nodes { last_100_trades{ ... } } } }
````

### Using Dataloaders

[Dataloaders ](https://gqlgen.com/reference/dataloaders/)basically try and do what we talked about above, but transparently in the backend. Given a query ```foos { bars { ... }}``` instead of calling `GetFoos()` and then `GetBar(foo1); GetBar(foo2)....` we inject some middleware which replaces that with `foos = GetFoos(); bars = GetBars(for_foos(foos))`

**Pros:**
- Transparent for frontend

**Cons**
- Doesn't solve nested pagination API strangeness
- Non-trivial to implement (this could be an understatement)

### Smarter complexity limits
[...]

PR #8161, opened 2023-04-18
chore: add docs for pagination protobuf message

### LINKED ISSUE #8000: Pagination has no doc-strings describing how it is used
### Issue

The proto message for pagination has no doc-strings at all apart from for this one field where it is pointlessly obvious:

```
// Newest records first, older records last, default is true
optional bool newest_first = 5;
```

There should be some doc-strings with some details. 

The pagination struct is used everywhere and so will bloat every API doc if there is too much, therefore it should be kept light; we will have a dedicated docs page talking about how to use pagination.

- [ ] Update the API docs as described above
- [ ] Discuss with @candida getting a new docs page created on how to use pagination (with examples?)

pscott31:
closes #8000

### DIFF
--- a/protos/sources/data-node/api/v2/trading_data.proto
+++ b/protos/sources/data-node/api/v2/trading_data.proto
@@ -749,11 +749,39 @@ service TradingDataService {
 
 // -- Pagination --
 
-// Cursor information for cursor based pagination requests
+// All data returned from the API is ordered in a well-defined manner.
+// The specific columns and the order in which the sorting is performed
+// depend on the API endpoint being called. However, the primary sorting
+// column is usually the timestamp of the block in which the data was last updated.
+//
+// To prevent excessively large response messages and to avoid overloading
+// database resources, the API employs a cursor-based pagination mechanism.
+//
+// This Pagination message can be optionally provided as part of the request to specify:
+// - The starting point within the total result set for beginning the page
+// - The size of the returned page
+// - The ordering of the data within that page
+//
+// If no Pagination message is provided, the API will return the first page of
+// data using the default page size. The default page size is 1000.
+//
+// To retrieve subsequent pages, the caller must examine the PageInfo structure
+// returned in the response to find a cursor string that uniquely identifies the
+// last row of that page. This cursor should then be passed in the 'after' field
+// of the Pagination message in a subsequent request.
+//
+// For paging backward, take the cursor for the first row of the page from PageInfo
+// and pass it in the 'before' field of the Pagination message.
+//
+// A Pagination message that uses both first/after and last/before is considered invalid.
 message Pagination {
+  // The number of records to be returned that sort greater than row identified by cursor supplied in 'after'.
   optional int32 first = 1;
+  // If paging forwards, the cursor string for the last row of the previous page.
   optional string after = 2;
+  // The number of records to be returned that sort less than row identified by cursor supplied in 'before'.
   optional int32 last = 3;
+  // If paging forwards, the cursor string for the first row of the previous page.
   optional string before = 4;
   // Newest records first, older records last, default is true.
   optional bool newest_first = 5;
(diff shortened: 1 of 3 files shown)


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -924,11 +924,40 @@ service TradingDataService {
...
+// To retrieve subsequent pages the caller needs to examine PageInfo structure returned in the
+// response to find a cursor string which uniquely identifies the last row of that page, and
+// then pass that in the 'after' field of the Pagination message in a subsequent request.
+//
+// To paging backwards, take the cursor for the first row of the page from PageInfo and pass that
+// in the 'before' field of the Pagination message.
+//
+// A pagination message that uses first/after as well as last/before is invalid.
 message Pagination {
+    // The number of records to be returned that sort greater than row identified by cursor supplied in 'after'
```
jeremyletang, line 954: This line needs to be terminated with a full stop to generate the docs appropriately.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -924,11 +924,40 @@ service TradingDataService {
...
     optional int32 first = 1;
+    // If paging forwards, the cursor string for the last row of the previous page.
     optional string after = 2;
+    // The number of records to be returned that sort less than row identified by cursor supplied in 'before'
```
jeremyletang, line 958: This line needs to be terminated with a full stop to generate docs appropriately.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -924,11 +924,40 @@ service TradingDataService {
 
 // -- Pagination --
 
-// Cursor information for cursor based pagination requests
+// Pagination information for cursor based pagination.
+//
+// All data returned from the API is ordered in some well defined manner. The details of which
+// columns and in which order the sorting is performed depend on which API is being endpoint is 
+// being called, but usually the timestamp of the block in which it data was last updated is the
+// primary sort column.
```
jeremyletang, line 932: `depend on which API is being endpoint is being called`: not sure this is making sense.
`the block in which it data`: this too.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -924,11 +924,40 @@ service TradingDataService {
...
+//
+// In order to prevent exessively large response messages, and to prevent large queries from
```
jeremyletang, line 934: maybe a typo on exessively?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -924,11 +924,40 @@ service TradingDataService {
...
+// starving database resources, the API uses a cursor based pagination mechanism.
+//
+// This pagination message is optionally provided as part of the the request to specify
+// - the starting point within a the total result set to begin the page
```
jeremyletang, line 938: `within a the total`?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -924,11 +924,40 @@ service TradingDataService {
...
+// - size of returned page
```
jeremyletang, line 939: the size of the returned page?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -749,11 +749,39 @@ service TradingDataService {
...
+// To prevent excessively large response messages and to avoid overloading
+// database resources, the API employs a cursor-based pagination mechanism.
+//
+// This Pagination message can be optionally provided as part of the request to specify:
+// - The starting point within the total result set for beginning the page
+// - The size of the returned page
+// - The ordering of the data within that page
+//
+// If no Pagination message is provided, the API will return the first page of
+// data using the default page size. The maximum page size is 1000.
```
wwestgarth, line 766: Is the maximum page size really 1000? Or was that supposed to say default page size is 1000?

The below gets me 2000 things in `["orders"]["edges"]`:
```
https://api.n00.devnet1.vega.xyz/api/v2/orders?pagination.first=2000
```

pscott31: IIRC we decided that the default should be 100 but the maximum 1000 - though you're right I don't see any code that enforces that!

pscott31: I see here we discussed about the page size, for graphql at least:

https://github.com/vegaprotocol/vega/issues/7291

I'll update the docs to reflect what actually happens at the moment and if/when we decide that we want a limit we can fix the docs up as well

wwestgarth (approved):

gordsport (approved):

PR #8163, opened 2023-04-18
fix: 7988 and 7997 clean up documentation about numbers expressed as …

### LINKED ISSUE #7988: Fields do not need a title in the API documentation
### Issue

Using titles for fields makes the formatting look odd, it also begs the question if a title is required its misplaced being in the doc-string against a field.

### Examples

- https://docs.vega.xyz/testnet/api/grpc/vega/vega.proto#delegation `Amount delegated`
- https://docs.vega.xyz/testnet/api/grpc/vega/vega.proto#deposit `The amount to be deposited`
- https://docs.vega.xyz/testnet/api/grpc/vega/vega.proto#fee

### Tasks

During investigations it could be that the 'title' is indeed the description and the rest has been added due to careless copy-paste. Each instance will need to be reviewed:

- [ ] Review the above examples and resolve
- [ ] Look for other instances throughout the docs and resolve these too ;-)

### LINKED ISSUE #7997: Clean up API comments when a returned value is signed/unsigned
### Issue

There are many entries that state that a value is signed or unsigned, from an initial triage it seems that many of these have been copy-pasted and in some cases seems to conflict if its signed or unsigned.

**For example:**
“Open volume for the position, value is **signed** +ve for long and -ve for short. This field is an **unsigned** integer passed as a string and needs to be scaled using the market's position decimal places.”

It seems highly likely the last sentence, in this example, was copy-pasted so there may be many more occurrences.

It also in many cases states that it needs to be scaled, should this really be in the documentation? Surely its up to the API user to decide what to do with the responce, the docs should be concerns with what the API/response is.

Task:

- [ ] Find all occurrences of descriptions of the value being value is signed/unsigned
- [ ] Ensure that the description is accurate 
- [ ] Correct where required

guoguojin:
…strings

Closes #7988 
Closes #7997

### DIFF
--- a/protos/sources/vega/chain_events.proto
+++ b/protos/sources/vega/chain_events.proto
@@ -10,8 +10,7 @@ message BuiltinAssetDeposit {
   string vega_asset_id = 1;
   // Vega party ID i.e. public key.
   string party_id = 2;
-  // Amount to be deposited. This field is an unsigned integer passed as a string
-  // and needs to be scaled using the asset's decimal places.
+  // Amount to be deposited. This field is an unsigned integer scaled to the asset's decimal places.
   string amount = 3;
 }
 
@@ -21,8 +20,7 @@ message BuiltinAssetWithdrawal {
   string vega_asset_id = 1;
   // Vega network party ID i.e. public key.
   string party_id = 2;
-  // The amount to be withdrawn. This field is an unsigned integer passed as a string
-  // and needs to be scaled using the asset's decimal places.
+  // The amount to be withdrawn. This field is an unsigned integer scaled to the asset's decimal places.
   string amount = 3;
 }
 
@@ -180,8 +178,7 @@ message StakeDeposited {
   string ethereum_address = 1;
   // Hex encoded public key of the party receiving the stake deposit.
   string vega_public_key = 2;
-  // Amount deposited as an unsigned base 10 integer. This field is passed as a string
-  // and needs to be scaled using the asset's decimal places.
+  // Amount deposited as an unsigned base 10 integer scaled to the asset's decimal places.
   string amount = 3;
   // Time at which the block was produced.
   // Will be used to inform the core at what time
@@ -194,8 +191,7 @@ message StakeRemoved {
   string ethereum_address = 1;
   // Hex encoded public key of the party from which to remove stake.
   string vega_public_key = 2;
-  // Amount removed as a base 10 unsigned integer.This field is passed as a string
-  // and needs to be scaled using the asset decimal places for the staking token.
+  // Amount removed as a base 10 unsigned integer scaled to the asset's decimal places.
   string amount = 3;
   // The time at which the block was produced
   // will be used to inform the core at what time
@@ -206,7 +202,6 @@ message StakeRemoved {
 message StakeTotalSupply {
   // Address of the staking asset
   string token_address = 1;
-  // Total supply observed for the token as an unsigned based 10 integer. This field is passed as a string
-  // and needs to be scaled using the asset decimal places for the staking token.
+  // Total supply observed for the token as an unsigned based 10 integer scaled to the asset's decimal places.
   string total_supply = 2;
 }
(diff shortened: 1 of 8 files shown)


protos/sources/vega/commands/v1/commands.proto
```diff
@@ -90,8 +90,8 @@ message OrderAmendment {
   optional int64 expires_at = 5;
   // Amend the time in force for the order, set to TIME_IN_FORCE_UNSPECIFIED to remain unchanged.
   vega.Order.TimeInForce time_in_force = 6;
-  // Amend the pegged order offset for the order
-  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  // Amend the pegged order offset for the order. This field is an unsigned integer passed as a string
+  // and needs to be scaled using the market's decimal places.
```
wwestgarth, line 94: Instead of saying `and needs to be scaled` can we not just say `in the market's precision`. So something like:
```
  // Amend the pegged order offset for the order. Must represent an unsigned integer in the market's precision.
```

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -31,10 +31,10 @@ message OrderSubmission {
   string market_id = 1;
   // Price for the order, the price is an integer, for example `123456` is a correctly
   // formatted price of `1.23456` assuming market configured to 5 decimal places,
-  // required field for limit orders, however it is not required for market orders
+  // required field for limit orders, however it is not required for market orders.
   // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
   string price = 2;
-  // Size for the order, for example, in a futures market the size equals the number of units, cannot be negative
+  // Size for the order, for example, in a futures market the size equals the number of units, cannot be negative.
```
wwestgarth, line 37: I don't think the size of an order has anything to do with the market's decimal places?

protos/sources/vega/chain_events.proto
```diff
@@ -20,8 +20,8 @@ message BuiltinAssetWithdrawal {
   string vega_asset_id = 1;
   // Vega network party ID i.e. public key.
   string party_id = 2;
-  // The amount to be withdrawn
-  // This field is an unsigned integer passed as a string and needs to be scaled using the asset's decimal places.
+  // The amount to be withdrawn. This field is an unsigned integer passed as a string
```
wwestgarth, line 23: Maybe @gordsport has an opinion, but for all of these can we go with a phrasing more like
"Must represent an unsigned integer scaled to the asset's decimal places." ? For me it flows a bit better.

guoguojin: I made the suggested edits and marked the other conversations resolved, the edits can be found in this commit https://github.com/vegaprotocol/vega/pull/8163/commits/6839a413141d2f91a205dedf35f6346084a65e31

protos/sources/vega/vega.proto
```diff
@@ -457,20 +449,15 @@ message Candle {
...
-  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  // Highest price for trading during the candle interval. This field is an unsigned integer scaled to the market's decimal places.
   string high = 3;
-  // Lowest price for trading during the candle interval
-  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  // Lowest price for trading during the candle interval. This field is an unsigned integer scaled to the market's decimal places.
   string low = 4;
-  // Open trade price
-  // This field is an unsigned integer passed as a string and needs to be scaled using the market's decimal places.
+  // Open trade price.This field is an unsigned integer scaled to the market's decimal places.
```
gordsport, line 456:
```suggestion
  // Open trade price. This field is an unsigned integer scaled to the market's decimal places.
```

gordsport (dismissed):

jeremyletang (dismissed):

jeremyletang (dismissed):

gordsport (approved):

jeremyletang (approved):

PR #8167, opened 2023-04-18
gRPC subscription rate limiter was limiting unary calls, not subscriptions

### LINKED ISSUE #8103: Error when querying for parties using graphql
**UPDATE:**
Looks like the below issue is sort of a feature rather than a bug. The below happens due to `complexity limiting` rather than rate limiting.

In view of the above, could there be an error message that better signposts what's happened, and what to do about it? 

**Issue Reported:**

Receive `ResourceExhausted` error when querying parties with nested queries. 

**Query:**

```
query Parties {
  partiesConnection {
    edges {
      node {
        id
        delegationsConnection {
          edges {
            node {
              party {
                id
              }
              node {
                id
              }
              epoch
            }
          }
        }
      }
    }
  }
}

```
**Error:**

```
{
  "errors": [
    {
      "message": "",
      "path": [
        "partiesConnection",
        "edges",
        52,
        "node",
        "delegationsConnection",
        "edges",
        342,
        "node",
        "node"
      ],
      "extensions": {
        "code": 8,
        "type": "ResourceExhausted"
      }
    },
And the above section repeats
```

pscott31:
Closes #8103 

A while ago @Tan committed [this pr](https://github.com/vegaprotocol/vega/issues/7923) to add a limit to the number of gRPC subscriptions a client could hold at once; however it was not quite right and was actually limiting the number of simultaneous unary calls an IP address could make.

It's quite easy for one graphql query to make many simultaneous gRPC calls.

### DIFF
--- a/datanode/gateway/middleware.go
+++ b/datanode/gateway/middleware.go
@@ -22,10 +22,9 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-
-	"google.golang.org/grpc"
 
 	"code.vegaprotocol.io/vega/datanode/contextutil"
 	"code.vegaprotocol.io/vega/datanode/metrics"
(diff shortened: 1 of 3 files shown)


guoguojin (dismissed):

datanode/gateway/middleware.go
```diff
@@ -193,34 +192,31 @@ func (s *SubscriptionRateLimiter) WithSubscriptionRateLimiter(next http.Handler)
...
-		_ *grpc.UnaryServerInfo,
-		handler grpc.UnaryHandler,
-	) (resp interface{}, err error) {
-		// Fish out IP address from context
-		addr, ok := contextutil.RemoteIPAddrFromContext(ctx)
-		if !ok {
+type ipGetter func(ctx context.Context, method string, log *logging.Logger) string
+
+func (s *SubscriptionRateLimiter) WithGrpcInterceptor(ipGetterFunc ipGetter) grpc.StreamServerInterceptor {
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
```
EVODelavega, line 198: non-blocking: `any` looks cleaner (to my eye at least) than `interface{}`

EVODelavega (approved):

datanode/gateway/middleware.go
pscott31: yes fair enough, that was copy-pasted from the definition inside grpc :)

PR #8174, opened 2023-04-19
Aggressor on trades when leaving auction should be unspecified

### LINKED ISSUE #8173: Core labels traders as the aggressors during auction
# Problem encountered
Core currently shows who the aggressor was on a trade that happened in auction uncrossing, see the [discussion](https://vegaprotocol.slack.com/archives/CFXA4EGN7/p1681289524255619)

# Expected behaviour
Auction matches should be marked with Aggressor: ~none~ unspecified,  so that Console can show something like Role: n/a on an auction fill

EVODelavega:
Fixes #8173

### DIFF
--- a/core/matching/orderbook_test.go
+++ b/core/matching/orderbook_test.go
@@ -3326,6 +3326,10 @@ func TestOrderBook_AuctionUncrossWashTrades(t *testing.T) {
 
 	// Get indicative trades
 	trades, err := book.ob.GetIndicativeTrades()
+	// because the trades are generated when calling LeaveAuction, the aggressor will be unspecified.
+	for _, tr := range trades {
+		tr.Aggressor = types.SideUnspecified
+	}
 	assert.NoError(t, err)
 	assert.Equal(t, len(trades), 1)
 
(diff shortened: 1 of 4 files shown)


core/matching/orderbook_test.go
```diff
@@ -3326,6 +3326,10 @@ func TestOrderBook_AuctionUncrossWashTrades(t *testing.T) {
 
 	// Get indicative trades
 	trades, err := book.ob.GetIndicativeTrades()
+	// because the trades are generated when calling LeaveAuction, the aggressor will be unspecified.
```
wwestgarth, line 3329: Why are we doing this and not just changing the assert to be:
```
assert.Equal(t, uncrossedOrders[0].Trades[0].Aggressor, vega.Side_SIDE_UNSPECIFIED)
```

EVODelavega: Because copying the loop from the `orderbook.go` file was quicker :smile:

jeremyletang (dismissed):

wwestgarth (approved):

mattrussell36 (approved):

PR #8180, opened 2023-04-20
fix: 8021 add better field descriptions where appropriate

### LINKED ISSUE #8021: Field descriptions that just state the field name or are empty
### Issue

There are many instance where the documentation the field name is just copied into the comment that gets displayed in the docs site.. This makes the documentation look incomplete.

An example can be seen here:
- https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-observe-orders

### Tasks

- [ ] Identify the APIs that we think will be most frequently used (and target these first)
- [ ] Update the information so its useful to the end user e.g., “Market identifier” -> “The market identifier of orders to subscribe to.”

guoguojin:
Closes #8021


protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -697,13 +697,13 @@ message PartyOrders {
 
 message LiquidityPartiesOrders {
   repeated PartyOrders party_orders = 1;
-  // Required for key generation
+  // ID of the market these orders apply to. Required for key generation
```
gordsport, line 700: Should we say something like "Market ID of the orders....."

this matches the format of other comments where it starts with XXXX ID...

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -697,13 +697,13 @@ message PartyOrders {
...
   string market_id = 2;
 }
 
 message LiquidityProvisions {
   repeated vega.LiquidityProvision liquidity_provisions = 1;
-  // Required for key generation
+  // ID of the market these provisions apply to. Required for key generation
```
gordsport, line 706: Should we say something like "Market ID of the orders....."

this matches the format of other comments where it starts with XXXX ID...

gordsport (dismissed):

jeremyletang (approved):

ValentinTrinque (approved):

PR #8182, opened 2023-04-20
Connection tokens survive reboot 

### LINKED ISSUE #7906: Connection tokens should survive reboot

ValentinTrinque:
Close #7906 

## Test protocol

1. Start the wallet service
2. Connect to a wallet using "hostname1", keep the token
3. Connect to the same wallet using "hostname2", keep the token
4. Restart the wallet service
5. Try to list keys using the first token
    - You should be able to restore the connection, and get the keys back (with permission list aproval)
7. Try to list keys using the second token
    - You should **not** be ask to unlock the wallet, and get the keys (with permission list aproval)

## Added CLI

```
vega wallet session list
```

## Output format

Screenshot 2023-05-03 at 10 43 08


wwestgarth (approved):
Had a little play, and its good!

PR #8183, opened 2023-04-20
chore: clean up documentation for list ledger entries

### LINKED ISSUE #7998: Verbose documentation in List Ledger Entries
### Issue

The List Ledger Entries API docs description needs to be reviewed and made more concise.

It mentions what fields can be filtered three times. Once in the initial description, then again in a bullet point list, then again in each field’s description. 

- https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-list-ledger-entries

### Task

- [ ] Review the description and make it more concise / useful

guoguojin:
Closes #7998


AnExsomnis (dismissed):

gordsport (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -113,7 +113,7 @@ service TradingDataService {
 
   // List ledger entries
   //
-  // Get a list of ledger entries by asset, market, party, account type and transfer type within the given date range.
+  // Get a list of ledger entries within the given date range.
   // This query requests and sums the number of ledger entries from a given subset of accounts, specified via the 'filter' argument.
   // It returns a time series - implemented as a list of AggregateLedgerEntry structs - with a row for every time
   // the summed ledger entries of the set of specified accounts changes.
```
wwestgarth, line 119: Can we shorten this:
"
Listed entries should be limited to a single party from each side only. If zero or more than one party is provided for each of the sides - sending and receiving accounts, the query returns an error.
"
to just:
"
Each account filter must contain no more than one party ID.

wwestgarth:
In the bit where we list the ways we can filter we change `Entries can be queried by:` -> `Entries can be filtered by:` then we can bin half of the repeated words in each bullet point and it becomes more managable to parse:
"
Entries can be filtered by:
   - the sending account (market ID, asset ID, account type)
   - the receiving account (market ID, asset ID, account type)
   - the sending AND receiving account
   - the transfer type either in addition to the above filters or as a standalone option
"

jeremyletang (dismissed):

jeremyletang (approved):

PR #8192, opened 2023-04-21
Remove TTL, use PoW instead

EVODelavega:
Reverts changes for #7969 to use PoW instead. Includes a minor change to validator commands in check & deliver Tx (skips the block height check)


core/blockchain/abci/tm_node.go
```diff
@@ -143,8 +143,6 @@ func overwriteConfig(config *config.Config) {
 	config.Consensus.CreateEmptyBlocks = true
 	// enforce using priority mempool
 	config.Mempool.Version = "v1"
-	// ensure rechecking tx is enabled
-	config.Mempool.Recheck = true
```
wwestgarth, line 147: Do we not still need this so that the checks in the PoW engine happen after every block? @ze97286

jeremyletang: yea we do.

davidsiska-vega (dismissed):
Looked at `defaults.go` and `keys.go`.

davidsiska-vega (dismissed):
Just default and keys.

core/blockchain/abci/tm_node.go
EVODelavega: Yes, got removed in the revert, and I've added it back in the last commit I pushed :)

davidsiska-vega (dismissed):

core/pow/engine.go
```diff
@@ -380,6 +380,17 @@ func (e *Engine) verify(tx abci.Tx) (byte, error) {
...
 	txHash := hex.EncodeToString(tx.Hash())
 
+	// check for replay attacks
+	if _, ok := e.seenTx[txHash]; ok {
+		e.log.Error("replay attack: txHash already used", logging.String("tx-hash", txHash), logging.String("tid", tx.GetPoWTID()), logging.String("party", tx.Party()), logging.String("command", tx.Command().String()))
+		return h, errors.New("transaction hash already used")
+	}
+
+	// validator commands skip PoW verification
+	if tx.Command().IsValidatorCommand() {
```
ze97286, line 390: just noting here that it is a change of behaviour which means validators block is not check. I don't feel strongly against it but I don't think it's necessary either.

jeremyletang (dismissed):

jeremyletang (approved):

davidsiska-vega (approved):

core/pow/engine.go
EVODelavega: Agreed, after looking at the [spec update PR](https://github.com/vegaprotocol/specs/pull/1673), I checked with @davidsiska-vega to make sure that everyone is aware of this change, and it's the intended behaviour.

ze97286 (approved):

gordsport (approved):

PR #8196, opened 2023-04-21
fix: handle time termination with "lessThan" condition

### LINKED ISSUE #8184: Market proposals are  not enacting or failing when oracle termination condition is set to LessThan (but time is > enactment)
When an oracle is used with a  term condition “Less Than” and a timestamp set to be after the enactment time the market state never moves to enacted/terminated/failed etc it just freezes at Passed.

Expecting the market to terminate immediately voting passes.

AnExsomnis:
close: #8184

### DIFF
--- a/core/governance/market.go
+++ b/core/governance/market.go
@@ -253,26 +253,25 @@ func validateFuture(future *types.FutureProduct, decimals uint64, assets Assets,
 		return types.ProposalErrorInvalidFutureProduct, ErrMissingDataSourceSpecForTradingTermination
 	}
 
-	if !et.shouldNotVerify {
-		filters := future.DataSourceSpecForTradingTermination.GetFilters()
+	filters := future.DataSourceSpecForTradingTermination.GetFilters()
+	for i, f := range filters {
+		if f.Key.Type == datapb.PropertyKey_TYPE_TIMESTAMP {
+			for j, cond := range f.Conditions {
+				v, err := strconv.ParseInt(cond.Value, 10, 64)
+				if err != nil {
+					return types.ProposalErrorInvalidFutureProduct, err
+				}
 
-		for i, f := range filters {
-			if f.Key.Type == datapb.PropertyKey_TYPE_TIMESTAMP {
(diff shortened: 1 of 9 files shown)


core/governance/market.go
```diff
@@ -264,6 +267,10 @@ func validateFuture(future *types.FutureProduct, decimals uint64, assets Assets,
 						return types.ProposalErrorInvalidFutureProduct, err
 					}
 
+					if cond.Operator == datapb.Condition_OPERATOR_LESS_THAN {
```
ze97286, line 270: shouldn't this also be LTE?

AnExsomnis: Yes, I think it should be

core/governance/market.go
```diff
@@ -493,8 +500,17 @@ func validateUpdateFuture(future *types.UpdateFutureProduct, et *enactmentTime)
...
```
ze97286, line 503: LTE too.

core/governance/market.go
AnExsomnis: This has been corrected, thanks!

core/governance/engine.go
```diff
@@ -998,7 +997,11 @@ func (e *Engine) updatedMarketFromProposal(p *proposal) (*types.Market, types.Pr
 		return nil, types.ProposalErrorUnsupportedProduct, ErrUnsupportedProduct
 	}
 
-	if perr, err := validateUpdateMarketChange(terms, &enactmentTime{current: p.Terms.EnactmentTimestamp}); err != nil {
+	if perr, err := validateUpdateMarketChange(
+		terms,
+		&enactmentTime{current: p.Terms.EnactmentTimestamp},
+		e.timeService.GetTimeNow().Unix(),
```
ze97286, line 1003: shouldn't this be `UnixNano()`?

EVODelavega (dismissed):

EVODelavega (dismissed):

ze97286 (approved):

EVODelavega (approved):

PR #8197, opened 2023-04-21
feat: don't include query string as part of Prometheus metric labels

### LINKED ISSUE #8194: Sanitize even more Prometheus requestType label by not using query args
A first attempt has been already done to only allow known queries and avoid getting the spammer/bruteforce random url attempts here: https://github.com/vegaprotocol/vega/issues/7469

This already greatly decreased the number of timeseries generated (and thus paid for) in prometheus. Unfortunately these metrics are still generating a lot (40k so far this month for example, only from our mainnet observers servers, costing us $500 just for these 2 metrics)  because it use the query arguments from each call (see screenshots bellow) .

image
image

Most affected metrics are:
- datanode_request_count_total
- datanode_request_time_total
- vega_request_count_total
- vega_request_time_total

pscott31:
Closes #8194

### DIFF
--- a/datanode/metrics/prometheus.go
+++ b/datanode/metrics/prometheus.go
@@ -16,6 +16,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


datanode/metrics/prometheus.go
```diff
@@ -898,8 +899,9 @@ func APIRequestAndTimeREST(method, request string, time float64) {
 	}
 
 	const (
-		invalid = "invalid route"
-		prefix  = "/api/v2/"
+		invalid     = "invalid route"
+		invalid_url = "invalid url"
```
EVODelavega, line 903: This conflicts with the CODING_STYLE document. This should be `invalidURL`

EVODelavega (dismissed):

PR #8215, opened 2023-04-25
Fix verbose documentation in list ledger entries

### LINKED ISSUE #7998: Verbose documentation in List Ledger Entries

zale144:
Closes #7998


candida-d:
@zale144 This has technically been done in a previous ticket, and some of the additions are actually repetitive because we have fields that are automatically labeled (optional, for example). We'll revisit some of these descriptions again in the future though. 

@gordsport Do you think we can close this PR?

gordsport:
Closing this PR as per the above comment.

We will have a re-review of the docs later and these can be cleaned up based on the new style guide once agreed.

PR #8224, opened 2023-04-26
fix: estimate position api cleanup

### LINKED ISSUE #8222: EstimatePosition needs the api-cleanup treatment
# Problem encountered

The new `EstimatePosition` API slipped through the api-cleanup work an inherited some some the previous bad practices. We need to give it the api-cleanup treatment.

Things to address:
- I can pass in a negative price, or a price with decimals, even though the docs say "This field is an unsigned integer"
- I can pass in an `OrderInfo` with an side `UNSPECIFIED` and it returns some numbers? This should either be a validation error, or if it has some meaning in this context, some docs explaining how it effects the calculations
- Setting a price to a nonsense string returns an error without a grpc error code
- Passing in an empty marketID is treated as `NotFound` and not `ErrEmptyMissingMarketID`
- The doc strings in `EstimatePositionRequest` have all the bad things we have worked to remove (use of parentheses, field descriptions that just state the field name etc etc)

# Observed behaviour

# Expected behaviour

# System response

# Steps to reproduce

## Manual

## Automation

# Evidence

## Logs

## Additional context

# Definition of Done

**Before Merging**

**After Merging**

guoguojin:
Closes #8222

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -2692,6 +2692,10 @@ func implyMarginLevels(maintenanceMargin num.Decimal, scalingFactors *vega.Scali
 func (t *TradingDataServiceV2) EstimatePosition(ctx context.Context, req *v2.EstimatePositionRequest) (*v2.EstimatePositionResponse, error) {
 	defer metrics.StartAPIRequestAndTimeGRPC("EstimatePosition")()
 
+	if req.MarketId == "" {
+		return nil, ErrEmptyMissingMarketID
+	}
+
 	var collateralAvailable num.Decimal
 	if req.CollateralAvailable != nil && len(*req.CollateralAvailable) > 0 {
 		var err error
@@ -2723,15 +2727,22 @@ func (t *TradingDataServiceV2) EstimatePosition(ctx context.Context, req *v2.Est
 		var price num.Decimal
 		p, err := num.DecimalFromString(o.Price)
 		if err != nil {
-			return nil, errors.Wrap(ErrInvalidOrderPrice, err.Error())
+			return nil, formatE(ErrInvalidOrderPrice, err)
 		}
+
+		if p.IsNegative() {
+			return nil, ErrInvalidOrderPrice
+		}
+
 		price = p.Mul(dPriceFactor)
 
-		if o.Side == types.SideBuy {
+		switch o.Side {
+		case types.SideBuy:
 			buyOrders = append(buyOrders, &risk.OrderInfo{Size: o.Remaining, Price: price, IsMarketOrder: o.IsMarketOrder})
-		}
-		if o.Side == types.SideSell {
+		case types.SideSell:
 			sellOrders = append(sellOrders, &risk.OrderInfo{Size: o.Remaining, Price: price, IsMarketOrder: o.IsMarketOrder})
+		default:
+			return nil, ErrInvalidOrderSide
 		}
 	}
 
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -2730,7 +2730,7 @@ func (t *TradingDataServiceV2) EstimatePosition(ctx context.Context, req *v2.Est
 			return nil, formatE(ErrInvalidOrderPrice, err)
 		}
 
-		if p.IsNegative() {
+		if p.IsNegative() || !p.IsInteger() {
 			return nil, ErrInvalidOrderPrice
 		}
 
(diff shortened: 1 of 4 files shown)


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3143,29 +3143,26 @@ message OrderInfo {
   vega.Side side = 1;
   // Price for the order. The price is an integer, for example `123456` is a correctly
```
candida-d, line 3144: If this is unsigned, we should still keep that detail in here. Not sure what to do about the fact that it's a string but that it's returned as a number - how can that be described clearly?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3143,29 +3143,26 @@ message OrderInfo {
...
   uint64 remaining = 3;
   // Boolean that indicates if it is a market order.
   bool is_market_order = 4;
 }
 
-// Request for the estimate of the margin level and, if available collateral is provided, liqudation price for the specified position.
+// Request for the estimate of the margin level and, if available collateral is provided, liquidation price for the specified position.
 message EstimatePositionRequest {
-  // Market ID.
+  // Market ID in which the position is held.
```
candida-d, line 3155:
```suggestion
  // ID for the market in which the position is held.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3143,29 +3143,26 @@ message OrderInfo {
...
   string market_id = 1 [(google.api.field_behavior) = REQUIRED];
-  // Open volume.
-  // This field is an signed integer (negative for short position) passed as a string and needs to be scaled using the market's position decimal places.
+  // Open volume. This field is a signed integer scaled to the market's position decimal places.
```
candida-d, line 3157:
```suggestion
  // Open volume. This field is a signed integer scaled to the market's position decimal places. A negative number denotes a short position.
```

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: We went through and tried to make everything consistent with XXX ID, and kept ID for/of XXX ... where it sounded more natural. With that in mind, how about `Market ID to estimate the position for`?

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: I think we removed the "expressed as a string" in most places because the type definition on the field dictates that it should be expressed as a string anyway, but we kept the description of how the integer should be expressed to indicate the data in the string should be an integer.

guoguojin: Happy to change it if you think it's not clear, but we would need to do it in a few other places too.

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: Okay, I'll add it back in.

datanode/api/trading_data_v2.go
```diff
@@ -2723,15 +2727,22 @@ func (t *TradingDataServiceV2) EstimatePosition(ctx context.Context, req *v2.Est
 		var price num.Decimal
 		p, err := num.DecimalFromString(o.Price)
 		if err != nil {
-			return nil, errors.Wrap(ErrInvalidOrderPrice, err.Error())
+			return nil, formatE(ErrInvalidOrderPrice, err)
+		}
+
+		if p.IsNegative() {
```
wwestgarth, line 2733: Given the docstring we say its an unsigned integer, I think we also need an `if !p.IsInteger{}` check here too?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3141,31 +3141,29 @@ message PingResponse {}
 message OrderInfo {
   // Side for the order, e.g. buy or sell.
   vega.Side side = 1;
-  // Price for the order. The price is an integer, for example `123456` is a correctly
+  // Price for the order. The price is an unsigned integer, for example `123456` is a correctly
```
candida-d, line 3144:
```suggestion
  // Price for the order. The price is an unsigned integer. For example `123456` is a correctly
```

candida-d (dismissed):

pscott31 (approved):

candida-d (approved):

PR #8230, opened 2023-04-28
feat: cleanup docker volumes at the end and beginning of pipeline

mkjmdski:


daniel1302 (commented):
We need the same logic in the system tests pipeline and probably other pipelines. It would be nice to have a dedicated function in the `jenkins-shared-library` and call the same function in multiple pipelines

daniel1302 (dismissed):

ze97286 (dismissed):

guoguojin (approved):

ze97286 (approved):

PR #8235, opened 2023-04-28
fix: cleanup list ledger documentation

### LINKED ISSUE #8221: Docs: List ledger entries can be clearer
The section below is hard to understand, it's assuming a lot of knowledge and also conflates different details. What is a "ledger entry" to an API user? What information will someone who doesn't know what a ledger entry is need to know in order to use this? 

```
// List ledger entries
  //
  // Get a list of ledger entries within the given date range.
  // Each account filter must contain no more than one party ID.
```

guoguojin:
Closes #8221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### 🛠 Improvements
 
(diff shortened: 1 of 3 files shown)


guoguojin:
API is going to be refactored and the documentation will be updated accordingly.

PR #8243, opened 2023-05-02
Order by id GraphQL docs

### LINKED ISSUE #8236: Fix docs for orderById
# Problem encountered
Since https://github.com/vegaprotocol/vega/issues/8017, you can no longer pass `0` to `orderById` to get the most recent version. However the docs say you still can. Bad docs.

# Observed behaviour
The docs lie

# Expected behaviour
The docs are honest

# Tasks
- [ ] Fix https://github.com/vegaprotocol/vega/blob/develop/datanode/gateway/graphql/schema.graphql#L621
- [ ] Check other docs

zale144:
Closes #8236


candida-d (dismissed):

datanode/gateway/graphql/schema.graphql
```diff
@@ -618,7 +618,7 @@ type Query {
     "ID for an order"
     id: ID!
 
-    "Version of the order (omitted or 0 for most recent; 1 for original; 2 for first amendment, etc)"
+    "Version of the order (omitted for most recent; 1 for original; 2 for first amendment, etc)"
```
candida-d, line 621: Approved but ... question. The protos says the following, and that doesn't seem exactly the same as this description does: 

> Version for the order, initial value is version 1 and is incremented after each successful amend

Is the protos unspecific about "most recent"?

candida-d, line 621:
```suggestion
    "Version of the order. Omitted for most recent; 1 for original; 2 for first amendment, etc."
```

zale144: I would say the difference is that the protos don't mention the fact that omitting the version would give us the most recent, but as for "1 for original", I believe it boils down to the same thing - the initial value, or the first version, before amending it. 
The phrase "2 for first amendment, etc" could be clearer, for sure. However, I believe it essentially boils down to "...and is incremented after each successful amend", meaning every time we update the order, the version will be incremented, thus after version 1 (initial/original) we have version 2, 3, 4...

zale144: I think if I replace it with something like:

"Order version number, starting at 1 for the original order and incrementing by 1 for each successful amendment (omitted for the most recent version)."

candida-d: @zale144 Thank you, that sounds good. Sorry, hadn't realised they weren't the same thing.

datanode/gateway/graphql/schema.graphql
```diff
@@ -617,8 +617,7 @@ type Query {
   orderByID(
     "ID for an order"
     id: ID!
-
-    "Version of the order (omitted or 0 for most recent; 1 for original; 2 for first amendment, etc)"
+    "Order version number, starting at 1 for the original order and incrementing by 1 for each successful amendment (omitted for the most recent version)."
```
candida-d, line 620:
```suggestion
    "Order version number, starting at 1 for the original order and incrementing by 1 for each successful amendment; omitted for the most recent version."
```

Just removing the brackets here, otherwise good.

edd (approved):

wwestgarth (approved):

PR #8245, opened 2023-05-03
fix: clean up listBalanceChanges docs

### LINKED ISSUE #8223: ListBalanceChanges needs the api-cleanup treatment
# Problem encountered

It looks like `ListBalanceChanges` slipped through the net and could use a bit of love.

Things noted:
- We are referencing proto structs that become confusing and meaningless in REST-land `// `ListBalanceChangesRequest` will return a list of `AggregatedBalance` records`
- Most of the docs is a bullet point dscription of what an account is, and I'm if this is the best place for it (maybe it should be against `message AccountFilter` instead?
- General confusing text `// Timestamp to of block the balance is referring`
- `with a row for each block at`: "row" is probably not the correct terminology at this level

# Observed behaviour

# Expected behaviour

# System response

# Steps to reproduce

## Manual

## Automation

# Evidence

## Logs

## Additional context

# Definition of Done

**Before Merging**

**After Merging**

AnExsomnis:
close: #8223

### DIFF
--- a/protos/sources/data-node/api/v2/trading_data.proto
+++ b/protos/sources/data-node/api/v2/trading_data.proto
@@ -145,16 +145,6 @@ service TradingDataService {
   // List balance changes
   //
   // Get a list of the changes in account balances over a period of time.
-  //
-  // An account is defined as a set of asset_id, type, party_id, and market_id.
-  // - Every account has an associated asset and type.
-  // - Certain account types such as the global reward pool for example, do not have an associated party.
-  //   These are denoted by the special party ID 'network'
-  // - Certain account types do not have an associated market such as the general party accounts, for example.
-  //   These are denoted by the special market ID '' (the empty string)
-  //
-  // `ListBalanceChangesRequest` will return a list of `AggregatedBalance` records,
-  // with a row for each block at which a given account's balance changes.
   rpc ListBalanceChanges(ListBalanceChangesRequest) returns (ListBalanceChangesResponse) {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Accounts"};
   }
@@ -836,15 +826,15 @@ message AccountBalance {
   vega.AccountType type = 6;
 }
 
-// Request that is sent when listing accounts
+// Request that is sent when listing accounts.
 message ListAccountsRequest {
-  // AccountFilter contains all filter conditions and values that can be applied to the listing.
+  // Account filter contains all filter conditions and values that can be applied to the listing.
   AccountFilter filter = 1;
   // Optional pagination control.
   optional Pagination pagination = 2;
 }
 
-// Response that is received from listing accounts query
+// Response that is received from listing accounts query.
 message ListAccountsResponse {
   // Page of accounts data and corresponding page information.
   AccountsConnection accounts = 1;
@@ -866,7 +856,7 @@ message AccountEdge {
   string cursor = 2;
 }
 
-// Request that is sent when getting a subscription to a stream of accounts
+// Request that is sent when getting a subscription to a stream of accounts.
 message ObserveAccountsRequest {
   // Market ID to filter accounts by. If empty, no markets will be filtered.
   string market_id = 1;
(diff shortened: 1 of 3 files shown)


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1285,7 +1284,7 @@ enum AccountField {
 // AggregatedBalance data contains the balance of the set of requested accounts combined with corresponding
 // party ID, market ID and account type details
 message AggregatedBalance {
-  // Timestamp to of block the balance is referring to, in nanoseconds.
+  // Timestamp of the block, in which the balance exists (in nanoseconds).
```
wwestgarth, line 1287: We're trying to avoid using `()` where possible. So maybe:
```suggestion
  // Timestamp in nanseconds of the block in which the balance exists.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -146,15 +146,7 @@ service TradingDataService {
...
-  // An account is defined as a set of asset_id, type, party_id, and market_id.
-  // - Every account has an associated asset and type.
-  // - Certain account types such as the global reward pool for example, do not have an associated party.
-  //   These are denoted by the special party ID 'network'
-  // - Certain account types do not have an associated market such as the general party accounts, for example.
-  //   These are denoted by the special market ID '' (the empty string)
-  //
-  // `ListBalanceChangesRequest` will return a list of `AggregatedBalance` records,
-  // with a row for each block at which a given account's balance changes.
+  // `ListBalanceChangesResponse` will return a paged list of `AggregatedBalance` records.
```
wwestgarth, line 149: I'm going to suggest we remove this line because we can't _really_ reference proto structs like this because that isn't necessarily how they look from the REST API docs. 
```suggestion

```

protos/sources/data-node/api/v2/trading_data.proto
AnExsomnis: All right, that is corrected, thanks!

protos/sources/data-node/api/v2/trading_data.proto
AnExsomnis: Yes, I removed the line now, thanks!

wwestgarth (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1260,7 +1250,14 @@ message AggregatedBalanceConnection {
   PageInfo page_info = 2;
 }
 
-// Filter that is used to filter account data by asset ID, party ID, market ID and account type
+// AccountFilter is used to filter account data.
```
candida-d, line 1253:
```suggestion
// Account filter is used to filter account data.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1260,7 +1250,14 @@ message AggregatedBalanceConnection {
...
+//
+// An account is defined as a set of asset ID, type, party ID, and market ID.
+// - Every account has an associated asset and type.
+// - Certain account types such as the global reward pool for example, do not have an associated party.
+//   These are denoted by the special party ID 'network'.
+// - Certain account types do not have an associated market such as the general party accounts, for example.
+//   These are denoted by the special market ID '' (the empty string).
```
candida-d, line 1260:
```suggestion
//   These are denoted by the special market ID '' (an empty string).
```

candida-d (dismissed):
Two minor suggestions but otherwise looks good!

candida-d (approved):

protos/sources/data-node/api/v2/trading_data.proto
AnExsomnis: This has been corrected as well as some other few places, thanks!

protos/sources/data-node/api/v2/trading_data.proto
AnExsomnis: Corrected, thanks!

wwestgarth (approved):

Issue #6951, opened 2022-11-25
Received ORDER_ERROR_INSUFFICIENT_ASSET_BALANCE instead of ORDER_ERROR_MISSING_GENERAL_ACCOUNT


claudiumilea:
# Problem encountered
The system is raising this error ORDER_ERROR_INSUFFICIENT_ASSET_BALANCE instead of ORDER_ERROR_MISSING_GENERAL_ACCOUNT

# Expected behaviour
Trying to test: https://github.com/vegaprotocol/specs/blob/master/protocol/0024-OSTA-order_status.md#0024-OSTA-028
After doing these steps I was expecting to the the missing general account error.
`Create market1 with asset1
Create asset2
Create party with asset2 (deposit=1, asset_to_deposit=asset2)
Create order with party for market1`

# Steps to reproduce
`Create market1 with asset1
Create asset2
Create party with asset2 (deposit=1, asset_to_deposit=asset2)
Create order with party for market1`

## Automation
`@pytest.mark.full
@pytest.mark.orderstatus
def test_missing_general_account(context):
    """
    Order reason of "ORDER_ERROR_MISSING_GENERAL_ACCOUNT"
    user does not have a general account with the asset required for the market (0024-OSTA-028)
    """
    traders_wo_general_account = marketsTrading.traders(context, count=1, asset_to_deposit='dummy_asset', deposit_amount=1, trader_type='trader_wo_general_account')
    trader_wo_general_account = traders_wo_general_account[0]

    tr1 = context['order_status_traders'][0]

    market_id = context['order_status_market']
    ordersMod.reset_market(context, market_id, traders=tr1, state_reqd='TRADING_MODE_CONTINUOUS')

    wallet_response = ordersMod.limit_order_submit(context=context, market=market_id,
                                                   side='Buy', price=1000, trader=trader_wo_general_account,
                                                   tforce='GTC', size=10, ord_ref='not_enough_asset_balance',
                                                   confirm_order=True)
    assert 'not_enough_asset_balance' in wallet_response
    ordersMod.find_and_verify_order_event(context,
                                          market=market_id,
                                          trader=tr1,
                                          size=10,
                                          price=1000,
                                          status="Rejected",
                                          reason='ORDER_ERROR_INSUFFICIENT_ASSET_BALANCE',
                                          ordRef='not_enough_asset_balance',
                                          tforce='GTC',
                                          ordtype="Limit")`

# Definition of Done
[...]

PR #8276, opened 2023-05-09
fix: set correct error reason for non-existing gen account

AnExsomnis:


core/execution/market.go
```diff
@@ -1267,7 +1267,7 @@ func (m *Market) validateAccounts(ctx context.Context, order *types.Order) error
 	if !m.collateral.HasGeneralAccount(order.Party, asset) {
 		// adding order to the buffer first
 		order.Status = types.OrderStatusRejected
-		order.Reason = types.OrderErrorInsufficientAssetBalance
+		order.Reason = types.OrderErrorMissingGeneralAccount
```
wwestgarth, line 1270: It looks like the spec was updated such that we don't need to do anything for this ticket as `0024-OSTA-028` was removed ?
https://github.com/vegaprotocol/vega/issues/6951#issuecomment-1364033964
https://github.com/vegaprotocol/specs/pull/1487

The comment against `ORDER_ERROR_INSUFFICIENT_ASSET_BALANCE` and the AC `0024-OSTA-027` mentions "having no general account"
```
// Party has an insufficient balance, or does not have
// a general account to submit the order (no deposits made
// for the required asset)
ORDER_ERROR_INSUFFICIENT_ASSET_BALANCE = 43;
```

EVODelavega: Hmm, that makes sense. An error saying _"the general account is missing"_ doesn't really make sense given that a party can have N general accounts (1 per asset). Functionally, there's no difference between a non-existent general balance for a given asset, and a general account with balance 0.

AnExsomnis: @wwestgarth  thanks, i didnt see the commen, closed the issue now

AnExsomnis: Not sure the ORDER_ERROR_INSUFFICIENT_ASSET_BALANCE is the best way of saying that no general account exists, just after checking if there was:
``` if !m.collateral.HasGeneralAccount(order.Party, asset) ... ```
Am I interpreting this wrong? Please let me know

EVODelavega: I don't think a different error message is going to help things long term. True, it makes sense in the conventional account/balance modal, but from a UX perspective, at least for the foreseeable future, parties will be transferring tokens from their crypto wallets. In that context, being told that _"you don't have a wallet for token X"_ means something very different to _"wallet has balance 0 for token X"_.

To address such an issue, the only thing a party has to do is transfer some balance, whether the account existed or its balance was zero makes no difference. To make a distinction between having no balance, or never having had any balance for a certain asset seems rather arbitrary, and IMO can only really cause confusion

PR #8300, opened 2023-05-15
feat: Implent successor markets - happy path

### LINKED ISSUE #8260: Successor market execution
# Feature Overview

In order for successful successor market proposals to be executed we need to update the execution engine to support this

## Specs
- [Spec Issue and linked PR](https://github.com/vegaprotocol/specs/issues/1440)

## Tasks
- [ ] Succeed market at governance enactment/market termination - starting with base balance / ELS from the parent market (preloaded from parent)
- [ ] Account for the Validation in https://github.com/vegaprotocol/vega/issues/8259

## Open questions:
- When do we need to get the data - unless its perps do we only read the oracle data when we close the market?
- If we have a cron job or similar - do we do this in "data sourcing" or in "filtering"?
- Is there any spam effect

EVODelavega:
This is the crude implementation (happy path only ATM) adding successor markets

Closes  #8260

### DIFF
--- a/core/execution/engine.go
+++ b/core/execution/engine.go
@@ -42,6 +42,9 @@ var (
 
 	// ErrInvalidOrderCancellation is returned when an incomplete order cancellation request is used.
 	ErrInvalidOrderCancellation = errors.New("invalid order cancellation")
+
+	// ErrSuccessorMarketDoesNotExists is returned when SucceedMarket call is made with an invalid successor market ID.
+	ErrSuccessorMarketDoesNotExist = errors.New("successor market does not exist")
 )
 
 // TimeService ...
@@ -67,6 +70,7 @@ type Collateral interface {
 	MarketCollateral
 	AssetExists(string) bool
 	CreateMarketAccounts(context.Context, string, string) (string, string, error)
+	SuccessorInsuranceFraction(ctx context.Context, successor, parent, asset string, fraction num.Decimal) *types.LedgerMovement
 }
 
 type StateVarEngine interface {
@@ -111,6 +115,13 @@ type Engine struct {
 
 	maxPeggedOrders        uint64
 	totalPeggedOrdersCount int64
+
+	marketCPStates map[string]*types.CPMarketState
+	// a map of all successor markets under parent ID
+	// used to manage pending markets once a successor takes over
+	successors      map[string][]string
+	isSuccessor     map[string]string
+	successorWindow time.Duration
 }
 
 type netParamsValues struct {
@@ -180,6 +191,9 @@ func NewEngine(
 		generatedProviders:    map[string]struct{}{},
 		stateVarEngine:        stateVarEngine,
 		marketActivityTracker: marketActivityTracker,
+		marketCPStates:        map[string]*types.CPMarketState{},
+		successors:            map[string][]string{},
+		isSuccessor:           map[string]string{},
 	}
 
 	// set the eligibility for proposer bonus checker
@@ -248,6 +262,19 @@ func (e *Engine) RejectMarket(ctx context.Context, marketID string) error {
 	}
 
 	e.removeMarket(marketID)
+	// a market rejection can have a knock-on effect for proposed markets which were supposed to succeed this market
+	// we should purge them here, and @TODO handle any errors
+	if successors, ok := e.successors[marketID]; ok {
+		delete(e.successors, marketID)
+		for _, sID := range successors {
+			_ = e.RejectMarket(ctx, sID)
+			delete(e.isSuccessor, sID)
+		}
+	}
+	// remove entries in succession maps
+	delete(e.isSuccessor, marketID)
+	// and clear out any state that may exist
+	delete(e.marketCPStates, marketID)
 	return nil
 }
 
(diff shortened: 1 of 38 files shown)


protos/sources/vega/checkpoint/v1/checkpoint.proto
```diff
@@ -224,3 +225,64 @@ message AssetAction {
...
+  // the party if of the LP.
+  string party_id = 1;
+  // their share as tracked by the ELS component (nominal amount).
+  string share = 2;
+  // the amount committed (the total commitment made - does not necessarily match the LP balance or share)
+  string supplied_stake = 3;
+  // the virtual stake.
+  string virtual_stake = 4;
+  // the average valuation
+  string avg = 5;
```
candida-d, line 246:
```suggestion
  // Party ID of the liquidity provider.
  string party_id = 1;
  // Liquidity provider's share as tracked by the ELS component; nominal amount.
  string share = 2;
  // Total amount committed. This does not necessarily match the LP balance or share
  string supplied_stake = 3;
  // Virtual stake.
  string virtual_stake = 4;
  // Average valuation.
  string avg = 5;
````

Comment (candida-d) [reply protos/sources/vega/checkpoint/v1/checkpoint.proto:246]: Can you add a little more info here about average valuation and virtual stake?

File: protos/sources/vega/checkpoint/v1/checkpoint.proto [hunk 2]
```diff
@@ -224,3 +225,64 @@ message AssetAction {
...
+
+  // Current trade value
+  string trade_value = 9;
+
+  // current fee window (used to calculate average trade value).
+  uint64 fee_window = 10;
+
+  // Market value proxy as used by ELS for growth rate calculation
+  string mvp = 11;
+  // TTL until the market state expires and cannot be applied to a successor market.
```
candida-d, line 281: What's TTL?

protos/sources/vega/checkpoint/v1/checkpoint.proto
EVODelavega: :D a little more information... These are internal values part of the equity like share calculations, it's all quite complex to document. For example: virtual stake is dependent on the market growth, whether you've increased or decreased your commitment, and how long you've held that stake, as outlined in the specs here: https://github.com/vegaprotocol/specs/blob/734c728dde0b4374d2dd2e7dc64ce6a8e11eb821/protocol/0042-LIQF-setting_fees_and_rewarding_lps.md?plain=1#L68

> i virtual stake <- max(LP i physical stake, (1 + r) x (LP i virtual stake)).

Where `r` is market growth over a given market window

protos/sources/vega/checkpoint/v1/checkpoint.proto
EVODelavega: Time To Live

core/processor/abci.go
```diff
@@ -1485,7 +1485,8 @@ func (app *App) DeliverPropose(ctx context.Context, tx abci.Tx, deterministicID
 	if toSubmit.IsNewMarket() {
 		nm := toSubmit.NewMarket()
 
-		if err := app.exec.SubmitMarket(ctx, nm.Market(), party); err != nil {
+		// @TODO pass in parent and insurance pool share if required
+		if err := app.exec.SubmitMarket(ctx, nm.Market(), party, toSubmit.ParentMarketID(), toSubmit.InsurancePoolFraction()); err != nil {
```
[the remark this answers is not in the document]
jeremyletang: Also, it might conflict with the call down there which actually is specific to successor

protos/sources/vega/markets.proto
```diff
@@ -283,6 +283,10 @@ message Market {
   string linear_slippage_factor = 13;
   // Quadratic slippage factor is used to cap the slippage component of maintenance margin - it is applied to the square of the slippage volume.
   string quadratic_slippage_factor = 14;
+  // The ID of the market this market succeeds
+  string parent_market_id = 15;
+  // The fraction of the insurance pool of the parent market this market inherits (range 0 through 1)
+  string insurance_pool_fraction = 16;
 }
```
guoguojin, line 290: Shouldn't we annotate these as optional for clarity, as not every market will have these?

guoguojin (dismissed):

core/execution/engine.go
```diff
@@ -262,8 +289,74 @@ func (e *Engine) StartOpeningAuction(ctx context.Context, marketID string) error
 	if !ok {
 		return ErrMarketDoesNotExist
 	}
+	err := mkt.StartOpeningAuction(ctx)
+	if err != nil {
+		return err
+	}
+	// @TODO ensure we keep these succession maps nice and tidy
+	// ATM we're likely deleting too much, or could do things at a better point in thime
```
AnExsomnis, line 297:
```suggestion
	// ATM we're likely deleting too much, or could do things at a better point in time
```

core/execution/engine_checkpoint.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (e *Engine) Load(ctx context.Context, data []byte) error {
+	wrapper := checkpoint.ExecutionState{}
+	if err := proto.Unmarshal(data, &wrapper); err != nil {
+		return err
+	}
+	e.marketCPStates = make(map[string]*types.CPMarketState, len(wrapper.Data))
+	// for now, restore all pending markets as though their state is valid for the full TTL window
+	for _, mcp := range wrapper.Data {
+		e.marketCPStates[mcp.Id] = types.NewMarketStateFromProto(mcp)
```
jeremyletang, line 44: anything to do with the maps of isSuccessor, etc that were added as well?

core/governance/engine.go
```diff
@@ -311,6 +313,10 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteCl
 
 	now := t.Unix()
 
+	// use slice for deterministic behaviour and event order
+	// succeededMarkets := []string{}
```
jeremyletang, line 317: ?

core/execution/engine_checkpoint.go
EVODelavega: Yes, this is needed in case a successor market is proposed during the successor window (grace period), after the market is settled. The settled market will be removed from the normal `e.markets` and `marketsCpy` fields as usual, but we keep settled markets around in case there is a successor market on the way. We also need to be able to create a successor market correctly from a checkpoint restore (which uses proposals, and therefore doesn't contain the necessary ELS data).

TL;DR: yes, it's related, and both are necessary

core/governance/engine.go
EVODelavega: Lots of comments in this PR left as things I want to work on/look in to further down the line. This is one of them

core/governance/market.go
```diff
@@ -430,6 +448,18 @@ func validateNewMarketChange(
 	return types.ProposalErrorUnspecified, nil
 }
 
+func validateParentMarket(prop *types.NewMarket, parent *types.Market) (types.ProposalError, error) {
+	return types.ProposalErrorUnspecified, nil
```
jeremyletang, line 452: what will we do here? ensure that the details of this market matches the parent or something?

AnExsomnis (approved):

core/governance/market.go
EVODelavega: yes: settlement assets have to match etc... spec PR with more details is in review currently

core/execution/engine.go
```diff
@@ -305,9 +399,22 @@ func (e *Engine) submitOrRestoreMarket(ctx context.Context, marketConfig *types.
 		e.log.Debug(msg, logging.Market(*marketConfig))
 	}
 
-	if err := e.submitMarket(ctx, marketConfig); err != nil {
+	parentMarketID := marketConfig.ParentMarketID
+	// propagate netparams unless this is a successor market
+	propagate := len(parentMarketID) == 0
+	if err := e.submitMarket(ctx, marketConfig, propagate); err != nil {
```
wwestgarth, line 405: Why do we not want to propagate the netparams for a successor market? Are we going to take the values from the parent or something?

candida-d (approved):

core/execution/engine.go
EVODelavega: This is something that will be addressed in the next PR. I need to go through the netparams, and see which ones should be propagated, and which may not be ideal to propagate. Initially, in this PR, the successor market was a copy of the parent to which only the changes unique to the successor market were applied, and propagating network parameters would've been unnecessary because the parent market would've already held all the correct network parameters anyway.

davidsiska-vega (approved):

PR #8310, opened 2023-05-16
Deprecate AggregatedLedgerEntry in favour of LedgerEntry type

### LINKED ISSUE #8221: Docs: List ledger entries can be clearer

zale144:
Closes #8221


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -115,16 +115,13 @@ service TradingDataService {
...
-  // the summed ledger entries of the set of specified accounts changes.
-  // Each account filter must contain no more than one party ID.
+  // This query returns all the ledger entries for a given subset of accounts, specified via the 'filter' argument.
   //
   // Entries can be filtered by:
-  //   - the sending account (market ID, asset ID, account type)
-  //   - receiving account (market ID, asset ID, account type)
-  //   - sending AND receiving account
+  //   - the sending account (party ID, market ID, asset ID, account type)
+  //   - receiving account (party ID, market ID, asset ID, account type)
```
candida-d, line 122:
```suggestion
  //   - sending account: party ID, market ID, asset ID, account type
  //   - receiving account: party ID, market ID, asset ID, account type
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1102,6 +1099,22 @@ message PositionUpdates {
...
+message LedgerEntriesEdge {
+  // 'LedgerEntry' data.
+  vega.LedgerEntry node = 1;
+  // Cursor that can be used to fetch further pages.
+  string cursor = 2;
+}
+
+// Page of ledger entries data and corresponding page information.
+message LedgerEntriesConnection {
+  // Page of 'LedgerEntry' data and their corresponding cursors.
```
candida-d, line 1112:
```suggestion
  // Page of ledger entry data and their corresponding cursors.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1102,6 +1099,22 @@ message PositionUpdates {
   repeated vega.Position positions = 1;
 }
 
+// Represents the ledger entry data with corresponding cursor for it
```
candida-d, line 1102:
```suggestion
// Represents the ledger entry data with corresponding cursor for it.
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3991,15 +3998,25 @@ type AggregatedLedgerEntry {
 }
 
 type AggregatedLedgerEntriesEdge {
-  node: AggregatedLedgerEntry!
-  cursor: String!
+node: AggregatedLedgerEntry!
+cursor: String!
```
candida-d, line 4002: is this one also deprecated?

datanode/gateway/graphql/schema.graphql
```diff
@@ -3991,15 +3998,25 @@ type AggregatedLedgerEntry {
 }
 
 type AggregatedLedgerEntriesEdge {
-  node: AggregatedLedgerEntry!
-  cursor: String!
+node: AggregatedLedgerEntry!
```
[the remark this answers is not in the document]
zale144: I hope it will, because in gql types cannot be deprecated

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -115,16 +115,13 @@ service TradingDataService {
...
+  // This query returns all the ledger entries for a given subset of accounts, specified via the 'filter' argument.
   //
   // Entries can be filtered by:
-  //   - the sending account (market ID, asset ID, account type)
-  //   - receiving account (market ID, asset ID, account type)
-  //   - sending AND receiving account
+  //   - sending account: party ID, market ID, asset ID, account type
+  //   - receiving account: party ID, market ID, asset ID, account type
   //   - transfer type either in addition to the above filters or as a standalone option
+  // Party ID is a required field, all other fields are optional.
```
wwestgarth, line 124: There is no field called Party ID:
```suggestion

It is required to filter by at least one party ID on either side of the transfer.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -115,16 +115,13 @@ service TradingDataService {
   // List ledger entries
   //
   // Get a list of ledger entries within the given date range.
-  // This query requests and sums the number of ledger entries from a given subset of accounts, specified via the 'filter' argument.
-  // It returns a time series - implemented as a list of AggregateLedgerEntry structs - with a row for every time
-  // the summed ledger entries of the set of specified accounts changes.
-  // Each account filter must contain no more than one party ID.
+  // This query returns all the ledger entries for a given subset of accounts, specified via the 'filter' argument.
```
wwestgarth, line 118: I don't think we need this line at all really because right below we say how entries can be filtered, and I think its obvious that the field called `filter` is used to filter the results.

```suggestion
```

guoguojin (dismissed):

candida-d (dismissed):

wwestgarth (dismissed):

datanode/entities/ledgerentry_filter.go
```diff
@@ -27,20 +27,12 @@ type LedgerEntryFilter struct {
...
-		var err error
-		filter.FromAccountFilter, err = AccountFilterFromProto(pbFilter.FromAccountFilter)
-		if err != nil {
-			return nil, err
-		}
-		filter.ToAccountFilter, err = AccountFilterFromProto(pbFilter.ToAccountFilter)
-		if err != nil {
-			return nil, err
-		}
+		filter.FromAccountFilter = AccountFilterFromProto(pbFilter.FromAccountFilter)
```
EVODelavega, line 34: nit-pick: will from _and_ to account filters be set at all times? I saw the `AccountFilterFromProto` wraps things in a nil check and returns an empty filter type if the input is `nil`, but IMO relying on these conversion functions checking for nil is a bit off, especially considering we have `pbFilter != nil` in this function, and this function previously returned `nil, err` in some cases, whereas now it will always return a non-nil pointer.

Like I said: a bit of a nit-pick, but overall this gives off the impression that we're more likely to silence potential errors, maybe to the point where invalid filter values are just going to end up being ignored and the entire data-set is going to be returned unfiltered.

datanode/api/trading_data_v2.go
```diff
@@ -267,11 +264,49 @@ func (t *TradingDataServiceV2) ListLedgerEntries(ctx context.Context, req *v2.Li
 		return nil, formatE(err)
 	}
 
+	leEedges := make([]*v2.LedgerEntriesEdge, 0, len(edges))
+	for _, agg := range edges {
+		if agg == nil || agg.Node == nil {
```
EVODelavega, line 269: this looks, to me at least, like we're accounting for a mistake elsewhere. If the data-set is empty, I'd expect `edges` to be empty, rather than `[]T{nil, T{Node: nil}}` or something

PR #8311, opened 2023-05-16
Initial supporting work for Ethereum oracles

### LINKED ISSUE #8247: Extend protos for Ethereum Oracles
# Feature Overview

In order to start building the APIs for users to specify ethereum oracles 
We will extend the required protos files

## Specs
- [Spec Issue and linked PR](https://github.com/vegaprotocol/specs/issues/881)

## Tasks
- [ ] Extend data source spec protos to specify data (ethereum contract address, ABI, Eth block height - types in the payload are TBC)
- [ ] Extend data chain event protos

pscott31:
Laying the groundwork for Ethereum oracles.

Nothing in enabled yet, but it would be nice to get this merged so that we've got a common base to work from, and to reduce the size of later commits.

Closes #8247


protos/sources/vega/data_source.proto
```diff
@@ -54,6 +55,41 @@ message DataSourceSpecConfiguration {
...
+  // Name of the method on the contract to call.
+  string method = 3;
+  // List of arguments to pass to method call.
+  // The protobuf 'Value' wraps an arbitaray JSON type, which is mapped to an ethereum type according to the ABI.
+  repeated google.protobuf.Value args = 4;
+  // Method for determining when to call the contract, and (optionally) at what interval
+  EthCallTrigger trigger = 5;  
+}
+
+// EthCallTrigger determines when the contract method should be called.
```
candida-d, line 74:
```suggestion
// Determines when the contract method should be called. 
```

protos/sources/vega/chain_events.proto
```diff
@@ -4,6 +4,18 @@ package vega;
...
 
+
+message EthContractCallEvent {
+  // The id of the DataSourceSpec that triggered this contract call
+  string spec_id = 1;
+  // Ethereum block height
+  uint64 block_height = 2;
+  // Ethereum block time (seconds since epoch)
+  uint64 block_time = 3;
+  // Result of contract call, packed according to the ABI stored in the associated DataSourceSpec
```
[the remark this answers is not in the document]
pscott31: Hi no it's not a vega epoch, its the UNIX timestamp epoch - 00:00 on 1 Jan 1970; it's [very standard](https://www.epochconverter.com/); but I'll be explicit about it. 

I think packed is the right word yes, it is the term used e.g. in the [Ethereum ABI specification](https://docs.soliditylang.org/en/v0.8.19/abi-spec.html)

protos/sources/vega/data_source.proto
```diff
@@ -54,6 +55,41 @@ message DataSourceSpecConfiguration {
...
+}
+
+// EthTimeTrigger is a trigger for an ethereum call based on the ethereum block timestamp. Can be one-off or repeating.
+message EthTimeTrigger {
+  // The first block in which ethereum time is greater or equal to this time (in seconds since the UNIX epoch)
+  // if not specified, begin repeating straight away.
+  optional uint64 initial = 1;  
+  // Optionally repeat the call every n seconds
+  optional uint64 every = 2;
+  // Optionally stop repeating the call once etherem time is greater than this time (in seconds since the UNIX epoch)
```
candida-d, line 88:
```suggestion
//  Trigger for an Ethereum call based on the Ethereum block timestamp. Can be one-off or repeating.
message EthTimeTrigger {
  // First block in which Ethereum time is greater or equal to this time, in seconds.
  // If not specified, begin repeating straight away
  optional uint64 initial = 1;  
  // Optionally repeat the call every n seconds
  optional uint64 every = 2;
  // Optionally stop repeating the call once Ethereum time is greater than this time, in seconds
```

pscott31: I think the UNIX epoch part is important - you can't measure absolute time in seconds without referencing an epoch. You could say something like 'encoded as a UNIX timestamp' but that requires you to know what that means

candida-d: I've edited my comment to say "in Unix seconds", which is the format we use in other doc-strings

pscott31: Ok thanks - I'll fold all those suggestions in

candida-d (dismissed):

protos/sources/vega/data_source.proto
```diff
@@ -54,6 +56,39 @@ message DataSourceSpecConfiguration {
   repeated vega.data.v1.Filter filters = 2;
 }
 
+// Specifies a data source that derives its content from calling a read method on an Ethereum contract.
+message DataSourceSpecEthCall {
```
AnExsomnis, line 60: I think we should be having something like this, no?
```
// EthCallSpecConfiguration type is to be used when handling ETH Oracle data source specs.
// The content of this type is derived from calling an Ethereum read contract.
message EthCallSpecConfiguration {
  // A deterministic id, created from hashing time+address
  string id = 1;
  uint64 time = 2;

  // Ethereum address of the contract
  string address = 3;
  // ABI of the Ethereum contract
  google.protobuf.ListValue abi = 4;
  // Name of the method to call on the contract.
  string method = 5;
  // Arguments used to pass to the method call.
  repeated google.protobuf.Value args = 6;

  // The signers of the data
  repeated vega.data.v1.Signer signers = 7;
  // Simple list of filters used as example, ETH most probably will not use these ones. // ?
  repeated vega.data.v1.EthereumSourceSpecFilter filters = 8;
}
```

protos/sources/vega/data_source.proto
```diff
@@ -54,6 +56,39 @@ message DataSourceSpecConfiguration {
...
+  string method = 3;
+  // List of arguments to pass to method call.
+  // Protobuf 'Value' wraps an arbitrary JSON type that is mapped to an Ethereum type according to the ABI.
+  repeated google.protobuf.Value args = 4;
+  // Conditions for determining when to call the contract method.
+  EthCallTrigger trigger = 5;
+}
+
+// Determines when the contract method should be called.
+message EthCallTrigger {
```
AnExsomnis, line 75: Can we call this simply a `Trigger`?

protos/sources/vega/data_source.proto
```diff
@@ -54,6 +56,39 @@ message DataSourceSpecConfiguration {
...
+    EthTimeTrigger time_trigger = 1;
+  }
+}
+
+// Trigger for an Ethereum call based on the Ethereum block timestamp. Can be one-off or repeating.
+message EthTimeTrigger {
+  // Trigger when the Ethereum time is greater or equal to this time, in Unix seconds.
+  // If not specified, begin repeating straight away.
+  optional uint64 initial = 1;
+  // Optionally repeat the call every n seconds
```
AnExsomnis, line 86:
```suggestion
  // Optionally repeat the call every n seconds.
```

protos/sources/vega/data_source.proto
pscott31: So one at a time:

```string id = 1;```

We already have an ID on the parent `DataSourceSpec`; or is there something else I'm missing here? e.g. we don't have one for the traditional oracles defined with `DataSourceSpecConfiguration`

```uint64 time = 2;```
What is this time? We've got the trigger mechanism for determining when to make the ethereum calls.

```repeated vega.data.v1.Signer signers = 7;```
Ethereum data sources don't have signers as I understand it. Is there some good reason why we'd want to include some sort of 'no signers' list of signers?

```// Simple list of filters used as example, ETH most probably will not use these ones. // ? repeated vega.data.v1.EthereumSourceSpecFilter filters = 8;
```
We will almost certainly want something like this but until I've got some code that actually makes use of it I figured it made sense to leave it out as it might take some as yet unexpected form (e.g. little scripts that run in an interpreter or what have you)

As for the name I am happy to change it to be consistent - but I can't quite see in my head what the difference is between Spec/Definition/Configuration that seem to be used sort of interchangeably at various levels. Is there some distinction to be made?

protos/sources/vega/data_source.proto
pscott31: Well we could I guess but it is kind of tied explicitly to Ethereum, e.g. `Trigger when the Ethereum time is greater or equal to this time`

protos/sources/vega/data_source.proto
pscott31: Sure, will fix

protos/sources/vega/data_source.proto
AnExsomnis: Yes, I thought we will be saving the time and the ID inside this object, but i might got it wrong.
The Ethereum data should have signers in all cases because it is external

AnExsomnis: The Definition of the data goes from the highest level as `DataSourceDefinition` that can be anything basically in the sub-levels.
`DataSourceSpecConfiguration` is the type left from the previous implementation (which I left for use from the pre-previous) and it is about the current oracles the VEGA system uses.
I suggested the one `EthCallSpecConfiguration` just to be consistent - because these two are going to be on the same level in the tree - just two types of oracles.
`DataSourceSpec` is again left name from the previous implementation and it represents the type of the data that is held by the instrument. If you go and see the Future type - it is `DataSourceSpecForSettlementData` and `DataSourceSpecForTermination` - these are DataSourceDefinition types that are wrapped further as ExternalDataSourceSpec and OracleDataSourceSpec.

protos/sources/vega/data_source.proto
AnExsomnis: All right

protos/sources/vega/data_source.proto
pscott31: OK i've renamed it DataSourceSpecConfigurationEthCall to match the others, though I think it's a bit of a mouthful!

I don't understand this:
> The Ethereum data should have signers in all cases because it is external

There aren't any signers for the data submitted by Ethereum oracles. Are you saying there's a good reason to put this in there even though it will always be empty?

AnExsomnis: Yes it is mouthful absolutely. Maybe a better name for both can be found?
Yes I am referring to this conversation- https://vegaprotocol.slack.com/archives/C056V8CBSCT/p1684239811515819
It just does not have to be signed with the ETH address;

core/evtforward/ethcall/engine.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (e *Engine) OnTick(ctx context.Context, vegaTime time.Time) {
+	block, err := e.client.BlockByNumber(ctx, nil)
+	if err != nil {
+		e.log.Errorf("failed to get current block header: %w", err)
+		return
+	}
+
+	e.log.Infof("tick: %v, blockNum:%v, ethTime: %v", vegaTime, block.Number(), time.Unix(int64(block.Time()), 0))
```
wwestgarth, line 68: Can we use structured logging please so that its consistent with the rest of core?
```suggestion
	e.log.Info("tick", logging.Time("vegaTime", vegaTime), logging.BigInt("block", block.Number()), logging.Uint64("blockTime", block.Time()))
```

protos/sources/vega/data_source.proto
```diff
@@ -54,6 +56,39 @@ message DataSourceSpecConfiguration {
...
+    EthTimeTrigger time_trigger = 1;
+  }
+}
+
+// Trigger for an Ethereum call based on the Ethereum block timestamp. Can be one-off or repeating.
+message EthTimeTrigger {
+  // Trigger when the Ethereum time is greater or equal to this time, in Unix seconds.
+  // If not specified, begin repeating straight away.
+  optional uint64 initial = 1;
+  // Optionally repeat the call every n seconds.
```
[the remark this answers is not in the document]
wwestgarth: I've just seen this in the title description "Can be one-off or repeating." which answers which way round it is! But I think its still worth having it explained against the field.

core/evtforward/ethcall/engine.go
pscott31: Yes of course - this was actually just a bit of debugging chaf that made it through but you are completely right

protos/sources/vega/data_source.proto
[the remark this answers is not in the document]
pscott31: Sure I'll rephrase

protos/sources/vega/data_source.proto
AnExsomnis: Right, so no signers for Ethereum oracles as discussed.

wwestgarth (approved):

candida-d (approved):

PR #8329, opened 2023-05-19
fix: stop deprecated netparams being in the default set

wwestgarth:
A follow on from: https://github.com/vegaprotocol/vega/pull/8291

@MuthuVega was testing this for me and noted that the `vega genesis etc.etc.` commands were still populating the genesis with the "deprecated-but-not-fully-removed" network parameters. So that has been fixed.

Also we realised that "hiding" the network-parameters from the API is just introducing difficulties with the difftool on the system tests. Thinking a bit deeper, it doesn't really make much sense to do this, or at least there isn't much benefit and it causes pain elsewhere. So I've removed that block.

So now a deprecated-network-parameter means:
- they are not added by any `vega genesis ...` commands
- you can't propose to change them
- they are allowed in a genesis file but do nothing (prevents it being a breaking change)


jeremyletang (dismissed):
looks good. there's that command to verify the genesis file, should that have any impact with it?

ze97286 (dismissed):

ze97286 (dismissed):

jeremyletang:
> > looks good. there's that command to verify the genesis file, should that have any impact with it?
> 
> Yea good idea. I've changed `vega verify genesis` so that it will point out deprecated network parameters that are still in the genesis file.

nice one. thanks.

ze97286 (approved):

jeremyletang (approved):

jeremyletang (approved):

PR #8331, opened 2023-05-19
Implement governance validation for successor markets + tests

### LINKED ISSUE #8330: Successor market validation
When a new market is proposed specifying a parent market id, the proposal needs to be validated to make sure the proposed new market can indeed be a successor to the specified market.

EVODelavega:
Closes #8330

### DIFF
--- a/core/governance/market.go
+++ b/core/governance/market.go
@@ -640,21 +640,17 @@ func validateSuccessorMarket(terms *types.NewMarket, parent *types.Market) (type
 }
 
 func validateParentProduct(prop *types.NewMarket, parent *types.Market) (types.ProposalError, error) {
-	asset, err := parent.GetAsset()
-	if err != nil {
-		// this shouldn't be possible, but let's handle all errors
-		return types.ProposalErrorInvalidSuccessorMarket, err
+	// make sure parent and successor are future markets
+	parentFuture := parent.GetFuture()
+	propFuture := prop.Changes.GetFuture()
+	if propFuture == nil || parentFuture == nil {
+		return types.ProposalErrorInvalidSuccessorMarket, fmt.Errorf("parent and successor markets must both be future markets")
 	}
-	qName, err := parent.TradableInstrument.Instrument.Product.GetQuoteName()
-	if err != nil {
-		// this shouldn't be possible, but let's handle all errors
-		return types.ProposalErrorInvalidSuccessorMarket, err
+	if propFuture.Future.SettlementAsset != parentFuture.Future.SettlementAsset {
+		return types.ProposalErrorInvalidSuccessorMarket, fmt.Errorf("successor market must use asset %s", parentFuture.Future.SettlementAsset)
 	}
-	if prop.Changes.Instrument.Product.Asset() != asset {
-		return types.ProposalErrorInvalidSuccessorMarket, fmt.Errorf("successor market must use asset %s", asset)
-	}
-	if prop.Changes.Instrument.Product.QuoteName() != qName {
-		return types.ProposalErrorInvalidSuccessorMarket, fmt.Errorf("successor market must use quote name %s", qName)
+	if propFuture.Future.QuoteName != parentFuture.Future.QuoteName {
+		return types.ProposalErrorInvalidSuccessorMarket, fmt.Errorf("successor market must use quote name %s", parentFuture.Future.QuoteName)
 	}
 	return types.ProposalErrorUnspecified, nil
 }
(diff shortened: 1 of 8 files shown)


core/types/market.go
```diff
@@ -413,6 +420,7 @@ type iProto interface {
 	iIntoProto() interface{}
 	getAsset() (string, error)
 	String() string
+	GetQuoteName() (string, error)
```
ze97286, line 423: do you mind calling it GetName rather than quoteName? quoteName sounds a bit confusing for spot.

ze97286: I hear you, but I hate it :) having getQuoteName on the product interface seems confusing to me, and I could be wrong but I don't think it's a standard name. In spot product I just call it Name. If there's anything better, happy to change to it.

EVODelavega: It's been a while since the future type was created, and quote name is used throughout. I do understand why it's jarring, considering a future doesn't have a quote name (at best it quotes a price, but that's besides the point). If memory serves, the `QuoteName` field here refers to the settlement asset (although I might be wrong, too). I think the idea was that a future like this would look something like this:

```
{
    SettlementAsset: "0xabcdef123456789", // asset ID
    QuoteName: "USDT", // human readable asset name
}
```

In that sense a future quote would could be easily read as 123 USDT rather than 123 0xabcdef12346789. Unless someone can point to what I'm missing here and correct me on what I think `QuoteName` refers to in this context, I'll change it to `GetAssetName()` or `GetSettlementAssetName()` or something

ze97286: GetSettlementAssetName is clearly worse from spots point of view. The documentation says:

```
"A Future product"
type Future {
  "The name of the asset (string)"
  settlementAsset: Asset!

  "String representing the quote (e.g. BTCUSD -> USD is quote)"
  quoteName: String!


```
so maybe lets stick with your current naming.

EVODelavega: ah, didn't check GQL for docs

core/governance/market.go
```diff
@@ -611,7 +607,56 @@ func validateNewMarketChange(
...
+			terms.Changes.RiskParameters = &types.NewMarketConfigurationLogNormal{
+				LogNormal: rm.LogNormalRiskModel.DeepClone(),
+			}
+		}
+	}
+	return types.ProposalErrorUnspecified, nil
+}
+
+func validateParentProduct(prop *types.NewMarket, parent *types.Market) (types.ProposalError, error) {
+	assets, err := parent.GetAssets()
```
ze97286, line 643: can we add verification that the product is a future product?

core/types/governance_new_spot_market.go
```diff
@@ -316,6 +316,10 @@ func (i InstrumentConfigurationSpot) String() string {
 	)
 }
 
+func (i InstrumentConfigurationSpot) QuoteName() string {
+	return i.Spot.QuoteAsset
```
ze97286, line 320: I don't think that's correct. Assuming the meaning is the name (vs ID of the asset). just return the i.Spot.Name

core/types/market.go
```diff
@@ -337,6 +337,14 @@ func (i InstrumentSpot) String() string {
 	)
 }
 
+// GetQuoteName must be implemented to satisfy the iProto interface.
+func (i InstrumentSpot) GetQuoteName() (string, error) {
+	if i.Spot == nil {
+		return "", ErrUnknownAsset
+	}
+	return i.Spot.QuoteAsset, nil
```
ze97286, line 345: same

core/types/governance_new_spot_market.go
EVODelavega: fixed

core/types/market.go
EVODelavega: fixed

core/governance/market.go
EVODelavega: added a check to make sure both successor and parent are future markets

core/types/governance_new_market.go
```diff
@@ -431,6 +433,14 @@ func (i InstrumentConfigurationFuture) Assets() []string {
 	return i.Future.Assets()
 }
 
+func (_ InstrumentConfigurationFuture) IsFuture() bool {
+	return true
```
ze97286, line 437: that doesn't feel right. Why not have a product enum instead?

EVODelavega: Agreed, will add something akin to what we have on the proposal type (getTermType I believe), along with getters for specific types. That should cut down significantly on this kind of interface bloat

AnExsomnis (approved):

ze97286 (approved):

PR #8333, opened 2023-05-19
feat: collateral support for spots

### LINKED ISSUE #8332: Spots collateral engine support
# Feature Overview

**In order to** be able to manage collateral transfers for spots, the collateral engine needs support for the following functions:
1. Create the accounts needed for spot market (LP fees and maker fees)
2. Clear spot market accounts when market is closed
3. Transfer funds from general account into holding account for a party
4. Release funds from holding account into party general account
5. Transfer spot from the general account of party a to party b.
6. Check if a party has sufficient funds in their account 

## Specs
- [[Spot spec](https://github.com/vegaprotocol/specs/blob/3ad0facad09ca18f225a2144b67d6b138ea46f7c/protocol/0080-SPOT-product_builtin_spot.md)](Spot spec WIP)

ze97286:
Closes #8332

### DIFF
--- a/core/collateral/engine.go
+++ b/core/collateral/engine.go
@@ -65,6 +65,8 @@ var (
 	ErrInvalidTransferTypeForFeeRequest = errors.New("an invalid transfer type was send to build a fee transfer request")
 	// ErrNotEnoughFundsToWithdraw a party requested to withdraw more than on its general account.
 	ErrNotEnoughFundsToWithdraw = errors.New("not enough funds to withdraw")
+	// ErrInsufficientFundsInAsset is returned if the party doesn't have sufficient funds to cover their order quantity.
+	ErrInsufficientFundsInAsset = errors.New("insufficient funds for order")
 )
 
 // Broker send events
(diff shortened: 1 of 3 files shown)


core/collateral/engine.go
```diff
@@ -2869,3 +2876,244 @@ func (e *Engine) GetRewardAccountsByType(rewardAcccountType types.AccountType) [
...
+		ToAccount:   []*types.Account{holdingAccount},
+	}
+
+	res, err := e.getLedgerEntries(ctx, req)
+	if err != nil {
+		e.log.Error("Failed to transfer funds", logging.Error(err))
+		return nil, err
+	}
+	for _, bal := range res.Balances {
+		if err := e.IncrementBalance(ctx, bal.Account.ID, bal.Balance); err != nil {
```
EVODelavega, line 2914: Not sure if we should call `IncrementBalance`. The `getLedgerEntries` updates the account balance with stuff like: `to.Account.Balance.AddSum(remainder)`, so wouldn't `UpdateBalance` make more sense?

In most cases we do use `IncrementBalance`, but the amount by which we increment is taken from `res.Entries[].Amount`

core/collateral/engine.go
```diff
@@ -2869,3 +2876,244 @@ func (e *Engine) GetRewardAccountsByType(rewardAcccountType types.AccountType) [
...
+		ToAccount:   []*types.Account{generalAccount},
+	}
+
+	res, err := e.getLedgerEntries(ctx, req)
+	if err != nil {
+		e.log.Error("Failed to transfer funds", logging.Error(err))
+		return nil, err
+	}
+	for _, bal := range res.Balances {
+		if err := e.IncrementBalance(ctx, bal.Account.ID, bal.Balance); err != nil {
```
EVODelavega, line 2952: Same here

ze97286:
> Apart from the `IncrementBalance` vs `UpdateBalance` thing, looks good to me. Perhaps a test that calls `ReleaseFromHoldingAccount` a couple of times to ensure that `IncrementBalance` isn't generating money out of thin air would be nice (ie holding account with balance of 1k, release in 2 tranches of 500 etc...)

I think it's fine, added a test as you suggested.

jeremyletang (approved):

PR #8336, opened 2023-05-19
successor enactment

### LINKED ISSUE #8334: Successor market: execution engine state and enacting successor markets
When a successor market is enacted the execution engine must:

Pass in the ELS state from the parent market if still valid, if no valid parent market state is present, the successor market must be enacted as a regular new market, no ELS data is transferred.
The market states with ELS data should be kept as long as they are relevant, once a successor market is enacted and leaves opening auction, the market state for the parent market is no longer relevant.

Network parameters must be propagated to successor markets correctly - in most cases this will not make much of a difference (assuming the parent market is not settled), but worth a look.

EVODelavega:
Closes #8334 

Note that this is based on the branch that closes 8330

### DIFF
--- a/core/execution/engine.go
+++ b/core/execution/engine.go
@@ -322,13 +322,17 @@ func (e *Engine) SucceedMarket(ctx context.Context, successor, parent string, in
 		return ErrMarketDoesNotExist
 	}
 	pm, ok := e.GetMarket(parent, true)
-	var parentM *types.Market
-	parentState, sok := e.marketCPStates[parent]
 	if !ok {
 		// so we have a successor market that has passed the vote, but the parent market either already was succeeded
 		// or the proposal vote closed when the parent market was still around, but it wasn't enacted until now
 		// and since then the parent market state expired. Should we reject this proposal or enact it as a non-successor?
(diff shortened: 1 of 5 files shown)


core/execution/engine.go
```diff
@@ -298,61 +298,67 @@ func (e *Engine) StartOpeningAuction(ctx context.Context, marketID string) error
...
+		assets, _ := mkt.mkt.GetAssets()
+		lm := e.collateral.SuccessorInsuranceFraction(ctx, successor, parent, assets[0], mkt.mkt.InsurancePoolFraction)
+		e.broker.Send(events.NewLedgerMovements(ctx, []*types.LedgerMovement{lm}))
+	}
+	// pass in the ELS and the like
+	mkt.InheritParent(ctx, parentState)
+	// if the parent market is in trading terminated state, then just remove it from the CP states
+	if parentM.State == types.MarketStateTradingTerminated {
+		delete(e.marketCPStates, parent)
+	}
```
jeremyletang, line 359: what if it's Settled, or whatever else which happens after TradingTerminated?>

jeremyletang (approved):

PR #8338, opened 2023-05-20
feat: ELS support for spots

### LINKED ISSUE #8337: [Feature]: ELS for spots
# Feature Overview

As the LP now has a different commitment amount on each side of the book, the following considerations must be made:

- Physical Stake:
  - An LPs `physical_stake` should be treated separately for each side of the book - call these the `buy_physical_stake` and the `sell_physical_stake`.
  - The current `physical_stake` for market stake calculations is the smaller of the two values, where the `sell_physical_stake` is converted into the `quote_asset` at the current `mark_price`.
- Virtual Stake:
  - An LPs `virtual_stake` should be treated separately for each side of the book - call these the `buy_virtual_stake` and `sell_virtual_stake`.
  - The same growth factor - as specified in the [LIQF spec](0042-LIQF-setting_fees_and_rewarding_lps.md) - derived from the `total value for fee purposes` in the quote asset is used to update both buy/sell virtual stakes (still in their respective assets).
  - The current `virtual_stake` for fee splitting is the smaller of the two values where the `sell_virtual_stake` is converted to `quote_asset` at the current `mark_price`.

From the above conditions, an LP is incentivised to provide a roughly equal value of liquidity on each side of the book at comparable levels of competitiveness in order to maximise their share of the liquidity fees.

## Specs
- [0080-SPOT-product_builtin_spot.md#submissions](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0080-SPOT-product_builtin_spot.md#submissions)

ze97286:
Closes #8337

### DIFF
--- a/core/execution/spot/equity_shares.go
+++ b/core/execution/spot/equity_shares.go
@@ -1,0 +1,300 @@
+// Copyright (c) 2022 Gobalsky Labs Limited
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE.VEGA file and at https://www.mariadb.com/bsl11.
+//
+// Change Date: 18 months from the later of the date of the first publicly
+// available Distribution of this version of the repository, and 25 June 2022.
+//
+// On the date above, in accordance with the Business Source License, use
+// of this software will be governed by version 3 or later of the GNU General
+// Public License.
+
+package spot
+
+import (
+	"fmt"
+
+	"code.vegaprotocol.io/vega/libs/num"
+)
+
+// lp holds LiquidityProvider stake and avg values.
+type lp struct {
+	// physical stake
+	buyStake      num.Decimal
+	sellStake     num.Decimal
+	physicalStake num.Decimal
+	// virtual stake
+	buyVStake    num.Decimal
+	sellVStake   num.Decimal
+	virtualStake num.Decimal
+
+	share num.Decimal
+	avg   num.Decimal
+}
+
+type EquityShares struct {
+	// market value proxy
+	mvp num.Decimal
+	// growth factor
+	r num.Decimal
+
+	totalVStake num.Decimal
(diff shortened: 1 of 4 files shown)


core/execution/spot/equity_shares.go
```diff
@@ -0,0 +1,300 @@
...
+	if es.r.IsZero() {
+		return
+	}
+	total := num.DecimalZero()
+	factor := num.DecimalOne().Add(es.r)
+	recalc := false
+	for _, v := range es.lps {
+		v.buyVStake = num.MaxD(v.buyStake, v.buyVStake.Mul(factor))
+		v.sellVStake = num.MaxD(v.sellStake, v.sellVStake.Mul(factor))
+		v.virtualStake = num.MinD(v.buyVStake, v.sellVStake.Mul(markPrice))
```
EVODelavega, line 80: Wouldn't this almost always guarantee that the virtual stake == sell virtual stake (except for when the sell virtual stake == 0). The buy virtual stake is either the physical stake or physical * (1 + r). With a mark price of, say, 10,000, the market growth would have to be insane, and even then, the virtual stake is scaled using the same growth factor.

ze97286: I don't follow. multiplying by mark price is to convert from base to quote, if the commitment is equal (when converting to quote space), there's no reason why one would be by definition greater than the other. Maybe I'm missing your point?

The way the above works is that the virtual stake for each is managed in the native token, so in quote for buy and in base for sell, then when calculating the virtual stake we choose the lower between the buy/sell virtual stake, having converted the sell to quote space.

EVODelavega: Ignore, just reading through the PR thinking about futures, not spot markets

EVODelavega (approved):

PR #8346, opened 2023-05-22
fix: fix internal ds listing

### LINKED ISSUE #8299: [Bug]: Console listing of market info does not show internal data source details
### Problem encountered

Listing of data for a market does not show internal data source properly

### Observed behaviour

Console listing of market info does not show internal data source details - behavior is fully described here - https://github.com/vegaprotocol/frontend-monorepo/issues/3765

### Expected behaviour

Expect to see internal data source details properly listed

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

0.71.4

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

AnExsomnis:
close: #8299


core/types/data_source.go
```diff
@@ -221,20 +218,52 @@ func (s DataSourceDefinition) DeepClone() DataSourceDefinition {
...
-			cpy = &DataSourceDefinition{
-				SourceType: &DataSourceDefinitionExternalx{
-					External: &DataSourceDefinitionExternal{
-						SourceType: &DataSourceSpecConfiguration{
-							Signers: SignersFromProto(dsn.External.GetOracle().Signers),
-							Filters: DataSourceSpecFiltersFromProto(dsn.External.GetOracle().Filters),
-						},
-					},
-				},
+			if dsn.External != nil {
```
EVODelavega, line 221: This looks quite repetitious, you have a similar type switch and nil checks for things like internal time a bit further down. Rather than having this `oneOfProto().(type)` switch wrapped around a `dsn.External.SourceTyp.(type)` switch, wouldn't it be possible to have an enum type of sorts, with some getters where you can do something like this:

```go
    switch s.SourceType {
    case SourceTypeExternal:
        if oracle := s.SourceType.oneOfProto().GetExternalOracle(); oracle != nil {
            signers := []*datapb.Signer{}
            // handle here
            return
        }
    case SourceTypeInternal:
        if tm := s.SourceType.oneOfProto().InternalTime(); tm != nil {
        }
    }
```

There still will be type assertions required in the actual getters, but this way, there would be less need to perform all the nil checks inside this many levels (we heve the outer `s.SourceType != nil`, a type switch on the `oneOfProto()`, with 2 nil checks, with another type switch. That's quite a lot of levels

datanode/entities/data_source_definition.go
```diff
@@ -0,0 +1,216 @@
...
+		Filters: []Filter{},
+	}
+
+	if s.DataSourceDefinition != nil {
+		if s.SourceType != nil {
+			switch dst := s.SourceType.(type) {
+			case *vega.DataSourceDefinition_External:
+				if dst.External != nil {
+					if dst.External.SourceType != nil {
+						switch dste := dst.External.SourceType.(type) {
```
EVODelavega, line 34: Same here, this many levels should be possible to avoid here

AnExsomnis: Right, I'll fix this everywhere, thanks!

datanode/sqlstore/migrations/0001_initial.sql
```diff
@@ -1260,8 +1260,7 @@ create table if not exists oracle_specs (
     id bytea not null,
     created_at timestamp with time zone not null,
     updated_at timestamp with time zone not null,
-    signers bytea[],
```
pscott31, line 1263: We'll need some migrations for this I think

core/types/data_source.go
AnExsomnis: This has been refactored everywhere, thanks!

datanode/sqlstore/migrations/0001_initial.sql
AnExsomnis: This has been resolved, thanks!

wwestgarth (approved):

PR #8348, opened 2023-05-23
Checkpoint restore for successor markets

### LINKED ISSUE #8347: Successor market state to be included in checkpoint data
When restoring from checkpoints, we restore markets based on the proposals in the checkpoint state. This can be used for successor markets all the same, but in addition to the proposal itself, the market will also need to hold the ELS data, which is tracked by the execution engine.

EVODelavega:
Closes #8347


core/execution/engine.go
```diff
@@ -298,21 +298,45 @@ func (e *Engine) StartOpeningAuction(ctx context.Context, marketID string) error
...
+
+func (e *Engine) restoreOwnState(ctx context.Context, mID string) (bool, error) {
+	mkt, ok := e.markets[mID]
+	if !ok {
+		return false, ErrMarketDoesNotExist
+	}
+	if state, ok := e.marketCPStates[mID]; ok {
+		// set ELS state and the like
+		mkt.InheritParent(ctx, state)
+		// if we had state of the market to restore, then check if this is a successor market
```
candida-d, line 319:
```suggestion
		// if there was state of the market to restore, then check if this is a successor market
```

core/execution/engine.go
```diff
@@ -321,8 +345,9 @@ func (e *Engine) SucceedMarket(ctx context.Context, successor, parent string, in
 		// nevertheless the proposal is still valid, and updated by the governance engine.
 		return ErrMarketDoesNotExist
 	}
+	// if this is a market restore, first check to see if we have some state already
```
candida-d, line 348:
```suggestion
	// if this is a market restore, first check to see if there is some state already
```

core/execution/engine.go
```diff
@@ -332,29 +357,37 @@ func (e *Engine) SucceedMarket(ctx context.Context, successor, parent string, in
 		return nil
 	}
 	var parentM *types.Market
+	if ok {
+		parentM = &pm
+	}
 	parentState, sok := e.marketCPStates[parent]
+	// we are restoring a market, but no state of the market nor parent market exists. Treat market as parent.
+	if restore && !sok && !ok {
+		// we are restoring a market, but the market state and parent market both are missing
```
candida-d, line 366:
```suggestion
		// network is restoring a market, but the market state and parent market both are missing
```

candida-d (dismissed):
some optional but preferred changes to the code comment docs, to remove the word "we" since it's not us doing any of these things :)

candida-d (dismissed):

jeremyletang (dismissed):

candida-d (dismissed):

candida-d (approved):

jeremyletang (approved):

PR #8350, opened 2023-05-23
BlockExplorer ListTransactions filter by tx types and parties

### LINKED ISSUE #7684: Block Explorer API: transfers/deposits to a user
# Task Overview

**As a** block explorer party trying to get an overview of the behaviour of one part
**I want** to see a user's deposits and withdrawals on the Party page
**So that** I can look at their inflows/outflows

## Current state
The party page on Explorer (e.g. https://stagnet3.explorer.vega.xyz/parties/121934387281812a2d5e6913e5d57c0f85a8f169e2752347ee2e23b52d46623c) shows an overview of the user's accounts and transactions. The transactions list comes from the block explorer API, querying for the submitter (https://docs.vega.xyz/testnet/api/rest/explorer/block-explorer-list-transactions). However for some events, including deposits, the party does not submit the event to the chain - validators do. This would also be true for transfers to a user.

My first thought was that we could add an additionalo filter to the listTransactions endpoint that would somehow allow us to filter on the recipient of a transfer or deposit. It could also be a new endpoint. I am indifferent, so either whoever picks up this ticket can decide or we can debate the endpoint structure here.

# Acceptance Criteria
- It is possible to list all of the transfers to and deposits to a specific public key using the Block Explorer API

# Test Scenarios
Detailed scenarios (1-3!) that can be executed as feature tests to verify that the feature has been implemented as expected.

GIVEN a user's first interaction with the system is a deposit
WHEN I query the new endpoint for the user
THEN I see one entry that lists the deposit

GIVEN a user's first interaction with the system is recieving a transfer
WHEN I query the new endpoint for the user
THEN I see one entry that lists the transfer

### LINKED ISSUE #8328: Block exporer API: Add filtering of transactions by types
Screenshot 2023-05-19 at 14 02 56

For the transaction list on Block Explorer, I would like to build a set of checkboxes that the user can use to filter the transaction list by transaction type. The current API supports passing a single filter, but does not support passing multiple filters.

# What we have now
https://be.vega.community/rest/transactions?limit=20&filters[cmd.type]=Withdraw - a single filter on transaction type can be passed in to the transaction list endpoint. Not specifying a filter returns all transactions. Specifying a filter on cmd.type returns only those transaction types.

✅ https://be.vega.community/rest/transactions?limit=20&filters%5Bcmd.type%5D=Validator%20Heartbeat

❌ https://be.testnet.vega.xyz/rest/transactions?limit=10&filters[cmd.type]=Validator%20Heartbeat&filters[cmd.type]=Withdraw
❌ https://be.testnet.vega.xyz/rest/transactions?limit=10&filters[cmd.type]=Validator%20Heartbeat,Withdraw
❌ https://be.testnet.vega.xyz/rest/transactions?limit=10&filters[cmd.type]=Validator%20Heartbeat|Withdraw
❌ https://be.testnet.vega.xyz/rest/transactions?limit=10&filters[cmd.type]=[Validator%20Heartbeat,Withdraw]

# What we need
There are two choices:
- Allow multiple `cmd.type` filters. 
    > A client that wants everything except validator heartbeat transactions must pass a list of the ~[23 cmd.types](https://github.com/vegaprotocol/vega/blob/97129d575716076693e70915f971b24ab32fe7ba/core/txn/command.go#L73-L97) that it wants.
- Add a `without` style filter, that filters out the specified type or types.
    > A client that wants everything except validator heartbeat transactions must pass in 'Validator Heartbeat' to the new `without` parameter.

# Task
- [ ] Update the documentation for https://docs.vega.xyz/mainnet/api/rest/explorer/block-explorer-list-transactions so that it includes documentation for how to use the filters
- [ ] Agree on which approach to use
- [ ] Implement it

zale144:
Deprecate BlockExplorer ListTransactions filter map, and add tx_type and party parameters.

#7684


protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -65,8 +65,18 @@ message ListTransactionsRequest {
   optional string before = 2;
   // Optional cursor to paginate the request
   optional string after = 3;
+  // Deprecated: We will remove this field in favor ov the fields below
```
edd, line 68:
```suggestion
  // Deprecated: We will remove this field in favour of the fields below
```

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -65,8 +65,18 @@ message ListTransactionsRequest {
...
+  // And we might add more filters to support the legacy ones, if needed
```
candida-d, line 69:
```suggestion
  // More filters may be added to support the legacy ones, if needed
```

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -65,8 +65,18 @@ message ListTransactionsRequest {
...
   // Filters to apply to the request
-  map<string, string> filters = 4;
+  map<string, string> filters = 4  [deprecated = true];
+  // TxTypeFilter (command or chain event)
```
candida-d, line 72:
```suggestion
  // Transaction type filter, for command or chain event
```

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -65,8 +65,18 @@ message ListTransactionsRequest {
...
+  repeated string tx_type = 5;
+  // PartyFilter (either sender or receiver)
```
candida-d, line 74:
```suggestion
  // Party ID filter, can be sender or receiver
```

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -65,8 +65,18 @@ message ListTransactionsRequest {
   optional string before = 2;
   // Optional cursor to paginate the request
   optional string after = 3;
+  // Deprecated: This field will be removed in favour of the fields below
+  // More filters may be added to support the legacy ones, if needed
   // Filters to apply to the request
-  map<string, string> filters = 4;
+  map<string, string> filters = 4  [deprecated = true];
```
pscott31, line 71: I don't have a problem with depreciating this but when I added it it seemed handy to be able to filter on arbitrary attributes; what is the logic behind removing it?

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -65,8 +65,18 @@ message ListTransactionsRequest {
...
+  // Transaction type filter, for command or chain event
+  repeated string tx_type = 5;
+  // Party ID filter, can be sender or receiver
+  repeated string party = 6;
+  // Sender of the transaction
+  repeated string sender = 7;
+  // Receiver of the transaction
```
pscott31, line 78: Dunno if it's maybe worth being a bit more explicit about this as there isn't really a 'receiver' for a transaction; but I know what you mean.. perhaps something like 'In the case of a deposit.. etc'. Maybe that's too wordy

protos/sources/blockexplorer/api/v1/blockexplorer.proto
zale144: The reason is that it would be nice to have a predefined set of parameters which we can document properly. Right now, the possible keys of this filter are not very obvious and are not documented. However, we won’t be removing the filter until we have all required parameters added and well documented.

blockexplorer/store/migrations/0004_txtype_index.sql
pscott31, line 1: I've got a little background worry that continuing to add triggers like this might slow things down beyond the ability of the indexer to keep up; but I guess we'll just have to try it  and see. each trigger adds a bit of time to an insert and I think we're pretty close to threshold.

pscott31 (dismissed):

protos/sources/blockexplorer/api/v1/blockexplorer.proto
zale144: To be fair, this PR is more about adding filtering by party and tx type (Deposit, Withdraw, Transfer...) than removing stuff. I don't really mind the filters, although it would be nice to have them documented.

protos/sources/blockexplorer/api/v1/blockexplorer.proto
zale144: Ah I see. We could say something like "where it applies" ?

blockexplorer/store/migrations/0004_txtype_index.sql
zale144: Yeah, it did occur to me as well, but I also though it is hard to predict. Also, I'm not sure what the alternatives are?

AnExsomnis (dismissed):

candida-d (dismissed):

blockexplorer/store/migrations/0004_txtype_index.sql
```diff
@@ -0,0 +1,81 @@
+
+-- +goose Up
+ALTER TABLE tx_results ADD column tx_type TEXT;
+ALTER TABLE tx_results ADD column sender TEXT;
+ALTER TABLE tx_results ADD column receiver TEXT;
+CREATE INDEX tx_results_tx_type_sender_index ON tx_results(tx_type, sender);
```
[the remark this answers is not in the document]
zale144: @EVODelavega Hey, I'm sorry, seems your comment somehow fell through the cracks... 
Look, those are all very valid points. However, taking into account that the Tendermint indexer will apparently be removed from Tendermint and left to be implemented at the app level, this PR brings but a workaround, until we get to implement it in the app level, which might happen sooner rather than later. I think it would be acceptable to make sure that this change will not get in the way and slow down the validator performance further, and we kind of did that by running performance tests (thnx @peterbarrow) and comparing the results to those of the current develop branch. So, I wouldn't look too much into it right now.

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,163 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+			}
+		}
+		if e20 := c.GetErc20(); e20 != nil {
+			if e20d := e20.GetDeposit(); e20d != nil {
+				txt.Type = TxTypeERC20Deposit
+				txt.Receiver = e20d.GetTargetPartyId()
+			}
+			if e20w := e20.GetWithdrawal(); e20w != nil {
+				txt.Type = TxTypeERC20Withdrawal
+				txt.Sender = tx.Party() // TODO: what else to do here?
```
pscott31, line 269: That seems ok to me if tx.Party() is the public key of the party whose funds are being withdrawn?

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,163 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+			}
+		}
+		if erc20ms := c.GetErc20Multisig(); erc20ms != nil {
+			if erc20msadd := erc20ms.GetSignerAdded(); erc20msadd != nil {
+				txt.Type = TxTypeERC20MultisigSignerAdded
+				txt.Sender = erc20msadd.GetNewSigner() // TODO: what else to do here?
```
pscott31, line 275: I guess this doesn't really make a huge amount of sense semantically but if it's something we want to be able to filter on in the block explorer it's not the worst hack; without adding attributes for just about every field any sort of tx can have

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,163 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+			}
+			if erc20msrem := erc20ms.GetSignerRemoved(); erc20msrem != nil {
+				txt.Type = TxTypeERC20MultisigSignerRemoved
+				txt.Sender = erc20msrem.GetOldSigner() // TODO: what else to do here?
```
pscott31, line 279: ditto

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,163 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+		txt.Type = TxTypeKeyRotateSubmission
+		txt.Sender = tx.Party()
+	case *commandspb.StateVariableProposal:
+		txt.Type = TxTypeStateVariableProposal
+		txt.Sender = tx.Party()
+	case *commandspb.CancelTransfer:
+		txt.Type = TxTypeCancelTransfer
+		txt.Sender = tx.Party()
+	// case *commandspb.ValidatorHeartbeat: TODO: do we need this?
+	// 	txt.Type = TxTypeValidatorHeartbeat
```
pscott31, line 338: I don't think so, maybe @edd has an opinon though?

pscott31 (dismissed):

core/blockchain/abci/abci.go
zale144: @pscott31 well, it's the same key as the submitter 🤷‍♀️

core/blockchain/abci/abci.go
zale144: yeah that was my thinking. We will end up having too many attributes recorded. We can always document it nicely

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,161 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+	TxTypeAnnounceNode                   = "AnnounceNode"
+	TxTypeNodeVote                       = "NodeVote"
+	TxTypeNodeSignature                  = "NodeSignature"
+	TxTypeOracleDataSubmission           = "OracleDataSubmission"
+	TxTypeDelegateSubmission             = "DelegateSubmission"
+	TxTypeUndelegateSubmission           = "UndelegateSubmission"
+	TxTypeKeyRotateSubmission            = "KeyRotateSubmission"
+	TxTypeStateVariableProposal          = "StateVariableProposal"
+	TxTypeCancelTransfer                 = "CancelTransfer"
+	// TxTypeValidatorHeartbeat                 = "ValidatorHeartbeat".
```
jeremyletang, line 223: why is this excluded?

jeremyletang:
Before this gets merge let's make sure this is exactly what's required by @edd / the work on the block explorer.

core/blockchain/abci/abci.go
edd: @jeremyletang agreed. There was a discussion on if heartbeats should be included. We all agreed they should, because *excluding* them was misleading.

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,161 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+	TxTypeProtocolUpgradeProposal     = "ProtocolUpgradeProposal"
+	TxTypeIssueSignatures             = "IssueSignatures"
+	TxTypeBatchMarketInstructions     = "BatchMarketInstructions"
+	TxTypeERC20MultisigSignerAdded    = "ERC20MultisigSignerAdded"
+	TxTypeERC20MultisigSignerRemoved  = "ERC20MultisigSignerRemoved"
+)
+
+type TxType struct {
+	Type     string
+	Sender   string
```
jeremyletang, line 234: Seems like Sender here, seeing how it's used my be a duplicate to the `submitter` label which is already added somewhere for every transaction isn't?

edd:
> Before this gets merge let's make sure this is exactly what's required by @edd / the work on the block explorer.

The two tickets I thought this would address are:
https://github.com/vegaprotocol/vega/issues/8328
https://github.com/vegaprotocol/vega/issues/7684

And I don't think it achieves this:
1. I can't manually exclude types. This would be OK, except
2. I can't see how to manually filter for multiple types. This is the most important part of #8328. 
3. I'm not sure if the party filter in this PR fixes #7684, or if the party filter *and* multiple types is how I would address this?
4. #8328 point 1 was about documenting the API endpoint properly. I don't think the documentation generated here is going to help me answer questions 2 and 3

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,161 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
 	return resp
 }
 
+const (
```
jeremyletang, line 198: Alright, I've just check with @edd and he doesn't need the details in here (e.g: StakeDeposited, StakeRemove, etc) in the ChainEvent, so all ChainEvent should have the type ChainEvent

jeremyletang:
> > Before this gets merge let's make sure this is exactly what's required by @edd / the work on the block explorer.
> 
> The two tickets I thought this would address are: #8328 #7684
> 
> And I don't think it achieves this:
> 
>     1. I can't manually exclude types. This would be OK, except
> 
>     2. I can't see how to manually filter for multiple types. This is the most important part of [Block exporer API: Add filtering of transactions by types #8328](https://github.com/vegaprotocol/vega/issues/8328).
> 
>     3. I'm not sure if the party filter in this PR fixes [Block Explorer API: new API endpoint to fetch party's inflows and outflows #7684](https://github.com/vegaprotocol/vega/issues/7684), or if the party filter _and_ multiple types is how I would address this?
> 
>     4. [Block exporer API: Add filtering of transactions by types #8328](https://github.com/vegaprotocol/vega/issues/8328) point 1 was about documenting the API endpoint properly. I don't think the documentation generated here is going to help me answer questions 2 and 3

1. correct you can't exclude type, and this should be part of this PR.
2. you are wrong, you should be able to pass in multiple type
3. you won't be able to fetch that reliably, the datanode API for deposits, withdrawals and transfers is what you are looking for here.
4. yes we need proper documentation here of some sort, because it's quite difficul / impossible for a user to know what to filter here.

edd:
> 3. you won't be able to fetch that reliably, the datanode API for deposits, withdrawals and transfers is what you are looking for here.

OK. I'm going to close ticket  https://github.com/vegaprotocol/vega/issues/7684 and use the data node for these queries. As I've said over there, I'll use data node for those queries. This isn't quite as neat as a block explorer endpoint would be, but for balances across chain resets there would be no TX anyway

core/blockchain/abci/abci.go
zale144: Are we sure we want to add the attributes for all the heartbeats to the attributes table? I mean, fine by me

core/blockchain/abci/abci.go
```diff
@@ -194,6 +195,161 @@ func AddCommonDeliverTxEvents(resp types.ResponseDeliverTx, tx Tx) types.Respons
...
+		if e20 := c.GetErc20(); e20 != nil {
+			if e20d := e20.GetDeposit(); e20d != nil {
+				txt.Type = TxTypeERC20Deposit
+				txt.Receiver = e20d.GetTargetPartyId()
+			}
+			if e20w := e20.GetWithdrawal(); e20w != nil {
+				txt.Type = TxTypeERC20Withdrawal
+				txt.Sender = tx.Party()
+			}
+		}
```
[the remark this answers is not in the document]
zale144: @jeremyletang we kind of need to be able to search by the specific transaction type, not just `tx.type=ChainEvent`, that's not granular enough - we need "Withdrawal", "Deposit", "Transfer", etc.

core/blockchain/abci/abci.go
zale144: @jeremyletang I'm fine merging those two if we can assume that the submitter will always be the same as the sender (for example in transfers). However, there is an exception in `GetErc20Multisig`. It wouldn't make sense to have an attribute for every single field, so we settle for the 3 we have now: `type`, `submitter`, `sender` and `receiver`. I'm open to suggestions here, I know this is not ideal

core/blockchain/abci/abci.go
zale144: @jeremyletang right, however I think we still need to be able to filter by say `Deposit`, which can only be found in the `ChainEvent`. Otherwise, we already have filtering by command type, which in this case is incomplete.

zale144:
> > 5\. you won't be able to fetch that reliably, the datanode API for deposits, withdrawals and transfers is what you are looking for here.
> 
> OK. I'm going to close ticket #7684 and use the data node for these queries. As I've said over there, I'll use data node for those queries. This isn't quite as neat as a block explorer endpoint would be, but for balances across chain resets there would be no TX anyway

@edd @jeremyletang Well this is the main issue, filtering by things like `Deposit`, `Transfer` and so on is quite tricky here, and as you saw I was able to get these attributes from the `ChainEvent` through triggers, which is not the most elegant way... Having said that, I do see the utility of filtering through transactions by any type, however with things like Transfers and Deposits, it's not that simple, since they can be initiated from different places (not command).

So the way I see it, we can choose from two solutions here:

1. Continue with this implementation, maybe get only the chain events that we care about (deposit...?), and add the exclusion filter, meaning filter of txs which we don't want to see
2. Forget about this whole thing, continue filtering by `cmd.type` and make the both filters a list (`[cmd.types]`, `[parties]`), and again exclusion filter like in no.1. As for transactions that didn't start from a command, use datanode

I think we are probably already on the same page: no 2.

zale144:
Closing in favour of https://github.com/vegaprotocol/vega/pull/8543

PR #8352, opened 2023-05-23
Make sure successor config is correctly set

EVODelavega:
Successor is optional, but non-nil pointers for empty values should be treated as optional field not being set.


ze97286:
can we have a unit test for this? something like market_cp_test or something.

core/execution/engine_checkpoint.go
```diff
@@ -34,6 +34,10 @@ func (e *Engine) Checkpoint() ([]byte, error) {
 }
 
 func (e *Engine) Load(ctx context.Context, data []byte) error {
+	if len(data) == 0 {
```
wwestgarth, line 37: I don't think this is necessary since in the checkpoint engine we only call `Load()` on an engine if data exists:
```
cpData := cp.Get(k)
if len(cpData) == 0 {
    continue
}
```
and the map is already set to non-nil in `execution.NewEngine()`.

core/governance/market_cp_restore_test.go
```diff
@@ -90,6 +93,41 @@ func TestMarketRestoreFromCheckpoint(t *testing.T) {
 	}
 }
 
+func TestMarketRestoreFromCheckpointWithEmptySuccessor(t *testing.T) {
+	t.Skipf("Skipping test as need to regenerate testcp/checkpoint.cp with appropriate values for LP - Zohar to fix")
```
wwestgarth, line 97: Probably want to unskip it?

core/types/governance_new_market.go
```diff
@@ -330,6 +330,10 @@ func NewMarketConfigurationFromProto(p *vegapb.NewMarketConfiguration) (*NewMark
 }
 
 func SuccessorConfigFromProto(p *vegapb.SuccessorConfiguration) *SuccessorConfig {
+	// successor config is optional, but make sure that, if provided, it's not set to empty parent market ID
+	if len(p.ParentMarketId) == 0 {
```
wwestgarth, line 334: How would a checkpoint file get into this state? In the protos the successor config is a new optional field, so for a checkpoint file taken from mainnet pre-all this code it would unmarshal into a nil-pointer.

And in new code any market proposal with a successor config and no parent ID would fail validation and never exist to be in a checkpoint.

core/execution/engine_checkpoint.go
EVODelavega: removed

core/governance/market_cp_restore_test.go
EVODelavega: indeed, I should've known things passing first time around was too good to be true :)

ze97286 (approved):

PR #8365, opened 2023-05-24
Add new liquidity engine with SLA support

### LINKED ISSUE #8367: Liquidity SLA: build liquidity engine (mocking the orders)
In order to get the liquid sla started we need to build the new liquidity engine. For this we should be mocking the orders, this will allow faster development and for the testing activities to be started as early as possible.

karlem:
close #8367


core/liquidity/v2/types.go
```diff
@@ -0,0 +1,138 @@
...
+func (l lpsByFee) Len() int           { return len(l) }
+func (l lpsByFee) Less(i, j int) bool { return l[i].Fee.LessThan(l[j].Fee) }
+func (l lpsByFee) Swap(i, j int)      { l[i], l[j] = l[j], l[i] }
+
+// sortByFee sorts in-place and returns the LiquidityProvisions for convenience.
+func (l Provisions) sortByFee() Provisions {
+	byFee := lpsByFee(l)
+	sort.Sort(byFee)
+	return Provisions(byFee)
+}
```
jeremyletang, line 58: Why not just do  this:
```suggestion
// sortByFee sorts in-place and returns the LiquidityProvisions for convenience.
func (l Provisions) sortByFee() Provisions {
	sort.Slice(l, func(i, j int) bool {return l[i].Fee.LessTna(l[j].Fee)})
	return Provisions(byFee)
}
```

karlem: Because it was a CP and Paste and I have not cleared that file yet mate.

core/liquidity/v2/sla.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	notionalVolume := num.DecimalZero()
+	orders := e.getAllActiveOrders(party)
+
+	for _, o := range orders {
+		price := o.Price.ToDecimal()
+		// this order is in range and does contribute to the volume on notional
+		if price.GreaterThanOrEqual(minPrice) && price.LessThanOrEqual(maxPrice) {
+			notionalVolume = notionalVolume.Add(price)
```
wwestgarth, line 164: should this not be adding up the orders sizes and not the price to get the LP's volume in that price range? i.e
```
notionalVolume.Add(o.TrueRemaining())
```

core/liquidity/v2/sla.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	periodAveragePenalty := num.DecimalZero()
+	for _, p := range previousPenalties[windowStart:] {
+		periodAveragePenalty = periodAveragePenalty.Add(p)
+	}
+
+	performanceHysteresisEpochsD := num.NewDecimalFromFloat(float64(e.performanceHysteresisEpochs))
+
+	periodAveragePenalty = periodAveragePenalty.Div(performanceHysteresisEpochsD)
```
wwestgarth, line 211: Wondering if we should be dividing by `len(previousPenalties[windowStart:])` here so thats its the average of whatever history we have. For example if this is only the second epoch for an LP, they had a penatly of `1` in the previous epoch and `performanceHysteresisEpochs = 10` we'll get their historic average being `1/10`.

This could just be my intrepretation of the spec though.

core/liquidity/v2/sla.go
```diff
@@ -0,0 +1,229 @@
...
+			commitment.start = now
+		}
+
+		commitment.s = 0
+	}
+
+	e.slaEpochStart = now
+}
+
+func (e *Engine) BeginBlock(txs []TX) {
```
wwestgarth, line 34: Not blocking this PR because I know you've not wired it in yet but: Can we currently get all the transaction for a block upfront to be able to do this? Wondering if we can just hash the block time instead, surely no one can guess that before it happens.

core/liquidity/v2/sla.go
karlem: I think you are right.

wwestgarth (dismissed):

ze97286 (commented):
what about the snapshot of the liquidity engine?

karlem:
@ze97286 there will be another PR later to add snapshot. For now it would be good to make everything work e2e and then add the snapshot.

ze97286:
> @ze97286 there will be another PR later to add snapshot. For now it would be good to make everything work e2e and then add the snapshot.

is there a ticket? just to make sure we don't forget

karlem:
> 

https://github.com/vegaprotocol/vega/issues/8505

karlem:
> > @ze97286 there will be another PR later to add snapshot. For now it would be good to make everything work e2e and then add the snapshot.
> 
> is there a ticket? just to make sure we don't forget

https://github.com/vegaprotocol/vega/issues/8505

jeremyletang (dismissed):

jeremyletang (approved):

PR #8375, opened 2023-05-25
chore: add successor markets API

### LINKED ISSUE #8303: Datanode to handle successor markets
With the introduction of successor market proposals, the governance side of the datanode will need to account for that new proposal type. 

Markets can now have a parent market ID (pointing to the market they replaced). We need to be able to store this data, and perhaps make it easy to query for a market + its ancestry and offspring (ie fetch market ID + its lineage).

guoguojin:
Closes #8303


datanode/gateway/graphql/schema.graphql
```diff
@@ -3361,10 +3384,19 @@ type NewMarket {
   positionDecimalPlaces: Int!
   "Liquidity Provision order price range"
   lpPriceRange: String!
-  "Linear slippage factor is used to cap the slippage component of maintainence margin - it is applied to the slippage volume"
+  "Linear slippage factor is used to cap the slippage component of maintenance margin - it is applied to the slippage volume"
   linearSlippageFactor: String!
-  "Quadratic slippage factor is used to cap the slippage component of maintainence margin - it is applied to the square of the slippage volume"
+  "Quadratic slippage factor is used to cap the slippage component of maintenance margin - it is applied to the square of the slippage volume"
   quadraticSlippageFactor: String!
+  "Successor configuration. If this proposal is meant to succeed a given market, then this should be set."
```
candida-d, line 3391:
```suggestion
  "Successor market configuration. If this proposed market is meant to succeed a given market, then this needs to be set."
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3361,10 +3384,19 @@ type NewMarket {
...
+  successorConfiguration: SuccessorConfiguration
+}
+
+type SuccessorConfiguration {
+  "The ID of the market this proposal should succeed"
```
candida-d, line 3396:
```suggestion
  "ID of the market this proposal will succeed"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3361,10 +3384,19 @@ type NewMarket {
...
+  parentMarketId: String!
+  "A decimal value between 0 and 1, specifying the fraction of the insurance pool balance is carried over from the parent market to the successor."
```
candida-d, line 3398:
```suggestion
  "Decimal value between 0 and 1, specifying the fraction of the insurance pool balance that is carried over from the parent market to the successor."
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -344,6 +344,14 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Markets"};
   }
 
+  // List successor markets
+  //
+  // Given a market id, return the full lineage of markets since inception, or all successor markets since and including
+  // the given market id.
```
candida-d, line 350:
```suggestion
  // Given a market ID, return the full lineage of markets since inception, or all successor markets since and including
  // the given market ID.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1981,22 @@ message MarketConnection {
   PageInfo page_info = 2;
 }
 
+// Request for listing successor markets
```
candida-d, line 1984:
```suggestion
// Request for listing successor markets.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1981,22 @@ message MarketConnection {
...
+message ListSuccessorMarketsRequest {
+  // Market ID that is a member of the succession chain. This can be the original market
+  // or any subsequent child market that succeeded.
```
candida-d, line 1987:
```suggestion
  // or any subsequent child market that succeeded it.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1981,22 @@ message MarketConnection {
...
+  string market_id = 1 [(google.api.field_behavior) = REQUIRED];
+  // Flag to indicate whether or not to include the full succession chain, or only list
+  // the children of the given market ID. If true, the full succession chain
```
candida-d, line 1990:
```suggestion
  // the children of the given market ID. If true, the full succession chain is included
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1981,22 @@ message MarketConnection {
...
+  bool full_history = 2;
+}
+
+// Response from listing successor markets request.
```
candida-d, line 1994:
```suggestion
// Response from a list successor markets request.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1981,22 @@ message MarketConnection {
...
+message ListSuccessorMarketsResponse {
+  // List of markets and corresponding page information.
```
candida-d, line 1996: "list of successor markets" specifically?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2411,6 +2435,8 @@ message ListGovernanceDataRequest {
     TYPE_NEW_FREE_FORM = 6;
     // Update asset proposals
     TYPE_UPDATE_ASSET = 7;
+    // Successor Markets
```
candida-d, line 2438:
```suggestion
    // Successor markets
```

candida-d (changes requested):

protos/sources/data-node/api/v2/trading_data.proto
wwestgarth: There is also no page information returned. I guess we've decided not to because it there will never been that many?

datanode/entities/market.go
```diff
@@ -133,6 +135,21 @@ func NewMarketFromProto(market *vega.Market, txHash TxHash, vegaTime time.Time)
 		quadraticSlippageFactor = &factor
 	}
 
+	parentMarketID := MarketID("")
+	if market.ParentMarketId != nil && *market.ParentMarketId != "" {
+		parent := MarketID(*market.ParentMarketId)
+		parentMarketID = parent
+	}
+
+	insurancePoolFraction := (*num.Decimal)(nil)
```
jeremyletang, line 144: Wouldn't that works nicer?
```suggestion
	var insurancePoolFraction *num.Decimal
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -344,6 +344,14 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Markets"};
   }
 
+  // List successor markets
+  //
+  // Given a market ID, return the full lineage of markets since inception, or all successor markets since and including
+  // the given market ID.
```
wwestgarth, line 350: Can we say something about the order of the markets in the returned array so users know what to expect? Something (assuming it is this way around) like. "The returned markets will be ordered such that market in position `i` will be the parent of market `i+1`"?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1981,22 @@ message MarketConnection {
...
 }
 
+// Request for listing successor markets.
+message ListSuccessorMarketsRequest {
+  // Market ID that is a member of the succession chain. This can be the original market
+  // or any subsequent child market that succeeded it.
+  string market_id = 1 [(google.api.field_behavior) = REQUIRED];
+  // Flag to indicate whether or not to include the full succession chain, or only list
+  // the children of the given market ID. If true, the full succession chain is in included.
+  bool full_history = 2;
```
wwestgarth, line 1991: In other APIs where we have booleans we name them like `include_settled`, `exclude_liquidity` and `live_only` (which should probably have been called something like `exclude_inactive`). I wonder if we should try to keep to something similar here too and call it `include_parents`?

@candida-d any opinions?

candida-d: I'd always vote for consistency. And now that you mention it, "full history" isn't the most obvious name for this boolean option. 

Would it be `include parent`, as presumably there's only one parent?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1973,6 +1982,22 @@ message MarketConnection {
   PageInfo page_info = 2;
 }
 
+// Request for listing successor markets.
+message ListSuccessorMarketsRequest {
+  // Market ID that is a member of the succession chain. This can be the original market
+  // or any subsequent child market that succeeded it.
+  string market_id = 1 [(google.api.field_behavior) = REQUIRED];
+  // Flag to indicate whether or not to include the full succession chain, or only list
+  // the children of the given market ID. If true, the full succession chain is in included.
```
candida-d, line 1991:
```suggestion
  // the children of the given market ID. If true, the full succession chain is included.
```

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: Well the API will return the market requested and its children only if full_history is false, and if true, then it will return the full successor chain, i.e. the full history. I'm happy to change the name, but I don't think include_parents is quite accurate either.

guoguojin: I suppose include_full_history would be more in line with what @wwestgarth is suggesting?

wwestgarth:
Just a question about the use-case for the `ListSuccessorMarket` API, where would it be used? is it something FE has asked for? If its just to check if a market already has a child before propsing one can we not just have a `child_market_id` on the market instead?

datanode/gateway/graphql/schema.graphql
```diff
@@ -677,6 +677,18 @@ type Query {
   "Get statistics about the Vega node"
   statistics: Statistics!
 
+  "List markets in a succession chain"
```
candida-d, line 680: On second thought, how about the "succession line" or another phrase that doesn't use the word chain?

guoguojin:
Well for example, if you proposed something like a BTC/USD Future with September 23 maturity, and proposals are made for all its successor markets which would expire quarterly for the next year for example, you would have:

BTC/USD Sept/23
BTC/USD Dec/23
BTC/USD Mar/24
BTC/USD Jun/24

Each child when the proposals passes and enacted would be pending and would go into auction at the appropriate time and eventually succeed it's parent.

wwestgarth:
> Well for example, if you proposed something like a BTC/USD Future with September 23 maturity, and proposals are made for all its successor markets which would expire quarterly for the next year for example, you would have:
> 
> BTC/USD Sept/23 BTC/USD Dec/23 BTC/USD Mar/24 BTC/USD Jun/24
> 
> Each child when the proposals passes and enacted would be pending and would go into auction at the appropriate time and eventually succeed it's parent.
> 
> From what I understand, the proposals for the children could be made at the same time as the original parent. You wouldn't necessarily wait until one is about to expire before proposing the next child.
> 
> Also, as a user, if I were writing a trading strategy and I wanted to get the historic market data for each of the markets in the succession line in order to do analytics, then I would need to know the succession line to fetch the data for previous markets.

Ok I see.

Maybe an edge case but given this AC:
`if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market not available".`

does it not mean that there will be a point in a markets lifecycle where a parent could have two childen? Even if one eventually gets cancelled if the first passes the vote first. What would this API return then?

guoguojin:
Yes, there could be a point where a market has more than one proposed children.
No, this API doesn't return proposed children, it will only return the enacted market successor line.

Do you think we need an API that will list proposed children too?

wwestgarth:
> 1. Yes, there could be a point where a market has more than one proposed children.
> 
>     2. No, this API doesn't return proposed children, it will only return the enacted market successor line.
> 
> 
> Do you think we need an API that will list proposed children too?

No idea 😄 I was more asking in case there was a bug, but I've spotted the bit in `0002_successor_markets.sql` now that handles it so its ok.

guoguojin:
I don't want to clutter the API with functions that we "think" people might want, but then never use. I've taken the view of give them something they can use to get started, and then they can tell us what they need, which will be dictated by what the business needs/wants I guess.

guoguojin:
Closing and replacing with https://github.com/vegaprotocol/vega/pull/8381 because the CI fails on pup check due to the PR branch being on a fork of the Vega repo.

PR #8381, opened 2023-05-26
Feat/8303 data node update for successor markets

### LINKED ISSUE #8303: Datanode to handle successor markets

guoguojin:
Closes #8303 

This PR replaces https://github.com/vegaprotocol/vega/pull/8375 which doesn't pass CI due to being from a forked repo and not all pipelines currently have support for forked repos.


candida-d (dismissed):

datanode/entities/proposal.go
```diff
@@ -56,6 +57,8 @@ func (p *ProposalType) String() string {
 		return "updateNetworkParameter"
 	case ProposalTypeNewFreeform:
 		return "newFreeform"
+	case ProposalTypeSuccessorMarket:
```
EVODelavega, line 60: Not sure about this one, a successor market is essentially a market proposal, just one with 2 fields not being nil. The dedicated proposal type was removed, so I don't think the datanode should differentiate between a market proposal with/without a parent market ID

datanode/gateway/graphql/new_market_resolver.go
```diff
@@ -96,3 +96,7 @@ func (r *newMarketResolver) RiskParameters(_ context.Context, obj *types.NewMark
 func (r *newMarketResolver) Metadata(_ context.Context, obj *types.NewMarket) ([]string, error) {
 	return obj.Changes.Metadata, nil
 }
+
+func (r *newMarketResolver) SuccessorConfiguration(ctx context.Context, obj *types.NewMarket) (*types.SuccessorConfiguration, error) {
+	return obj.Changes.Successor, nil
```
EVODelavega, line 101: same here, successor no longer exists

datanode/sqlstore/markets.go
```diff
@@ -233,3 +237,41 @@ func (m *Markets) GetAllPaged(ctx context.Context, marketID string, pagination e
...
+	if !fullHistory {
+		lineageFilter = "and vega_time >= (select vega_time from lineage_root)"
+	}
+
+	query := fmt.Sprintf(`
+		with lineage_root(root_id, vega_time) as (
+            select root_id, vega_time
+            from market_lineage
+            where market_id = %s
+        ), lineage(market_id, parent_id, root_id) as (
```
EVODelavega, line 259: I don't really see something here that creates a market history in a sort of doubly-linked list way. On this side of things (ie datanode) markets could (And should) have the ID of their parent and the market that succeeded them.

datanode/entities/proposal.go
guoguojin: Yes, sorry, this should have been removed after the removal of a separate successor market proposal but I missed it. Will fix.

datanode/gateway/graphql/new_market_resolver.go
guoguojin: No, this is the mapping for the protobuf SuccessorConfiguration:

```
// Configuration required to turn a new market proposal in to a successor market proposal.
message SuccessorConfiguration {
  // ID of the market that the successor should take over from.
  string parent_market_id = 1;
  // A decimal value between or equal to 0 and 1, specifying the fraction of the insurance pool balance that is carried over from the parent market to the successor.
  string insurance_pool_fraction = 2;
}
```

datanode/sqlstore/markets.go
guoguojin: The `market_lineage` table has a mapping for the market_id, it's parent_market_id, and the `root_id` which is the basically the first market in the succession line, I guess a better name for it could be lineage_id, but basically you can get the root_id for any market, and from that get the full succession line, or you can use the parent_market_id to get the child market_id, or vice versa.

```sql
CREATE TABLE IF NOT EXISTS market_lineage (
    market_id BYTEA NOT NULL PRIMARY KEY,
    parent_market_id BYTEA, -- the root market in the lineage chain will not have a parent, but all subsequent children will
    root_id BYTEA NOT NULL, -- market id of the first market in the lineage chain
    vega_time TIMESTAMP WITH TIME ZONE NOT NULL
);
```

EVODelavega: OK, so the comment doesn't apply here, but I think we should return markets with a successor_id and parent_id field, in case the market was requested without the full history, we still would want to be able to fetch the previous/next market in line easily

guoguojin: The parent_market_id is already a member of the market objects that are returned, but you're right, it would be useful to include the successor too.

EVODelavega (approved):

candida-d (approved):

PR #8386, opened 2023-05-26
Bridge snapshot fix

### LINKED ISSUE #8385: Prepare release with patched bridge addresses
Allow core to load snapshot state, but change the bridge addresses

EVODelavega:
Closes #8385 

Early review, the commit hash may be different after restore, will check and fix if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 
 ### 🐛 Fixes
 - [8251](https://github.com/vegaprotocol/vega/issues/8251) - Fix bug in expired orders optimisation resulting in non deterministic order sequence numbers
+- [8385](https://github.com/vegaprotocol/vega/issues/8385) - Add code patches to swap out addresses of `ETH` bridges.
 
 
 ## 0.71.3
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 
 ### 🐛 Fixes
 - [8251](https://github.com/vegaprotocol/vega/issues/8251) - Fix bug in expired orders optimisation resulting in non deterministic order sequence numbers
+- [8385](https://github.com/vegaprotocol/vega/issues/8385) - Add code patches to swap out addresses of `ETH` bridges.
```
gordsport, line 26: should this be under a new heading of 0.71.5 if we are going to patch?

wwestgarth:
> the commit hash may be different after restore

I think we _might_ be ok because when tendermint asks us "what is your commit hash" after we've loaded we just hand it the hash of the snapshot we loaded from, and not a re-hash of the whole state of core. We just give TM this:
```
e.lastSnapshotHash = e.snapshot.Hash // set the engine's "last snapshot hash" field to the hash of the snapshot we've just loaded
```

gordsport:
This change was deployed as a patch, so the state itself has been fixed, therefore there is not a requirement to merge this fix itself into the codebase

PR #8387, opened 2023-05-26
feat: ethereum oracles query verifier

### LINKED ISSUE #8380: [Feature]: Ethereum oracles - verify chain events and dispatch oracle data
# Feature Overview
Verify the chain event using the witness process and dispatch normalised oracle event after verification complete

ettec:
closes #8380  For review - answers some question, raises some others that need a bit more thought (the TODOs) - but something to start working with to make coordination easier.


core/evtforward/ethcall/normaliser.go
```diff
@@ -0,0 +1,59 @@
...
+	"golang.org/x/crypto/sha3"
+)
+
+type Normaliser interface {
+	Normalise(callResult []byte) (map[string]string, error)
+	ToProto() *vega.EthNormaliser
+	Hash() []byte
+}
+
+func NormaliserFromProto(proto *vega.EthNormaliser) (Normaliser, error) {
```
AnExsomnis, line 17: Seems to me that this part should be in core/types

AnExsomnis (commented):
Some initial comments

protos/sources/vega/data_source.proto
```diff
@@ -88,6 +96,24 @@ message EthTimeTrigger {
...
+
+message EthNormaliser {
+  oneof normaliser {
+    EthDecimalsNormaliser decimals_normaliser = 1;
+  }
+
+  //  This is where decimals for chain link would go? -
+}
+
+message EthDecimalsNormaliser {
```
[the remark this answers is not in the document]
ettec: same as below, it's not yet clear to me that the normaliser will be something reusable across datasource, or specifc to the particular type, so, for now at least, its eth specific.

protos/sources/vega/data_source.proto
```diff
@@ -88,6 +96,24 @@ message EthTimeTrigger {
   optional uint64 until = 3;
 }
 
+message EthFilter {
+  // Filters describes which source data are considered of interest
+  repeated vega.data.v1.Filter filters = 1;
```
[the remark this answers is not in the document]
ettec: yeah, I know these are defined at a level outside the external def spec as well,  the structure as it confuses me a little and I'm not sure if that's due to my misunderstanding, but I wanted to raise this, we should discuss and figure out

AnExsomnis (commented):
Left some more comments about the code placing.
Oracle engine should have some changes, but that can be done in the next step.

AnExsomnis (approved):

pscott31 (approved):

protos/sources/vega/data_source.proto
```diff
@@ -69,6 +69,14 @@ message EthCallSpec {
   repeated google.protobuf.Value args = 4;
   // Conditions for determining when to call the contract method.
   EthCallTrigger trigger = 5;
+  // Number of confirmations required before the query is considered verified
+  uint64 required_confirmations = 6;
+
+  // Below - is this common across all sources? so should be on DataSourceDefinitionExternal?
```
candida-d, line 75: Is this a comment directed at the core team?

protos/sources/vega/commands/v1/validator_commands.proto
```diff
@@ -89,6 +89,8 @@ message NodeVote {
     TYPE_SIGNER_THRESHOLD_SET = 11;
     // Node vote to validate a new assert governance proposal
     TYPE_GOVERNANCE_VALIDATE_ASSET = 12;
+    // Node vote for an ethereum contract call result
```
candida-d, line 92:
```suggestion
    // Node vote for an Ethereum contract call result
```

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+
+	witness           Witness
+	timeService       TimeService
+	oracleEngine      OracleDataBroadcaster
+	ethCallSpecSource EthCallSpecSource
+	ethContractCaller ethereum.ContractCaller
+	ethConfirmations  EthereumConfirmations
+
+	pendingCallEvents    []*pendingCallEvents
+	finalizedCallResults []*types.EthContractCallEvent
```
wwestgarth, line 68: `finalizedCallResults` and `hashes` will need to be in the snapshot too. Happy for you to make a new ticket for it to address later, but we really must not forget.

core/evtforward/ethcall/spec_test.go
```diff
@@ -32,7 +34,13 @@ func TestSpec(t *testing.T) {
...
 	}
 
-	originalSpec := ethcall.NewSpec(call, trigger)
+	normaliser := ethcall.EthDecimalsNormaliser{
+		Decimals: 5,
+	}
+
+	filter := ethcall.CallResultFilter{Filters: types.DataSourceSpecFilters{}}
+
+	originalSpec := ethcall.NewSpec(call, trigger, filter, normaliser)
```
wwestgarth, line 43: Do we also need to pass in a `requiredConfirmations` here to flex that field going to/from protos?

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+		return fmt.Errorf("failed confirmations check: %w", err)
+	}
+
+	return nil
+}
+
+func (s *EthereumOracleVerifier) removePendingCallEvent(id string) error {
+	for i, v := range s.pendingCallEvents {
+		if v.GetID() == id {
+			s.pendingCallEvents = s.pendingCallEvents[:i+copy(s.pendingCallEvents[i:], s.pendingCallEvents[i+1:])]
```
wwestgarth, line 186: Not sure if we want to be iterating over a slice while also removing things from it. Does `pendingCallEvents` need to be ordered? if it was a map instead and then this is just `delete(s.pendingCallEvents, id)`.

protos/sources/vega/data_source.proto
ettec: Yeah, at the others working on this feature, something we need to figure out, note this PR is to the feature branch, not to develop, this comment wouldn't make it to develop.

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (s *EthereumOracleVerifier) onCallEventVerified(event interface{}, ok bool) {
+	pv, isPendingCallEvent := event.(*pendingCallEvents)
+	if !isPendingCallEvent {
+		s.log.Error("oracle verifier received invalid event")
+	}
+
+	if err := s.removePendingCallEvent(pv.GetID()); err != nil {
+		s.log.Error("could not remove pending stake deposited event", logging.Error(err))
```
wwestgarth, line 200: If we can't remove it because its not found as pending, I shouldn't we return or something? Currently we might be adding this unexpected event we don't know about into `finalizedCallResults`.

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+			return nil
+		}
+	}
+	return fmt.Errorf("invalid pending call event hash: %s", id)
+}
+
+func (s *EthereumOracleVerifier) onCallEventVerified(event interface{}, ok bool) {
+	pv, isPendingCallEvent := event.(*pendingCallEvents)
+	if !isPendingCallEvent {
+		s.log.Error("oracle verifier received invalid event")
```
wwestgarth, line 196: Shouldn't we return here instead of continuing?

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+	} else {
+		s.log.Error("failed to verify call event")
+	}
+}
+
+func (s *EthereumOracleVerifier) OnTick(ctx context.Context, t time.Time) {
+	for _, callResult := range s.finalizedCallResults {
+		spec, err := s.ethCallSpecSource.GetSpec(callResult.SpecId)
+		if err != nil {
+			s.log.Errorf("failed to get spec for call result: %w", err)
```
wwestgarth, line 214:
```suggestion
			s.log.Error("failed to get spec for call result", logging.Error(err))
```

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+			continue
+		}
+
+		normalisedData, err := spec.Normalise(callResult.Result)
+		if err != nil {
+			s.log.Errorf("failed to normalise oracle data: %w", err)
```
wwestgarth, line 220:
```suggestion
			s.log.Error("failed to normalise oracle data", logging.Error(err))
```

core/oracles/ethereum_oracle_verifier.go
ettec: perhaps, I've based this code on what the stake verifier does, I'm guessing at this point there is little you can do, the signature of the callback doesn't have error IIRC

core/protocol/all_services.go
```diff
@@ -198,6 +201,15 @@ func newServices(
 	svcs.protocolUpgradeEngine = protocolupgrade.New(svcs.log, svcs.conf.ProtocolUpgrade, svcs.broker, svcs.topology, version.Get())
 	svcs.witness = validators.NewWitness(svcs.log, svcs.conf.Validators, svcs.topology, svcs.commander, svcs.timeService)
 
+	specEngine, err := ethcall.NewEngine(svcs.log, ethcall.Config{}, nil, svcs.eventForwarder)
+	if err != nil {
+		svcs.log.Error("unable to initialise eth call engine", logging.Error(err))
+		return nil, err
+	}
+
+	svcs.ethereumOraclesVerifier = oracles.NewEthereumOracleVerifier(svcs.log, svcs.witness, svcs.timeService, svcs.oracle,
```
wwestgarth, line 210: I think you need to add this engine to `registerTimeServiceCallbacks()` for the `OnTick()` callback to happen.

core/oracles/ethereum_oracle_verifier.go
wwestgarth: Ok I've convinced myself that `finalizedCallResults` is ok not being in the snapshot provided that the `OnTick` on the witness engine happens before this engine's `OnTick`.

I'm still unsure about `hashes` though.

core/oracles/ethereum_oracle_verifier.go
wwestgarth: If the type assertion fails `pv` will be nil and lower down `pv.callEvent` will panic.

ettec: yeah, sorry I was mixing this up with the removal failure

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -0,0 +1,268 @@
...
+	finalizedCallResults []*types.EthContractCallEvent
+
+	mu     sync.Mutex
+	hashes map[string]struct{}
+
+	// snapshot data
+	snapshotState *ethereumOracleVerifierSnapshotState
+}
+
+type pendingCallEvents struct {
```
wwestgarth, line 77: Should this be `pendingCallEvent` singular? Seems weird to have the slice in the engine named the same thing as the struct `pendingCallEvents    []*pendingCallEvents`

core/oracles/ethereum_oracle_verifier.go
wwestgarth: hmm doesn't seem to be much point in `pendingCallEvents` if all we do it add events to the slice, then remove an event for the slice, but if an event it not in there it doesn't matter anyway.

ettec: true, I guess at least you get a log line where an event is processed that was not pending :) - but yeah its feels more like debug logic

core/oracles/ethereum_oracle_verifier.go
ettec: yeah, I think probably they should be, after snapshot restores the witness checks from the pending events, with the timing issues inherent in the way the ethereum oracle data is polled, its possible that we could get a second submission on the next block for the same event, which would result in kicking off a 2nd witness process for the same event without correctly restored `hashes`

core/oracles/ethereum_oracle_verifier.go
ettec: ...but of course the other reason we have them is to restore the witness's after snapshot restore

core/oracles/ethereum_oracle_verifier.go
ettec: actually, ignore that, its already handled in the snapshot code:

```
for _, callEvent := range results {
		// this populates the id/hash structs
		if !s.ensureNotDuplicate(callEvent.Hash()) {
			s.log.Panic("pendingCallEvents's unexpectedly pre-populated when restoring from snapshot")
		}
```

ettec:
eek - I was addressing some review comments!

PR #8396, opened 2023-05-30
feat: implement iceberg orders in core

### LINKED ISSUE #8301: Implement iceberg orders
## Issue

On centralised exchanges, icebergs are a type of order that enables a trader to make an order with a relatively small visible "display quantity" and a larger hidden total size. Like an iceberg, most of the order is not visible to other traders.

After the full size of the visible portion of the order has traded away, the order is "refreshed" and reappears with its maximum display quantity.

This is repeated until the order is cancelled or expires, or its full volume trades away. Some platforms also allow the display quantity to be randomised on each refresh to further frustrate traders trying to identify the existence of the iceberg order).

### Tasks

- [ ] Review the specification for iceberg orders
- [ ] Implement the changes required ensuring the spec is aligned with implementation
- [ ] Add any additional ACs required (noting where best to have these covered unit/feature/system tests)

### Spec PR:

- https://github.com/vegaprotocol/specs/pull/1685/files

wwestgarth:
closes #8301 

Basic system test which:
- submits an iceberg
- submits another order which trades bringing the iceberg below its minimum peak, it refreshes
- submits another order which trades but iceberg remains above the minimum peak, doesn't refresh
- submits another order with takes out all of the iceberg's peak, refresh happens
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/72101/pipeline/

Know issues being fixed in https://github.com/vegaprotocol/vega/issues/8398:
- pegged-icebergs don't fully work
- cancelling an iceberg order doesn't always return potential buy/sell to 0

Also there is this ticket to weave it into spots markets, when that becomes possible:
- https://github.com/vegaprotocol/vega/issues/8399

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -110,8 +110,9 @@ type Market struct {
 
 	as common.AuctionState
 
-	peggedOrders   *common.PeggedOrders
-	expiringOrders *common.ExpiringOrders
+	peggedOrders     *common.PeggedOrders
+	expiringOrders   *common.ExpiringOrders
+	icebergRefreshes []*types.Order
 
 	// Store the previous price values so we can see what has changed
 	lastBestBidPrice *num.Uint
@@ -285,6 +286,7 @@ func NewMarket(
 		lpPriceRange:              mkt.LPPriceRange,
 		linearSlippageFactor:      mkt.LinearSlippageFactor,
 		quadraticSlippageFactor:   mkt.QuadraticSlippageFactor,
+		icebergRefreshes:          []*types.Order{},
 	}
 
 	assets, _ := mkt.GetAssets()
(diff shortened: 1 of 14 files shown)


core/execution/future/market.go
```diff
@@ -1001,6 +1006,16 @@ func (m *Market) leaveAuction(ctx context.Context, now time.Time) {
 				logging.String("OrderID", uncrossedOrder.Order.ID))
 		}
 
+		if uncrossedOrder.Order.NeedsRefreshing() {
+			icebergRefreshes = append(icebergRefreshes, uncrossedOrder.Order)
+		}
+		for _, po := range uncrossedOrder.PassiveOrdersAffected {
+			if !po.NeedsRefreshing() {
```
ze97286, line 1013: why not:
```
if po.NeedsRefreshing() {
    icebergRefreshes = append(icebergRefreshes, po)
}
```

core/matching/side.go
```diff
@@ -133,6 +133,33 @@ func (s *OrderBookSide) BestStaticPriceAndVolume() (*num.Uint, uint64, error) {
 	return num.UintZero(), 0, errors.New("no non pegged orders found on the book")
 }
 
+func (s *OrderBookSide) amendIcebergOrder(amendOrder *types.Order, oldOrder *types.Order, priceLevelIndex int, orderIndex int) (uint64, error) {
+	if amendOrder.Remaining > oldOrder.Remaining {
+		// iceberg amend should never increase the visiable remaining
```
ze97286, line 138: how come the linter is not complaining about visiable?

wwestgarth: looks like I spelt it _too_ wrong: https://github.com/client9/misspell/blob/master/words.go#L27308

PR #8416, opened 2023-06-01
feat: spots execution + liquidity SLA 

### LINKED ISSUE #8118: Spots trading
# Feature Overview

In order to trade spots need to:
1. implement collateral engine handling of a spot trade (i.e. transfer of quote and base assets accordingly)
2. implement fees transfers according to spec 
3. implement execution engine/market/snapshot for spot trading

## Specs
- [Spot trading](https://github.com/vegaprotocol/specs/blob/3ad0facad09ca18f225a2144b67d6b138ea46f7c/protocol/0080-SPOT-product_builtin_spot.md#sell-orders) 

# Acceptance Criteria
TBD

ze97286:
Closes #8118


core/execution/engine.go
```diff
@@ -250,33 +265,39 @@ func (e *Engine) RejectMarket(ctx context.Context, marketID string) error {
 		e.log.Debug("reject market", logging.MarketID(marketID))
 	}
 
-	mkt, ok := e.markets[marketID]
-	if !ok {
-		return ErrMarketDoesNotExist
-	}
+	if mkt, ok := e.markets[marketID]; ok {
```
jeremyletang, line 268: Not returning an error anymore here if the market doesn't exists?

jeremyletang: ah no doint it at the end.

core/execution/engine.go
```diff
@@ -710,26 +911,41 @@ func (e *Engine) SubmitStopOrders(
 		market = submission.RisesAbove.OrderSubmission.MarketID
 	}
 
-	mkt, ok := e.markets[market]
-	if !ok {
-		return nil, types.ErrInvalidMarketID
-	}
+	if mkt, ok := e.markets[market]; ok {
```
jeremyletang, line 914: Not a big deal, but there's a few method where the same code is duplicated for both maps of markets if not mistaken. I would have probably done something like:
```
type CommonMarket interface {
   SubmitOrder(...)
   CancelOrder(...)
   ... 
}

func (e *Engine) getMarket(id string) CommonMarket {
   if m := e.markets[id]; m != nil {
      return m
   }

   return e.spotMatkets[id]
}  
```

jeremyletang: actual looking at the rest of the file, it would seem that we do that on pretty much all method, seems like a source of future bug where we'd change things for one type of market and not the other

ze97286: it's true for many but not all cases as there's a few places related to successor market where the future call has more/different logic than the spot one. still worth doing.

core/types/snapshot.go
```diff
@@ -88,6 +88,7 @@ const (
 	PoWSnapshot                    SnapshotNamespace = "pow"
 	ProtocolUpgradeSnapshot        SnapshotNamespace = "protocolUpgradeProposals"
 	SettlementSnapshot             SnapshotNamespace = "settlement"
+	HoldingAccountTrackerSnapshot  SnapshotNamespace = "holdingaAccountTracker"
```
candida-d, line 91:
```suggestion
	HoldingAccountTrackerSnapshot  SnapshotNamespace = "holdingAccountTracker"
```

core/processor/abci.go
```diff
@@ -1554,7 +1559,24 @@ func (app *App) DeliverPropose(ctx context.Context, tx abci.Tx, deterministicID
 			app.log.Debug("unable to submit new market with liquidity submission",
 				logging.ProposalID(nm.Market().ID),
 				logging.Error(err))
-			// an error happened when submitting the market + liquidity
+			// an error happened when submitting the market
+			// we should cancel this proposal now
+			if err := app.gov.RejectProposal(ctx, toSubmit.Proposal(), types.ProposalErrorCouldNotInstantiateMarket, err); err != nil {
+				// this should never happen
+				app.log.Panic("tried to reject an non-existing proposal",
```
candida-d, line 1566:
```suggestion
				app.log.Panic("tried to reject a nonexistent proposal",
```

core/execution/common/interfaces.go
```diff
@@ -208,6 +240,11 @@ type MarketLiquidityEngine interface {
 	OnMaximumLiquidityFeeFactorLevelUpdate(num.Decimal)
 	OnStakeToCcyVolumeUpdate(stakeToCcyVolume num.Decimal)
 	StopAllLiquidityProvision(context.Context)
+	IsPoTInitialised() bool
```
candida-d, line 243: same question as above

core/execution/common/interfaces.go
```diff
@@ -174,31 +204,33 @@ type LiquidityEngine interface {
...
 	EndBlock(*num.Uint, *num.Uint, num.Decimal)
-	UpdateMarketConfig(liquidity.RiskModel, PriceMonitor, *types.LiquiditySLAParams)
+	UpdateMarketConfig(liquidity.RiskModel, liquidity.PriceMonitor, *types.LiquiditySLAParams)
 	OnNonPerformanceBondPenaltySlopeUpdate(num.Decimal)
 	OnNonPerformanceBondPenaltyMaxUpdate(num.Decimal)
 	OnMinProbabilityOfTradingLPOrdersUpdate(num.Decimal)
 	OnProbabilityOfTradingTauScalingUpdate(num.Decimal)
 	OnMaximumLiquidityFeeFactorLevelUpdate(num.Decimal)
 	OnStakeToCcyVolumeUpdate(stakeToCcyVolume num.Decimal)
+	IsPoTInitialised() bool
```
candida-d, line 222: Is this right? Looks weird, like it should be IsSpotInitialised

core/types/banking.go
```diff
@@ -140,13 +140,13 @@ func GovernanceTransferFromProto(g *checkpointpb.GovernanceTransfer) *Governance
 
 func (g *GovernanceTransfer) IntoEvent(amount *num.Uint, reason *string) *eventspb.Transfer {
 	// Not sure if this symbology gonna work for datanode
-	from := "0000000000000000000000000000000000000000000000000000000000000000"
+	from := "*"
```
jeremyletang, line 143: Is that change expected?

jeremyletang (commented):
Is there anything to be done to clear accounts on checkpoints as well?

core/types/banking.go
ze97286: doesn't seem to belong here, probably sneaked in in some merge

core/execution/spot/special_orders.go
```diff
@@ -0,0 +1,150 @@
...
+	// send new order events
+	m.broker.SendBatch(evts)
+
+	return updatedOrders
+}
+
+func (m *Market) repriceAllSpecialOrders(
+	ctx context.Context,
+	changes uint8,
+	orderUpdates []*types.Order,
```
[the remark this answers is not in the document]
ze97286: not for pegged orders I don't think

core/execution/common/interfaces.go
ze97286: what? no, it's probability of trading

core/execution/common/interfaces.go
ze97286: same answer :)

core/execution/spot/market.go
```diff
@@ -0,0 +1,3017 @@
...
+			return fmt.Errorf("party does not have sufficient balance to cover the new size")
+		}
+	}
+
+	return nil
+}
+
+// applyOrderAmendmentSizeIceberg update the orders reservedRemaining fields of an iceberg order
+// given the delta change.
+func applyOrderAmendmentSizeIceberg(order *types.Order, delta int64) {
```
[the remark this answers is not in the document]
ze97286: yes

core/execution/spot/market.go
```diff
@@ -0,0 +1,3017 @@
...
+
+func (m *Market) StopSnapshots() {
+	m.matching.StopSnapshots()
+	m.tsCalc.StopSnapshots()
+	// TODO when liquidity snapshot is merged @karel
+	// m.liquidity.StopSnapshots()
+}
+
+// applyOrderAmendment assumes the amendment have been validated before.
+func (m *Market) applyOrderAmendment(existingOrder *types.Order, amendment *types.OrderAmendment) (order *types.Order, err error) {
```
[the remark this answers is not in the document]
ze97286: yes

core/execution/spot/market.go
```diff
@@ -0,0 +1,3017 @@
...
+	m.log.Panic(
+		"invalid amend did not match any amendment combination",
+		logging.String("amended-order", amendedOrder.String()),
+		logging.String("existing-order", amendedOrder.String()),
+	)
+
+	return nil, nil, types.ErrEditNotAllowed
+}
+
+func (m *Market) validateOrderAmendment(order *types.Order, amendment *types.OrderAmendment) error {
```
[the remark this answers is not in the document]
ze97286: yes

core/execution/common/interfaces.go
candida-d: Ahh ok - sorry! Why not spell it out like it is in the ones above it?

ze97286: mostly cause it's long and I'm lazy :)

core/execution/engine.go
```diff
@@ -187,6 +197,17 @@ func NewEngine(
 	return e
 }
 
+// organises common markets copy such that it is always storing all future markets order by their slice then spot markets order by theirs.
```
wwestgarth, line 200: Whats the reason `commonMarketsCpy` it needing to be ordered like this? Given `markets` and `spotMarkets` are ordered by when they were submitted it feels nature for the common one to be like that too.

core/execution/engine.go
```diff
@@ -683,18 +851,45 @@ func (e *Engine) propagateInitialNetParamsToFutureMarket(ctx context.Context, mk
 
 func (e *Engine) removeMarket(mktID string) {
 	e.log.Debug("removing market", logging.String("id", mktID))
-
-	delete(e.markets, mktID)
-	for i, mkt := range e.marketsCpy {
+	delete(e.commonMarkets, mktID)
+	for i, mkt := range e.commonMarketsCpy {
+		m := e.commonMarketsCpy[i]
+		println(m.GetID())
```
wwestgarth, line 857: Probably want to remove the `println`

core/execution/engine.go
ze97286: because unless we save this order on its own on the snapshot it would get lost when loading from snapshot.

core/execution/engine.go
ze97286: I feel like a doctor who forgot the scissors in the stomach of the patient :)

core/execution/engine.go
wwestgarth: We should save the order in the snapshot then, and not have it dictate the code outside of snapshots. We're just going to expose ourselves to horrible bug where for whatever reason we forget in some place, somewhere, to call the regenerate function and then some snapshot will fail and we'll have to hunt all through the execution engine for it.

Its going to be a pain because we'll have to to an upgrade path thing, but at a minimum something like that guesses an upgrade has happened:
```
if len(commonMarkets) == 0 && len(spotMarkets) == 0 {
    // we've come from an old snapshot, populate common from the future's slice
}
```

or we could be fancy and do it properly and throw a version number into the snapshot and then we can know for sure we've come from an older version.

core/execution/spot/holding_account_tracker.go
```diff
@@ -0,0 +1,259 @@
...
+	le, err := hat.collateral.ReleaseFromHoldingAccount(ctx, transfer)
+	if err != nil {
+		return nil, err
+	}
+	return le, err
+}
+
+func (hat *HoldingAccountTracker) ReleaseQuantityHoldingAccount(ctx context.Context, orderID, party, asset string, quantity *num.Uint, fee *num.Uint) (*types.LedgerMovement, error) {
+	total := num.Sum(quantity, fee)
+	if fee != nil && !fee.IsZero() {
```
wwestgarth, line 119: if `fee` is allowed to be `nil` like this line suggests, we should probably also check it before doing `total := num.Sum()` above.

core/execution/spot/holding_account_tracker.go
```diff
@@ -0,0 +1,259 @@
...
+		lockedFee, ok := hat.orderIDToFee[orderID]
+		if !ok || lockedFee.LT(fee) {
+			return nil, fmt.Errorf("insufficient locked fee to release for order %s", orderID)
+		}
+		hat.orderIDToFee[orderID] = num.UintZero().Sub(lockedFee, fee)
+	}
+	lockedQuantity, ok := hat.orderIDToQuantity[orderID]
+	if !ok || lockedQuantity.LT(quantity) {
+		return nil, fmt.Errorf("insufficient locked quantity to release for order %s", orderID)
```
wwestgarth, line 128: If we fail here, we've already done `hat.orderIDToFee[orderID] = num.UintZero().Sub(lockedFee, fee)` above but won't make the transfer to actually do the release so the tracking is going to me out-of-sync with the collateral engine.

core/execution/engine.go
ze97286: I'm not convinced this is really necessary given that there are exactly 2 places where regenerate is called (new market and restore market). I also feel like there could be functional reasons to process one set of markets before the others but there is currently no such requirement.

core/execution/spot/holding_account_tracker_test.go
```diff
@@ -0,0 +1,158 @@
+package spot_test
```
wwestgarth, line 1: No snapshot tests?

core/execution/engine.go
wwestgarth: The reasons are:
1) having the snapshot rely on an assumption always goes wrong
2) there is benefit to keeping the ordering of _all_ the slices consistent, namely, ordered by when they were submitted
3) we're letting awkwardness of snapshots dictate code outside of snapshots, to cut a corner and make things easier _right now_

core/execution/engine.go
```diff
@@ -1137,6 +1290,16 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) {
 		}
 		evts = append(evts, events.NewMarketDataEvent(ctx, mkt.GetMarketData()))
 	}
+
+	for _, mkt := range e.spotMarketsCpy {
```
wwestgarth, line 1294: I guess theres not any successor market's stuff in this PR, there is a very brief mention of it in the successor market spec. We should open a ticket for it so its not forgotten.

ze97286: successor markets are not going to be supported for spots at this point. It was agreed to be pushed back.

core/execution/engine.go
ze97286: I think it all comes down to your assumption that the order of iteration must reflect the order of insertion, to which I don't see a compelling reason to so far. 

1. I don't think saving duplicate information in the snapshot is a good idea. One can argue that this is not duplicate if the creation order across types *must* be maintained, but I haven't seen an argument for why it must yet.
2. what is this benefit?
3. It's true that if there wasn't for the snapshot we'd just have it in the insertion order because it's the easy thing, but i don't see how rearranging it by market type - then insertion order is cutting a corner or making things easier _right now_. 

anyways, I concede, I'll add it to the snapshot.

core/limits/snapshot.go
```diff
@@ -41,6 +41,7 @@ func (e *Engine) serialiseLimits() ([]byte, error) {
 				CanProposeAsset:          e.canProposeAsset,
 				GenesisLoaded:            e.genesisLoaded,
 				ProposeMarketEnabled:     e.proposeMarketEnabled,
+				ProposeSpotMarketEnabled: e.proposeSpotMarketEnabled,
```
wwestgarth, line 44: I think the new field is missing from `restoreLimits()`. Could do with a unittest for this.

ze97286: yep and added.

core/execution/spot/market.go
```diff
@@ -0,0 +1,2875 @@
...
+	// Send the aggressive order into matching engine
+	confirmation, err := m.matching.SubmitOrder(order)
+	if err != nil {
+		return nil, nil, m.unregisterAndReject(ctx, order, err)
+	}
+
+	// if the order is not finished and remaining is non zero, we need to transfer the remaining base/quote from the general account
+	// to the holding account for the market/asset. If an auction is on-going we also need to account for potential fees (applicable for buy orders only)
+	if !order.IsFinished() && order.Remaining > 0 {
+		err := m.transferToHoldingAccount(ctx, order)
```
wwestgarth, line 1351: Maybe I can't see it, but if we have a pegged order submitted while we're in auction, it gets parked right away, and we won't do any holding account shuffling. But then I can't see any call to `transferToHoldingAccount` when we leave auction and the parked pegged orders are resubmitted?

core/execution/spot/market.go
```diff
@@ -0,0 +1,2875 @@
...
+		}
+	}
+	// transfer quote (potentially minus fees) to the seller
+	transfer, err = m.collateral.TransferSpot(ctx, trade.Buyer, trade.Seller, m.quoteAsset, sellerPayoutAmt)
+	if err != nil {
+		m.log.Panic("failed to complete spot transfer", logging.Trade(trade))
+	}
+	transfers = append(transfers, transfer)
+
+	// now pay fees - always from the buyer
```
witgaw, line 2695: I find this comment quite confusing, it's the aggressive party that pays the fees, not necessary the buyer. 
I get that there are difference between buyer and seller todo with base and quote asset, but that should've falsify the statement I made above.

Fees are the most confusing part of this PR for me, it looks from the tests that everything is fine and they are implemented correctly, but I struggle to convince myself of that from just reading the code alone.

witgaw (commented):
Looks good to me overall, just a bunch of nitpicks and open questions below:

core/execution/spot/market.go
ze97286: ok, do you understand the idea though? the aggressor always covers the fee, but this is about who it's taken from. If the aggressor is the buyer, then the fee is additive to the cost of the trade, i.e. buyer pays (in quote) the cost of the trade + fee. If the aggressor is a seller, then they still pay the fee, but by way of getting less of the trade's value so the fees are taken from what the buyer is supposed to pay the seller.

core/execution/spot/market.go
```diff
@@ -0,0 +1,2875 @@
...
+		m.broker.Send(endEvt)
+		m.nextMTM = m.timeService.GetTimeNow().Add(m.mtmDelta)
+	} else {
+		// revert to old mark price if we're not leaving the auction after all
+		m.markPrice = previousMarkPrice
+	}
+}
+
+// validatePeggedOrder validates pegged order.
+func (m *Market) validatePeggedOrder(order *types.Order) types.OrderError {
```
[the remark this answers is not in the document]
ze97286: yes, it needs to be removed from here.

core/execution/spot/holding_account_tracker.go
ze97286: I'll remove the nil check for now and verify that it can't actually be.

core/execution/engine.go
```diff
@@ -61,6 +62,13 @@ type Engine struct {
 
 	markets    map[string]*future.Market
 	marketsCpy []*future.Market
+
+	spotMarkets    map[string]*spot.Market
+	spotMarketsCpy []*spot.Market
+
+	commonMarkets    map[string]common.CommonMarket
+	commonMarketsCpy []common.CommonMarket
```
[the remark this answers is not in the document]
ze97286: I'm fine with that

core/execution/engine.go
```diff
@@ -250,19 +274,22 @@ func (e *Engine) RejectMarket(ctx context.Context, marketID string) error {
...
 	if err := mkt.Reject(ctx); err != nil {
 		return err
 	}
 
 	// send market data event so market data and markets API are consistent.
 	e.broker.Send(events.NewMarketDataEvent(ctx, mkt.GetMarketData()))
-
 	e.removeMarket(marketID)
+
+	if _, ok := e.markets[marketID]; !ok {
```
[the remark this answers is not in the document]
ze97286: no. this is checking if this is a futures market, and if so, follows up with the successor market stuff, so this is an early return for a spot market.

core/execution/engine.go
```diff
@@ -451,21 +476,55 @@ func (e *Engine) submitOrRestoreMarket(ctx context.Context, marketConfig *types.
...
 	}
 
 	// keep state in pending, opening auction is triggered when proposal is enacted
 	mkt := e.markets[marketConfig.ID]
+	e.publishNewMarketInfos(ctx, mkt.GetMarketData(), *mkt.Mkt())
+	return nil
+}
 
-	e.publishNewMarketInfos(ctx, mkt)
+func (e *Engine) submitOrRestorSpotMarket(ctx context.Context, marketConfig *types.Market, proposer string, isNewMarket bool, oos time.Time) error {
```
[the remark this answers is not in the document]
ze97286: can't even spell.

core/execution/engine.go
```diff
@@ -451,21 +476,55 @@ func (e *Engine) submitOrRestoreMarket(ctx context.Context, marketConfig *types.
...
+	if err := e.submitSpotMarket(ctx, marketConfig, oos); err != nil {
+		return err
+	}
+
+	if isNewMarket {
+		assets, err := marketConfig.GetAssets()
+		if err != nil {
+			e.log.Panic("failed to get asset from market config", logging.String("market", marketConfig.ID), logging.String("error", err.Error()))
+		}
+		e.marketActivityTracker.MarketProposed(assets[1], marketConfig.ID, proposer)
```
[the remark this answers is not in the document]
ze97286: ok

core/execution/engine.go
```diff
@@ -923,69 +1086,67 @@ func (e *Engine) CancelOrder(
...
+		}
+		metrics.OrderGaugeAdd(-confirmed, market)
+		return confirmations, nil
 	}
-	confirmations, err := mkt.CancelAllOrders(ctx, party)
-	if err != nil {
+	return nil, types.ErrInvalidMarketID
+}
+
+func (e *Engine) cancelAllPartyOrdersForMarket(ctx context.Context, ID, party string, f func(context.Context, string) ([]*types.OrderCancellationConfirmation, error)) ([]*types.OrderCancellationConfirmation, error) {
```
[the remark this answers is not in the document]
ze97286: it's used. what makes you wonder?

core/execution/spot/market_snapshot.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	liquidity.SetGetStaticPricesFunc(market.getBestStaticPricesDecimal)
+	for _, p := range em.Parties {
+		market.parties[p] = struct{}{}
+	}
+
+	if em.Closed {
+		market.closed = true
+		stateVarEngine.UnregisterStateVariable(baseAsset+"_"+quoteAsset, mkt.ID)
+	}
```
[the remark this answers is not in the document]
ze97286: good catch.

core/integration/features/spots/SpotFunctionalTests/TC-0012-Spot-Infra-FeeCalculation.feature
```diff
@@ -0,0 +1,79 @@
...
+    Then "party2" should have holding account balance of "0" for asset "BTC"
+    Then "party1" should have holding account balance of "0" for asset "ETH"
+
+    Then the parties should have the following account balances:
+      | party  | asset | market id | general |
+      | party1 | ETH   | BTC/ETH   | 9999000 |
+      | party1 | BTC   | BTC/ETH   | 0       |
+      | party2 | ETH   | BTC/ETH   | 1000    |
+      | party2 | BTC   | BTC/ETH   | 100000  |
+
```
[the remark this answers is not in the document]
ze97286: it needs to be covered in FTs as well for sure. I believe @Jiajia-Cui was starting on that. Not sure where it's at just now.

core/integration/features/spot/accounts_and_fees/trading_fees_spot.feature
```diff
@@ -0,0 +1,172 @@
...
+
+    Then the market data for the market "BTC/ETH" should be:
+      | mark price | trading mode            | auction trigger             |
+      | 1500       | TRADING_MODE_CONTINUOUS | AUCTION_TRIGGER_UNSPECIFIED |
+
+    Then the following transfers should happen:
+      | from   | to     | from account            | to account                       | market id | amount | asset |
+      | party1 | market | ACCOUNT_TYPE_GENERAL    | ACCOUNT_TYPE_FEES_MAKER          | BTC/ETH   | 10     | ETH   |
+      | party1 | market | ACCOUNT_TYPE_GENERAL    | ACCOUNT_TYPE_FEES_INFRASTRUCTURE | BTC/ETH   | 30     | ETH   |
+      | market | party1 | ACCOUNT_TYPE_FEES_MAKER | ACCOUNT_TYPE_GENERAL             | BTC/ETH   | 10     | ETH   |
```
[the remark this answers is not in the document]
ze97286: it's just because of the mechanics I explained to your other general misunderstanding about fees. It's a bit ugly but it's how it's implemented. 

party 2 is a seller, they are the aggressor. They don't have nor expected to have a general account in the quote balance. Party 1 is the buyer, they're not *really* paying the fees. Lets say that cost of the trade is 1000, and the fees on that are 10, then party1 pays 9990 to party2, then pays 10 to the fees. Party1 paid exactly how much the trade cost and paid not fees, and party 2, the aggressor got 9990, i.e. paid 10 in fees. 

as party1 was the maker on this trade so they get the maker fee back. 

I guess this could be made less ugly by just not making the make fee transfer, but this will also be slightly confusing so not sure what's better.

core/execution/engine.go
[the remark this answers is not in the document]
witgaw: It's fine, it's crypto, nobody can.

core/execution/engine.go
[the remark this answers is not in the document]
witgaw: It's fine, it's crypto, nobody can.

core/execution/engine.go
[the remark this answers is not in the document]
ze97286: this will be done some day. not just now. reference: https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1687432743445819

core/execution/engine.go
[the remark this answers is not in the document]
witgaw: Just the fact that something like that seems to work perfectly fine:
```
func (e *Engine) cancelAllPartyOrdersForMarket(ctx context.Context, mkt common.CommonMarket, party string) ([]*types.OrderCancellationConfirmation, error) {
	confs, err := mkt.CancelAllOrders(ctx, party)
	if err != nil && err != common.ErrTradingNotAllowed {
		return nil, err
	}
	var confirmed int
	for _, conf := range confs {
		if conf.Order.Status == types.OrderStatusCancelled {
			confirmed++
		}
	}
	metrics.OrderGaugeAdd(-confirmed, mkt.GetID())
	return confs, nil
}

func (e *Engine) cancelAllPartyOrders(ctx context.Context, party string) ([]*types.OrderCancellationConfirmation, error) {
	confirmations := []*types.OrderCancellationConfirmation{}

	for _, mkt := range e.commonMarketsCpy {
		confs, err := e.cancelAllPartyOrdersForMarket(ctx, mkt, party)
		if err != nil && err != common.ErrTradingNotAllowed {
			return nil, err
		}
		confirmations = append(confirmations, confs...)
	}
	return confirmations, nil
}
```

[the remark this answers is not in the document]
ze97286: yes, 100%, you should have been more clear, there's no way I could have understood that from the original comment...

core/integration/features/spot/accounts_and_fees/trading_fees_spot.feature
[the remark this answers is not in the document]
witgaw: But what's stopping from first transferring everything to party2 (10ETH for 1 BTC) and then paying their fees (0.4ETH) from that account? That'd keep the transfers clean and probably save many raised eyebrows and user queries in the future?

core/execution/spot/market.go
witgaw: Yes I get the idea, but it's just very difficult for me to work it out from the code. What you explained above would make for a good comment in the code I think.

core/integration/features/spot/accounts_and_fees/trading_fees_spot.feature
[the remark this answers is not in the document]
cdummett: I agree with Witolds suggestions if its an easy change 👍  I think cleaner and a little easier to trace back, I'm not sure what TransferType that first transfer of 10  from party1 to party2 is using but it's probably misleading right?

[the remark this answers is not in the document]
ze97286: ok, I'm perfectly fine with making the full payment from the buyer to the seller and then making the fee payment from the seller in this case. @cdummett this has not functional meaning, but it affects what the ledger looks like, i.e. as in the PR right now it looks like:
Given the seller (party2) is the aggressor and the buyer is party1 (only the quote asset transfers):
* party1 transfers the trade amount - fees to party2 
* party1 transfers the fees to the fees accounts

as opposed to what's @witgaw suggesting which is:
* party1 transfers the trade amount to party2
* party2 transfers the fees to the fee accounts

I think questions gonna be asked either way but I'm happy to change this to the latter.

ze97286:
> Is there anything to be done to clear accounts on checkpoints as well?

yes.

core/execution/spot/holding_account_tracker_test.go
ze97286: added

core/execution/spot/market.go
ze97286: this should be fixed now.

wwestgarth (dismissed):

candida-d (dismissed):

witgaw (dismissed):

commands/liquidity_provision_submission.go
```diff
@@ -73,117 +71,6 @@ func checkLiquidityProvisionSubmission(cmd *commandspb.LiquidityProvisionSubmiss
 			errs.AddForProperty("liquidity_provision_submission.fee", ErrMustBePositive)
 		}
 	}
-
```
jeremyletang, line 76: Should  we have an explicit error returned if the users have been setting the shape? Probably not so useful, just to at least notify users that the shapes have been deprecated?

core/execution/common/liquidity_provision.go
```diff
@@ -396,10 +400,10 @@ func (m *MarketLiquidity) makePerPartyAccountsAndTransfers(ctx context.Context,
...
 	}
 
-	_, err = m.collateral.CreatePartyMarginAccount(ctx, party, m.marketID, m.asset)
-	if err != nil {
-		return err
-	}
+	// _, err = m.collateral.CreatePartyMarginAccount(ctx, party, m.marketID, m.asset)
+	// if err != nil {
+	// 	return err
+	// }
```
jeremyletang, line 406: ?

core/execution/engine_snapshot.go
```diff
@@ -52,6 +55,105 @@ func (e *Engine) marketsStates() ([]*types.ExecMarket, []types.StateProvider) {
...
+				logging.MarketID(marketConfig.ID),
+				logging.String("asset-id", asset),
+				logging.Error(err),
+			)
+			return nil, err
+		}
+		assetDetatils = append(assetDetatils, ad)
+	}
+
+	nextMTM := time.Unix(0, em.NextMTM)
```
jeremyletang, line 112: any chance for nextMTM to be the zero time? That would be an issue to serialise it like this then.

core/collateral/checkpoint.go
```diff
@@ -130,11 +130,12 @@ func (e *Engine) getCheckpointBalances() []*checkpoint.AssetBalance {
 		}
 		switch acc.Type {
```
jeremyletang, line 131: Shouldn't the holding account be added here somehow?

core/execution/common/liquidity_provision.go
ze97286: this should be here. @karlem left it in here. I'm leaving it for him to choose to remove or do whatever is right. Uncomment is definitely not what's right.

jeremyletang: sure I was just wondering why was it commented etc.

commands/liquidity_provision_submission.go
ze97286: well, for spot there is one, not at this level. In here we should add one on @karlem PR when deprecating the shapes, but as of this PR they're still valid for futures.

core/execution/engine_snapshot.go
ze97286: indeed.

core/execution/common/liquidity_provision.go
ze97286: because it is there, and shouldn't have been :)

core/collateral/checkpoint.go
ze97286: it should

core/execution/engine_snapshot.go
ze97286: what do you mean?

commands/liquidity_provision_submission.go
jeremyletang: hum, fair point.

jeremyletang (approved):

witgaw (approved):

PR #8432, opened 2023-06-02
chore: improve ledger csv export

### LINKED ISSUE #8353: Update ledger entry CSV export query to make clearer and add the maker rebate
## Issue

During initial testing of the first mainnet market it has been identified that we need to make some updates to the ledger entry CSV export:

Required for 0.71.x patch:

- remove the `\x` from the IDs in the CSV
- add the following missing fields to the CSV:
  - MTM gains and losses
  - Any original deposits of assets
  - maker rebate (thats not shown)
- update the account types as these are enums and ideally should be user readable as account types

Other points to consider:

- **NICE TO HAVE (possible new ticket to be raised - if not simple/low risk for the patch)**: Its not particularly easy to be able to work out the in/out accounting for auditing purposes. Can any enhancements be made to help users do their accounting e.g., positive and negative numbers so summing is easier.  
  -  can we have a column e.g., `net` that shows positive/negative values or 0 if the transfer is between accounts on the same public key

pscott31:
closes #8353


zale144 (approved):
lgtm

PR #8447, opened 2023-06-07
greatly simplify data source type marshalling

pscott31:
not done but gives a good idea


protos/sources/vega/data_source.proto
```diff
@@ -75,9 +75,9 @@ message EthCallSpec {
 
   // Below - is this common across all sources? so should be on DataSourceDefinitionExternal?
   // Filters describes which source data are considered of interest
-  EthFilter filter = 7;
+  repeated vega.data.v1.Filter filters = 7;
 
-  EthNormaliser normaliser = 8;
+  // EthNormaliser normaliser = 8;
```
candida-d, line 80: Could you expand on this one a bit more?

pscott31: don't worry this is a work in progress feature branch, this will get improved shortly :)

AnExsomnis (approved):

PR #8455, opened 2023-06-07
Add block parameter for epoch query

### LINKED ISSUE #7592: Add `blockHeight` parameter to `epoch` query
# Feature Overview

**In order to** know what epoch a block was in
**We will** add a `blockHeight` parameter to the `epoch` query
**So that** there is a way to correlate a block with an epoch

In GraphQL there is a `epoch(id:"string") {` and in REST there is [Get Epoch ](https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-get-epoch). You can get the current Block Height from the statistics. But there is no way to know if that block is in a specified epoch, other than to fetch the `firstBlock` and `lastBlock` from every epoch, then compare them to your block height.

This ticket is for an update to the epoch query, to add a new parameter, `blockheight`. If specified for a block in the past, it will return the epoch for that block height.

# Acceptance Criteria

- It is possible to call an API with a block number (e.g. `10`) and know that it was in epoch `1`
- If the block height is 100, and the user requests the epoch for block height 200, return an error. 

# Notes
- We could *guess* which epoch a future block is in, but I don't think there is any benefit at this point.

zale144:
Resolve #7592


datanode/api/trading_data_v2.go
```diff
@@ -2434,10 +2433,12 @@ func (t *TradingDataServiceV2) GetEpoch(ctx context.Context, req *v2.GetEpochReq
 		epoch entities.Epoch
 		err   error
 	)
-	if req.GetId() == 0 {
-		epoch, err = t.epochService.GetCurrent(ctx)
+	if req.GetId() > 0 {
```
wwestgarth, line 2436: Should we just make it an error if a block and an ID are given at the same time?

zale144: We could. Right now we just ignore the block if id was provided. I guess returning an error would let the client know that they don't have to bother with both params. On the other hand it's not really hurting anyone

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2642,6 +2642,8 @@ message NodeSignaturesConnection {
 message GetEpochRequest {
   // Epoch ID to get, if provided. If omitted, data for the current epoch is returned.
   optional uint64 id = 1;
+  // Block Height that is within the range of the epoch to get, if provided. If omitted, data for the current epoch is returned.
```
candida-d, line 2645:
```suggestion
  // Block height. If provided, returns the epoch that the given block is in. If omitted, data for the current epoch is returned.
```

datanode/api/trading_data_v2.go
wwestgarth: Sure. Can you make sure that behaviour is documented in the protos docstring somewhere.

guoguojin (dismissed):

candida-d (dismissed):

wwestgarth (approved):

candida-d (approved):

jeremyletang (approved):

jeremyletang (approved):

PR #8480, opened 2023-06-09
feat: test margin for order type GTT/GTC/GFN

Jiajia-Cui:
Create feature test to cover the following ACs which are better covered in feature test rather than system test. 

- Limit GTT (0011-MARA-007)
- Limit GTC (0011-MARA-008)
- Limit GFN (0011-MARA-009)

### DIFF
--- a/core/integration/features/verified/0011-MARA-006.feature
+++ b/core/integration/features/verified/0011-MARA-006.feature
@@ -17,6 +17,7 @@ Feature: check margin account with partially filled order
       | name                                    | value |
       | market.auction.minimumDuration          | 1     |
       | network.markPriceUpdateMaximumFrequency | 0s    |
+  And the average block duration is "1"
 
   Scenario: 001 If an order is partially filled and if this leads to a reduced position and reduced riskiest long / short then the margin requirements are seen to be reduced and if margin balance is above release level then the excess amount is transferred to the general account.0011-MARA-006. 0011-MARA-008
     Given the parties deposit on asset's general account the following amount:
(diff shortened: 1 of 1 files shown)


vega-paul (dismissed):

wwestgarth (dismissed):

core/integration/features/verified/0011-MARA-006.feature
```diff
@@ -98,3 +97,298 @@ Feature: check margin account with partially filled order
...
+    Then the network moves ahead "7" blocks
+    And the parties should have the following margin levels:
+      | party   | market id | maintenance | search | initial | release |
+      | trader4 | ETH/DEC20 | 601         | 721    | 901     | 1202    |
+
+    Then the parties should have the following account balances:
+      | party   | asset | market id | margin | general |
+      | trader4 | USD   | ETH/DEC20 | 901    | 89099   |
+
+  Scenario: 003 check margin for GFN order type 0011-MARA-009
```
witgaw, line 255: Wouldn't it make sense to trigger an auction and check that the order gets cancelled and margin level is updated? Not much to test in continuous trading I think.

core/integration/features/verified/0011-MARA-006.feature
```diff
@@ -98,3 +97,298 @@ Feature: check margin account with partially filled order
...
+
+    And the parties should have the following margin levels:
+      | party   | market id | maintenance | search | initial | release |
+      | trader2 | ETH/DEC20 | 33636       | 40363  | 50454   | 67272   |
+
+    Then the parties should have the following account balances:
+      | party   | asset | market id | margin | general |
+      | trader2 | USD   | ETH/DEC20 | 50454  | 39546   |
+
+Scenario: 004 check margin for GTT order type.0011-MARA-007
```
witgaw, line 328: Is this scenario really necessary? I don't see much value over scenario 002. Sure, the GTT order in one market expires and the GTC one in another market doesn't, but does it really give us much?

witgaw (commented):
Please see my comments below. 

Also, there's no point adding the "reference" field to every order (especially not with a duplicated reference), unless you plan on amending or cancelling any of them further down the line in the scenario.

witgaw (approved):

PR #8504, opened 2023-06-13
Liquidity SLA integrate to spot

### LINKED ISSUE #8368: Integrate the new liquidity SLA with the product spot
As the new product type spot will only be developed to work with the new liquidity mechanism (Liquidity SLA) we need to integrate sla liquidity into spot.

karlem:
close #8368 .


core/liquidity/v2/sla.go
```diff
@@ -1,14 +1,18 @@
...
 
+	"code.vegaprotocol.io/vega/core/types"
 	"code.vegaprotocol.io/vega/libs/crypto"
 	"code.vegaprotocol.io/vega/libs/num"
 )
 
+const maximumNumberOfHistoricalPenalties = 100000
+
 // TODO karel - real Tendermint txs should be used.
 type TX struct {
```
ze97286, line 17: lets get rid of it.

core/liquidity/v2/sla.go
```diff
@@ -35,14 +49,14 @@ func (e *Engine) BeginBlock(txs []TX) {
 	e.kSla = e.GenerateKSla(txs)
 }
 
-func (e *Engine) TxProcessed(txCount int) {
+func (e *Engine) TxProcessed(txCount int, markPrice *num.Uint, positionFactor num.Decimal) {
 	// Check if the k transaction has been processed
-	if e.kSla != txCount {
+	if e.kSla != txCount && e.auctionState.IsOpeningAuction() {
```
ze97286, line 54: let's remove the randomness/k/ txProcessed and just do it at the end of block as agreed.

core/liquidity/v2/sla.go
```diff
@@ -1,14 +1,18 @@
...
 	"encoding/binary"
 	"math/rand"
 	"time"
 
+	"code.vegaprotocol.io/vega/core/types"
 	"code.vegaprotocol.io/vega/libs/crypto"
 	"code.vegaprotocol.io/vega/libs/num"
 )
 
+const maximumNumberOfHistoricalPenalties = 100000
```
ze97286, line 14: what's the point in keeping `maximumNumberOfHistoricalPenalties` where the maximum number of epochs ever used is performanceHysteresisEpochs, in which case you can just keep a cyclic buffer with len performanceHysteresisEpochs as you just always calculate the mean of it.

core/liquidity/v2/sla.go
```diff
@@ -80,26 +96,41 @@ func (e *Engine) CalculateSLAPenalties(now time.Time) {
...
+		AllPartiesHaveFullFeePenalty: partiesWithFullFeePenaltyCount == len(penaltiesPerParty),
+		PenaltiesPerParty:            penaltiesPerParty,
 	}
 
-	e.slaPenalties = penaltiesPerParty
+	return e.slaPenalties
 }
 
-func (e *Engine) GetSLAPenalties() map[string]*SlaPenalty {
+func (e *Engine) GetSLAPenalties() SlaPenalties {
```
ze97286, line 133: commenting here for line 137 - getMidPrice() - that doesn't seem like something the SLA should be doing. It can be passed to it where needed.

core/liquidity/v2/sla.go
```diff
@@ -154,21 +189,28 @@ func (e *Engine) doesLPMeetsCommitment(party string) bool {
 		maxPrice = one.Add(e.priceRange).Mul(midD)
 	}
 
-	notionalVolume := num.DecimalZero()
+	notionalVolumeBuys := num.DecimalZero()
```
ze97286, line 192: commenting on line 171/2: couldn't e.orderBook.GetLastTradedPrice() be nil?

separately, should the sla engine interact with the orderbook directly?

ze97286 (commented):
GenerateKSla - I think it's sufficient there to take just the block hash and number of transactions if you want to support the randomness or get rid of the whole thing as discussed.

core/liquidity/v2/sla.go
```diff
@@ -123,7 +154,11 @@ func (e *Engine) getMidPrice() (*num.Uint, error) {
 	return midPrice, nil
 }
 
-func (e *Engine) doesLPMeetsCommitment(party string) bool {
+func (e *Engine) doesLPMeetsCommitment(
+	party string,
+	markPrice *num.Uint,
+	positionFactor num.Decimal,
+) bool {
```
ze97286, line 161: re `one.Sub(e.priceRange)` and `one.Add(e.priceRange)` consider keeping these instead of the net param to avoid recalculating every block.

core/liquidity/v2/sla.go
karlem: The point is that the performanceHysteresisEpochs is a parameter that can challenge any time to larger number. And this is the max limit.

ze97286: sure, it can change, then you increase and decrease the size of the slice, so what?

karlem: Wait. What if we keep 5 items in slice but the period increases for 30 then you miss the previous data.

ze97286: then you will be using 30 periods once you have 30 periods. I think that's sensible. It makes less sense to me to keep 100000 entries just for the odd chance that maybe at some point the community will change it from 5 to 100000.

karlem: Well. As long as research is aligned with it. This is implemented based on their input.

ze97286: the spec doesn't say what is the valid range of this *market* parameter. What value do we expect it to have? @davidsiska-vega. I don't know. In my mind it makes sense to keep a circular buffer of size `performanceHysteresisEpochs` rather than a buffer of size up to 100k just for the case someone ever updates the market to increase the size. Which is another point - have these params been added to the spot market proposal?

core/liquidity/v2/sla.go
```diff
@@ -80,26 +96,41 @@ func (e *Engine) CalculateSLAPenalties(now time.Time) {
...
 			Bond: bondPenalty,
 			Fee:  e.calculateHysteresisFeePenalty(feePenalty, previousPenalties),
 		}
+
+		if penaltiesPerParty[party].Fee.Equal(one) {
+			partiesWithFullFeePenaltyCount++
+		}
+	}
+
+	e.slaPenalties = SlaPenalties{
```
ze97286, line 125: do we actually need to store it for future use? I don't see where it's used beyond as the return value. There's an API below that returns it but nothing is currently using it so not sure.

core/liquidity/v2/sla.go
karlem: I honestly have no problem (design wise) with the engine getting data from orderbook. As long as it protected by interface and only get data - not creates orders.

core/liquidity/v2/engine.go
```diff
@@ -225,11 +226,54 @@ func (e *Engine) SubmitLiquidityProvision(
...
+	}
+
+	provision.Status = types.LiquidityProvisionStatusPending
+	e.pendingProvisions[party] = provision
+	return false, nil
+}
+
+func (e *Engine) ApplyPendingProvisions(ctx context.Context, now time.Time) map[string]*types.LiquidityProvision {
+	updatedProvisionsPerParty := make(map[string]*types.LiquidityProvision, len(e.pendingProvisions))
+	for party, provision := range e.pendingProvisions {
```
ze97286, line 246: should probably sort this by party first so that the events are sent in deterministic order

core/execution/common/liquidity_provision.go
```diff
@@ -0,0 +1,671 @@
...
+	liquidityEngine       LiquidityEngine
+	collateral            Collateral
+	broker                Broker
+	orderBook             liquidity.OrderBook
+	equityShares          EquityLikeShares
+	marketActivityTracker *MarketActivityTracker
+	fee                   *fee.Engine
+
+	marketID        string
+	settlementAsset string
```
ze97286, line 46: can we just call it asset or something?

core/execution/common/liquidity_provision.go
```diff
@@ -0,0 +1,671 @@
...
+	marketState types.MarketState,
+) error {
+	m.idGen = idgeneration.New(deterministicID)
+	defer func() { m.idGen = nil }()
+
+	if !m.canSubmitCommitment(marketState) {
+		return ErrCommitmentSubmissionNotAllowed
+	}
+
+	if err := m.liquidityEngine.ValidateLiquidityProvisionAmendment(lpa); err != nil {
```
ze97286, line 469: this isn't implemented by v2 is it?

core/liquidity/v2/engine.go
karlem: Defo

core/execution/common/liquidity_provision.go
karlem: The function is implemented in V2 engine. Or what do you mean?

ze97286 (dismissed):

protos/sources/vega/vega.proto
```diff
@@ -872,6 +878,18 @@ enum TransferType {
...
+  // Allocate liquidity fee among liquidity providers to their network controlled liquidity fee account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_ALLOCATE = 30;
+  // Distributes net fee from liquidity provider's fee account to their general account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_NET_DISTRIBUTE = 31;
+  // Applies SLA penalty by moving funds from party bond account to market insurance pool.
+  TRANSFER_TYPE_SLA_PENALTY_BOND_APPLY = 32;
+  // Applies SLA penalty by moving funds from the liquidity provider's fee account to market insurance pool.
+  TRANSFER_TYPE_SLA_PENALTY_LP_FEE_APPLY = 33;
+  // Collects unpaid rewards from the liquidity provider's fee account to the market bonus distribution account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_UNPAID_COLLECT = 34;
```
candida-d, line 890: I think the name for this could be clearer, I'm not sure if unpaid_collect explains what's going on here.

protos/sources/vega/vega.proto
```diff
@@ -872,6 +878,18 @@ enum TransferType {
...
   TRANSFER_TYPE_SUCCESSOR_INSURANCE_FRACTION = 29;
+  // Allocate liquidity fee among liquidity providers to their network controlled liquidity fee account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_ALLOCATE = 30;
+  // Distributes net fee from liquidity provider's fee account to their general account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_NET_DISTRIBUTE = 31;
+  // Applies SLA penalty by moving funds from party bond account to market insurance pool.
+  TRANSFER_TYPE_SLA_PENALTY_BOND_APPLY = 32;
+  // Applies SLA penalty by moving funds from the liquidity provider's fee account to market insurance pool.
+  TRANSFER_TYPE_SLA_PENALTY_LP_FEE_APPLY = 33;
+  // Collects unpaid rewards from the liquidity provider's fee account to the market bonus distribution account.
```
candida-d, line 889:
```suggestion
  // Collects penalties from the liquidity provider's fee account before the fee revenue is paid, and transfers it to the market's bonus distribution account.
```

candida-d: Trying out this description based on our chat

protos/sources/vega/vega.proto
```diff
@@ -872,6 +878,18 @@ enum TransferType {
...
+  TRANSFER_TYPE_LIQUIDITY_FEE_UNPAID_COLLECT = 34;
+  // Distributes performance bonus from market bonus to liquidity provider's general account.
```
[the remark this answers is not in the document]
karlem: Should be "market bonus account"

protos/sources/vega/vega.proto
```diff
@@ -872,6 +878,18 @@ enum TransferType {
...
+  TRANSFER_TYPE_LIQUIDITY_FEE_ALLOCATE = 30;
+  // Distributes net fee earnings from liquidity provider's fee account to their general account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_NET_DISTRIBUTE = 31;
+  // Applies SLA penalty by moving funds from party's bond account to market's insurance pool.
+  TRANSFER_TYPE_SLA_PENALTY_BOND_APPLY = 32;
+  // Applies SLA penalty by moving funds from the liquidity provider's fee account to market insurance pool.
+  TRANSFER_TYPE_SLA_PENALTY_LP_FEE_APPLY = 33;
+  // Collects penalties from the liquidity provider's fee account before the fee revenue is paid, and transfers it to the market's bonus distribution account.
+  TRANSFER_TYPE_LIQUIDITY_FEE_UNPAID_COLLECT = 34;
+  // Distributes performance bonus from market bonus to liquidity provider's general account.
```
candida-d, line 891:
```suggestion
  // Distributes performance bonus from market bonus account to liquidity provider's general account.
```

candida-d (approved):

ze97286 (approved):

PR #8534, opened 2023-06-19
chore: rename iceberg proto fields

### LINKED ISSUE #8524: Rename initial and minimum peak for icebergs in API
For the iceberg order API fields, term 'initial peak' is a little misleading as it's the actual peak amount of the iceberg, where initial implies it's likely to be changed, and 'minimum peak' doesn't actually make much sense language-wise.

@wwestgarth and I propose we change those field names:
`initial_peak_size` would become `peak_size`
`minimum_peak_size` would become `minimum_visible_size`

wwestgarth:
closes #8524 

This renames the following iceberg order fields:
- `initial_peak_size -> peak_size`
- `minimum_peak_size -> minimum_visible_size`
this is to make the use of the work "peak" fit with its proper meaning in english.

Note that I am **not** marking this as breaking-change.

This is because iceberg fields have not been in a release, and are not on fairground. I have still included the rename in the database migration files to help with deployment onto stagnet1. I'm imagining we will squash it down into one migrations file before release at which point the rename will not exist?

Updated system tests still working:
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/76171/pipeline


jeremyletang (dismissed):

protos/sources/vega/vega.proto
```diff
@@ -176,10 +176,11 @@ message PeggedOrder {
...
 // Details of an iceberg order
 message IcebergOrder {
-  // Size of the order that will be made visible if the iceberg order is refreshed after a trade has occurred.
-  uint64 initial_peak_size = 1;
-  // Threshold at which the order's visible remaining size will be refreshed back to its initial peak size.
-  uint64 minimum_peak_size = 2;
+  // Size of the order that will be made visible if the iceberg order is replensished after trading.
+  uint64 peak_size = 1;
+  // If the visible size of the order falls below this value it will be replensished back to the peak size using the 
+  // reserved amount back to the peak size.
```
candida-d, line 182:
```suggestion
  // If the visible size of the order falls below this value, it will be replenished back to the peak size using the reserved amount.
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -27,10 +27,10 @@ type PeggedOrder {
 
 "Details of the iceberg order"
 type IcebergOrder {
-  "Size of the order that will be made visible if the iceberg order is refreshed after a trade has occurred"
-  initialPeakSize: String!
-  "Threshold at which the order's visible remaining size will be refreshed back to its initial peak size"
-  minimumPeakSize: String!
+  "Size of the order that will be made visible if the iceberg order is replensished after trading"
+  peakSize: String!
+  "If the visible size of the order falls below this value it will be replensished back to the peak size using the reserved amount back to the peak size"
```
candida-d, line 32:
```suggestion
  "Size of the order that will be made visible if the iceberg order is replenished after trading"
  peakSize: String!
  "If the visible size of the order falls below this value, it will be replenished back to the peak size using the reserved amount"
```

candida-d (changes requested):

candida-d (dismissed):

jeremyletang (approved):

candida-d (approved):

PR #8555, opened 2023-06-20
Add notional volume to candles

### LINKED ISSUE #8409: [Feature]: add notional value of candle volume
# Feature Overview

**In order to** show the notional value of last 24h traded volume
**We will** provide the notional traded volume for a candle
**So that** Its possible for apps to get the notional traded volume without tracking all trades on a market

Feature ticket for console: https://github.com/vegaprotocol/frontend-monorepo/issues/3982

zale144:
Closes #8409


protos/sources/vega/vega.proto
```diff
@@ -534,8 +534,10 @@ message Candle {
   string close = 6;
   // Total trading volume during the candle interval.
   uint64 volume = 7;
+  // Total notional value (size * price) traded during the candle interval.
+  uint64 notional = 8;
```
wwestgarth, line 538: We need to add new fields at the end with a new number, otherwise it'll be a breaking change.

wwestgarth:
Can we add a test to `sqlstore/candles_test.go` which checks that the notional is correct after pushing in some trades?

zale144:
> Can we add a test to `sqlstore/candles_test.go` which checks that the notional is correct after pushing in some trades?

Yes I keep trying to figure out a way to get the expected notional value but I'm kind of lost there

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1587,6 +1587,8 @@ message Candle {
   string close = 6;
   // Total trading volume during the candle interval.
   uint64 volume = 7;
+  // Total notional value (size * price) traded during the candle interval.
```
candida-d, line 1590:
```suggestion
  // Total notional value traded during the candle interval.
```

protos/sources/vega/vega.proto
```diff
@@ -541,6 +541,8 @@ message Candle {
   uint64 volume = 7;
   // Time interval for the candle.
   Interval interval = 8;
+  // Total notional value (size * price) traded during the candle interval.
```
candida-d, line 544:
```suggestion
  // Total notional value traded during the candle interval.
```

candida-d (commented):
Two small changes to the proto docs

wwestgarth (approved):

candida-d (approved):

PR #8569, opened 2023-06-22
Update protos and core

### LINKED ISSUE #8567: Add growth and virtual stake to market data
LPs should be able to track their virtual stake and keep track of the growth factor of the market (`r` in the ELS spec). `MarketData` already contains a field called `LiquidityProviderFeeShares`, which contains the ELS, but not the virtual stake. A new field should be added to the `LiquidityProviderFeeShare` type.
The `MarketData` type itself should contain a field with the `market_growth` as a decimal (represented as a string in the protos).

EVODelavega:
Closes #8567


AnExsomnis (dismissed):

protos/sources/vega/vega.proto
```diff
@@ -1063,6 +1065,8 @@ message LiquidityProviderFeeShare {
   string average_entry_valuation = 3;
   // Average liquidity score.
   string average_score = 4;
+  // The virtual stake of this liquidity provider (float).
```
candida-d, line 1068: What does (float) mean here?

datanode/gateway/graphql/schema.graphql
```diff
@@ -336,6 +338,8 @@ type LiquidityProviderFeeShare {
   averageEntryValuation: String!
   "The average liquidity score"
   averageScore: String!
+  "The virtual stake for this liquidity provider (float)"
```
candida-d, line 341: Same question here as in the protos

protos/sources/vega/vega.proto
EVODelavega: string representing a decimal. I included it here (and in the graphQL schema) because one of the other fields had that bit included.

AnExsomnis (dismissed):

datanode/sqlstore/testdata/marketdata.csv
```diff
@@ -1,551 +1,551 @@
...
-53035a92e297d29b6729bcd5cf032e675efb0e3df8ff4a0f0e876edde7faf3eb,2022-02-11 10:02:39.997006 +00:00,2022-02-11 10:02:41.000000 +00:00,1000034006,1000029421,1,1000034006,2,1000029421,1,1000034006,2,1000031713,1000031713,11,0,0,0,0,TRADING_MODE_CONTINUOUS,AUCTION_TRIGGER_UNSPECIFIED,AUCTION_TRIGGER_UNSPECIFIED,32501105195,50000000000,null,327667631512506.800000000000000,"[{""party"": ""af2bb48edd738353fcd7a2b6cea4821dd2382ec95497954535278dfbfff7b5b5"", ""equity_like_share"": ""1"", ""average_entry_valuation"": ""50000000000"", ""average_score"": ""123""}]",STATE_ACTIVE
-53035a92e297d29b6729bcd5cf032e675efb0e3df8ff4a0f0e876edde7faf3eb,2022-02-11 10:02:41.315951 +00:00,2022-02-11 10:02:42.000000 +00:00,1000034006,1000029421,1,1000034006,2,1000029421,1,1000034006,2,1000031713,1000031713,11,0,0,0,0,TRADING_MODE_CONTINUOUS,AUCTION_TRIGGER_UNSPECIFIED,AUC
```
wwestgarth, line 551: Setting them all to zero isn't sufficient. If the `entities.MarketData.MarketGrowth` is _**NOT**_ being saved into the database, when we populate a `entities.MarketData` struct when reading from the database the `MarketGrowth` field would default to `0` and the test would still pass.

wwestgarth (approved):

datanode/sqlstore/testdata/marketdata.csv
EVODelavega: #NotAllZeroes There's a couple of them (literally 2 I think :P) that have non-zero values

wwestgarth (dismissed):

candida-d (approved):

ValentinTrinque (approved):

PR #8574, opened 2023-06-23
Add snapshots to liquidity engine v2

### LINKED ISSUE #8505: Add snapshot support for liquidity engine
Add snapshot support for liquidity v2 (SLA) engine.

ValentinTrinque:
This snapshots the provisions, the SLA performance data, the data related to the supplied engine and the scores. The unit tests only cover very simple use cases.

The snapshot engine is not wired to anything except the liquidity engine v2 as the wiring is not yet done with the spot market.

The pending penalties are not yet merged and needs to be integrated afterward. Also, once the engine is fully implemented and merged, we want to improve the snapshot tests to account for all cases.

Close #8505


core/types/snapshot_nodes.go
```diff
@@ -809,6 +825,14 @@ func PayloadFromProto(p *snapshot.Payload) *Payload {
 		ret.Data = PayloadLiquiditySuppliedFromProto(dt)
 	case *snapshot.Payload_LiquidityTarget:
 		ret.Data = PayloadLiquidityTargetFromProto(dt)
+	case *snapshot.Payload_LiquidityV2Provisions:
```
ze97286, line 828: shouldn't there be a similar bit for the intoProto?

core/liquidity/v2/snapshot.go
```diff
@@ -0,0 +1,461 @@
...
+		},
+	}).Key()
+
+	e.scoresKey = (&types.PayloadLiquidityV2Scores{
+		Scores: &snapshotpb.LiquidityV2Scores{
+			MarketId: market,
+		},
+	}).Key()
+
+	e.hashKeys = append([]string{}, e.provisionsKey, e.suppliedKey, e.scoresKey)
```
ze97286, line 405: where's the performancesKey gone?

core/liquidity/v2/snapshot.go
```diff
@@ -0,0 +1,461 @@
...
+func (e *SnapshotEngine) Stopped() bool {
+	return e.stopped
+}
+
+func (e *SnapshotEngine) StopSnapshots() {
+	e.log.Debug("market has been cleared, stopping snapshot production", logging.MarketID(e.marketID))
+	e.stopped = true
+}
+
+func (e *SnapshotEngine) OnMinProbabilityOfTradingLPOrdersUpdate(d num.Decimal) {
```
ze97286, line 80: given that the underlying engine is defined as *Engine I don't think you need all of those wrappers.

core/liquidity/v2/snapshot.go
```diff
@@ -0,0 +1,461 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if e.stopped {
+		return nil, nil
+	}
+
+	if !changed {
```
ze97286, line 148: can we not do this please - changed is by construction always true so can we remove it from the serialiseXXX, the marshal, and basically everywhere? I wonder if the linter isn't gonna tell you this before me...

core/liquidity/v2/snapshot.go
```diff
@@ -0,0 +1,461 @@
...
+			RegisteredPenaltiesPerEpoch:      registeredPenaltiesPerEpochSnapshot,
+		}
+		performancePerPartySnapshot = append(performancePerPartySnapshot, partyPerformanceSnapshot)
+	}
+
+	payload := &snapshotpb.Payload{
+		Data: &snapshotpb.Payload_LiquidityV2Performances{
+			LiquidityV2Performances: &snapshotpb.LiquidityV2Performances{
+				MarketId:            e.market,
+				EpochStartTime:      e.slaEpochStart.String(),
```
wwestgarth, line 218: For saving times in snapshots we usually just save the `.UnixNano()`. We should do that here too 
(the docs for `time.String()` say "The returned string is meant for debugging" and mention stuff about not using it for stable serialization)

core/liquidity/v2/snapshot.go
```diff
@@ -0,0 +1,461 @@
...
+
+	return e.marshalPayload(payload)
+}
+
+func (e *SnapshotEngine) serialiseSupplied() ([]byte, bool, error) {
+	v1Payload := e.suppliedEngine.Payload()
+
+	// Dirty hack to support serialization of a mutualized supplied engine between
+	// liquidity engine version 1 and 2.
+	supplied := v1Payload.GetLiquiditySupplied()
```
wwestgarth, line 232: Do we need a V2 payload if `e.suppliedEngine` isn't changing?

candida-d (dismissed):

core/liquidity/v2/snapshot.go
ValentinTrinque: Nice catch.

core/liquidity/v2/snapshot.go
ValentinTrinque: That is true. I just stupidly copy / pasted the liquidity snapshot engine V1, thinking it's the way we do things.

core/liquidity/v2/snapshot.go
ValentinTrinque: The linter didn't catch it, but still, you are right. Nice catch.

core/liquidity/v2/snapshot.go
ValentinTrinque: Oh, right.

core/liquidity/v2/snapshot.go
ValentinTrinque: Because engine and payload namespaces are tied, and we can't reuse them across one another

core/types/snapshot_nodes.go
ValentinTrinque: Nice catch.

libs/time/time.go
```diff
@@ -0,0 +1,10 @@
+package time
+
+import "time"
+
+func UnixNanoToTime(timestampNano int64) time.Time {
+	seconds := timestampNano / 1e9
+	nanos := timestampNano % 1e9
+
+	return time.Unix(seconds, nanos)
```
ze97286, line 9: this makes no sense... isn't it just `time.Unix(0, timestampNano)`?

candida-d (dismissed):

jeremyletang (dismissed):

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -125,6 +125,10 @@ message Payload {
     BankingScheduledGovernanceTransfers banking_scheduled_governance_transfers = 61;
     EthContractCallResults eth_contract_call_results = 62;
     EthOracleVerifierLastBlock eth_oracle_verifier_last_block = 63;
+    LiquidityV2Provisions liquidity_v2_provisions = 64;
```
ze97286, line 128: what about pending provision?

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -836,6 +840,48 @@ message LiquidityScore {
   string party_id = 2;
 }
 
+message LiquidityV2Provisions {
+  // Market ID these liquidity parameters apply to. Required for key generation
+  string market_id = 1;
+
+  repeated vega.LiquidityProvision liquidity_provisions = 2;
```
ze97286, line 847: I don't think this makes sense anymore. We should use a LiquidityProvisionV2 which doesn't have all of the reference, shape nonsense.

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -836,6 +840,48 @@ message LiquidityScore {
...
+
+  int64 epoch_start_time = 2;
+
+  repeated LiquidityV2PerformancePerParty performance_per_party = 3;
+}
+
+message LiquidityV2PerformancePerParty {
+  string party = 1;
+
+  string elapsed_time_meeting_sla_during_epoch = 2;
```
ze97286, line 862: why save a duration (int64) as string?

protos/sources/vega/snapshot/v1/snapshot.proto
jeremyletang: we should keep the LiquidityProvision type as it is now, in the state serialization not settign the shap is no cost. And we keep the benefit of not having to change the APIs anywhere by keep the same types.

ze97286: this type is already v2, what's the point in building new types and reusing existing types that will never be valid for it?

jeremyletang: I'm not sure I get it, there's no new LiquidityProvision type.

ze97286:
```
message **LiquidityV2Provisions** {
  // Market ID these liquidity parameters apply to. Required for key generation
  string market_id = 1;

  repeated vega.LiquidityProvision liquidity_provisions = 2;
```

LiquidityV2Provisions is a new type, the existing type vega.LiquidityProvision will *never* be right for it, so why use it?

ze97286: why confuse the API with fields that can never be non empty?

jeremyletang: well, the only thing which is not right is the fields for the shapes. But they still exists in the historic tables etc. we can make them nullable now and deprecate it in the long term.

ze97286: this is the *snapshot* what it's got to do with any of that? anyways, doesn't really matter, just seems unclean.

jeremyletang: Doesn't have much to do with it yes, but it's just adding one more type, with convertion at some point from one to the other, which is unecessary right now I'd say.

karlem (approved):

karlem (approved):

wwestgarth (approved):

Issue #8533, opened 2023-06-19
[Feature]: Ethereum Oracles Error Handling


ettec:
Figure out and implement ethereum oracles error handling

ettec:
merged

PR #8595, opened 2023-06-26
feat: ethereum oracle error handling refactor

ettec:
close: #8533 
Change to the error handling such that an error that occurs consistently across all validators will result in an oracle data event with error attached.  All other errors (node specific) are logged but do not result in an oracle data event.

### DIFF
--- a/core/oracles/ethereum_oracle_verifier.go
+++ b/core/oracles/ethereum_oracle_verifier.go
@@ -140,22 +140,6 @@ func (s *EthereumOracleVerifier) ProcessEthereumContractCallResult(callEvent typ
 		return ErrDuplicatedEthereumCallEvent
 	}
 
-	if callEvent.Error != nil {
-		dataProto := vegapb.OracleData{
-			ExternalData: &datapb.ExternalData{
-				Data: &datapb.Data{
-					MatchedSpecIds: []string{callEvent.SpecId},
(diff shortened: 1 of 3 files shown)


AnExsomnis (approved):

core/oracles/ethereum_oracle_verifier.go
```diff
@@ -180,7 +164,13 @@ func (s *EthereumOracleVerifier) checkCallEventResult(callEvent types.EthContrac
 	// Maybe TODO; can we do better than this? Might want to cancel on shutdown or something..
 	ctx := context.Background()
 	checkResult, err := s.ethEngine.CallSpec(ctx, callEvent.SpecId, callEvent.BlockHeight)
-	if err != nil {
+	if callEvent.Error != nil {
```
wwestgarth, line 167: I think the case where `chainEvent.Error == nil` and `err != nil` is missing here. We should fail with:
```
return fmt.Errorf("failed to execute call event spec: %w", err)
```
but instead we're going to fall down and do the bytes equal bit.

ettec: yep, also the case where the error is nil is a bit messed up as it will log nil error with failed to execute message.  Will rework this.

pscott31 (approved):

PR #8601, opened 2023-06-27
feat: 8528 add data node support for stop orders

### LINKED ISSUE #8528: feat: Add data node APIs duplicating the existing orders API for Stop Orders
Currently we have the following APIs in data node for Orders:

- GetOrder - retrieves specific order given the order id
- ListOrders - retrieves a paginated list of orders that match a filter criteria
- ListOrderVersions - retrieves a list of all versions of an order given the order id
- ObserveOrders - subscribe for order events that match a filter criteria

We should provide the same functionality to:

- GetStopOrder
- ListStopOrders

as a first step from the data node.

guoguojin:
Closes #8528


wwestgarth (dismissed):

datanode/sqlstore/migrations/0009_stop_orders.sql
```diff
@@ -0,0 +1,63 @@
+-- +goose Up
+
+create type stop_order_expiry_strategy as enum('EXPIRY_STRATEGY_UNSPECIFIED', 'EXPIRY_STRATEGY_CANCELS', 'EXPIRY_STRATEGY_SUBMIT');
+create type stop_order_trigger_direction as enum('TRIGGER_DIRECTION_UNSPECIFIED', 'TRIGGER_DIRECTION_RISES_ABOVE', 'TRIGGER_DIRECTION_FALLS_BELOW');
+create type stop_order_status as enum('STATUS_UNSPECIFIED', 'STATUS_PENDING', 'STATUS_CANCELLED', 'STATUS_STOPPED', 'STATUS_TRIGGERED', 'STATUS_EXPIRED');
```
jeremyletang, line 5: this is missing the REJECTED enum

guoguojin: Thanks, that's been fixed and rebased now.

gordsport (dismissed):
Approved on behalf of API docs styling

mattrussell36:
@jeremyletang is there a plan to add a subscription for stop orders? Currently the pattern the frontends use to stay in-sync once a tx has been submitted is to open up a subscription and wait for the resulting data to be emitted

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -88,6 +88,20 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Orders"};
   }
 
+  // Get Stop order
```
candida-d, line 91:
```suggestion
  // Get stop order
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1020,6 +1034,64 @@ message OrderUpdates {
...
+message ListStopOrdersRequest {
+  // Optional pagination control.
+  optional Pagination pagination = 2;
+  // Order filter contains all filtering conditions and values that are applied to the orders listing.
+  optional StopOrderFilter filter = 5;
+}
+
+// Stop Order Filter
+//
+// Stop Order filter that contains all filtering conditions and values that are applied to the stop orders listing.
```
candida-d, line 1059:
```suggestion
// Stop order filter that contains all filtering conditions and values that are applied to the stop orders listing.
```

candida-d (dismissed):

jeremyletang:
> @guoguojin @jeremyletang is there a plan to add a subscription for stop orders? Currently the pattern the frontends use to stay in-sync is that once a TX has been submitted we open up a subscription and wait for the resulting data to be emitted.
> 
> We could use the TransactionResult bus event subscription which would allow us to be notified if the TX was successful, but as we also plan on showing stop orders within the order list, it would be nice if we could just open up a subscription and apply any incoming data to our order store

There might be in the future. For now we focused on the minimal required API. Now if really we see a need we may do the subscribtion etc.

candida-d (approved):

jeremyletang (approved):

wwestgarth (approved):

PR #8604, opened 2023-06-27
fix: ensure the asset used for asset for metrics exists in transfers

### LINKED ISSUE #8335: Incomplete validation on recurring transfer TX
### Problem encountered

Look at [this TX](https://explorer.fairground.wtf/txs/AB519655B8B3EF5D764AEF3E95723DD42C812C96AE1686C873205629729E7668) 

The assetForMetric is displayed as a text string and you cant click through to the asset ... this is because it doesnt exist! But despite this the TX was accepted and the BE displays it as a "success". 

image

This seems bad.  See [slack thread](https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1684513216119899?thread_ts=1684494600.856379&cid=CD4AKTFQT) for full details.

### Observed behaviour

On further investigation Paul was able to determine that both the assetID and the assetForMetric are subject to basic format checks and will reject if not valida. 

However:

If the assetID does not exist you get the 

'{"jsonrpc":"2.0","error":{"code":1070,"message":"Network error","data":"the network could not process the transaction: could not transfer funds, asset does not exist (ABCI code 70)"},"id":"request"}'

But if the assetForMetric does not exist the TX is processed without issue, leading to the case above.

### Expected behaviour

I would expect both validations to behave the same.  If the asset doesn not exist the TX should be rejected. 

### Steps to reproduce

```bash
1. Submit the transaction below on fairground 

./vegawallet_71_4 transaction send --wallet market-creator --pubkey 69464e35bcb8e8a2900ca0f87acaf252d50cf2ab2fc73694845a16b7c8a0dc6f --network fairground '{                   
    "transfer":{
        "fromAccountType": "ACCOUNT_TYPE_GENERAL",
        "toAccountType": "ACCOUNT_TYPE_REWARD_MAKER_PAID_FEES",
        "to":"0000000000000000000000000000000000000000000000000000000000000000",
        "asset":"fc7fd956078fb1fc9db5c19b88f0874c4299b2a7639ad05a47a28c0aef291b55",
        "amount":"1720000000000000000",
        "recurring":{
            "startEpoch": 6780,
            "endEpoch": 7100,
            "factor": "1",
            "dispatchStrategy": {
                "assetForMetric": "e6561f69c2a76858866aab2896eeb529b46040614566e0665602d67bc682c31f",
                "metric": "DISPATCH_METRIC_MAKER_FEES_PAID",
                "markets": ["e6561f69c2a76858866aab2896eeb529b46040614566e0665602d67bc682c31f"]
            }
        }
    }
}'



2. it will be accepted
3. View the TX and you will see the asset was invalid
```

### Software version

0.71.4

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

jeremyletang:
close #8335

### DIFF
--- a/core/banking/recurring_transfers.go
+++ b/core/banking/recurring_transfers.go
@@ -49,6 +49,19 @@ func (e *Engine) recurringTransfer(
 		return fmt.Errorf("could not transfer funds: %w", err)
 	}
 
+	if transfer.DispatchStrategy != nil {
+		hasAsset := len(transfer.DispatchStrategy.AssetForMetric) > 0
+		// ensure the asset transfer is correct
+		if hasAsset {
+			_, err := e.assets.Get(transfer.DispatchStrategy.AssetForMetric)
+			if err != nil {
+				transfer.Status = types.TransferStatusRejected
+				e.log.Debug("cannot transfer funds, invalid asset for metric", logging.Error(err))
+				return fmt.Errorf("could not transfer funds, invalid asset for metric: %w", err)
+			}
+		}
+	}
+
 	if err := transfer.IsValid(); err != nil {
 		transfer.Status = types.TransferStatusRejected
 		return err
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@
 - [8586](https://github.com/vegaprotocol/vega/issues/8586) - Fix cancel governance transfer proposal validation
 - [8597](https://github.com/vegaprotocol/vega/issues/8597) - Enact governance transfer cancellation
 - [8428](https://github.com/vegaprotocol/vega/issues/8428) - Add missing `LastTradedPrice` field in market data
+- [8335](https://github.com/vegaprotocol/vega/issues/8335) - Validate asset for metrics in transfers to be an actual asset
 
 ## 0.71.0
 


core/banking/recurring_transfers.go
```diff
@@ -49,6 +49,16 @@ func (e *Engine) recurringTransfer(
 		return fmt.Errorf("could not transfer funds: %w", err)
 	}
 
+	if transfer.DispatchStrategy != nil {
+		// ensure the asset transfer is correct
+		_, err := e.assets.Get(transfer.DispatchStrategy.AssetForMetric)
```
ze97286, line 54: I don't think this is correct. This may be legit empty if the dispatch strategy is market proposer

jeremyletang: right. let me fix.

core/banking/recurring_transfers.go
```diff
@@ -49,6 +49,27 @@ func (e *Engine) recurringTransfer(
 		return fmt.Errorf("could not transfer funds: %w", err)
 	}
 
+	if transfer.DispatchStrategy != nil {
+		hasAsset := len(transfer.DispatchStrategy.AssetForMetric) > 0
+		isMarketProposer := transfer.DispatchStrategy.Metric == vegapb.DispatchMetric_DISPATCH_METRIC_MARKET_VALUE
+
+		if !hasAsset && !isMarketProposer {
```
ze97286, line 56: this seems unnecessary because we're already checking this in checkTransfer

jeremyletang: fair enough.

ze97286 (approved):

PR #8614, opened 2023-06-28
check triggers against every eth block

pscott31:
instead of just once per tick

### DIFF
--- a/core/evtforward/ethcall/engine.go
+++ b/core/evtforward/ethcall/engine.go
@@ -3,6 +3,7 @@ package ethcall
 import (
 	"context"
 	"fmt"
+	"math/big"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


wwestgarth (approved):

core/evtforward/ethcall/engine.go
```diff
@@ -87,14 +81,43 @@ func (e *Engine) Start() {
 	}
 
 	e.poller.Loop(func() {
-		e.OnTick(ctx, time.Now())
+		e.Poll(ctx, time.Now())
 	})
 }
 
 func (e *Engine) UpdatePreviousEthBlock(height uint64, time uint64) {
 	e.mu.Lock()
```
ettec, line 89: This is going to happen at some point after the component is started at the moment I think, which means it will start with latest block height initially and then get reset to something lower, which means I think there is possiblilty to trigger twice for example.  Ideally this  engines start method would be passed the prev eth height to use which would solve this and remove the need for locking around the heights

pscott31: yes that's a good idea, once we figure out how to do that i'll add it in but i reckon this is still better than what we had so i'll merge it for now

PR #8627, opened 2023-06-28
fix: 8545 blockexplorer does not pagination backward correctly

### LINKED ISSUE #8545: [Bug]: Block Explorer API does not support pagination properly
### Problem encountered

The Block Explorer `listTransactions` API has cursors and limits. These are not implemented in a way that allows for paginating backwards through a set of data.

https://docs.vega.xyz/testnet/api/rest/explorer/block-explorer-list-transactions is the current endpoint

### Observed behaviour

Given a cursor, I can easily move forwards in data:

```
1 2 3 4 5 [6] 7 8 9 10
```

* If I've got the cursor 6, and I request limit: 1, after: 6 I'll get 7
* If I've got the cursor 6 and I request limit: 1, before 6 I'll get 1

### Expected behaviour

# Replace `limit` with `first` & `last`
This is how the GraphQL APIs work

* If I've got the cursor 6, and I request first: 1, after: 6 I'll get 7
* If I've got the cursor 6, and I request last: 1, before: 6 I'll get 5

### Steps to reproduce

```bash
Try using `listTransactions` to go back a page
```

### Software version

0.71.4

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

guoguojin:
Closes #8545


edd:
Looks good to me, thanks for using the deprecation flag but leaving limit in. I'll make another ticket when I've transitioned to first/last.

I'll leave ticking it to someone else but I am happy for this to go in 👍

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -73,6 +74,12 @@ message ListTransactionsRequest {
   repeated string exclude_cmd_types = 6;
   // Party IDs filter, can be sender or receiver
   repeated string parties = 7;
+  // Number of transaction to be returned from the blockchain. Use in conjunction with the `after` cursor to paginate forwards.
+  // On its own, this will return the first `first` transactions.
+  uint32 first = 8;
+  // Number of transaction to be returned from the blockchain. Use in conjunction with the `before` cursor to paginate backwards.
+  // On its own, this will return the last `last` transactions.
+  uint32 last = 9;
```
candida-d, line 82: I didn't want to leave a suggestion but I just couldn't let this one go. 

```suggestion
  // Number of transactions to be returned from the blockchain. Use in conjunction with the `after` cursor to paginate forwards.
  // On its own, this will return the first `first` transactions.
  uint32 first = 8;
  // Number of transactions to be returned from the blockchain. Use in conjunction with the `before` cursor to paginate backwards.
  // On its own, this will return the last `last` transactions.
  uint32 last = 9;
```

candida-d (approved):

PR #8628, opened 2023-06-28
Use vega home to locate snapshot databases 

ValentinTrinque:


jeremyletang (dismissed):

core/snapshot/databases/metadata/database.go
```diff
@@ -0,0 +1,60 @@
...
+
+type Adapter interface {
+	Save(version []byte, state []byte) error
+	Load(version []byte) (state []byte, err error)
+	Close() error
+	Clear() error
+}
+
+type Database struct {
+	adapter Adapter
```
EVODelavega, line 19: nit-pick, but can't we just embed the adapter here? Saves us from having to add the `Close` and `Clear` functions

ValentinTrinque: Sure thing

EVODelavega (approved):

PR #8638, opened 2023-06-29
feat: allow update market proposal submission with only ELS

### LINKED ISSUE #8635: [Feature]: Allow token holders to change a market without being an LP
# Feature Overview

We used to require both the governance token AND a minimum ELS for a party to be able to submit a market change proposal. This change is to use OR instead. This will bring it in line with rules for the voting (a party with sufficient ELS, but no governance tokens is still able to successfully cast a vote).

## Spec change PR
- https://github.com/vegaprotocol/specs/pull/1768

jeremyletang:
close #8635


EVODelavega:
This change concerns me a bit, as it seems to go against one of the AC's, specifically 0028-GOVE-072 (https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0028-GOVE-governance.md#0028-GOVE-072):

> A market change that's to modify any parameters on a market in pending state (i.e. voting has successfully completed on the market creation and the market is in the opening auction) will run voting rules the same as market creation proposals i.e. LPs don't get a vote.

EVODelavega (approved):

PR #8659, opened 2023-07-03
feat: Allow to terminate, suspend/resume market via governance

### LINKED ISSUE #8415: [Feature]: Terminate/Suspend market
# Feature Overview

**In order to** terminate a spot market or suspend/resume any market a governance proposal is required. 
**We will** create a new governance proposal for terminating/suspending/resuming a market. 
**So that** a spot market can be terminated and any market can be suspended and resumed. 

## Specs
- https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0028-GOVE-governance.md#6-change-market-state

## Tasks
- [ ] Implement the governance proposal 
- [ ] Implement the market termination for spots
- [ ] Implement the market suspend for any market
- [ ] Implement the market resume for any market

ze97286:
Closes #8415


witgaw (commented):
LGTM

protos/sources/vega/vega.proto
```diff
@@ -155,6 +155,8 @@ enum AuctionTrigger {
   AUCTION_TRIGGER_LIQUIDITY_TARGET_NOT_MET = 5;
   // Liquidity auction due to not being able to deploy LP orders because there's nothing to peg on one or both sides of the book
   AUCTION_TRIGGER_UNABLE_TO_DEPLOY_LP_ORDERS = 6;
+  // Market is suspended and put into auction via governance
+  AUCTION_TRIGGER_GOVERNANCE_SUSPENSION = 7;
```
wwestgarth, line 159: I think we're going to need to add these new types to some datanode migration somewhere given they are defined as enums here:
https://github.com/vegaprotocol/vega/blob/develop/datanode/sqlstore/migrations/0001_initial.sql#L504-L506

protos/sources/vega/markets.proto
```diff
@@ -241,6 +241,8 @@ message Market {
     STATE_TRADING_TERMINATED = 8;
     // Settlement triggered and completed as defined by product
     STATE_SETTLED = 9;
+    // Market has been suspended via governance
+    STATE_SUSPENDED_VIA_GOVERNANCE = 10;
```
wwestgarth, line 245: Can't we reuse `STATE_SUSPENDED` and use the auction trigger `AUCTION_TRIGGER_GOVERNANCE_SUSPENSION = 7;` distinguish between the reason?

protos/sources/vega/vega.proto
ze97286: ugh

protos/sources/vega/markets.proto
ze97286: technically we could, but why?  I'd prefer it to be explicit unless there's any benefit for reusing the existing state_suspended.

wwestgarth: Can we at least change the trading mode one to `TRADING_MODE_SUSPENDED_VIA_GOVERNANCE`? I just find it odd that we can have a market state of `STATE_SUSPENDED` and a trading mode of `TRADING_MODE_SUSPENDED`, but a market state of `SUSPENDED` =/=> a trading mode of `SUSPENDED`.

ze97286: sure

core/execution/future/market.go
```diff
@@ -997,6 +997,31 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
 	return parked
 }
 
+func (m *Market) UpdateMarketState(ctx context.Context, changes *types.MarketStateUpdateConfiguration) error {
+	if changes.UpdateType == types.MarketStateUpdateTypeTerminate {
+		// terminate and settle
+		m.tradingTerminated(ctx, true)
+		n := num.Numeric{}
+		n.SetUint(changes.SettlementPrice)
+		m.settlementData(ctx, &n)
```
wwestgarth, line 1006: Do we not want to set the market state to `CLOSED`? In the market lifecycle spec its distinct from `SETTLE`:
https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0043-MKTL-market_lifecycle.md#closed

And also a protos value that already exists for it ready an waiting:
```
// Governance vote to close (Not currently implemented)
    STATE_CLOSED = 7;
```

Theres also mention of uncrossing auctions before closing?

core/processor/abci.go
```diff
@@ -1912,6 +1914,20 @@ func (app *App) enactCancelTransfer(ctx context.Context, prop *types.Proposal) {
 	}
 }
 
+func (app *App) enactMarketStateUpdate(ctx context.Context, prop *types.Proposal) {
+	prop.State = types.ProposalStateEnacted
+	changes := prop.Terms.GetMarketStateUpdate().Changes
+	if err := app.exec.VerifyUpdateMarketState(changes); err != nil {
+		app.log.Error("failed to enact governance market state update", logging.String("proposal", prop.ID), logging.String("error", err.Error()))
+		prop.FailWithErr(types.ProporsalErrorFailedGovernanceTransferCancel, err)
```
wwestgarth, line 1922: `ProporsalErrorFailedGovernanceTransferCancel` copy and paste error?

core/execution/future/market.go
ze97286: idk, because we're settling the market at least for future so not sure this is still as was originally designed. It's a question for @witgaw, same about the uncrossing auctions.

core/processor/abci.go
ze97286: yes

core/governance/engine.go
```diff
@@ -1036,6 +1039,26 @@ func (e *Engine) validateCancelGovernanceTransfer(transferID string) (types.Prop
...
+	marketState, err := e.markets.GetMarketState(marketID)
+	if err != nil {
+		return types.ProposalErrorInvalidMarket, err
+	}
+
+	// if the market is already terminated or not yet started or settled
+	if marketState == types.MarketStateTradingTerminated || marketState == types.MarketStateSettled || marketState == types.MarketStateProposed {
+		return types.ProposalErrorInvalidMarket, ErrMarketStateUpdateNotAllowed
+	}
+
```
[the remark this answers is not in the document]
ze97286: I don't think we need to do anything - if a proposal for termination gets enacted, it will get the market into a terminal state and then any further proposal getting pre-enacted will fail here and rejected.

wwestgarth:
For this AC:
Verify that a party with 0 balance of the governance token, but with sufficient ELS can submit a market change proposal successfully. ([0028-GOVE-117](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0028-GOVE-governance.md#0028-GOVE-117))

I think you're going to want to do something in `validateOpenProposal` to handle it the same ways as market-update proposals.

datanode/gateway/graphql/schema.graphql
```diff
@@ -3352,6 +3354,9 @@ enum MarketTradingMode {
 
   "No trading allowed"
   TRADING_MODE_NO_TRADING
+
+  "Special auction mode triggered via governance market suspension"
```
candida-d, line 3358:
```suggestion
  "Special auction mode triggered via market suspension governance action"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3694,6 +3699,26 @@ type UpdateNetworkParameter {
...
+  "Terminate the market"
+  MARKET_STATE_UPDATE_TYPE_TERMINATE
+  "Suspend the market"
+  MARKET_STATE_UPDATE_TYPE_SUSPEND
+  "Resume a suspended market"
+  MARKET_STATE_UPDATE_TYPE_RESUME
+}
+
+type UpdateMarketState {
+  "The market whose state we want to update"
```
candida-d, line 3714:
```suggestion
  "The market for which to update the state"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -4044,6 +4070,8 @@ enum AuctionTrigger {
   AUCTION_TRIGGER_LIQUIDITY_TARGET_NOT_MET
   "Liquidity monitoring due to not being able to deploy LP orders because there's nothing to peg on one or both sides of the book"
   AUCTION_TRIGGER_UNABLE_TO_DEPLOY_LP_ORDERS
+  "Auction triggered by governance market suspension"
```
candida-d, line 4073:
```suggestion
  "Auction triggered by market suspension governance action"
```

core/governance/engine.go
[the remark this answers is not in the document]
wwestgarth: What about the case where I submit a update-proposal before its the first one has been enacted? So that market hasn't changed state yet.

protos/sources/vega/governance.proto
```diff
@@ -263,6 +263,8 @@ message ProposalTerms {
     NewTransfer new_transfer = 109;
     // Cancel a governance transfer.
     CancelTransfer cancel_transfer = 110;
+    // Proposal change for updating the state of a market.
```
candida-d, line 266: Is this to "change a proposal" or a "proposal to change/update the state of a market"?

protos/sources/vega/governance.proto
```diff
@@ -486,6 +490,27 @@ message Vote {
...
+message UpdateMarketState {
+  // Configuration for governance-initiated change of a market state
+  UpdateMarketStateConfiguration changes = 1;
+}
+
+enum MarketStateUpdateType {
+  MARKET_STATE_UPDATE_TYPE_UNSPECIFIED = 0;
+  MARKET_STATE_UPDATE_TYPE_TERMINATE = 1;
+  MARKET_STATE_UPDATE_TYPE_SUSPEND = 2;
+  MARKET_STATE_UPDATE_TYPE_RESUME = 3;
```
candida-d, line 502: docs please :)

protos/sources/vega/governance.proto
```diff
@@ -486,6 +490,27 @@ message Vote {
...
+}
+
+message UpdateMarketStateConfiguration {
+  // ID of the market
+  string market_id = 1;
+  // The type of the market update
```
candida-d, line 508:
```suggestion
  // Type of market update
```

protos/sources/vega/governance.proto
```diff
@@ -486,6 +490,27 @@ message Vote {
...
+  MarketStateUpdateType update_type = 2;
+  // a settlement price. relevant only for market termination for futures
```
candida-d, line 510:
```suggestion
  // Settlement price, relevant only for market termination for futures markets
```

core/governance/engine.go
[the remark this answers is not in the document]
ze97286: still, the check on VerifyUpdateMarketState would fail the proposal before it gets enacted. 

```
state := market.GetMarketState()
if state == types.MarketStateCancelled || state == types.MarketStateClosed || state == types.MarketStateRejected || state == types.MarketStateSettled || state == types.MarketStateTradingTerminated {
			return fmt.Errorf("invalid state update request. Market is already in a terminal state")
		}
```

core/execution/future/market.go
```diff
@@ -997,6 +997,31 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
 	return parked
 }
 
+func (m *Market) UpdateMarketState(ctx context.Context, changes *types.MarketStateUpdateConfiguration) error {
+	if changes.UpdateType == types.MarketStateUpdateTypeTerminate {
+		// terminate and settle
+		m.tradingTerminated(ctx, true)
```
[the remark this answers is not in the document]
ze97286: that's a good point. I think we'd want to use the price from the governance proposal but what matters is what @witgaw thinks.

protos/sources/vega/governance.proto
ze97286: it's the latter, it's just following terminology of this proto to refer to the action as the "change", you'll see what I mean if you scroll up in that file.

core/execution/future/market.go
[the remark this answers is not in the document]
witgaw: Yes, we should use the proposal price in that case. I'll clarify that in the spec and add an AC.

core/execution/future/market.go
witgaw: `STATE_CLOSED` would be ideal and yes, we'd want to uncross any auctions there may be first - will clarify both these points in the spec.

core/governance/engine.go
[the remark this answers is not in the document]
wwestgarth: Yea you're right, Cool.

ze97286:
> For this AC: Verify that a party with 0 balance of the governance token, but with sufficient ELS can submit a market change proposal successfully. ([0028-GOVE-117](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0028-GOVE-governance.md#0028-GOVE-117))
> 
> I think you're going to want to do something in `validateOpenProposal` to handle it the same ways as market-update proposals.

I think it's similar to marketUpdate, so added, please take a look.

core/execution/future/market.go
ze97286: done

core/execution/future/market.go
```diff
@@ -997,6 +997,67 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
 	return parked
 }
 
+func (m *Market) uncrossOrderAtAuctionEnd(ctx context.Context) {
+	if !m.as.InAuction() {
```
wwestgarth, line 1001: As per the spec update, we also don't want to uncross if we are in an opening auction.

core/execution/future/market.go
```diff
@@ -997,6 +997,67 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
...
+		m.tradingTerminatedWithFinalState(ctx, types.MarketStateClosed)
+	} else if changes.UpdateType == types.MarketStateUpdateTypeSuspend {
+		m.mkt.State = types.MarketStateSuspendedViaGovernance
+		m.mkt.TradingMode = types.MarketTradingModeSuspendedViaGovernance
+		m.as.StartGovernanceSuspensionAuction(m.timeService.GetTimeNow())
+		m.enterAuction(ctx)
+		m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+	} else if changes.UpdateType == types.MarketStateUpdateTypeResume && m.mkt.State == types.MarketStateSuspendedViaGovernance {
+		m.as.EndGovernanceSuspensionAuction()
+		m.checkAuction(ctx, m.timeService.GetTimeNow(), m.idgen)
```
wwestgarth, line 1051: I think we need to call `leaveAuction()` somewhere otherwise we won't get an `AuctionState` event saying we've left, and also unpark pegged orders etc etc.

wwestgarth: Also I can't see anything in `checkAuction` that would re-enter us into an opening-auction if the suspension happened while in one (and then was resumed while still in the period)?

core/execution/update_market_state_test.go
```diff
@@ -0,0 +1,124 @@
...
+		SettlementPrice: num.NewUint(100),
+	}
+	require.NoError(t, exec.engine.UpdateMarketState(context.Background(), config))
+
+	// after governance suspension
+	state, err = exec.engine.GetMarketData(mkt.ID)
+	require.NoError(t, err)
+	require.Equal(t, types.MarketStateSuspendedViaGovernance, state.MarketState)
+	require.Equal(t, types.MarketTradingModeSuspendedViaGovernance, state.MarketTradingMode)
+
```
wwestgarth, line 95: Can we throw an `OnTick()` somewhere in here between the suspension and resume to make sure it stays in the correct state for a block before resuming?

core/governance/engine.go
```diff
@@ -794,7 +795,7 @@ func (e *Engine) validateOpenProposal(proposal *types.Proposal) (types.ProposalE
 
 	checkProposerToken := true
 
-	if proposal.IsMarketUpdate() {
+	if proposal.IsMarketUpdate() || proposal.IsMarketStateUpdate() {
 		proposalError, err := e.validateMarketUpdate(proposal, params)
```
wwestgarth, line 799: This is going to panic. When we call `validateMarketUpdate()` if its an MarketStateUpdate proposal.

core/monitor/auction.go
```diff
@@ -110,6 +111,26 @@ func (a *AuctionState) StartPriceAuction(t time.Time, d *types.AuctionDuration)
 	a.stateChanged = true
 }
 
+func (a *AuctionState) StartGovernanceSuspensionAuction(t time.Time) {
+	a.mode = types.MarketTradingModeSuspendedViaGovernance
+	a.trigger = types.AuctionTriggerGovernanceSuspension
+	a.start = true
+	a.stop = false
+	a.begin = &t
+	a.end = &types.AuctionDuration{Duration: math.MaxInt64}
```
wwestgarth, line 120: I don't think we need to set this duration to `MaxInt64`, you can set it to `0` and it gets treated as forever. Eg in `ExpiresAt()`:
```
if a.end == nil || a.end.Duration == 0 { // not time limited
	return nil
}
```

core/execution/future/market.go
ze97286: indeed, returning to an opening auction is a bit of a problem with this implementation. need to think about it.

core/execution/future/market.go
```diff
@@ -997,6 +999,78 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
...
+			m.leaveAuction(ctx, m.timeService.GetTimeNow())
+		} else {
+			if m.as.GetState().Trigger == types.AuctionTriggerOpening {
+				m.mkt.State = types.MarketStatePending
+				m.mkt.TradingMode = types.MarketTradingModeOpeningAuction
+			} else {
+				m.mkt.State = types.MarketStateSuspended
+				m.mkt.TradingMode = types.MarketTradingModeMonitoringAuction
+			}
+			m.checkAuction(ctx, m.timeService.GetTimeNow(), m.idgen)
```
jeremyletang, line 1067: What calls this method? isn't m.idgen going to be nil here?

core/governance/engine.go
jeremyletang, line 1: In this file, there's a switch for getProposalParams, which hasn't been updated? do we also need to introduce new params for this specific new governance proposal? Maybe the UpdateMarket one are OK?

Also a simple test to enact such proposal could be nice, as I would think these proposal would not even go through right now maybe?

core/execution/future/market.go
```diff
@@ -997,6 +999,86 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
...
+				m.broker.Send(evt)
+			}
+		} else {
+			m.as.StartGovernanceSuspensionAuction(m.timeService.GetTimeNow())
+			m.enterAuction(ctx)
+			m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+		}
+	} else if changes.UpdateType == types.MarketStateUpdateTypeResume && m.mkt.State == types.MarketStateSuspendedViaGovernance {
+		if m.as.GetState().Trigger == types.AuctionTriggerGovernanceSuspension {
+			m.as.EndGovernanceSuspensionAuction()
```
witgaw, line 1057: Just because the auction was triggered with a governance vote doesn't imply that we can always leave it with a resume vote.

Some other triggers might be active now (the volume in the auction could be such that the liquidity trigger is now active or the uncrossing price could be triggering the price monitoring auction).

This should go through the regular `checkAuction` flow (but it'd need to get modified first). I think changing the market state to `Suspended` (from `GovernanceSuspended`) upon enactment of the resume proposal and then just going through `checkAuction` might just work.

witgaw (dismissed):

core/execution/future/market.go
```diff
@@ -997,6 +999,86 @@ func (m *Market) parkAllPeggedOrders(ctx context.Context) []*types.Order {
...
+			evt := m.as.AuctionExtended(ctx, m.timeService.GetTimeNow())
+			if evt != nil {
+				m.broker.Send(evt)
+			}
+		} else {
+			m.as.StartGovernanceSuspensionAuction(m.timeService.GetTimeNow())
+			m.enterAuction(ctx)
+			m.broker.Send(events.NewMarketUpdatedEvent(ctx, *m.mkt))
+		}
+	} else if changes.UpdateType == types.MarketStateUpdateTypeResume && m.mkt.State == types.MarketStateSuspendedViaGovernance {
```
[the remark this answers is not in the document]
ze97286: I don't think that is correct. m.mkt.State will be MarketStateSuspendedViaGovernance regardless of if governance suspension is the trigger or extension trigger as it is set when we suspend in line 1042, unless I'm missing something?

core/execution/future/auction.go
```diff
@@ -31,6 +31,10 @@ func (m *Market) checkAuction(ctx context.Context, now time.Time, idgen common.I
 		return
 	}
 
+	if m.mkt.State == types.MarketStateSuspendedViaGovernance {
+		m.as.ExtendAuctionSuspension(types.AuctionDuration{Duration: int64(m.minDuration)})
```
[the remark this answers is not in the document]
ze97286: not sure I follow. If I do, you mean we should only extend here if it expires, which is fine. is that what you mean though?

core/execution/future/market.go
ze97286: the idea is that if the trigger was suspension and there is not extension, we should just EndGovernanceSuspensionAuction and leave the auction (line 1056 is missing the check for the extension trigger, but imagine it's there)

core/execution/future/auction.go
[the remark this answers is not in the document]
witgaw: Yes. (I meant that if we extend every second by say 5 minutes then the time to expiry will keep growing which might seem weird if someone's not familiar with the details)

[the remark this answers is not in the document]
ze97286: ok

core/execution/future/market.go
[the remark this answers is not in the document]
witgaw: Ah, yes, you're right.

core/execution/future/market.go
witgaw: But the principle we want to follow is that we never leave an auction if we'd immediately re-enter another auction type, we just extend the original auction instead. 

So the trigger could be a suspension, but we there might be an extension when we try to resume the market.

Say:
- market is in continuous trading mode, price monitoring bounds are [95,105],
- market gets suspended via governance,
- orders get put in whilst market is in (governance) auction mode,
- orders are such that the uncrossing price is 106,
- in that case the original (governance suspension) auction should get extended by the price monitoring auction.

ze97286: again, no. I'm saying we enter this code only if the trigger is governance and there is no extension. 

```
if m.as.GetState().Trigger == types.AuctionTriggerGovernanceSuspension && m.as.GetState().Extension == types.AuctionTriggerUnspecified {
			m.as.EndGovernanceSuspensionAuction()
			m.leaveAuction(ctx, m.timeService.GetTimeNow())
		} 
```

witgaw: Aren't you referring to local changes by any chance? Can't find the snipped you've pasted above anywhere.

ze97286: I'm referring to the concept and I did say:  **(line 1056 is missing the check for the extension trigger, but imagine it's there)** - so now you don't need to imagine. I didn't want to push the changes before you agree that it makes sense.

witgaw: But when will that extension trigger get a chance to get set if:
https://github.com/vegaprotocol/vega/blob/81400c8187318b7db7e688dcdca4c3ad49481c46/core/execution/future/auction.go#L34-L36
?

witgaw: Yes, perfect.

Sorry, I somehow manage to convince myself there's a return here:
https://github.com/vegaprotocol/vega/blob/81400c8187318b7db7e688dcdca4c3ad49481c46/core/execution/future/auction.go#L34-L36

witgaw (dismissed):

witgaw (dismissed):

wwestgarth:
Just thought, in the data-node when you ask to list market's by default it excludes rejected and settled ones. We probably want to also exclude closed ones now. Awkwardly we have this `includeSettled` flag which probably should also include closed too now (but then its name is all wrong)? I don't know. Happy for that to be a spin-off ticket though because it might need some FE involvement depending on how they use `includeSettled`.

ze97286:
> includeSettled



> Just thought, in the data-node when you ask to list market's by default it excludes rejected and settled ones. We probably want to also exclude closed ones now. Awkwardly we have this `includeSettled` flag which probably should also include closed too now (but then its name is all wrong)? I don't know. Happy for that to be a spin-off ticket though because it might need some FE involvement depending on how they use `includeSettled`.

added.

candida-d (approved):

wwestgarth (approved):

PR #8667, opened 2023-07-04
Rework snapshot engine for better testability

### LINKED ISSUE #8661: Strengthen snapshot engine unit testing
The snapshot engine is missing unit test coverage. Some bugs slipped through. This is mostly due to the API misuse of the engine in the unit tests.

One of the culprit is the method `ClearAndInitialize()` method. It is not accessed from outside in production code, but is in unit tests. As a result, the code path tested is not representative of the production code path, that leads to different results, and make the tests themselves useless _to some extent_.

The goal of this issue is to close access to the method, and rewrite the unit tests that depends on it to match the production code path.

### LINKED ISSUE #8680: Move loading snapshots to initialization phase
The snapshots are loaded when the `Info()` method is called on the ABCI. This is not the right place to do this. It has been done initially because we couldn't guess which mode the snapshot was loadign from (state-sync, or local storage). Once the configuration is sorted, the one set in it will be the one used, which allow the snapshot engine to load the snapshots right during the initialization phase.

The `Info()` ABCI won't trigger data loading anymore.

The intialization workflow must respect the following conditions:
```golang
if localSnapshots { // so ignoring state-sync
    if startHeight == 0 {
        // Loading from latest local snapshot,
    } else {
	// Try reloading snapshot from specified height.
	// Error -> No snapshot for version XXX
    }
} else {
    if startHeight == 0 {
         // Replay the chain or state-sync if enabled. Up to tendermint to decide.
    } else {
	// Wait for state-sync to offer expected snapshot for height
    }
}
```

### LINKED ISSUE #8682: Restoring the snapshots from specific height should use the metadata

ValentinTrinque:
Close #8661 
Close #8680 
Close #8682


core/snapshot/engine.go
```diff
@@ -24,226 +22,165 @@ import (
...
+
+		providers: map[string]types.StateProvider{},
+		appState:  &types.AppState{},
+
+		// By default, a snapshot is triggered after each commit.
+		intervalBetweenSnapshots: 1,
+		// This must not be set to 0, otherwise a snapshot will be generated
+		// right after the state has been reload, leading effectively to a
+		// duplicated snapshot.
+		commitsLeftBeforeSnapshot: 1,
```
wwestgarth, line 153: I believe this (previously known as "current") was set to -1 to mean "unset" so on the first call to `OnSnapshotIntervalUpdate` it'll stop us taking a snapshot on block 1 e.g if the interval is set to to 1000 you'd expect snapshots on `1000, 2000, 3000`, but we were seeing snapshots on `1, 1001, 2002`. I have a vague memory of snapshots on block height 1 interacting badly with checkpoint restores.

Anyway, it would be good to just double check that we still do not snapshot on height 1 when `intervalBetweenSnapshots > 1`.

ValentinTrinque: Normally, I kept the behavior as before.

Here is the way I understand the previous behavior, assuming `intervalBetweenSnapshots` is set to `1000`.

It's initialized to `-1`. From there 2 paths:
1. a snapshot is applied (state-sync or local, it doesn't matter), once the state is restored, the `commitsLeftBeforeSnapshot` is set to `intervalBetweenSnapshots`. And it decreases (becomes `999`) when Tendermint calls the ABCI method `Commit`. It's not less or equal to `0` so it doesn't snapshot. We don't want to create a snapshot.
2. no snapshot is applied, the `commitsLeftBeforeSnapshot` is left at `-1`. So, when the ABCI method `Commit` is called, it's already below `0`, it get reset and the snapshot is taken. So, I believe it will take a snapshot on the first block.

In our case, setting the `commitsLeftBeforeSnapshot` to the value of `intervalBetweenSnapshots` (instead of `-1`) should yield the same result when a snapshot is applied. When loading from the local storage, there is effectively no difference between setting the value before or after the state restoration. When loading from state-sync, it _shouldn't_ change anything as well as Tendermint shouldn't call the ABCI method `Commit()` on the node when state-sync is enabled. Therefore, it's akin to restoring for the local storage. However, it differs from the previous implementation when nothing is restored, and we are starting from a brand new chain. It would still need to go through the counter to take the snapshot, whereas, in the previous implementation, we were already below `0`, triggering the snapshot on the first call to `Commit()`. Since it didn't feel right, I changed it so it's consistent.

ValentinTrinque:
The nightly run
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/24541/pipeline

ValentinTrinque:
Nightly run
https://jenkins.ops.vega.xyz/job/common/job/system-tests/24758/

ValentinTrinque:
Nightly run https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/24860/pipeline

ValentinTrinque:
Nightly build:
https://jenkins.ops.vega.xyz/job/common/job/system-tests/24906/

ValentinTrinque:
Nightly run
https://jenkins.ops.vega.xyz/job/common/job/system-tests/24933/

jeremyletang (dismissed):

core/netparams/defaults.go
```diff
@@ -217,7 +217,7 @@ func defaultNetParams() map[string]value {
 		NetworkCheckpointTimeElapsedBetweenCheckpoints: NewDuration(DurationGT(0 * time.Second)).Mutable(true).MustUpdate("1m"),
 		// take a snapshot every 1000 blocks, ~20 minutes
 		// if we assume a block time of anything between 1 to 2 seconds
-		SnapshotIntervalLength: NewInt(gteI0).Mutable(true).MustUpdate("1000"),
+		SnapshotIntervalLength: NewUint(gteU1).Mutable(true).MustUpdate("1000"),
```
EVODelavega, line 220: Though in practice we probably never want to do anything of the sort, but by restricting this value to be >= 1, we effectively make it impossible to snapshot every block should we ever want to do that

pscott31 (approved):

PR #8685, opened 2023-07-04
Ethereum oracle network parameter

pscott31:
Defaults to disabled. When disabled, don't start the eth call engine and reject market proposals that contain data sources that require ethereum oracles.

### DIFF
--- a/core/governance/market.go
+++ b/core/governance/market.go
@@ -54,6 +54,8 @@ var (
 	ErrInvalidRiskParameter = errors.New("invalid risk parameter")
 	// ErrInvalidInsurancePoolFraction is returned if the insurance pool fraction parameter is outside of the 0-1 range.
 	ErrInvalidInsurancePoolFraction = errors.New("insurnace pool fraction invalid")
+	// ErrEthOraclesDisabled is returned if ethereum oracles are not enabled via network parameter and someone tries to propose a market that uses them
+	ErrEthOraclesDisabled = errors.New("ethereum oracles are disabled by network parameter")
 )
 
 func assignProduct(
@@ -588,6 +590,9 @@ func validateNewMarketChange(
 	if perr, err := validateNewInstrument(terms.Changes.Instrument, terms.Changes.DecimalPlaces, assets, etu, deepCheck); err != nil {
 		return perr, err
 	}
+	if perr, err := validateUseOfEthOracles(terms, netp); err != nil {
+		return perr, err
+	}
 	// verify opening auction duration, works the same for successor markets
 	if perr, err := validateAuctionDuration(openingAuctionDuration, netp); !etu.cpLoad && err != nil {
 		return perr, err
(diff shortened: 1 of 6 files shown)


core/protocol/all_services.go
```diff
@@ -574,6 +570,15 @@ func (svcs *allServices) setupNetParameters(powWatchers []netparams.WatchParam)
 				return svcs.eventForwarderEngine.SetupEthereumEngine(svcs.ethClient, svcs.eventForwarder, svcs.conf.EvtForward.Ethereum, ethCfg, svcs.assets)
 			},
 		},
+		{
+			Param: netparams.EthereumOraclesEnabled,
+			Watcher: func(ctx context.Context, isEnabled int64) error {
```
ettec, line 575: Is this so we don't run a poller thread when no enable ethereum oracles?  IMO it seems preferable just to always run it and remove the additional locking, if there are no specs if will not call out.  But maybe I have misunderstood?

core/governance/market.go
```diff
@@ -615,6 +620,24 @@ func validateNewMarketChange(
...
+	}
+
+	ethOracleEnabled, _ := netp.GetInt(netparams.EthereumOraclesEnabled)
+
+	switch product := terms.Changes.Instrument.Product.(type) {
+	case *types.InstrumentConfigurationFuture:
+		terminatedWithEthOracle := !product.Future.DataSourceSpecForTradingTermination.GetEthCallSpec().IsZero()
+		settledWithEthOracle := !product.Future.DataSourceSpecForSettlementData.GetEthCallSpec().IsZero()
+		if (terminatedWithEthOracle || settledWithEthOracle) && ethOracleEnabled != 1 {
+			return types.ProposalErrorUnsupportedProduct, ErrEthOraclesDisabled
```
ettec, line 635: Should the error here be something like "Eth Oracle Disabled" to make it clearer why validation failed  (I imagine this will only happen a few times but still...)

pscott31: err, it is right? ```ErrEthOraclesDisabled = errors.New("ethereum oracles are disabled by network parameter")```

core/protocol/all_services.go
pscott31: well yes, that was my thought; no chance if it being able to mess anything up somehow if it's not running - but I do take your point.

core/governance/market.go
```diff
@@ -615,6 +620,24 @@ func validateNewMarketChange(
 	return types.ProposalErrorUnspecified, nil
 }
 
+func validateUseOfEthOracles(terms *types.NewMarket, netp NetParams) (types.ProposalError, error) {
```
jeremyletang, line 623: This validation could probably be done here in the processor: https://github.com/vegaprotocol/vega/blob/develop/core/processor/abci.go#L1107

This is where we were doing the CanPropose ones, the advantage is that as the check is done in the tendermint checkTx, the transaction is rejected as soon as it get submitted, and never even get onto a block.

PR #8689, opened 2023-07-05
Improve message for Pow ban

### LINKED ISSUE #8544: Improve error message for PoW ban
An integrator hit a PoW ban because they were using a block number in the future. The error messages they got were not helpful enough to get them to resolve the issue without asking for advice. 

Could these error messages be more helpful so that a user can self-serve and interrogate the problem themselves?

```
SubmitTransactionResponse {
    code: 89,
    data: "unknown block height for tx:d06ea5d08e3c93fa06383aa5e1c5af818acbdd05345374f633cd3c23c04e1d71, command:Batch Market Instructions 
...
```

```
SubmitTransactionResponse {
    code: 89,
    data: "party is banned from sending transactions",
    height: "0",
    log: "",
    success: false,
```

Additionally, can code 89 be documented somewhere like this? https://docs.vega.xyz/mainnet/api/grpc/vega/vega.proto#ordererror

zale144:
Closes #8544


zale144:
Here we need t lo,be fill clear with what info we will provide the client with, taking into account that it may or may not be a client with malicious intent. I propose here that when caching the banned party we also include some of the metadata in the naughty cache, so that it can be used to return more info for the subsequent requests where we tell the party that they have been banned and why. Otherwise, I also propose we add another field to the submit Transaction response, which will be an enum field wrapping the code with a textual value for the mapped error.

jeremyletang:
Ban are being removed with the integration of tendermin 0.38

PR #8697, opened 2023-07-05
Disallow snapshot state-sync if local snapshots exist

### LINKED ISSUE #8679: Disallow snapshot state-sync if local snapshots exist
To improve the snapshot initialization workflow, we first need to make the loading mode (state-sync or local store) in configuration much more strict. It should be either the state-sync, either from local storage. If both are configured then we should throw an error.

This work prepares the next step that will make the snaphsot engine load the snapshots right upon initialization.

ValentinTrinque:
Close #8679


core/snapshot/engine.go
```diff
@@ -220,6 +231,12 @@ func New(ctx context.Context, vegaPath paths.Paths, conf Config, log *logging.Lo
 		return nil, fmt.Errorf("could not initialize databases: %w", err)
 	}
 
+	if eng.metadb.ContainsMetadata() {
+		eng.reloadFromLocalStorage = true
+	} else {
+		eng.reloadFromLocalStorage = false
+	}
```
ValentinTrinque, line 238: Might look silly but it will change in the next PR.

core/snapshot/config.go
```diff
@@ -39,16 +46,18 @@ func NewDefaultConfig() Config {
...
 	}
 }
 
 // Validate checks the values in the config file are sensible.
 func (c *Config) Validate() error {
-	// default to min 1 version, just so we don't have to account for nil slice.
-	// A single version kept in memory is pretty harmless.
-	if c.KeepRecent < 1 {
-		c.KeepRecent = 1
+	if c.KeepRecent < 1 || c.KeepRecent > 10 {
```
wwestgarth, line 55: I'm not sure of this restriction because it is useful forto able to keep more than 10 locally (even if 10 is the limit for listing during statesync). For example the `snapshot-soak-tests` rely on being able to keep hundreds of snapshots locally.

core/snapshot/config.go
```diff
@@ -39,16 +46,18 @@ func NewDefaultConfig() Config {
...
+		return ErrKeepRecentMustBeBetween1And10
+	}
+
+	if c.StartHeight < 0 {
```
wwestgarth, line 59: This is going to cause a lot of mayhem given that the system-tests/devops scripts etc will all be using `-1` at the moment. Maybe we can do:
```
if c.StartHeight < 0 {
    log.Warn("some warning to tell people to update their configs")
    c.StartHeight = 0
}
```
as a first step?

ValentinTrinque: I don't think anybody is monitoring the logs for such things. A more effective way would be to run some kind of migration / update script before or upon start up. But I don't think we have any of this, so it's very likely we are just going to delay the inevitable.

wwestgarth: Yea thats a fair point. I've kicked off a full system-test so we can see which tests will be affected:
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/23783/pipeline

ValentinTrinque: Killing it and rerunning it based on latest changes.

https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/23792/pipeline

ValentinTrinque:
```{
"system-tests": fix/align-snapshot-config-change"
}```

wwestgarth (dismissed):

jeremyletang (approved):

PR #8733, opened 2023-07-08
chore: sla interface cleanup

ze97286:
No functional changes (apart from adding a missing event emission for amend) - this is just moving things around to make it suitable for integration with the execution package.


core/liquidity/v2/engine.go
```diff
@@ -430,19 +410,18 @@ func (e *Engine) CalculateSuppliedStake() *num.Uint {
...
 	return supplied
 }
 
 func (e *Engine) IsPoTInitialised() bool {
 	return e.suppliedEngine.IsPoTInitialised()
 }
 
-func (e *Engine) UpdateMarketConfig(model risk.Model, monitor PriceMonitor) {
+func (e *Engine) UpdateMarketConfig(model RiskModel, monitor PriceMonitor, slaParams *types.LiquiditySLAParams) {
+	e.onSLAParamsUpdate(slaParams)
```
karlem, line 424: Why have you decided that the only way to change the SLA params should be through the `UpdateMarketConfig` method?

karlem (commented):
Hey @ze97286, just couple of questions. Otherwise it looks good!

core/liquidity/v2/amendments.go
```diff
@@ -16,23 +16,21 @@ import (
...
-) error {
+func (e *Engine) AmendLiquidityProvision(ctx context.Context, lpa *types.LiquidityProvisionAmendment, party string) error {
 	if err := e.CanAmend(lpa, party); err != nil {
 		return err
 	}
 
 	// LP exists, checked in the previous func.
 	lp, _ := e.provisions.Get(party)
 	updatedLp := e.createAmendedProvision(lp, lpa)
+	e.broker.Send(events.NewLiquidityProvisionEvent(ctx, updatedLp))
```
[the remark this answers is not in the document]
ze97286: well, it should be triggered with pending for sure, also I'll change it a bit because it may be applied immediately. I'll also fix submit because in there it's emitted with active immediately rather than pending

core/liquidity/v2/engine.go
ze97286: because it's **market** parameters. how else can it be changed?

core/liquidity/v2/engine.go
```diff
@@ -488,10 +453,14 @@ func (e *Engine) OnNonPerformanceBondPenaltyMaxUpdate(nonPerformanceBondPenaltyM
...
 }
 
-func (e *Engine) OnPerformanceHysteresisEpochsUpdate(performanceHysteresisEpochs uint) {
-	e.performanceHysteresisEpochs = performanceHysteresisEpochs
-
-	for _, performance := range e.slaPerformance {
-		performance.previousPenalties.ModifySize(e.performanceHysteresisEpochs)
+func (e *Engine) onSLAParamsUpdate(slaParams *types.LiquiditySLAParams) {
+	one := num.DecimalOne()
+	e.openPlusPriceRange = one.Add(slaParams.PriceRange)
```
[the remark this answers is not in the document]
ze97286: yes. it's a market parameter and it's validated on the way in. it's passed as string, and zero is not a valid value anyways. 

```
lppr, err := num.DecimalFromString(config.PriceRange)
if err != nil {
   errs.AddForProperty(fmt.Sprintf("%s.price_range", parent), ErrIsNotValidNumber)
} else if lppr.IsNegative() || lppr.IsZero() {
   errs.AddForProperty(fmt.Sprintf("%s.price_range", parent), ErrMustBePositive)
} else if lppr.GreaterThan(num.DecimalFromInt64(100)) {
   errs.AddForProperty(fmt.Sprintf("%s.price_range", parent), ErrMustBeAtMost100)
}
```

core/liquidity/v2/engine.go
ze97286: point is the only way to change any of these params that is a market parameter is through a governance update which must include everything (the risk model, price monitor, and sla params, etc.) so there's not really a point in having separate functions to change each one individually.

karlem (approved):

PR #8771, opened 2023-07-11
fix: cancel all orders on market rejection

### LINKED ISSUE #8739: [Bug]: orders not cancelled when successor market is rejected
### Problem encountered

please see the [feature test](https://github.com/vegaprotocol/vega/pull/8736):

in this test, market ETH20 and ETH21 are both successor market after parent ETH19; ETH20 get out of opening auction first; ETH21 is rejected, hence:

- LP commitment is released….which is correct
- orders placed during opening auction, asset is released from margin account….which is correct
- **orders placed during opening auction are cancelled…the behaviour is: the order is still STATUS_ACTIVE, which is a bug** 

### Observed behaviour

Orders are still **active** when market is rejected 

### Expected behaviour

orders should be **STATUS_STOPPED** which is inline with when market is settled 

### Steps to reproduce

```bash
please see the feature test
```

### Software version

v0.71.6

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #8739

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -3918,6 +3918,17 @@ func (m *Market) cleanupOnReject(ctx context.Context) {
 			logging.Error(err))
 	}
 
+	// cancel all pending orders
+	orders := m.matching.CancelOrdersOnReject()
+	evts := make([]events.Event, 0, len(orders))
+	for _, o := range orders {
+		evts = append(evts, events.NewOrderEvent(ctx, o))
+	}
+	if len(evts) > 0 {
+		m.broker.SendBatch(evts)
+	}
+
+	// release margin balance
 	tresps, err := m.collateral.ClearMarket(ctx, m.GetID(), m.settlementAsset, parties, false)
 	if err != nil {
 		m.log.Panic("unable to cleanup a rejected market",
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -3919,10 +3919,23 @@ func (m *Market) cleanupOnReject(ctx context.Context) {
 	}
 
 	// cancel all pending orders
-	orders := m.matching.CancelOrdersOnReject()
-	evts := make([]events.Event, 0, len(orders))
-	for _, o := range orders {
+	orders := m.matching.Settled()
+	// stop all parkedPeggedOrders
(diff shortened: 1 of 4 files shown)


core/execution/future/market.go
```diff
@@ -3918,6 +3918,17 @@ func (m *Market) cleanupOnReject(ctx context.Context) {
...
+	// cancel all pending orders
+	orders := m.matching.CancelOrdersOnReject()
+	evts := make([]events.Event, 0, len(orders))
+	for _, o := range orders {
+		evts = append(evts, events.NewOrderEvent(ctx, o))
+	}
+	if len(evts) > 0 {
+		m.broker.SendBatch(evts)
+	}
+
```
wwestgarth, line 3930: We also need do parked-pegged orders and also probably stop orders too.

core/matching/orderbook.go
```diff
@@ -704,6 +704,26 @@ func (b *OrderBook) RemoveOrder(id string) (*types.Order, error) {
 	return order, nil
 }
 
+func (b *OrderBook) CancelOrdersOnReject() []*types.Order {
+	buys, sells := b.getSide(types.SideBuy).getLevels(), b.getSide(types.SideSell).getLevels()
+	ret := []*types.Order{}
+	for _, lvl := range buys {
+		for _, o := range lvl.orders {
+			o, _ = b.DeleteOrder(o)
+			o.Status = types.OrderStatusCancelled
```
wwestgarth, line 713: I don't think CANCELLED is thr correct status. Both in the spec at the protos definitions, CANCELLED is related to a user actively cancelling their order and STOPPED is for when the network makes that decision:
```
    // Used for orders cancelled by the party that created the order
    STATUS_CANCELLED = 3;
    // Used for unfilled FOK or IOC orders, and for orders that were stopped by the network
    STATUS_STOPPED = 4;
```

also the orderbook has the `Settled()` method cleans up orders already, so I don't think we need two ways to clean up all orders and set a particular status.

jeremyletang: stopped would be the correct

wwestgarth (approved):

PR #8776, opened 2023-07-12
fix: normalisation of eth calls that return structs

### LINKED ISSUE #8417: Ethereum Oracles: Normaliation
An Ethereum contract call returns a list of arbitrarily typed objects

The data sourcing/filtering system needs a map of string, string key value pairs.

Extend the Ethereum call spec with a way of extracting named bits of data from the call result, probably using something like [jsonpath](https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html).

pscott31:
closes #8417

.

### DIFF
--- a/core/evtforward/ethcall/result.go
+++ b/core/evtforward/ethcall/result.go
@@ -1,10 +1,15 @@
 package ethcall
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
+	"text/scanner"
 
 	"code.vegaprotocol.io/vega/core/types"
+	"github.com/PaesslerAG/gval"
 	"github.com/PaesslerAG/jsonpath"
 )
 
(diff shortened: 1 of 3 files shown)


core/evtforward/ethcall/result.go
```diff
@@ -49,18 +51,43 @@ func newResult(call Call, bytes []byte) (Result, error) {
...
 			return nil, fmt.Errorf("unable to normalise key %v: %v", key, err)
 		}
-		valueJson, err := json.Marshal(value)
-		if err != nil {
-			return nil, fmt.Errorf("unable to serialse normalised value: %v", err)
+		switch v := value.(type) {
+		case json.Number:
+			res[key] = v.String()
+		case float64:
+			res[key] = strconv.FormatFloat(v, 'f', -1, 64)
```
jeremyletang, line 85: wouldn't that be eventual source of non determinism, can we instead just get the string of the float64 directly?

pscott31: ah well spotted; so this is for the very specific case where a 'static' value is given in the path itself. any numbers that are in the eth call result get unmarshalled as a json.Number, which is what we want. However, you can give a reques a 'path' to a static value like

```normalisers = {"number_one":1}```

which on the face of it seems rather useless but was actually handy when doing the demo; because as things are set up currently you can only settle a market based on (iirc) a boolean or a timestamp.. 

Our current data processing pipeline has no way for us to actually change the values; only filter them; so if you set up e.g. a termination signal like

```json
                                        "filters": [
                                            {
                                                "key": {
                                                    "name": "myproperty",
                                                    "type": "TYPE_INTEGER"
                                                },
                                                "conditions": [
                                                    {
                                                        "operator": "OPERATOR_GREATER_THAN",
                                                        "value": "100"
                                                    }
                                                ]
                                            },
```

you can't use myproperty as a termination signal, because it is an integer.. so I hacked around that by adding a 'static' json path:

```json
                                        "normalisers": {
                                            "myproperty": "$[0]",
                                            "shouldTerminate": "true"
                                        },
```
and using `shouldTerminate` as the termination property; which works because only if all the filters match does it get broadcast.

That whole thing is a bit grotty, I think we really want some pipeline of 'transformers', so e.g. when we filter we can put the boolean result of the filter as a new property in its own right or something.. but even then, I thought it might be handy to be able to add static values as paths; I dunno maybe an identifier or scaling factor or somthing that is unique to the contract but perhaps not part of the contract call.

so.. back to the actual issue you raised; I don't think it's non deterministic because you only get to the `case float64:` here if the number is generated by the JSONPath itself; all numbers in the ethcall result are kept as strings, wrapped in json.Number because of

```go
	d := json.NewDecoder(bytes.NewBuffer(valuesJson))
	// Keep numbers as a json.Number, which holds the original string representation
	// otherwise all numbers get cast to float64 which is no good for e.g uint256.
	d.UseNumber()
```

Though perhaps I could make that clearer in the code; if you are happy in principle I'll add a comment

pscott31: OK I came up with a different way to do it involving no floats and a bit of poking around in the jsonpath interpreter

jeremyletang (approved):

PR #8795, opened 2023-07-13
fix: token permissions and versions on PM actions

gordsport:
This PR updates the GH action tokens and version numbers of external actions


wwestgarth (approved):

daniel1302 (approved):
Looks good action worked for that PR, so I think it's safe to merge.

Issue #8753, opened 2023-07-11
[Feature]: Perps - Governance


wwestgarth:
# Feature Overview

**In order to** Allow markets with a perpetuals product to be proposed
**We will** extend the `NewMarket` proposals and relevant proto-definitions
**So that** a valid proposal can be constructed and submitted to create a perpetual's market

## Specs
- [Link](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0053-PERP-product_builtin_perpetual_future.md) to spec or section within a spec

## Tasks
A checklist of the tasks that are needed to develop the feature and meet the acceptance criteria and feature test scenarios.
- [ ]
- [ ]

## Feature test scenarios
Detailed scenarios that can be executed as feature tests to verify that the feature has been implemented as expected.

GIVEN (setup/context) 
WHEN (action) 
THEN (assertion) For example...
See [here](https://github.com/vegaprotocol/vega/tree/develop/core/integration) for more format information and examples.

### Additional Details (optional)
Any additional information that provides context or gives information that will help us develop the feature.

Issue #8763, opened 2023-07-11
[Feature]: Perps - Datanode API: add a new endpoint to view periodic settlement data


wwestgarth:
# Feature Overview

**In order to** allow uses of Vega to see how periodic settlement funding payments are being calculated
**We will** add an new API / extend existing API
**So that** this data is available from any data node

Note: we should get input from FE early.

All data-points and subsequent calculated funding payments should be stored in the datanode and queryable from a new datanode API. I'm imagining (although it doesn't have to look like this at all) something that returns a list of:
```
{ funding_intervals: [
    [
        {
        seq: 13
        period_start: 20
        period_end: "" # because it hasn't finished yet
        funding_payment: ""
        data_points : [
              {
                  price: 100,
                  type: external
              } ....
          ]
        }
        {
        seq: 12
        period_start: 10
        period_end: 20
        funding_payment: 100
        funding_rate: 102
        data_points : [
              {
                  price: 110
                  type: internal
              } ....
          ]
        }
}
```
Although I don't know how we would paginate if its like that, so something to think about.

## Specs
- [Link](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0048-DSRI-data_source_internal.md) to spec or section within a spec

## Tasks
A checklist of the tasks that are needed to develop the feature and meet the acceptance criteria and feature test scenarios.
- [ ]
- [ ]

## Feature test scenarios
Detailed scenarios that can be executed as feature tests to verify that the feature has been implemented as expected.

GIVEN (setup/context) 
WHEN (action) 
THEN (assertion) For example...
See [here](https://github.com/vegaprotocol/vega/tree/develop/core/integration) for more format information and examples.

### Additional Details (optional)
Any additional information that provides context or gives information that will help us develop the feature.

PR #8799, opened 2023-07-14
feat: introduce perpetual and funding payment calculations

### LINKED ISSUE #8754: [Feature]: Perps - Settlement engine funding payment calculations
# Feature Overview

**In order to** calculate the transfers for each party in a perpetuals market
**We will** store incoming data-points in the settlement engine
**So that** when needed, the funding payment can be calculated and the necessary transfers returned sent to the collateral engine

This will happily sit alongside the MTM functionality in the settlement engine.

We need to add functionality that will handle the incoming periodic settlement data-points. For each settlement data-point that is received we need to save it, and then for each settlement schedule event we need to calculate the funding rate for each party with a position.

Any MTM that happens within settlement-period will need to create a new data-point using the "spot" price from any previous data-point.

All data-points and funding payment calculation results should be emitted as an event ready for a new datanode API.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
closes #8754 

Introduces the perpetual products, but not plumbed into anything at all. This PR gives us:
- the funding rate calculations based on the twap difference between mark-prices and index-prices
- the events sent to the data node related to funding-periods and their data-points
- a function on the settlement engine which takes the funding-payment and positions, and returns win/loss transfers to be made

Some more plumbing from the otherside is coming in: https://github.com/vegaprotocol/vega/issues/8753

Also note there is a current spec update to add clamped-interest to the funding payment. Its still in the review stage but will simply be a few extra product-parameters and a ` + clamp(e^{r*t_d}...)` thing at the end of the payment calcuation. I won't be and a major overhaul or anything.


core/events/bus.go
```diff
@@ -141,6 +141,8 @@ const (
 	ExpiredOrdersEvent
 	DistressedPositionsEvent
 	StopOrderEvent
+	FundingPeriodEvent
```
ze97286, line 144: don't we also need to add the events to the datanode/broker/convert.go?

libs/num/uint.go
```diff
@@ -111,6 +111,15 @@ func UintFromDecimal(d Decimal) (*Uint, bool) {
 	return &Uint{*u}, ok
 }
 
+// UintFromUint64 allows for the conversion from uint64.
+func UintFromUint64(ui uint64) *Uint {
```
jeremyletang, line 115: How does it compare with NewUint()?

libs/num/numeric.go
```diff
@@ -7,12 +7,13 @@ import (
...
 	asUint    *Uint
 	asDecimal *Decimal
 }
 
 func (n *Numeric) Clone() *Numeric {
-	if n.asUint != nil {
+	if n.asUint != nil || n.asInt != nil {
 		nn := n
 		return nn
 	}
```
jeremyletang, line 19: that doesn't Clone does it? I think we should be calling the .Clone() function on those types to actualy clone.

libs/num/uint.go
wwestgarth: Removed it and used `NewUint()` instead. Didn't see it originally because I was looking for a `UintFromBlah()` named function.

core/products/products.go
```diff
@@ -52,16 +58,23 @@ type Product interface {
 	UnsubscribeTradingTerminated(ctx context.Context)
 	UnsubscribeSettlementData(ctx context.Context)
 	RestoreSettlementData(*num.Numeric)
+	OnLeaveOpeningAuction(context.Context, int64)
+
+	// tell the product about an internal data-point such as a the current mark-price
+	SubmitDataPoint(context.Context, *num.Uint, int64) error
```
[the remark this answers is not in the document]
ze97286: it's just a bunch of hacks and patches. It needs to be looked at with a fresh eye I think.

core/events/bus.go
```diff
@@ -141,6 +141,8 @@ const (
...
+	FundingPeriodDataPointEvent
```
ze97286, line 145: also, do we need to add it to the gql model in gqlgen.yml?

wwestgarth: Probably, but I think we're slowly creeping into this ticket really:
https://github.com/vegaprotocol/vega/issues/8763

ze97286: yeah, sure it doesn't need to be in this PR.

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,45 @@ import "vega/vega.proto";
 
 option go_package = "code.vegaprotocol.io/vega/protos/vega/events/v1";
 
+// Event notifying the details of a funding interval of a perpetuals market
```
candida-d, line 16:
```suggestion
// Event notifying on the details of a funding interval for a perpetuals market.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,45 @@ import "vega/vega.proto";
...
+message FundingPeriod {
+  // ID of the market for which this funding period relates to.
+  string market_id = 1;
+  // Sequence number of the funding period
```
candida-d, line 20:
```suggestion
  // Sequence number of the funding period.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,45 @@ import "vega/vega.proto";
...
+  uint64 seq = 2;
+  // Time in Unix nanosecond when the funding period started.
+  int64 start = 3;
+  // Time in Unix nanosecond when the funding period ended.
```
candida-d, line 24: Will this only ever be seen in the past?

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,45 @@ import "vega/vega.proto";
...
+  optional int64 end = 4;
+  // Funding payment for this period as the difference between the time-weighted-average-price of the external and internal data point.
```
candida-d, line 26:
```suggestion
  // Funding payment for this period as the difference between the time-weighted average price of the external and internal data point.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,45 @@ import "vega/vega.proto";
...
+  optional string funding_payment = 5;
+  // Percentage difference between the time-weighted-average-price of the external and internal data point.
```
candida-d, line 28:
```suggestion
  // Percentage difference between the time-weighted average price of the external and internal data point.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,45 @@ import "vega/vega.proto";
...
+  optional string funding_rate = 6;
+}
+
+// Event notifying a data-point for a funding period.
```
candida-d, line 32:
```suggestion
// Event notifying a data point for a funding period.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -883,6 +928,10 @@ message BusEvent {
     DistressedPositions distressed_positions = 157;
     // A stop order event
     StopOrderEvent stop_order = 158;
+    // Start or end of a funding period.
+    FundingPeriod funding_period = 159;
+    // Data point withint a funding period.
```
candida-d, line 933:
```suggestion
    // Data point within a funding period.
```

core/products/perpetual.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+		return carry
+	}
+
+	// carry over data-points at times > endAt and the first data-points that is <= endAt
+	p.external = carryOver(p.external)
+	p.internal = carryOver(p.internal)
+
+	// send events for all the data-points that were carried over
+	for _, dp := range p.external {
```
EVODelavega, line 213: would be cleaner to create a slice and `SendBatch` all the events here

EVODelavega (dismissed):

protos/sources/vega/events/v1/events.proto
EVODelavega: This will only be set if the funding period has ended, so yes

EVODelavega (dismissed):

EVODelavega (approved):

candida-d (approved):

PR #8802, opened 2023-07-14
fix: stop orders api always returns null for expiresAt

### LINKED ISSUE #8800: Stop orders API is returning null for expiresAt when expiryStrategy is set
https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1689324365991729 for reference.

Investigation suggests this appears to be a data node issue as the `expiresAt` property is not being mapped properly before being inserted into the database.

guoguojin:
Closes #8800 

This was caused by the parsing of the event from the event bus before being inserted into the database. The value of expiresAt was not checked and updated if it isn't null, so it was always being inserted into the database as null.

### DIFF
--- a/datanode/entities/stop_orders.go
+++ b/datanode/entities/stop_orders.go
@@ -172,6 +172,10 @@ func StopOrderFromProto(so *pbevents.StopOrderEvent, vegaTime time.Time, seqNum 
 		ocoLinkID = StopOrderID(*so.StopOrder.OcoLinkId)
 	}
 
+	if so.StopOrder.ExpiresAt != nil {
+		expiresAt = ptr.From(NanosToPostgresTimestamp(*so.StopOrder.ExpiresAt))
+	}
+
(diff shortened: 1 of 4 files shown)


datanode/entities/stop_orders.go
```diff
@@ -172,6 +172,13 @@ func StopOrderFromProto(so *pbevents.StopOrderEvent, vegaTime time.Time, seqNum
 		ocoLinkID = StopOrderID(*so.StopOrder.OcoLinkId)
 	}
 
+	if so.StopOrder.ExpiresAt != nil {
+		expiresAt = ptr.From(NanosToPostgresTimestamp(*so.StopOrder.ExpiresAt))
+		if expiresAt.Before(vegaTime) {
```
jeremyletang, line 177: Can we remove that if? That would return an error once the stop order get updated at the time it expires

jeremyletang (approved):

PR #8833, opened 2023-07-18
fix: enforce transaction age on validator transactions and fix pow ca…

### LINKED ISSUE #8827: [Investigation + fix]: proof of work in the snapshots
### Problem encountered

The issue is that in the original mainnet snapshot in the leaf proofOfwork.tidAtHeight I can see the latest transactions for the height 7007898 , but those transactions(and about 92 earlier transactions) are missing in the new checkpoint.

### Observed behaviour

Is this behavior expected? I am not sure how this pow should behave.
In the attachment, I am attaching those 2 different sections. You can see also the diff on the attached screen.

image

### Expected behaviour

TBC

### Steps to reproduce

```bash
1. I am downloading the mainnet snapshot(from our data node)
2. I am preparing the null-chain network(single-node validator)
3. I am loading the mainnet snapshot at a given height- on the latest available snapshot (here it is StartHeight = 7007900)
4. I am moving the network forward for a few hundred blocks( at the moment my network is on the block: "blockHeight": "7008560",  - it means 660 blocks after mainnet snapshot)
5. I am getting the latest produced snapshot from my null chain blocks, then comparing the proofOfWork section between the two above snapshots (original mainnet snapshot snapshot-before.json and new null-chain network snapshot snapshot-after.json
```

### Software version

from 0.71.3 to 0.71.4

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
…che cleanup

Closes #8827 

1. one off - clean old transactions from pow cache whose block height refers to inactive param set
2. enforce transaction age constraint on validator commands
3. every 1000 blocks make sure no old transactions remain in cache


core/pow/engine.go
```diff
@@ -157,6 +158,26 @@ func (e *Engine) EndOfBlock() {
...
+					delete(e.seenTid, v)
+				}
+				delete(e.heightToTx, block)
+				delete(e.heightToTid, block)
+			}
+		}
+	}
+
 	toDelete := []int{}
 	// iterate over parameters set and clear then out if ther's not relevant anymore.
```
candida-d, line 182:
```suggestion
	// iterate over parameters set and clear then out if they're not relevant anymore.
```

I realise this is out of scope of the PR but this typo hurts my eyes

ze97286: I hear you :)

candida-d (dismissed):

jeremyletang (dismissed):

candida-d (dismissed):

AnExsomnis (approved):

EVODelavega (approved):

candida-d (approved):

Issue #8837, opened 2023-07-18
[Bug]: Soak test fail on successors map data


EVODelavega:
### Problem encountered

A snapshot containing successors data when the parent market has been rejected, or its state has expired incorrectly restores the successors map

### Observed behaviour

When restoring a snapshot that contains a parent/successor reference involving a cancelled/rejected/expired parent market will cause consensus failures if restored

### Expected behaviour

The successors map should not cause consensus failures

### Steps to reproduce

```bash
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsnapshot-soak-tests/detail/snapshot-soak-tests/8893/pipeline/
```

### Software version

develop

PR #8841, opened 2023-07-18
fix: skip parent markets that are no more

EVODelavega:
Cherry-picked fix for #8837

### DIFF
--- a/core/execution/engine_snapshot.go
+++ b/core/execution/engine_snapshot.go
@@ -162,6 +162,9 @@ func (e *Engine) serialise() (snapshot []byte, providers []types.StateProvider, 
 	})
 	successors := make([]*types.Successors, 0, len(e.successors))
 	for pid, ids := range e.successors {
+		if _, ok := e.GetMarket(pid, true); !ok {
+			continue
+		}
 		successors = append(successors, &types.Successors{
 			ParentMarket:     pid,
 			SuccessorMarkets: ids,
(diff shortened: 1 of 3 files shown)


core/execution/engine_snapshot.go
```diff
@@ -162,6 +162,9 @@ func (e *Engine) serialise() (snapshot []byte, providers []types.StateProvider,
 	})
 	successors := make([]*types.Successors, 0, len(e.successors))
 	for pid, ids := range e.successors {
+		if _, ok := e.GetMarket(pid, true); !ok {
```
ze97286, line 165: how is it possible that we have to do it when serialising? are we not removing from the successors properly if/when a market is gone? wouldn't there be other potential side effects if we're trying to access the market of a successor before the snapshot block and it's not there?

EVODelavega: It seems to be related to parent markets getting rejected. The successor map gets updated for cancelled markets, but only `if m, ok := e.futureMarkets[id]; ok && m.Mkt().State == types.MarketStateCancelled`. If the parent market is rejected, the successor map doesn't get updated immediately (happens next time update).
If the successor market itself gets rejected, or stays in proposed state until the next snapshot is taken, then the successor map doesn't seem to get updated in time, and the next snapshot after a restore will likely contain the outdated successor entries.

There's quite a few things that have to happen for this issue to manifest itself (and they can, of course, but due to the nature of the system tests, they're more likely to be seen there - which is a good thing). We could add a lot of checks in the `OnTick` and `removeMarket` calls throughout, but I felt like that would add a lot of complexity and bloat for a very specific issue that can quite easily be addressed with a simple check here

PR #8851, opened 2023-07-19
chore: implement internal time trigger

### LINKED ISSUE #8847: Implement TimeTrigger internal data source
Implement time triggered internal data source as per:
https://github.com/vegaprotocol/specs/blob/master/protocol/0048-DSRI-data_source_internal.md#12-time-triggered

Use existing time based type as an example/base for the types.

AnExsomnis:
close: #8847


protos/sources/vega/data/v1/spec.proto
```diff
@@ -68,3 +72,12 @@ message Condition {
...
   }
 }
+
+// Trigger for an internal time data source.
+message InternalTimeTrigger {
+  // Trigger when the vega time is greater or equal to this time, in Unix seconds.
+  optional string initial = 1;
+  // Repeat the trigger every n seconds after the initial. If no time for
+  // initial was specified, begin repeating immediately.
+  string every = 2;
```
jeremyletang, line 82: int64 here.

AnExsomnis: This has been address, thanks!

core/datasource/common/time.go
```diff
@@ -0,0 +1,144 @@
...
+		Initial: i.Initial,
+		Every:   i.Every,
+	}
+}
+
+func (i InternalTimeTrigger) IsTriggered(now time.Time) (bool, error) {
+	// Initial is not set - it will never be triggered.
+	if i.Initial == "" {
+		return false, nil
+	}
```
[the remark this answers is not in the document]
AnExsomnis: Yes, removed that step.
Thanks!

core/datasource/common/time.go
```diff
@@ -0,0 +1,144 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	datapb "code.vegaprotocol.io/vega/protos/vega/data/v1"
+)
+
+type InternalTimeTrigger struct {
+	Initial string
+	Every   string
```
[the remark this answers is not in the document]
AnExsomnis: This has been addressed, thanks!

protos/sources/vega/data_source.proto
```diff
@@ -24,12 +24,21 @@ message DataSourceSpecConfigurationTime {
   repeated vega.data.v1.Condition conditions = 1;
 }
 
+// Internal data source used for emitting timestamps automatically by predefined intervals and conditions.
```
candida-d, line 27:
```suggestion
// Internal data source used for emitting timestamps automatically using predefined intervals and conditions.
```

AnExsomnis: Addressed.

core/datasource/errors/errors.go
```diff
@@ -39,4 +39,8 @@ var (
 
 	// ErrDuplicatedEthereumCallEvent is returned when there is a duplicated Ethereum event.
 	ErrDuplicatedEthereumCallEvent = errors.New("duplicated call event")
+
+	// ErrDuplicatedEthereumCallEvent is returned when no deterministic time is given to set the next time trigger
+	// repetition.
+	ErrMissingTimeForSettingTriggerRepetiotion = errors.New("missing time for setting trigger repetition")
```
jeremyletang, line 45:
```suggestion
	ErrMissingTimeForSettingTriggerRepetition = errors.New("missing time for setting trigger repetition")
```

core/datasource/definition/definition.go
```diff
@@ -116,7 +125,7 @@ func (s *Definition) Content() interface{} {
 
 // FromProto tries to build the Definiition object
 // from the given proto object.
-func FromProto(protoConfig *vegapb.DataSourceDefinition) (common.DataSourceType, error) {
+func FromProto(protoConfig *vegapb.DataSourceDefinition, tm ...time.Time) (common.DataSourceType, error) {
```
jeremyletang, line 128: Why are we passing a variadic list of time.Time here?

AnExsomnis: Because DataSourceDefinition is the top level data source object that contains all of the others, and here it needs also to handle the `timetrigger.SpecConfigurationFromProto(dtp, tm)` method which needs time provided in order to update the nextTrigger value

core/datasource/common/time.go
```diff
@@ -0,0 +1,173 @@
...
+func (i InternalTimeTrigger) IsTriggered(timeNow time.Time) bool {
+	if i.NextTrigger.Before(timeNow) {
+		*i.NextTrigger = i.NextTrigger.Add(time.Duration(i.Every) * time.Second)
+		return true
+	}
+
+	return false
+}
+
+func (i InternalTimeTrigger) SetNextTrigger(timeNow time.Time) {
```
jeremyletang, line 81: this should probably ne private, and called only on the constructor / FromProto method.

AnExsomnis: In that case there will be a getter for this field, otherwise it cannot be tested - is this what you mean ?

core/datasource/errors/errors.go
AnExsomnis: Corrected, thanks!

protos/vega/data_source.go
```diff
@@ -36,6 +36,13 @@ func NewDataSourceSpec(sc *DataSourceDefinition) *DataSourceSpec {
 		case *DataSourceDefinition_Internal:
 			in := sc.GetInternal()
 			if in != nil {
+				switch in.SourceType.(type) {
+				case *DataSourceDefinitionInternal_Time:
+					//t := in.GetTime()
+
+				case *DataSourceDefinitionInternal_TimeTrigger:
+					//t := in.GetTimeTrigger()
```
jeremyletang, line 44: why are those commented out?

protos/vega/spec_definition.go
```diff
@@ -76,6 +77,20 @@ func (d DataSourceDefinition) GetFilters() []*datapb.Filter {
 					)
 				}
 			}
+		case *DataSourceSpecConfigurationTimeTrigger:
+			if tp != nil {
+				if len(tp.Conditions) > 0 {
+					filters = append(filters,
+						&datapb.Filter{
+							Key: &datapb.PropertyKey{
+								Name: "vegaprotocol.builtin.timetrigger",
```
jeremyletang, line 86: we could probably use the exported const from the timetrigger package hjere.

core/datasource/common/time.go
jeremyletang: well, IsTriggered is the way to test it isn't it?

core/datasource/definition/definition.go
jeremyletang: yea but you need only one time isn't it? which is the current time.

protos/vega/spec_definition.go
AnExsomnis: That would create a cycle import because the timetrigger already imports this package as ``vegapb "code.vegaprotocol.io/vega/protos/vega"``

protos/vega/data_source.go
AnExsomnis: Because they are not used anywhere yet, this method is created for testing purposes.

core/datasource/spec/spec.go
```diff
@@ -71,6 +71,13 @@ func New(originalSpec datasource.Spec) (*Spec, error) {
 		}
 	}
 
+	builtInTrigger := false
+	for _, f := range filtersFromSpec {
+		if strings.HasPrefix(f.Key.Name, "vegaprotocol.timetrigger") && f.Key.Type == datapb.PropertyKey_TYPE_TIMESTAMP {
```
jeremyletang, line 76: Is that correct or should it be vegaprotocol.builtin.timetrigger ?

AnExsomnis: Yes, this is incorrect, it should be ` vegaprotocol.builtin.timetrigger `, missed that. 
Thanks

core/datasource/common/time.go
AnExsomnis: As testing the boolean result yes, i meant the setting of the actual value of this trigger
Do we want to test that as well ?

jeremyletang: hum I don't think it matters really, at the end of the day you want to test the overall behaviour of the type, not the actual value it contains, which is meant to be abstracted away I suppose.

core/datasource/definition/definition.go
AnExsomnis: Yes it takes only the first position.
The idea was to have the option to send `nil`; when you reach the step of the timetrigger you have this
```
func SpecConfigurationFromProto(protoConfig *vegapb.DataSourceSpecConfigurationTimeTrigger, tm []time.Time) (SpecConfiguration, error) {
	if tm == nil || len(tm) < 1 {
		return SpecConfiguration{}, errors.ErrMissingTimeForSettingTriggerRepetition
	}
	if protoConfig == nil {
		return SpecConfiguration{}, nil
	}

	return SpecConfiguration{
		Triggers:   common.InternalTimeTriggersFromProto(protoConfig.Triggers, tm[0]), // Use one position for now
		Conditions: common.SpecConditionsFromProto(protoConfig.Conditions),
	}, nil
}
```

core/datasource/common/time.go
AnExsomnis: All right, I'll remove it then

AnExsomnis: This has been corrected now, as well as the tests.

core/datasource/spec/spec.go
AnExsomnis: Corrected now

jeremyletang (approved):

PR #8866, opened 2023-07-20
fix: liquidity provision bug fixes and wiring of target stake

### LINKED ISSUE #8853: [Bug]: Target stake is wrong, and LP amendment is not working (SLA in Spot market)
### Problem encountered

please see the [feature file](https://github.com/vegaprotocol/vega/pull/8828/files)

- Target stake is wrong on line 70
- LP can not reduce the LP commitment on line 83

### Observed behaviour

- Target stake is always 0 on line 70 (when I change scaling factor, target stake stays at 0)
- LP can not reduce the LP commitment on line 83 (when reduce LP commitment from 2000 to 20, LP ends up at 1000)

### Expected behaviour

- Target stake should be non-zero for this scenario is scaling factor is non-zero
- LP should be able to reduce LP commitment as long as supplied stake is over target stake when market is in continuous trading mode. 

### Steps to reproduce

```bash
please see the feature test
```

### Software version

v0.71.6

### Failing test

https://github.com/vegaprotocol/vega/pull/8828/files

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
Closes #8853


karlem (approved):
Thanks for the fix!

PR #8869, opened 2023-07-20
fix: 0004_oracle_specs.sql migration script needs default value for n…

### LINKED ISSUE #6361: ERROR unable to get remove expired orders	{"market-id": "2a1da427eb0817d7e24f2eb63be586dc1969adb294d6443fb75088b494b80e41", "error": "trading not allowed"}
These log messages occur multiple times every seconds in core logs,  on every node

for e.g. http://18.134.160.20:4646/ui/allocations/409c8f4d-e4bd-4acf-70c4-b77bd5d3dc46/vega-full-2/logs

Can we look at reducing these or removing them. Its obvious a condition is failing somewhere but if it occurs so frequently, should be just either remove the check or change to INFO?

If we log an ERROR then its something validators should actually care about.

guoguojin:
…ot null

close #6361

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
 - [8853](https://github.com/vegaprotocol/vega/issues/8853) - Liquidity provision amendment bug fixes
 - [8862](https://github.com/vegaprotocol/vega/issues/8862) - Fix settlement via governance
 - [8772](https://github.com/vegaprotocol/vega/issues/8772) - Checkpoint panic on successor markets.
+- [8837](https://github.com/vegaprotocol/vega/issues/8837) - Remove successor entries from snapshot if they will be removed next tick.
 
 
 ## 0.72.1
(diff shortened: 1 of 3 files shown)


datanode/sqlstore/migrations/0004_oracle_specs.sql
```diff
@@ -1,4 +1,9 @@
 -- +goose Up
 ALTER TABLE oracle_specs DROP COLUMN IF EXISTS signers;
 ALTER TABLE oracle_specs DROP COLUMN IF EXISTS filters;
-ALTER TABLE oracle_specs ADD COLUMN data JSONB NOT NULL;
\ No newline at end of file
+ALTER TABLE oracle_specs ADD COLUMN data JSONB NOT NULL default '{}';
```
jeremyletang, line 4: Any chance this could break queries to the database if the entities have non nullable fields?

guoguojin: I think if the database migration script wanted the column to be not null, then the Go could would expect the value to be not null and therefore handle it correctly. All we're doing is assigning an empty JSON object to the field if we are given a null value. The problem would be elsewhere, i.e. why are we being given a null value for this data

Issue #8810, opened 2023-07-17
SLA Feature tests: 0044-LIME


gordsport:
Spec:

- https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0044-LIME-lp_mechanics.md

ACs:

- [ ] 0044-LIME-013
- [ ] 0044-LIME-014
- [ ] 0044-LIME-015
- [ ] 0044-LIME-016
- [ ] 0044-LIME-018
- [ ] 0044-LIME-019
- [ ] 0044-LIME-020
- [ ] 0044-LIME-021
- [ ] 0044-LIME-030
- [ ] 0044-LIME-031
- [ ] 0044-LIME-022
- [ ] 0044-LIME-023
- [ ] 0044-LIME-030 - review if this is already covered and if so remove AC or tag test twice
- [ ] 0044-LIME-024
- [ ] 0044-LIME-031 - review if this is already covered and if so remove AC or tag test twice
- [ ] 0044-LIME-025
- [ ] 0044-LIME-026
- [ ] 0044-LIME-027
- [ ] 0044-LIME-028
- [ ] 0044-LIME-029
- [ ] 0044-LIME-043
- [ ] 0044-LIME-044
- [ ] 0044-LIME-045
- [ ] 0044-LIME-046
- [ ] 0044-LIME-047
- [ ] 0044-LIME-048
- [ ] 0044-LIME-049
- [ ] 0044-LIME-050
- [ ] 0044-LIME-051
- [ ] 0044-LIME-052
- [ ] 0044-LIME-053

PR #8899, opened 2023-07-25
feat: add feature test for SLA spot market 

Jiajia-Cui:
link to https://github.com/vegaprotocol/vega/issues/8810


Jiajia-Cui:
Did not sign the commit with GPG signature

PR #8921, opened 2023-07-26
Add business logic for team management

### LINKED ISSUE #8913: Implement business logic for team management
- Creating a team
- Updating a team
- Disbanding a team
- Joining a team

All this with basic tracking of the epoch, and parties and volumes.

ValentinTrinque:
Close #8913


core/teams/models.go
```diff
@@ -0,0 +1,52 @@
...
+
+type TeamID string
+
+type Party string
+
+type Team struct {
+	ID TeamID
+
+	Leader    Party
+	Followers []Party
```
jeremyletang, line 15: I would name those as they are in the spec to be fair.

core/teams/models.go
```diff
@@ -0,0 +1,52 @@
+package teams
```
jeremyletang, line 1: Not sure why this is not in the types package? I can imagine. there will be some ToProto / FromProto method used for snapshots or events etc.

core/teams/engine.go
```diff
@@ -0,0 +1,145 @@
...
+
+	e.teams[id].Name = name
+	e.teams[id].TeamURL = teamURL
+	e.teams[id].AvatarURL = avatarURL
+	e.teams[id].EnableRewards = enableRewards
+
+	return nil
+}
+
+func (e *Engine) JoinTeam(id TeamID, party Party) error {
```
jeremyletang, line 70: Should we implement the events being sent when a new team is joined, or created, etc?

ValentinTrinque: This is another issue.

core/teams/models.go
ValentinTrinque: Alright.

core/teams/engine.go
jeremyletang: fair enough.

core/teams/models.go
ValentinTrinque: Sure.

core/teams/engine.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	epochEngine.NotifyOnEpoch(engine.OnEpoch, engine.OnEpochRestore)
+
+	return engine
+}
+
+func errNoTeamMatchesID(id types.TeamID) error {
+	return fmt.Errorf("no team matches ID %q", id)
+}
```
jeremyletang, line 143: I usually create an error like this for this kind of things:
```suggestion
var ErrNoTeamMatchesID = func(id types.TeamID) error {
	return fmt.Errorf("no team matches ID %q", id)
}
```

So you can still export the error and use it in your test like this: `ErrNoTeamMatchesID("myteamid")`

core/teams/engine.go
```diff
@@ -0,0 +1,148 @@
...
+	// referees, by their current team ID.
+	allTeamMembers map[types.PartyID]types.TeamID
+
+	// membersToMove tracks all the parties that switch teams. The switch only
+	// happens by the end of the epoch.
+	membersToMove map[types.PartyID]teamMove
+}
+
+func (e *Engine) CreateTeam(referrer types.PartyID, name string, teamURL string, avatarURL string, enableRewards bool) (types.TeamID, error) {
+	newTeamID := e.newUniqueTeamID()
```
jeremyletang, line 34: This is fine for now, but I would think that later on this will be passed as a parameter and be a deterministic ID extract from the transaction as for every ID in the system.

ValentinTrinque: alright

jeremyletang (approved):

PR #8930, opened 2023-07-27
feat: start setting up settlement handlers

### LINKED ISSUE #8756: [Feature]: Perps - Basic creation in the execution engine
# Feature Overview

**In order to** allow perpetual markets to exist in the most basic state
**We will** plumb in the results from a governance proposal containing a perpetual as its product
**So that** we can test non-perp-specific functionality

Once we have https://github.com/vegaprotocol/vega/issues we should use the result to create a functioning market in the execution engine.

This should not necessarily include wiring in the callbacks from oracles since we will first need:
https://github.com/vegaprotocol/vega/issues/8755
https://github.com/vegaprotocol/vega/issues/8754

The idea here is that the only difference between a fixed-expiry futures market, and a perpetual market, is the data-sourcing and settlement. Almost everything else should function the same, even margin calculations if `margin_funding_factor == 0`. So getting to this point early means we can open up a load of testing an trying to leverage what we already have for fixed-expiry futures.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

EVODelavega:
Closes #8756


protos/sources/vega/vega.proto
```diff
@@ -895,6 +895,10 @@ enum TransferType {
   TRANSFER_TYPE_LIQUIDITY_FEE_UNPAID_COLLECT = 34;
   // Distributes performance bonus from market bonus to liquidity provider's general account.
   TRANSFER_TYPE_SLA_PERFORMANCE_BONUS_DISTRIBUTE = 35;
+  // Funds deducted from margin account after a perps funding loss.
+  TRANSFER_TYPE_PERP_FUNDING_LOSS = 36;
+  // Funds added to margin account after a perps funding gain.
```
candida-d, line 900:
```suggestion
  // Funds deducted from margin account after a perpetuals funding loss.
  TRANSFER_TYPE_PERP_FUNDING_LOSS = 36;
  // Funds added to margin account after a perpetuals funding gain.
```

guoguojin (dismissed):

core/collateral/engine.go
```diff
@@ -1073,9 +1073,25 @@ func (e *Engine) getMTMPartyAccounts(party, marketID, asset string) (gen, margin
...
+	if len(transfers) == 0 {
+		return nil, nil, nil
+	}
+	return e.mtmOrFundingSettlement(ctx, marketID, transfers, asset, types.TransferTypePerpFundingWin, round)
+}
+
 // MarkToMarket will run the mark to market settlement over a given set of positions
 // return ledger move stuff here, too (separate return value, because we need to stream those).
 func (e *Engine) MarkToMarket(ctx context.Context, marketID string, transfers []events.Transfer, asset string) ([]events.Margin, []*types.LedgerMovement, error) {
+	if len(transfers) == 0 {
```
jeremyletang, line 1088: this is also done as the first thing in mtmOrFundingSettlement not sure it's needed in both those function toow?

guoguojin (approved):

jeremyletang (approved):

PR #8931, opened 2023-07-27
feat: add perpetual product snapshot state

### LINKED ISSUE #8765: [Feature]: Perps - snapshots
# Feature Overview

**In order to** ensure that new state relating to perpetual markets is stored in a snapshot
**We will** include all new state into the snapshot
**So that** restored nodes do not fall out of consensus

Probably places where new state will be introduced:
- settlement engine: where we will store periodic settlement data points
- oracle engine (?): where we might have to store state about where we got up to in the reccuring time triggers.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

jeremyletang:
close #8765


EVODelavega (commented):
the perps product is probably going to change quite a lot, so most of this won't matter, but ok

candida-d (approved):

EVODelavega (approved):

PR #8939, opened 2023-07-27
Feat/8763 periodic settlement data endpoint

### LINKED ISSUE #8763: [Feature]: Perps - Datanode API: add a new endpoint to view periodic settlement data

guoguojin:
Closes #8763


datanode/gateway/graphql/schema.graphql
```diff
@@ -1943,6 +1943,14 @@ type Market {
 
   "Optional: Market ID of the successor to this market if one exists"
   successorMarketID: ID
+
+  "Funding Periods for perpetual markets"
```
candida-d, line 1947:
```suggestion
  "Funding periods for perpetual markets"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -5188,3 +5196,82 @@ type Entities {
...
+  pageInfo: PageInfo!
+}
+
+# Details of a funding interval for a perpetual market.
+type FundingPeriod {
+  # Market the funding period relates to.
+  marketId: ID!
+  # Sequence number of the funding period.
+  seq: Int!
+  # Time when the funding period started.
```
candida-d, line 5236: how is this time measured/expressed?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -713,6 +713,20 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Explorer"};
   }
 
+  // List funding periods
+  //
+  // Get a list of funding periods for a market
```
candida-d, line 718:
```suggestion
  // Get a list of funding periods for a perpetual market
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -713,6 +713,20 @@ service TradingDataService {
...
+  rpc ListFundingPeriods(ListFundingPeriodsRequest) returns (ListFundingPeriodsResponse) {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Markets"};
+  }
+
+  // List funding period data points
+  //
+  // Get a list of data points for a market's funding periods
```
candida-d, line 725:
```suggestion
  // Get a list of data points for a perpetual market's funding periods
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3312,6 +3326,72 @@ message ListEntitiesResponse {
   repeated vega.events.v1.ProtocolUpgradeEvent protocol_upgrade_proposals = 28;
 }
 
+// Request to list a market's funding periods
```
candida-d, line 3329:
```suggestion
// Request to list a perpetual market's funding periods
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3312,6 +3326,72 @@ message ListEntitiesResponse {
...
+message ListFundingPeriodsRequest {
+  // Market ID to get funding periods for.
+  string market_id = 1 [(google.api.field_behavior) = REQUIRED];
+  // Optional funding period sequence number to limit results to a specific funding period.
```
candida-d, line 3333: would it be clear to someone using these APIs how to get the funding period sequence number? (This is a question for all the instances that mention this sequence number)

datanode/entities/fund_period.go
```diff
@@ -0,0 +1,202 @@
...
+	v2 "code.vegaprotocol.io/vega/protos/data-node/api/v2"
+
+	"code.vegaprotocol.io/vega/libs/ptr"
+
+	events "code.vegaprotocol.io/vega/protos/vega/events/v1"
+
+	"code.vegaprotocol.io/vega/libs/num"
+)
+
+type FundingPeriod struct {
```
witgaw, line 17: This should also include:
1. Market price TWAP (time-weighted average price coming from the perps market on Vega
2. Reference price TWAP (time-weighed average price coming from the oracle)

datanode/entities/fund_period.go
```diff
@@ -0,0 +1,202 @@
+package entities
```
witgaw, line 1: Could you rename the file to `funding_period.go`?

We might have fund accounts eventually and they might have some periods, so this might end up being confusing.

witgaw (changes requested):

datanode/gateway/graphql/schema.graphql
edd: I agree - this would be good to have in the field description. Based on https://github.com/vegaprotocol/vega/blob/97980ebf5ba285076bf3a5eb5a4e0c23706df08f/datanode/gateway/graphql/schema.graphql#L287-L288 I guess it's a RFC3339Nano?

datanode/gateway/graphql/schema.graphql
```diff
@@ -1943,6 +1943,14 @@ type Market {
 
   "Optional: Market ID of the successor to this market if one exists"
   successorMarketID: ID
+
+  "Funding Periods for perpetual markets"
+  fundingPeriods(
```
edd, line 1948: **question (blocking)**: Checking other platforms, it looks like funding periods typically take a date parameter or two:

(dydx v4 uses `.../v4/historicalFunding/BTC-USD?effectiveBeforeOrAt=2023-07-06T21:59:59.899Z&null=2023-07-02T18:59:59.899Z`, okx `.../funding-rate-history-new?t=1690480575735&ccy=BTC`) . Is there a reason we don't provide that? 

Presumably this would be used for charting the funding rates. Date controls would help there.

**Basic suggestion**: An optional new parameter, `endDate`, that can be used to fetch `n` funding periods up until a dateTime (where `n` is defined by pagination or a sensible default)

**Fancier suggestion**: Two optional parameters, `startDate` and `endDate`.

Is there arny reason we can't get those?

edd (changes requested):
Feature request: date/date range for fundingPeriods

datanode/entities/fund_period.go
guoguojin: Yes, sorry, that was a typo, I will fix.

datanode/entities/fund_period.go
guoguojin: This isn't being provided by core, only the data points. The question is should the TWAP market/reference price be provided by core, or do you want this to be calculated from the data points by data node? The entity here represents the database object that we map from the core event we receive so this may or may not be updated depending on the answer.

datanode/gateway/graphql/schema.graphql
guoguojin: There's no reason we can't add it, happy to do so.

datanode/entities/fund_period.go
guoguojin: Confirmed with @witgaw that the TWAP information should be emitted from core.

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: I don't think so, but @edd has asked for a date range as input which makes much more sense so I'm just going to remove this option in the request

datanode/gateway/graphql/schema.graphql
```diff
@@ -1943,6 +1943,14 @@ type Market {
 
   "Optional: Market ID of the successor to this market if one exists"
   successorMarketID: ID
+
+  "Funding periods for perpetual markets"
+  fundingPeriods(
```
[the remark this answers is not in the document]
guoguojin: @mattrussell36 do you want the funding periods data points API which I nested as a sub-query under funding periods to be a separate top level query too, as that also allows for pagination of the data.

candida-d (dismissed):

witgaw (dismissed):
LGTM

datanode/gateway/graphql/schema.graphql
```diff
@@ -5188,3 +5222,79 @@ type Entities {
...
+"Details of a funding interval for a perpetual market."
+type FundingPeriod {
+  "Market the funding period relates to."
+  marketId: ID!
+  "Sequence number of the funding period."
+  seq: Int!
+  "RFC3339Nano time when the funding period started."
+  startTime: Timestamp!
+  "RFC3339Nano time when the funding period ended."
+  endTime: Timestamp
```
mattrussell36, line 5267: @EVODelavega @jeremyletang 

At what point is a funding event created? And at what point is the endTime value set?

We need to be sure UIs can show when the current funding period will end (or the next one will start). Either by using the endTime field here, or by using the startTime of the next period.

mattrussell36: I've spoken with @jeremyletang in person and I think we are good, we'll be using a time trigger from the products data source, adding the interval duration to the current funding periods startTime

wwestgarth (dismissed):

mattrussell36 (dismissed):

bglownia (dismissed):

candida-d (dismissed):

edd (dismissed):
Thanks for the date ranges 👍

wwestgarth (approved):

candida-d (approved):

PR #8948, opened 2023-07-28
feat: implement (and tidy up) oracle bindings for perps

### LINKED ISSUE #8769: [Feature]: Perps -  timestamp selection in settlement data-point data-sourcing
# Feature Overview

**In order to** allow the origin of the timestamp to chosen for incoming settlement data-points
**We will** extend the data-sourcing configurations when proposing a market
**So that** we remain flexible when users are specifying their data sources

The idea is that for a given data-source a proposer of a market should be able to choose whether the timestamp associated with that data should be"
- the block-time of the chain the data is taken from
- a field that represents the timestamp in the payload itself
- to just use vega's timestamp from when the data-source is received

There is little bit of chat about it here: https://vegaprotocol.slack.com/archives/C9E6XA8GK/p1688569952202169

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

EVODelavega:
implements parts of #8957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
 - [8918](https://github.com/vegaprotocol/vega/issues/8918) - Implement commands for team management.
 - [8756](https://github.com/vegaprotocol/vega/issues/8756) - Settlement and margin implementation for `PERPS`.
 - [8887](https://github.com/vegaprotocol/vega/pull/8887) - Remove differences for snapshot loading when the `nullchain` is used instead of `tendermint`
-- [8769](https://github.com/vegaprotocol/vega/issues/8769) - Oracle bindings for `PERPS`.
+- [8957](https://github.com/vegaprotocol/vega/issues/8957) - Oracle bindings for `PERPS`.
 
 ### 🐛 Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - [8918](https://github.com/vegaprotocol/vega/issues/8918) - Implement commands for team management.
 - [8756](https://github.com/vegaprotocol/vega/issues/8756) - Settlement and margin implementation for `PERPS`.
 - [8887](https://github.com/vegaprotocol/vega/pull/8887) - Remove differences for snapshot loading when the `nullchain` is used instead of `tendermint`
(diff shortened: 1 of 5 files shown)


wwestgarth:
Maybe I'm missing something but this ticket is about implementing this part of the spec:
https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0053-PERP-product_builtin_perpetual_future.md#41-periodic-settlement-data-point-received

but by the looks of it in `receiveDataPoint` we still only read the external data point time from the Ethereum meta data:
```
// add price point with "eth-block-time" as time
	pTime, err := strconv.ParseUint(data.MetaData["eth-block-time"], 10, 64)
	if err != nil {
		p.log.Error("Could not parse the eth block time",
			logging.String("eth-block-time", data.MetaData["eth-block-time"]),
			logging.Error(err),
		)
		return err
	}
	// now add the price
	p.addExternalDataPoint(ctx, assetPrice, int64(pTime))
```

jeremyletang:
> > Maybe I'm missing something but this ticket is about implementing this part of the spec: https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0053-PERP-product_builtin_perpetual_future.md#41-periodic-settlement-data-point-received
> > but by the looks of it in `receiveDataPoint` we still only read the external data point time from the Ethereum meta data:
> > ```
> > // add price point with "eth-block-time" as time
> > 	pTime, err := strconv.ParseUint(data.MetaData["eth-block-time"], 10, 64)
> > 	if err != nil {
> > 		p.log.Error("Could not parse the eth block time",
> > 			logging.String("eth-block-time", data.MetaData["eth-block-time"]),
> > 			logging.Error(err),
> > 		)
> > 		return err
> > 	}
> > 	// now add the price
> > 	p.addExternalDataPoint(ctx, assetPrice, int64(pTime))
> > ```
> 
> Ok so I hear that we're delaying the choice of where the time comes from until later to help get testing going, which makes sense. So can this PR _not_ close #8769 and instead can we have a new ticket which describes what this change is (a refactor?).

Update the ticket description, should not clos the ticket anymore. @EVODelavega could you fix the linting errors please?

wwestgarth (dismissed):

wwestgarth (approved):

PR #8958, opened 2023-08-01
chore: rename perps to perpetual in protos

wwestgarth:
This is only an update to the protos to change use of `Perps -> Perpetual`. It currently stands out a lot because for the other products we do not shorten their name, and also use the singular for all the other products. Now looks like:
```
oneof product {
    // Future.
    Future future = 100;
    // Spot.
    Spot spot = 101;
    // Perpetual.
    Perpetual perpetual = 102;
  }
```

Doing it now before its too late.


protos/sources/vega/vega.proto
```diff
@@ -1156,7 +1156,7 @@ message NetworkLimits {
   // Are spot market proposals allowed at this point in time.
   bool can_propose_spot_market = 10;
   // Are perpetuals market proposals allowed at this point in time.
```
candida-d, line 1158:
```suggestion
  // Are perpetual market proposals allowed at this point in time.
```

candida-d (commented):
This is such a nice improvement. I made a small doc string suggestion to bring it in line with singular perpetual.

ze97286 (approved):

candida-d (approved):

PR #8961, opened 2023-08-02
Pegged iceberg bug

witgaw:
Price of a pegged iceberg order fails to refresh when the reference price moves.


witgaw:
Addressed by https://github.com/vegaprotocol/vega/pull/8965

PR #8977, opened 2023-08-03
feat: wire in internal time triggers

### LINKED ISSUE #8968: [Bug]: Successor has no reference to a settled parent after checkpoint
### Problem encountered

If the parent of a successor market has been settled, then we do a checkpoint restart, the parent is forgotten, and no longer referred to on the child, even as a reference. Per https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0081-SUCM-successor_markets.md#0081-SUCM-031 , it should be. We should either do this or update the spec if it is not desired.

### Observed behaviour

No parent can be found after checkpoint, either via the market API searching for the parent, or as a reference on the child market.

### Expected behaviour

AC says parent should still be available post restart since it has children, and as a reference on the child.

### Steps to reproduce

```bash
1. Check out this branch: https://github.com/vegaprotocol/system-tests/pull/2382 
2. Ensure test is not skipped (probably will be shortly, while this issue is open)
3. pytest -k test_settled_succeeded_parent_remembered
```

### Software version

ee22fd47cd44a334c9dcb47353318628c8b30755

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #8969: [Feature]: Perps - fully wire in internal time-trigger
# Feature Overview

**In order for** time-triggered internal data-sources to function
**We will** wire them in fully to the data-source spec engine
**So that** when running a system-test end-to-end settlement cues for perperuals are received properly

Currently the time-triggered data-source code exists, but `IsTriggered()` is not called anywhere. It needs weaving into the data-source spec engine somehow such that when the builtin-datasource `OnTick()` is broadcast it is matched, filtered against the conditions, and then assessed for whether it is triggerable.

Something like that anyway.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
closes #8969  

Wires in the internal time trigger into the spec engine so that the cue's get broadcast time the perp product. There was also an issue where times received internally were in unixnano, but everything external is in seconds so for now I've scaled the external seconds to nanoseconds so that they are all at least all consistent. We might want to think a bit harder about this later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 - [8765](https://github.com/vegaprotocol/vega/issues/8765) - Implement snapshots state for `PERPS`.
 - [8918](https://github.com/vegaprotocol/vega/issues/8918) - Implement commands for team management.
 - [8960](https://github.com/vegaprotocol/vega/issues/8960) - Improve wiring perpetual markets through governance.
+- [8968](https://github.com/vegaprotocol/vega/issues/8968) - Improve wiring of internal time triggers for perpetual markets.
 - [8756](https://github.com/vegaprotocol/vega/issues/8756) - Settlement and margin implementation for `PERPS`.
 - [8887](https://github.com/vegaprotocol/vega/pull/8887) - Remove differences for snapshot loading when the `nullchain` is used instead of `tendermint`
 - [8957](https://github.com/vegaprotocol/vega/issues/8957) - Oracle bindings for `PERPS`.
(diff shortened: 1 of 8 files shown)


EVODelavega:
This isn't linked to the correct issue I don't think

wwestgarth:
> This isn't linked to the correct issue I don't think

The classic off-by-one error. Fixed!

jeremyletang (approved):

PR #8980, opened 2023-08-03
fix: ensure parent ID is not erased

### LINKED ISSUE #8968: [Bug]: Successor has no reference to a settled parent after checkpoint

EVODelavega:
Closes #8968

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@
 - [8908](https://github.com/vegaprotocol/vega/issues/8908) - A rejected parent market should result in all successors getting rejected, too.
 - [8953](https://github.com/vegaprotocol/vega/issues/8953) - Fix reward distribution when validator delegate to other validators
 - [8898](https://github.com/vegaprotocol/vega/issues/8898) - Fix auction uncrossing handling for spots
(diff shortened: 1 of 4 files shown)


guoguojin (approved):

ValentinTrinque (dismissed):
Is there a way we have proper test to verify that change, to proves it works ?

PR #8996, opened 2023-08-07
Support amend/cancel pending liquidity provision

### LINKED ISSUE #8988: [Bug]: It's not possible to cancel or amend a liquidity commitment that's pending
It's not possible to submit a liquidity amendment or cancellation transaction while the commitment is in a PENDING state. 

As an aside: PENDING state is not consistent with previous behaviour, before we'd have that state only if liquidity provision was submitted during an auction. Now it remains PENDING until start of the new epoch for new commitments, not sure how amendments are treated. It's an internal concept (though it's visible externally) not covered by specs so I don't really care, just flagging it in case it has some implications for other bits of core or the frontend. 

When fixing this please assure the bond amount eventually gets released after successfully cancelling the commitment. 

See https://github.com/vegaprotocol/vega/pull/8987 to reproduce.

ze97286:
Closes #8988

### DIFF
--- a/core/execution/common/liquidity_provision.go
+++ b/core/execution/common/liquidity_provision.go
@@ -487,12 +487,13 @@ func (m *MarketLiquidity) AmendLiquidityProvision(
 		}
 	}
 
-	if !m.liquidityEngine.IsLiquidityProvider(party) {
+	existingAmendment := m.liquidityEngine.PendingProvisionByPartyID(party)
+	if !m.liquidityEngine.IsLiquidityProvider(party) && existingAmendment == nil {
 		return ErrPartyNotLiquidityProvider
 	}
 
 	lp := m.liquidityEngine.LiquidityProvisionByPartyID(party)
-	if lp == nil {
+	if lp == nil && existingAmendment == nil {
 		return fmt.Errorf("cannot edit liquidity provision from a non liquidity provider party (%v)", party)
 	}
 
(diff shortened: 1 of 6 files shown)


core/execution/common/liquidity_provision.go
```diff
@@ -512,22 +513,28 @@ func (m *MarketLiquidity) AmendLiquidityProvision(
...
 
 			return err
 		}
+		proposedCommitmentVariation = existingAmendment.CommitmentAmount.ToDecimal().Sub(lpa.CommitmentAmount.ToDecimal())
+	} else {
+		if lp != nil {
+			proposedCommitmentVariation = lp.CommitmentAmount.ToDecimal().Sub(lpa.CommitmentAmount.ToDecimal())
+		} else {
+			proposedCommitmentVariation = existingAmendment.CommitmentAmount.ToDecimal().Sub(lpa.CommitmentAmount.ToDecimal())
+		}
```
jeremyletang, line 535: why do we need the ToDecimal conversions her?

ze97286: I guess because it could be negative

witgaw (approved):
Looks good to me, I'm not sure about the cases where they should be penalised (we need to assure that amending first and then cancelling doesn't let them circumvent the pending penalty), but that can be explored (and fixed if need be) outside of this PR.

core/execution/common/liquidity_provision.go
```diff
@@ -487,12 +487,13 @@ func (m *MarketLiquidity) AmendLiquidityProvision(
 		}
 	}
 
-	if !m.liquidityEngine.IsLiquidityProvider(party) {
+	existingAmendment := m.liquidityEngine.PendingProvisionByPartyID(party)
+	if !m.liquidityEngine.IsLiquidityProvider(party) && existingAmendment == nil {
```
karlem, line 491: I think the `existingAmendment` is not needed since then `IsLiquidityProvider` now checks for pending as well correct?

ze97286: correct - what I wasn't sure of is if having IsLiquidityProvider account for pending has any unexpected negative consequences unrelated to here. you can leave it and remove after merging.

karlem:
I agree with @witgaw  that might be out of scope of this PR but I think cancelling should be essentially the same like amending to 0 - which means penalties need to be applied.

PR #9005, opened 2023-08-07
Wire the teams engine to the ABCI

ValentinTrinque:


jeremyletang (dismissed):

ze97286 (commented):
are there any spam protection constraints for team creation/update?

ze97286 (commented):
do we not need to save teams data in checkpoints? or are we ok with losing them in this case?

core/processor/abci.go
```diff
@@ -2154,3 +2172,34 @@ func (app *App) DeliverEthereumKeyRotateSubmission(ctx context.Context, tx abci.
...
 }
+
+func (app *App) CreateTeam(ctx context.Context, tx abci.Tx) error {
+	params := &commandspb.CreateTeam{}
+	if err := tx.Unmarshal(params); err != nil {
+		return fmt.Errorf("could not deserialize CreateTeam command: %w", err)
+	}
+
+	// The party that initiated this command has to retrieve the team ID through
+	// the API.
```
wwestgarth, line 2183: Can we not use `addDeterministicID()` here to create the teamID from using the hash of the signature, like we do for proposals, orders etc. etc. Then any FE built against all this can find the teamID from their submission in the same way.

ValentinTrinque:
@ze97286:

> are there any spam protection constraints for team creation/update?

The implementation is not finished. We are missing a whole bunch of checks regarding the constraints to create a team, such as minimum tokens, etc. This PR has been split into small chunks. This will come later. The main goal right now is to lay the foundation for the end-to-end scenario.

> do we not need to save teams data in checkpoints? or are we ok with losing them in this case?

Not familiar with the checkpoints, and the data they are supposed to restore. Is it about the generated events ?

core/processor/abci.go
ValentinTrinque: Yes we can. We are just implementing this feature incrementally. I can add it now.

ze97286:
> > do we not need to save teams data in checkpoints? or are we ok with losing them in this case?
> 
> Not familiar with the checkpoints, and the data they are supposed to restore. Is it about the generated events ?

it's about preserving teams across checkpoint restarts. As of now, whenever we restart from a checkpoint the whole teams information is going to be lost.

jeremyletang:
> > > do we not need to save teams data in checkpoints? or are we ok with losing them in this case?
> > 
> > 
> > Not familiar with the checkpoints, and the data they are supposed to restore. Is it about the generated events ?
> 
> it's about preserving teams across checkpoint restarts. As of now, whenever we restart from a checkpoint the whole teams information is going to be lost.

No need for checkpoint for this I would say. They are really mostly to secure against any financial lost.

ze97286:
> No need for checkpoint for this I would say. They are really mostly to secure against any financial lost.

if there's no requirement, there's no requirement...

wwestgarth (approved):

wwestgarth (approved):

PR #9013, opened 2023-08-08
fix: clear insurance pools in a determinisitc order

### LINKED ISSUE #9004: [Bug]: network falling out of consensus during a full-run
### Problem encountered

This full run:
https://jenkins.ops.vega.xyz/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/89943/pipeline/551/

kicked off against a system-test branch that fixes a successor market test:
https://github.com/vegaprotocol/system-tests/pull/2411

Has shown that the network falls out of consensus:
```
{"err": "wrong Block.Header.AppHash.  Expected 3E0A9EAD4993C5E148146886B0F1ACB97F77E692BE94E21CFC303F2745F6BDF8, got F574CA43B9D671909A5AE09591713089EBAEC7DAE85CC956BA266C92924A530F"}
```

It was also picked up by the snapshot soak tests when it tried to replay the chain to generate snapshots.

The suspected cause is successor markets.

### Observed behaviour

Node's fall out of consensus.

### Expected behaviour

Nodes stay in consensus.

### Steps to reproduce

```bash
Run the snapshot soak tests locally.
```

### Software version

76720bbb79eee8d24135d25b84b1ed2491485709

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #9004 

We have two markets that are getting their insurance pools cleared during an OnTick:
`market1` has balance `0`
`market2` has balance `432400000000000000000`
there are 8 other markets not being cleared but will receive the cleared balance split equally.

If we process `market1` first:
- `market1` has balance `0` and is removed
- `market2`then has its balance split between 8
- each other market gets `54050000000000000000`

If we process `market2` first:
- `market2`'s balance is split between 9, including `market1`
- each get `48044444444444444444`
- then we process `market1` which now has a balance of `48044444444444444444`
- this is split between 8 and each get `6005555555555555555`

The end result is the same, except that because it one case we have to divide twice, we have to handle the remainer twice and the crumbs end up in with different markets.

### DIFF
--- a/core/execution/engine.go
+++ b/core/execution/engine.go
@@ -1330,9 +1330,20 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) {
 		}
 		e.removeMarket(id)
 	}
+
+	// sort the marketCPStates by ID since the order we clear insurance pools
+	// changes the division when we split it across remaining markets and
+	// who the remainder ends up with if it doesn't divide equally.
+	allIDs := []string{}
+	for id := range e.marketCPStates {
+		allIDs = append(allIDs, id)
+	}
+	sort.Strings(allIDs)
+
 	// find state that should expire
-	for id, cpm := range e.marketCPStates {
+	for _, id := range allIDs {
 		// market field will be nil if the market is still current (ie not closed/settled)
+		cpm := e.marketCPStates[id]
 		if !cpm.TTL.Before(t) {
 			// CP data has not expired yet
 			continue
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@
 - [8874](https://github.com/vegaprotocol/vega/issues/8874) - Database migration can fail when rolling back and migrating up again.
 - [8855](https://github.com/vegaprotocol/vega/issues/8855) - Preserve reference to parent market when restoring checkpoint data
 - [8909](https://github.com/vegaprotocol/vega/issues/8909) - initialise id generator for all branches of market state update
+- [9004](https://github.com/vegaprotocol/vega/issues/9004) - Clear insurance pools in a deterministic order in successor markets.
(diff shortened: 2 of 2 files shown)


ValentinTrinque (dismissed):

jeremyletang (dismissed):

core/execution/engine.go
```diff
@@ -1330,9 +1330,20 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) {
...
+	// who the remainder ends up with if it doesn't divide equally.
+	allIDs := []string{}
+	for id := range e.marketCPStates {
+		allIDs = append(allIDs, id)
+	}
+	sort.Strings(allIDs)
+
 	// find state that should expire
-	for id, cpm := range e.marketCPStates {
+	for _, id := range allIDs {
```
ze97286, line 1344: couldn't you just iterate the allMarketsCpy for the id and skip those not in the marketCPStates?

wwestgarth: I don't think so, `marketCPStates` contans closed/settlement markets waiting to be succeeeded so they won't be in `allMarketsCpy`.

core/execution/engine.go
```diff
@@ -1330,9 +1330,20 @@ func (e *Engine) OnTick(ctx context.Context, t time.Time) {
 		}
 		e.removeMarket(id)
 	}
+
+	// sort the marketCPStates by ID since the order we clear insurance pools
+	// changes the division when we split it across remaining markets and
+	// who the remainder ends up with if it doesn't divide equally.
+	allIDs := []string{}
```
ze97286, line 1337: it's probably not a big deal now because it's a small number but maybe it's worth keeping a marketCPStatesCpy as a slice because this is called on tick.

ze97286 (approved):

PR #9017, opened 2023-08-08
test: add test tool for json comparison of requests

dalebennett1992:
Test tool for testing if an external application can communicate with vega proto definitions without needing a system integration test. The idea is if we can 'generate' a contract via unit tests or a script (it could also be hand cranked) in the consuming apps repo (in this instance it will be browser wallet) we can then replay them against this test tool to check the data sent will retain all information once decoded by vega. This will ensure we are crafting our requests correctly and that our protos remain up to date. This means we can test our interactions with vega in seconds instead of test runs that take an hour or so and it is much more scalable

**Transaction files:**
The tool currently accepts a directory of transaction.json files (will scale this back so we can just fire a decoded string at it too in another PR) the traction json should have the below top level keys:

**Transaction:** Raw json generated by the consuming application (e.g browser wallet), this is for comparison against the decoded json
**InputData:** Raw input data json generated by the consuming application, this is for comparison against the decoded json
**EncodedData:** Encoded data generated by the consuming application, this is for the test tool to decode using vega protos and used to create json to compare against the Transaction and InputData keys above.

Here is an example transaction.json:

`{
  "Transaction": {
    "from": {
      "pubKey": "mykey"
    },
    "inputData": "CHsQyAOiPyUIARIEZGF2ZRgCIgp0ZXN0IGFzc2V0KgMxMjMyCHRlc3QgcmVm",
    "signature": {
      "algo": "testAlgo",
      "value": "74657374536967",
      "version": 3
    },
    "version": "this will cause a diff"
  },
  "InputData": {
    "blockHeight": "456",
    "command": {
      "transfer": {
        "from_account_type": 1,
        "to": "dave",
        "to_account_type": 2,
        "asset": "test asset",
        "amount": "123",
        "reference": "test ref",
        "Kind": null
      }
    },
    "nonce": "123"
  },
  "EncodedData": "Ci0IexDIA6I/JQgBEgRkYXZlGAIiCnRlc3QgYXNzZXQqAzEyMzIIdGVzdCByZWYSHAoONzQ2NTczNzQ1MzY5NjcSCHRlc3RBbGdvGAOAfQPSPgVteWtleQ=="
}`

example command:
`vega tools check-tx --transactions=/your/dir/with/transaction/files.json --diffs=/your/dir/for/diffs`

results should be outputted in the diff directory. You will see an html diff that will render like this
image
there is also a json file outputted to compare core json vs your json in full.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ integration.test
 .testCoverage.txt
 *.tmp
 tmp/
+cmd/vegatools/testdata/*
 
 # Exclude files related to proto
 proto/*-re
--- a/cmd/vegatools/checktx.go
+++ b/cmd/vegatools/checktx.go
@@ -1,0 +1,62 @@
+package tools
+
+import (
+	"fmt"
+
+	inspecttx2 "code.vegaprotocol.io/vega/vegatools/inspecttx"
+
+	"github.com/sirupsen/logrus"
+
+	"code.vegaprotocol.io/vega/core/config"
+)
+
+var (
+	transactionDiffs     int
+	transactionsAnalysed int
+	transactionsPassed   int
+)
+
+type checkTxCmd struct {
+	config.OutputFlag
+	Diffs        string `description:"The output directory for detailed reporting on diffs" long:"diffs"        short:"d"`
+	Transactions string `description:"Path to the transaction json files"                   long:"transactions" required:"true" short:"t"`
+}
+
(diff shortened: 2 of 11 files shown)


dalebennett1992:
In draft until we chat about the json marshalling, so hopefully we can nuke the reflection hack in this PR @jeremyletang @emilbayes

dalebennett1992:
Out of scope for now after chat with @emilbayes

PR #9050, opened 2023-08-11
Terminating perps market through governance

EVODelavega:
Closes #8761

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -1029,13 +1029,6 @@ func (m *Market) uncrossOrderAtAuctionEnd(ctx context.Context) {
 }
 
 func (m *Market) UpdateMarketState(ctx context.Context, changes *types.MarketStateUpdateConfiguration) error {
-	settleData := m.getLastTradedPrice()
-	// a perpetual market cannot be terminated with a given settlement price as such
-	if m.perp {
-		changes.SettlementPrice = num.UintZero()
-	} else {
-		settleData = num.UintZero().Mul(changes.SettlementPrice, m.priceFactor)
-	}
 	_, blockHash := vegacontext.TraceIDFromContext(ctx)
 	// make deterministic ID for this market, concatenate
 	// the block hash and the market ID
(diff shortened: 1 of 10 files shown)


core/execution/future/market.go
```diff
@@ -1025,6 +1029,10 @@ func (m *Market) uncrossOrderAtAuctionEnd(ctx context.Context) {
 }
 
 func (m *Market) UpdateMarketState(ctx context.Context, changes *types.MarketStateUpdateConfiguration) error {
+	// a perpetual market cannot be terminated with a given settlement price as such
```
wwestgarth, line 1032: Perps are settlement with a settlement price. The idea is that we do a final funding payment *and then* do settlement as usual. From the spec:
"""
Should a perpetual futures market get closed using the [governance proposal](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0028-GOVE-governance.md#61-move-market-to-a-closed-state) an final funding payment should be calculated using the data available at that time and exchanged **_right before the final settlement using the price contained in the proposal is carried out._**
"""

EVODelavega: Alright, made the settlement price required in the integration test, so terminating the perps market is handled pretty much the same way as a regular future. The final funding payment was already implemented, and is triggered from when calling the `UnsubscribeTradingTerminated` func on the oracle.

wwestgarth (approved):

core/execution/future/market.go
jeremyletang: Not sure to understand, there shouldbe no trading terminated callback used on the perps market right? Not sure it make sense to expect it to the UnsubscribeTradingTerminated ever on a perps market then.

EVODelavega: There is no direct callback as such. The normal flow when a futures market is terminated is to call `UnsubscribeTradingTerminated` on the oracle. Before this PR, the perps oracle just emitted a panic as being not implemented. Instead, a perpetual oracle will use this call as the signal to trigger the final funding payment, before the market is settled as per spec. There is no trading terminated oracle binding, so in that sense, callbacks aren't set up. We could change the `UnsubscribeTradingTerminated` call to something like:

```go
if m.perp {
    m.tradableInstrument.Instrument.Product.TriggerFundingPayment(ctx)
} else {
    m.tradableInstrument.Instrument.Product.UnsubscribeTradingTerminated(ctx)
}
```

But to do so, we'd have to add yet another call to the `Product` interface, that will only be implemented by the perps product. It seemed cleaner, to my eye at least, to use the functions that are already there

PR #9135, opened 2023-08-21
feat: add funding rate to market data

### LINKED ISSUE #9076: [Perps]: Add API for ongoing / current funding rate
Add the missing API for the F

- [ ] add the ongoing / current funding rate as part of the market data event-  this is used to calculate settlement

guoguojin:
Closes #9076


candida-d (dismissed):

core/products/perpetual.go
```diff
@@ -365,6 +365,15 @@ func (p *Perpetual) handleSettlementCue(ctx context.Context, t int64) {
 	p.startNewFundingPeriod(ctx, t)
 }
 
+func (p *Perpetual) GetFundingRate(t int64) *num.Decimal {
+	internalTWAP := twap(p.internal, p.startedAt, t)
```
wwestgarth, line 369: Can you do a `if !p.haveData(t){}` return `0` at the start of this function?

datanode/sqlstore/migrations/0025_market_data_add_funding_rate.sql
```diff
@@ -0,0 +1,83 @@
+-- +goose Up
+
+alter table market_data add column if not exists funding_rate numeric;
+alter table current_market_data add column if not exists funding_rate numeric;
```
jeremyletang, line 4: Can / should we make them both nullable, and default to NULL? would make it easily compatible for all existing future market.

datanode/gateway/graphql/schema.graphql
```diff
@@ -256,6 +256,8 @@ type MarketData {
   marketGrowth: String!
   "The last traded price (an unsigned integer)"
   lastTradedPrice: String!
+  "The current funding rate. This applies only to a perpetual market"
+  fundingRate: String!
```
mattrussell36, line 260: Firstly, what will the value be if this market is not a perp. I would expect a nullable field here?

Secondly, will there be another pr to have this field added to ObservableMarketData too? This is key for frontend requirements

datanode/sqlstore/migrations/0025_market_data_add_funding_rate.sql
mattrussell36: I mentioned this in my comment also.. I think returning null over the api would be better than an empty string which is what I would expect given the current schema change.

guoguojin: The nullable and default null value here is already implicit in the declaration. The table is too big to set a not null default, the alter command would take forever to complete on archive nodes.

jeremyletang (dismissed):

wwestgarth (dismissed):

bglownia (dismissed):

datanode/gateway/graphql/schema.graphql
```diff
@@ -320,6 +322,8 @@ type ObservableMarketData {
   marketGrowth: String!
   "The last traded price (an unsigned integer)"
   lastTradedPrice: String!
+  "The current funding rate. This applies only to a perpetual market"
+  fundingRate: String
```
edd, line 326: **chore (non-blocking)** In the future, let's tidy this object up so that fields that don't apply to all markets don't apply to all markets. I've marked this as non blocking as there'll almost certainly be more than just this field. I'll file a ticket.

ValentinTrinque (approved):

jeremyletang (approved):

PR #9169, opened 2023-08-23
fix: asset id on ledger entry export

### LINKED ISSUE #8944: [Bug] Issue when using assetID in the query for the ledger entry export
### Issue

When running the query for the ledger entry CSV export it is required to specify an asset ID, however, when the output is returned it looks to have ignored the asset ID and retruns everything.

This needs to be fixed so the correct data is returned as requested, also as an additional enhancement the asset ID should be optional.

### Tasks:

- [ ] fix the bug where the specified asset ID seems to be being ignored
- [ ] make the asset ID optional  

breaking because Asset ID field on the `ExportLedgerEntriesRequest gRPC API` for exporting ledger entries has changed type to make it optional; so users of this API will need to regenerate their protobuf bindings.

pscott31:
Filter was previously ignored. Also make asset id optional.

close #8944 

breaking because Asset ID field on the `ExportLedgerEntriesRequest gRPC API` for exporting ledger entries has changed type to make it optional; so users of this API will need to regenerate their protobuf bindings.


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1276,7 +1276,7 @@ message ExportLedgerEntriesRequest {
   // Restrict exported ledger entries to those relating to the given party ID.
   string party_id = 1 [(google.api.field_behavior) = REQUIRED];
   // Restrict exported ledger entries to those relating to the given asset ID.
-  string asset_id = 2 [(google.api.field_behavior) = REQUIRED];
+  optional string asset_id = 2;
```
candida-d, line 1279: Is this already reflected in the GraphQL?

pscott31: good question, but you can't access this API via GraphQL, it's for bulk downloading CSV files of ledger entries

CHANGELOG.md
```diff
@@ -116,6 +116,7 @@
 - [8945](https://github.com/vegaprotocol/vega/issues/8945) - Expose missing order from the stop order `GraphQL` endpoint.
 - [9034](https://github.com/vegaprotocol/vega/issues/9034) - Add missing transfer types to `GraphQL` schema definition.
 - [9075](https://github.com/vegaprotocol/vega/issues/9075) - `Oracle Specs API` can fail when oracle data is null.
+- [8944](https://github.com/vegaprotocol/vega/issues/8944) - Fix ignoring of asset `ID` in ledger export, and make it optional.
```
wwestgarth, line 119: I think this is also a breaking change right? because the type in the gRPC was a `string` but nows its a `*string`. So can we make sure we include it in the right changelog section and add the relevant labels to the PR and ticket?

pscott31:
> I think this is also a breaking change right? because the `assetID` type in the gRPC was a `string` but nows its a `*string`. So can we make sure we include it in the right changelog section and add the relevant labels to the PR and ticket? 

happy to do that but I don't see how it's a breaking change.. If I am not mistaken you can do an extra thing which you couldn't do before (not include an asset ID in a grpc request).. could you give me an example of a thing that would work before but not now?

CHANGELOG.md
pscott31: the previous behaviour was that you had to provide an asset ID, but it ignored it and gave you all the assets anyway :)

protos/sources/data-node/api/v2/trading_data.proto
candida-d: Ah, forgot that! Thank you :)

candida-d (dismissed):

CHANGELOG.md
wwestgarth: Its breaking in the sense the the gRPC is no longer backwards-compatible, i.e someone wanted to _start_ using `ExportLedgerEntries` but hadn't yet update their client-side-generate-from-the-protos, they'll get an unmarshaling error when their message hits the datanode because the type has changed.

Equally the person who submitted the bug might be like "its fixed now, I can just re-run my script" and they'll get an unmarshalling error too. So from their point-of-view its going from "I supply an assetId and it doesn't filter" to "now it doesn't work at all".

pscott31: ah I see, I initially thought you meant breaking as in breaking usage of the API

You mean breaking in the sense that the generated go bindings for grpc are no longer compatible? because I think they actually would be; it's my understanding protobuf wire format doesn't change by making a field optional; the zero value and a non-present optional are both (as I understand it) encoded as simply that record not being present in the encoded message..

 though I can't find any definitive source to back that up [this post](https://github.com/golang/protobuf/issues/225) touches upon it

wwestgarth: Ok, well if thats true then its fine as it is!

PR #9170, opened 2023-08-23
fix: duplicate ethereum oracle chain events

### LINKED ISSUE #8631: [Bug]: Possible to get duplicate eth oracle chain events at startup
### Problem encountered

Possible to get duplicate eth oracle chain events at startup as the prev block height is set after the eth call poller is started, which means it can be set to block already processed. Not a blocker for the merge to dev branch but should be addressed shortly thereafter.

### Observed behaviour

"

### Expected behaviour

"

### Steps to reproduce

```bash
"
```

### Software version

eth-oracles-feature

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ettec:
closes #8631

### DIFF
--- a/core/datasource/external/ethverifier/verifier_snapshot.go
+++ b/core/datasource/external/ethverifier/verifier_snapshot.go
@@ -128,9 +128,11 @@ func (s *Verifier) LoadState(ctx context.Context, payload *types.Payload) ([]typ
 }
 
 func (s *Verifier) OnStateLoaded(ctx context.Context) error {
-	// tell the eth call engine what the last block seen was so it does not retrigger calls
+	// tell the eth call engine what the last block seen was, so it does not re-trigger calls
 	if s.lastBlock != nil && s.lastBlock.Height > 0 {
-		s.ethEngine.UpdatePreviousEthBlock(s.lastBlock.Height, s.lastBlock.Time)
+		s.ethEngine.StartAtHeight(s.lastBlock.Height, s.lastBlock.Time)
+	} else {
+		s.ethEngine.Start()
 	}
 
 	return nil
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,7 @@
 - [9105](https://github.com/vegaprotocol/vega/issues/9105) - Truncate virtual stake decimal places
 - [8800](https://github.com/vegaprotocol/vega/issues/8800) - `expiresAt` is always null in the Stop Orders `API`.
 - [8796](https://github.com/vegaprotocol/vega/issues/8796) - Avoid updating active proposals slice while iterating over it.
+- [8631](https://github.com/vegaprotocol/vega/issues/8631) - Prevent duplicate Ethereum call chain event after snapshot start
 - [8679](https://github.com/vegaprotocol/vega/issues/8679) - Disallow snapshot state-sync if local snapshots exist
 - [8364](https://github.com/vegaprotocol/vega/issues/8364) - Initialising from network history not working after database wipe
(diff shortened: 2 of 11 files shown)


jeremyletang (commented):
Could you add a bunch of details on how this is fixing the issue? It seems that this basically moves the start of the engine to a later place in the code, but what was causing the issue?

jeremyletang:
@ettec this doesn't pass the CI, could you fix it please?

ettec:
> Could you add a bunch of details on how this is fixing the issue? It seems that this basically moves the start of the engine to a later place in the code, but what was causing the issue?

Sure, before the change the eth call engine was being started before snapshot state was restored, hence the last block was not set, this meant the eth call engine would could effectively re-read old blocks and submit chain events that would have already been submitted.  The change here is to only start the eth call engine after snapshot state is restored.

jeremyletang:
> > Could you add a bunch of details on how this is fixing the issue? It seems that this basically moves the start of the engine to a later place in the code, but what was causing the issue?
> 
> Sure, before the change the eth call engine was being started before snapshot state was restored, hence the last block was not set, this meant the eth call engine would could effectively re-read old blocks and submit chain events that would have already been submitted. The change here is to only start the eth call engine after snapshot state is restored.

Right makes sense. cheers.

wwestgarth:
> > > Could you add a bunch of details on how this is fixing the issue? It seems that this basically moves the start of the engine to a later place in the code, but what was causing the issue?
> > 
> > 
> > Sure, before the change the eth call engine was being started before snapshot state was restored, hence the last block was not set, this meant the eth call engine would could effectively re-read old blocks and submit chain events that would have already been submitted. The change here is to only start the eth call engine after snapshot state is restored.
> 
> Right makes sense. cheers. Although could @ValentinTrinque / @wwestgarth confirm that at that point, in any cases, either the snapshot from local or tendermint gossip will be fully loaded? if not that might not be then an accurate fix?

Its still starting too early for the state-sync case. You can see in the logs for the snapshot-test in the smoke run the ethcall engine is ticking away while we're still discovering new snapshots:
https://jenkins.ops.vega.xyz/job/common/job/system-tests-wrapper/98955/artifact/testnet/logs/testnet-nodeset-validators-3-validator/vega-validator-3.stdout-2023-08-25T15%3A16%3A54Z.log

What'll you'll probably need to do is to start the ethcall engine during an `OnStateLoaded()` call, and then for the starting-from-genesis case start it during `InitChain` in `processor.abci.go`.

core/datasource/external/ethverifier/verifier_snapshot.go
```diff
@@ -128,9 +128,11 @@ func (s *Verifier) LoadState(ctx context.Context, payload *types.Payload) ([]typ
 }
 
 func (s *Verifier) OnStateLoaded(ctx context.Context) error {
-	// tell the eth call engine what the last block seen was so it does not retrigger calls
+	// tell the eth call engine what the last block seen was, so it does not re-trigger calls
 	if s.lastBlock != nil && s.lastBlock.Height > 0 {
-		s.ethEngine.UpdatePreviousEthBlock(s.lastBlock.Height, s.lastBlock.Time)
+		s.ethEngine.StartAtHeight(s.lastBlock.Height, s.lastBlock.Time)
+	} else {
+		s.ethEngine.Start()
```
ettec, line 135: Not entirely convinced this else clause is necessary, in what way would it be possible to load from a snapshot and have these values as the defaults?  seems improbable but not impossible

ettec:
eth oracle tests run-> https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/5569/

ettec:
snapshots test run-> https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/5619/

wwestgarth (approved):

PR #9176, opened 2023-08-23
chore: add statement_timeout to postgres connection config

pscott31:
close vegaprotocol/devops-infra#1932


jeremyletang (commented):
this struct is initialised in NewDefaultConfig in blockexplorer/store/config.go which will create a default value for it, is 0 the default we want here or is there any good default value that we consider?

pscott31:
> this struct is initialised in NewDefaultConfig in blockexplorer/store/config.go which will create a default value for it, is 0 the default we want here or is there any good default value that we consider?

yeah, 0 means no timeout which is the current behaviour.. the thing is this code is shared with datanode proper as well and I didn't want to introduce any change there.. I could always make it block explorer specific

jeremyletang (approved):

PR #9191, opened 2023-08-24
feat: move market param to net

### LINKED ISSUE #9063: Add upper bound for calculationTimeStep
Upper bound of `calculationTimeStep` should be validators epochs length.

karlem:
Close #9063


protos/sources/vega/markets.proto
```diff
@@ -245,11 +245,11 @@ message LiquidityMonitoringParameters {
 }
 
 message LiquiditySLAParameters {
+  reserved 3; // Deprecated "providers_fee_calculation_time_step"
```
candida-d, line 248: Should this be deprecated rather than removed? If yes deprecated, then could you also deprecate in GQL?

jeremyletang (dismissed):

PR #9199, opened 2023-08-25
Consider running notional volume to determine referers and referees tier

### LINKED ISSUE #9152: Account for referral set running taker volume in reward factor selection
## Tier properties
- `window_length`: the number of epochs over which to evaluate a referral sets running notional taker volume
- `minimum_running_notional_taker_volume`: the required `referral_set_running_notional_taker_volume` in quantum units for parties to access this tier

### Party volumes

The network must now track the cumulative notional volume of taker trades for each party in an epoch, call this value party_epoch_notional_taker_volume. Note, trades generated by auction uncrossing are not counted. Each time a eligible trade is generated, the network should increment a parties party_epoch_notional_taker_volume by the quantum notional volume of the trade. For a spot market, the quantum is the quantum of the asset used to express the price (i.e. the [quote_asset](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0080-SPOT-product_builtin_spot.md/#1-product-parameters)).

```
party_epoch_notional_taker_volume = party_epoch_notional_taker_volume + (trade_price * trade_size * settlement_asset_quantum)
```

At the end of an epoch, the party_epoch_notional_taker_volume is stored by the network and each parties party_epoch_notional_taker_volume is reset to 0 ready for the next epoch.

### Referral set volumes

At the end of an epoch, for each referral set, a `referral_set_epoch_notional_taker_volume` is calculated by summing the `party_epoch_notional_taker_volume` of each party in the referral set (include both referrers and referees). The amount a party can contribute to their referral set is capped by the network parameter `referralProgram.maxPartyNotionalVolumeByQuantumPerEpoch`. Note this cap is not applied directly to `party_epoch_notional_taker_volume` in case the network parameter is updated during an epoch.

```
referral_set_epoch_notional_taker_volume = sum[min(party_epoch_notional_taker_volume, referralProgram.maxPartyNotionalVolumeByQuantumPerEpoch) for each party in team]
```

After the values are calculated, the `referral_set_epoch_notional_taker_volume` is stored by the network.

The network can then calculate the sets `referral_set_running_notional_taker_volume` by summing the sets `referral_set_epoch_notional_taker_volume` values over the last n epochs where n is the window_length set in the [governance proposal](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0083-RFPR-on_chain_referral_program.md#governance-proposals).

### Setting benefit factors

Whilst a referral program is `STATUS_ACTIVE`, at the start of an epoch (after pending `ApplyReferralCode` transactions have been processed) the network must set the referral_reward_factor and referral_discount_factor for each referee.
Setting the referral reward factor

The `referral_reward_factor` should be set by identifying the "highest" benefit tier where the following conditions are fulfilled.

`referral_set_running_notional_taker_volume` is greater than or equal to the tiers `minimum_running_notional_taker_volume`.

The referees `referral_reward_factor` is then set to the `referral_reward_factor` defined in the selected benefit tier.

### LINKED ISSUE #9173: Change `Membership.NumberOfEpoch` into `Membership.StartingAtEpoch`
The `NumberOfEpoch` is the wrong type of property in the `Membership` property

Taking a step back, it should be a `StartingAtEpoch` because we won’t have any other way to update the `NumberOfEpoch` than send gazillion of event to the DataNode. While with the `StartingAtEpoch`, a front-end would easily be able to compute the number of epoch spent in a team by applying the following calculation:

```
CurrentEpoch - StartingAtEpoch
```

ValentinTrinque:
Close #9152 
Close #9173


core/execution/common/market_activity_tracker.go
```diff
@@ -40,27 +40,54 @@ type marketTracker struct {
...
 	}
+
+	mat.clearNotionalTakerVolume()
+
 	epochEngine.NotifyOnEpoch(mat.onEpochEvent, mat.onEpochRestore)
 	return mat
 }
 
-// GetProposer returns the proposer of the market or empty string if the market doesn't exist.
+func (mat *MarketActivityTracker) TrackNotionalTakerVolume(party types.PartyID, volumeToAdd *num.Uint) {
```
ze97286, line 70: this should definitely not be merged. Can you not just use a dummy tracker and pass it to there referral engine? 

why?
1. the market_activity_tracker is radically changing in the reward PR - I can add it in there properly
2. it's missing here snapshot support
3. it's missing here the integration to the market which feeds it

ValentinTrinque: Noted ! Will use a dummy.

protos/sources/vega/events/v1/events.proto
```diff
@@ -725,6 +727,8 @@ message RefereeJoinedTeam {
   string referee = 2;
   // Time in Unix nanoseconds when the party joined a team.
   int64 joined_at = 3;
+  // Epoch at which the party joined the set.
```
wwestgarth, line 730:
```suggestion
  // Epoch at which the party joined the team.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -716,6 +716,8 @@ message RefereeSwitchedTeam {
   // Time in Unix nanoseconds when the party switched team. This time acts as
   // the joining time.
   int64 switched_at = 4;
+  // Epoch at which the party joined the set.
```
wwestgarth, line 719:
```suggestion
  // Epoch at which the party switched teams.
```

core/execution/engine.go
```diff
@@ -721,23 +721,7 @@ func (e *Engine) submitSpotMarket(ctx context.Context, marketConfig *types.Marke
...
-		e.timeService,
-		e.broker,
-		mas,
-		e.stateVarEngine,
-		e.marketActivityTracker,
-		bad,
-		qad,
-		e.peggedOrderCountUpdated,
-	)
+	mkt, err := spot.NewMarket(e.log, e.Matching, e.Fee, e.Liquidity, e.collateral, marketConfig, e.timeService, e.broker, mas, e.stateVarEngine, e.marketActivityTracker, bad, qad, e.peggedOrderCountUpdated)
```
ze97286, line 724: is there actually a change here other than formatting (which I like on the left more)?

core/referral/engine.go
```diff
@@ -255,47 +307,64 @@ func (e *Engine) loadReferralSetsFromSnapshot(sets *snapshotpb.ReferralSets) {
...
 			)
 		}
 
-		e.sets[set.Id] = newSet
+		e.sets[setID] = newSet
 	}
 }
 
-func NewEngine(epochEngine EpochEngine, broker Broker, timeSvc TimeService) *Engine {
+func (e *Engine) computeReferralSetsNotionalRunningVolume(epoch types.Epoch) {
```
ze97286, line 341: for this to work you need to make sure the referral engine is notified on end of epoch *after* market activity tracker

core/referral/notional_volumes.go
```diff
@@ -0,0 +1,90 @@
...
+	// runningVolumesBySet tracks the running notional volume by referralSets.
+	runningVolumesBySet map[types.ReferralSetID][]*notionalVolume
+}
+
+type notionalVolume struct {
+	epoch uint64
+	value *num.Uint
+}
+
+func (vs *runningVolumes) Add(epoch uint64, setID types.ReferralSetID, volume *num.Uint) {
```
ze97286, line 34: does it ever get cleaned up? seems to grow indefinitely unless I've missed the remove.

core/referral/snapshot.go
```diff
@@ -196,9 +199,9 @@ func (e *SnapshottedEngine) buildHashKeys() {
 	e.hashKeys = append([]string{}, e.currentProgramKey, e.newProgramKey, e.referralSetsKey)
 }
 
-func NewSnapshottedEngine(epochEngine EpochEngine, broker Broker, timeSvc TimeService) *SnapshottedEngine {
+func NewSnapshottedEngine(epochEngine EpochEngine, broker Broker, timeSvc TimeService, mat MarketActivityTracker) *SnapshottedEngine {
 	se := &SnapshottedEngine{
-		Engine:  NewEngine(epochEngine, broker, timeSvc),
+		Engine:  NewEngine(epochEngine, broker, timeSvc, mat),
```
ze97286, line 204: don't we have the referralSetsNotionalVolumes in the snapshot?

core/execution/engine.go
ValentinTrinque: The context was not needed, I can reformat to what is was. It will be clearer

core/referral/engine.go
ValentinTrinque: Yes.

core/referral/notional_volumes.go
ValentinTrinque: It grows indefinitely for now, we can and should limit. However we don't have a limit do the window_length. I will start a conversation about that.

core/referral/notional_volumes.go
```diff
@@ -0,0 +1,90 @@
...
+		vs.runningVolumesBySet[setID] = []*notionalVolume{
+			{
+				epoch: epoch,
+				value: volumeToAdd.Clone(),
+			},
+		}
+		return
+	}
+
+	for _, notionalVolume := range runningVolumeForSet {
```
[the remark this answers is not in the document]
ValentinTrinque: I wouldn't say it's extremely inefficient (~`O(n)`), but indeed, we can do what you suggest. That would be smarter.

core/referral/notional_volumes.go
ze97286: there should definitely be a limit on the window length. for example for the metrics we set it to 100 (to cover a quarter) or so.

core/referral/notional_volumes.go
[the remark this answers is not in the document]
ze97286: well o(n) vs o(1) :)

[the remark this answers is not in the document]
ValentinTrinque: Well CPU iterates fast. And `O(1)` is definitely better, I don't disagree. I just wanted to nuance the _extremely_ part.

core/referral/engine.go
ValentinTrinque: That's already the case but I think linking an engine to the epoch service during its initialization reduces flexibility in terms of lifecycle. I will move it outside and add a comment, so we don't forget.

jeremyletang (approved):

PR #9235, opened 2023-08-30
feat: add support for transfers out of the vested account

jeremyletang:


core/collateral/engine.go
```diff
@@ -2323,6 +2321,17 @@ func (e *Engine) getTransferFundsTransferRequest(ctx context.Context, t *types.T
 			// we always pay onto the pending transfers accounts
 			toAcc = e.GetPendingTransfersAccount(t.Amount.Asset)
 
+		case types.AccountTypeVestedRewards:
+			fromAcc = e.GetOrCreatePartyVestedRewardAccount(ctx, t.Owner, t.Amount.Asset)
+			if err != nil {
```
ze97286, line 2326: where is err coming from? you're creating the account if it doesn't exist so I'm not sure what you're doing here with the check.

jeremyletang: yea bad copy pasta of the previous lines.

ze97286 (approved):

PR #9253, opened 2023-08-31
feat: add transaction id to orders returned by api

### LINKED ISSUE #9221: [Feature]: Add transaction ID to orders returned by API
# Feature Overview

**In order to** link between client applications and the block explorer
**We will** add a transaction ID field to Orders
**So that** we can determine which transaction created an order

## Tasks
- [ ] In REST, the `Order` object should have a new `txId` field that contains a transaction ID
- [ ] In GraphQL, the `Order` object should have a new `txId` field that contains a transaction ID
- [ ] In both cases, the `txId` field should contain the hash of the transaction that created the order

## Feature test scenarios

GIVEN a `submitOrder` TX with a hash of `0497b51e5b30694edc1b031bad4c6165a925b270da16445eade20cce5522fad0`
AND that `submitOrder` creates a order with the deterministic ID of `49acf2f39ebbe92d708c313138dab3b70180b42a33e420db8553536fee2fab07`
WHEN I hit the [Get Order REST endpoint](https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-get-order)
THEN the Order should be returned
AND that order should contain a field called `txId`
And that `txId` field should contain `0497b51e5b30694edc1b031bad4c6165a925b270da16445eade20cce5522fad0`

guoguojin:
Closes #9221 

Not sure if the change in proto object would constitute a breaking change in the gRPC API, but labelling it as such to err on the side of caution, but REST and GQL should be fine.


jeremyletang:
closing this for now as well'll have this properly spec'd and implemented in a more general way over anything stored in the database.

PR #9269, opened 2023-09-01
Adding test for LIMIT, PEGGED and ICEBERG commitment

peterbarrow:
Adding tests and support function for SLA lp commitment checks.


core/integration/features/liquidity-provision/0044-LIME-028.feature
```diff
@@ -0,0 +1,122 @@
...
+      | mark price | trading mode            | target stake | supplied stake | open interest | best static bid price | static mid price | best static offer price |
+      | 1000       | TRADING_MODE_CONTINUOUS | 1000         | 1000           | 10            | 900                   | 1000             | 1100                    |
+
+    # Place ICEBERG orders to cover our commitment
+    When the parties place the following iceberg orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     | peak size | minimum visible size | only |
+      | party1 | ETH/DEC21 | buy  | 20000  | 999   | 0                | TYPE_LIMIT | TIF_GTC | 100       | 10                   | post |      
+      | party1 | ETH/DEC21 | sell | 20000  | 1001  | 0                | TYPE_LIMIT | TIF_GTC | 100       | 10                   | post |      
+
+    # Move forward an epoch and make sure the accounts do not change as we have the full epoch covered with ICEBERG orders
```
wwestgarth, line 68: If the iceberg peak of 100 covers the commitment, so does the full iceberg volume so if by accident the full iceberg volume is being taken into account the test will still pass. So maybe we need the test to be such that their commitment *isn't* covered by the iceberg peaks but is by the full volume, and then check they are penalised.

peterbarrow: Added test to cover this case and it found a bug. I've made the change to fix it in this PR. Thanks Will.

core/liquidity/v2/sla.go
```diff
@@ -149,7 +149,7 @@ func (e *Engine) doesLPMeetsCommitment(
 		price := o.Price.ToDecimal()
 		// this order is in range and does contribute to the volume on notional
 		if price.GreaterThanOrEqual(minPrice) && price.LessThanOrEqual(maxPrice) {
-			orderVolume := num.UintZero().Mul(markPrice, num.NewUint(o.TrueRemaining())).ToDecimal().Div(positionFactor)
+			orderVolume := num.UintZero().Mul(markPrice, num.NewUint(o.VisibleRemaining())).ToDecimal().Div(positionFactor)
```
wwestgarth, line 152: I have a feeling that "visible peak" in the AC just means `o.Remaining`? I also don't really know why only the visible peak counts, so I might ask protocol design.

peterbarrow: Spec is changed to mean all available volume, so reverting to the old code/

jeremyletang (approved):

wwestgarth (approved):

PR #9353, opened 2023-09-11
feat: add more product specific data to market-data

### LINKED ISSUE #9319: [Perps]: API insufficient to satisfy `0053-PERP-019`
# Feature Overview

The AC is:
```
For the ongoing period the following data is available via the API: funding period start time, estimate time, funding rate estimate, funding payment estimate, external (spot) price TWAP to-date, internal (mark) price TWAP to-date.
```

The data we can currently get for an _on-going_ funding period is:
- the funding-rate at every block, from the market data
- the internal TWAP _when the last internal point was received_, from the list-data-points API
- the external TWAP _when the last external point was received_, from the list-data-points API
- funding payment, nowhere

I think what we need to do is to extend the market data events to include `{funding payment, funding rate, the TWAPs}` instead of just the funding rate. Then all the data is available at every block. We have all the other data to hand anyway when we set the funding rate in the market data, so we just need to add them to the event (and the datanode).

Might also be worth having it all in a `oneof` field for like `product-data` within the market-data so that all these perps only fields just aren't there for other market types.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
closes #9319 

Extended the market-data API to expose all the necessary perptual "in progress" funding infomation as stated in `0053-PERP-019`. It adds a `productData` field to the market data which can hold a one-of depending on product type so that all the new fields do not clutter up non-perpetual markets.
REST:
Screenshot 2023-09-11 at 13 57 27

graphQL (one perp, one future)
Screenshot 2023-09-11 at 13 50 14


jeremyletang (dismissed):

core/products/perpetual.go
```diff
@@ -46,6 +46,13 @@ const (
 	dataPointSourceInternal dataPointSource = eventspb.FundingPeriodDataPoint_SOURCE_INTERNAL
 )
 
+type fundingData struct {
+	fundingPayment *num.Int
+	fundingRate    *num.Decimal
```
jeremyletang, line 51: Why is fundingRate a pointer? usally we neved use them as pointer.
Also I would have made the fundingPayment a Uint, it sound weird to me to have a negative amount to be paid, I guess you can figure where the payment is going basd on the fundingRate being negative or positive?

jeremyletang: althought I realise it's already un Int everywhere, I'm just having this thoughts about it now.

wwestgarth: I don't know why its a pointer, I'm made it not one.

The funding payment is signed because by definition it is the amount being paid and the direction, and the funding-rate is just the funding payment scaled so I think it would be more weird to not have them both signed:
```
fundingPayment = (f_twap - s_twap)
fundingRate = funding_payment /  s_twap
```

jeremyletang (dismissed):

jeremyletang (dismissed):

protos/sources/vega/vega.proto
```diff
@@ -1138,6 +1138,25 @@ message MarginLevels {
...
+  string funding_payment = 1;
+  // Current funding rate for the in-progress funding period.
+  string funding_rate = 2;
+  // Time-weighted-average the internal data-points for the in-progress funding period.
+  string internal_twap = 3;
+  // Time-weighted-average the external data points for the in-progress funding period.
+  string external_twap = 4;
+}
+
+// Represets market data specific to a particular product type.
```
gordsport, line 1153: should it be this?

```suggestion
// Represents market data specific to a particular product type.
```

jeremyletang (dismissed):

mattrussell36 (dismissed):

jeremyletang (approved):

PR #9373, opened 2023-09-12
feat: prevent malicious validator from submitting additional ethereum call oracle chain events

### LINKED ISSUE #9343: [Bug]: Historic blocks of an eth-contract are not rejected allowing manipulation of settlement price
### Problem encountered

As a malicious validator I can submit a `ContractCall` chain events to the network for a valid contract call, but for an _historic_ ethereum block and they are accepted and propagated to the product.

For example, if I propose a market with an eth-call data source that looks like:
```
{
    "EthTimeTrigger": {
        "initial": [6 months from now],
        "every: "10",
    }
}
```

I can manually submit a chain event with a valid ethereum call for a block height from _a year ago_ and it will be verified and used to terminate/settle a market. So if I find any price that a like from the history of that contract, provided it passes the filters, I can manipulate the market's settlement to be that price.

### Observed behaviour

Contract values from any point in history can be used to settle a market that uses etherem oracles.

### Expected behaviour

Only contract values from any point _after_ the creation of the market, or the `initial` value in the trigger options can be used to settle a market.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ettec:
closes #9343 

ethereum oracles system test run-> https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/8178/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,7 @@
 - [8545](https://github.com/vegaprotocol/vega/issues/8545) - Block Explorer pagination does not order correctly.
 - [8748](https://github.com/vegaprotocol/vega/issues/8748) - `ListSuccessorMarkets` does not return results.
 - [8749](https://github.com/vegaprotocol/vega/issues/8749) - Ensure stop order expiry is in the future
+- [9343](https://github.com/vegaprotocol/vega/issues/9343) - Prevent malicious validator from submitting additional Ethereum call oracle chain events
 - [9337](https://github.com/vegaprotocol/vega/issues/9337) - Non deterministic ordering of vesting ledger events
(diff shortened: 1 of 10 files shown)


wwestgarth:
I don't really like having this strong link between the the ethcall engine and the verifier. I think its moving away from the concept of our data-sources where core received _something_ from the outside and then it filters it down. It comes back down to the other discussion where multiple different specs with the same contract at its root should be viewed as all having the same data-source. We don't do that currently, and I think this is moving even further away from that.

I think its sufficient to just have a `if blockHeight < spec.initial { reject }` check, and the case of a validator sending in manual events more frequently than specified is much less of an issue that we shouldn't worry about it.

PR #9381, opened 2023-09-13
chore: better metrics for snapshots

### LINKED ISSUE #8775: Better snaspshot engine metrics
The metrics timing the elapsed time for each snapshot provider forward the same time for every one of them. Seems to be buggy. We need to correct that. 

This is because there is a single metric to account for all engine. The metric has probably be thought to be used like this:

```go
m := newMetric("provider-1")

SnapshotProvider1()

m.flush()

m = newMetric("provider-2")

SnapshotProvider2()

m.flush()
```

However, this is not how the code work.

First, the snapshot engine deserializes the entire state for all namespaces, then handle it. It doesn't handle each namespace individually, so the above logic (in the code block) doesn't apply.

Second, it's done in parallel, and multiple tasks might be snapshotting the same namespace, so the metric should only account for the time spent by the worker on this very specific namespace.

A more sensible approach would be:

```go
metricsByNamespaces := map[string]Metric{
  "provider-1": NewMetric(),
  "provider-2": NewMetric()
}

func SnapshotNow() {
  for _, tk := range treeKeys {
     go worker(tk)
  }
}

func worker(tk string) {
  startSnapshot := time.Now()
  defer func() {
     endSnapshot := time.Now()
     n := namespaceByTreeKeys[tk]
     metricsByNamespaces[n].Add(endRestoration - startRestoration)
  }
  
  // Snapshot some state.
}

### PR DESCRIPTION
**Pull request by jeremyletang**
close #8775

### DIFF
--- a/core/metrics/prometheus.go
+++ b/core/metrics/prometheus.go
@@ -56,7 +56,11 @@ var (
 	// Total time counters for each request type per API.
 	apiRequestTimeCounter *prometheus.CounterVec
 	// Total time spent snapshoting.
-	snapshotTimeCounter *prometheus.CounterVec
+	snapshotTimeGauge *prometheus.GaugeVec
+	// Size of the snapshot per namespace.
+	snapshotSizeGauge *prometheus.GaugeVec
+	// Height of the last snapshot.
+	snapshotBlockHeightCounter prometheus.Gauge
 	// Core HTTP bindings that we will check against when updating HTTP metrics.
 	httpBindings *protos.Bindings
 )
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: core/metrics/prometheus.go [hunk 1]
```diff
@@ -482,19 +486,53 @@ func setupMetrics() error {
 	// snapshots times
 	h, err = addInstrument(
 		Counter,
-		"snapshot_time_total",
+		"snapshot_time_total_seconds",
```
fkondej, line 489: it is no longer counter, so maybe we should remove `total` from name?

core/metrics/prometheus.go
```diff
@@ -482,19 +486,53 @@ func setupMetrics() error {
...
 	if err != nil {
 		return err
 	}
-	snapshotTimeCounter = snap
+	snapshotTimeGauge = snap
+
+	// snapshots sizes
+	h, err = addInstrument(
+		Counter,
+		"snapshot_size_total_bytes",
```
fkondej, line 506: this is also a gauge, so maybe remove `total` from name

wwestgarth (approved):

ValentinTrinque (approved):

ValentinTrinque (approved):

PR #9384, opened 2023-09-13
feat: reject call events with block time prior to the specifications initial time

### LINKED ISSUE #9343: [Bug]: Historic blocks of an eth-contract are not rejected allowing manipulation of settlement price

ettec:
closes #9343.  This change also modifies the behaviour of the of an eth oracle spec trigger that has an 'every' period and no 'inital' time so that the every period is now relative to the enactment time, previously the trigger would fire when Mod(period, ethereum time) == 0 which could be an arbitrary interval after enactment.

### DIFF
--- a/commands/proposal_submission.go
+++ b/commands/proposal_submission.go
@@ -1362,9 +1362,8 @@ func checkDataSourceSpec(spec *vegapb.DataSourceDefinition, name string, parentP
 
 				if ethOracle.Trigger != nil &&
 					ethOracle.Trigger.GetTimeTrigger() != nil &&
-					(ethOracle.Trigger.GetTimeTrigger().Initial == nil || *ethOracle.Trigger.GetTimeTrigger().Initial == 0) &&
-					(ethOracle.Trigger.GetTimeTrigger().Every == nil || *ethOracle.Trigger.GetTimeTrigger().Every == 0) {
-					errs.AddForProperty(fmt.Sprintf("%s.%s.external.ethoracle.trigger.timetrigger.(initial|every)", parentProperty, name), ErrIsRequired)
+					(ethOracle.Trigger.GetTimeTrigger().Initial == nil || *ethOracle.Trigger.GetTimeTrigger().Initial == 0) {
+					errs.AddForProperty(fmt.Sprintf("%s.%s.external.ethoracle.trigger.timetrigger.initial", parentProperty, name), ErrIsRequired)
 				}
 			} else {
 				errs.AddForProperty(fmt.Sprintf("%s.%s.external.oracle", parentProperty, name), ErrIsRequired)
--- a/commands/proposal_submission.go
+++ b/commands/proposal_submission.go
@@ -145,11 +145,11 @@ func checkProposalChanges(terms *protoTypes.ProposalTerms) Errors {
 
 	switch c := terms.Change.(type) {
 	case *protoTypes.ProposalTerms_NewMarket:
-		errs.Merge(checkNewMarketChanges(c))
+		errs.Merge(checkNewMarketChanges(c, terms.EnactmentTimestamp))
 	case *protoTypes.ProposalTerms_UpdateMarket:
-		errs.Merge(checkUpdateMarketChanges(c))
+		errs.Merge(checkUpdateMarketChanges(c, terms.EnactmentTimestamp))
 	case *protoTypes.ProposalTerms_NewSpotMarket:
-		errs.Merge(checkNewSpotMarketChanges(c))
+		errs.Merge(checkNewSpotMarketChanges(c, terms.EnactmentTimestamp))
 	case *protoTypes.ProposalTerms_UpdateSpotMarket:
 		errs.Merge(checkUpdateSpotMarketChanges(c))
 	case *protoTypes.ProposalTerms_UpdateNetworkParameter:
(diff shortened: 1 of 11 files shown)


commands/proposal_submission.go
```diff
@@ -1361,9 +1361,8 @@ func checkDataSourceSpec(spec *vegapb.DataSourceDefinition, name string, parentP
 
 				if ethOracle.Trigger != nil &&
 					ethOracle.Trigger.GetTimeTrigger() != nil &&
-					(ethOracle.Trigger.GetTimeTrigger().Initial == nil || *ethOracle.Trigger.GetTimeTrigger().Initial == 0) &&
```
ettec, line 1364: Seems reasonable to require an initial time, having the option to specify 'every' without an initial can lean to some arguably unexpected behaviour - e.g if 'every' is set to 1000 and  current block is 999, it will trigger immediately on the next block which I think would be unexpected behaviour, probably better to err on the side of caution and have the initial time as mandatory to avoid any doubt.   Also makes this fix easy as we now have a time against which we can verify eth call events to prevent submission of historical events.

ettec:
...pending eth oracle system tests before moving out of draft

commands/proposal_submission.go
jeremyletang: what will be the value of every if it's not set here? I don't get how every can be optional and not Initial, I would have expected the inverse.
If initial is not set, set it at time of enactment of the proposal. 
If every is not set, set it to what arbitrary value?

jeremyletang: Alright I guess, I still think the Initial should be optional and set dynamically to the enactment date of the proposal if not specified, @wwestgarth any thoughts on this whole thing?

wwestgarth: I feel like setting it based off the enactment data is the more correct thing to do. Making it non-optional is definitely easier, but would need signing off with the spec team.

ettec: If its set to the enactment date/time of the proposal, then depending on how in/out of sync a nodes etherium client is it may trigger immediately? so I think we would have to then add some offset to prevent this?  seems a bit messy

ettec: perhaps enactment data plus the every interval?

ettec: ....summary of chat:  The spec activation will always occur sufficiently far ahead of the enactment time for the boundary condition of an eth client's block time being slightly ahead/behind of the vega time resulting in unpredicatable initial trigger behaviour to not be an issue.

ettec:
Pending successful runs of the below tests and merging of sys test branch fixes:

Eth oracle tests against fixed up sys test branch-> https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/10036/

Full run against fixed up sys test branch-> https://jenkins.vega.rocks/job/common/job/system-tests/1979/

Merging of sys test PR -> https://github.com/vegaprotocol/system-tests/pull/2706

commands/proposal_submission.go
```diff
@@ -1362,9 +1364,9 @@ func checkDataSourceSpec(spec *vegapb.DataSourceDefinition, name string, parentP
...
-					(ethOracle.Trigger.GetTimeTrigger().Every == nil || *ethOracle.Trigger.GetTimeTrigger().Every == 0) {
-					errs.AddForProperty(fmt.Sprintf("%s.%s.external.ethoracle.trigger.timetrigger.(initial|every)", parentProperty, name), ErrIsRequired)
+					(ethOracle.Trigger.GetTimeTrigger().Initial == nil || *ethOracle.Trigger.GetTimeTrigger().Initial == 0) {
```
wwestgarth, line 1367: I don't love that we're doing this in the command validation because I think its unexpected that a validation step would change values. For example this will be called by the CLI/Desktop wallet, so if I send in a proposal with `initial == nil` the transaction that will appear in the wallet to verify manually will have `initial` set to something and the transaction will look different than I asked it to send.

As I linked to you in the other PR, we do the exact same thing of setting `initial` to `enactmentTime` when its `nil` for internal time triggered oracles, and that happens in the governance engine: https://github.com/vegaprotocol/vega/blob/develop/core/governance/market.go#L583 We should just so that same here too.

jeremyletang: yes please, let's not change the transaction in this layer, this should be only validation, and the initial time if null should be applied in the governance engine. @ettec

ettec:
New eth oracle tests-> https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/10974/

ettec:
Moved to draft, needs sys tests

wwestgarth (dismissed):

ettec:
eth oracle Sys test run with new test-> https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/11424/

ettec:
latest full run-> https://jenkins.vega.rocks/job/common/job/system-tests/2274/

jeremyletang (approved):

PR #9395, opened 2023-09-13
feat: add endpoint to get transfer by ID

### LINKED ISSUE #7461: New REST/GraphQL endpoint: Fetch transfer by ID
# Problem encountered
REST and GraphQL both have methods for fetching a *list* of transfers, with optional filters by time or party. There is no way to get a transfer by ID.

# Expected behaviour
Either `transfersConnection` has a new filter, `id`, or a new endpoint/query is added for `transferByID`

EVODelavega:
Closes #7461


jeremyletang (commented):
could you also add the rest binding please.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -224,6 +224,13 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Transfers"};
   }
 
+  // Get transfer by ID
+  //
+  // Get a specific transfer by ID
+  rpc GetTransferByID(GetTransferByIDRequest) returns (GetTransferByIDResponse) {
```
wwestgarth, line 230: This should either be just `GetTransfer` so that its named the same as all the other `GetSomething` API that retrieve something by their ID, or (which is something more recent thats being done with new API recently, see `ListReferralSets`) adding a `optional string transfer_id` to the `ListTransfersRequest`.

datanode/sqlstore/transfers.go
```diff
@@ -174,6 +174,13 @@ func (t *Transfers) GetByTxHash(ctx context.Context, txHash entities.TxHash) ([]
 	return transfers, nil
 }
 
+func (t *Transfers) GetByID(ctx context.Context, id string) (entities.Transfer, error) {
```
wwestgarth, line 177: Can you add a unit-test please?

protos/sources/data-node/api/v2/trading_data.proto
EVODelavega: I ended up adding this as a separate call because this is only expected to return nothing but the current version of the specific transfer, so the whole `ListTransfers` pagination stuff is just noise, or might suggest that providing the ID would return the the historical transfer data (pending record, final record, etc...). I can change it, though

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1613,6 +1620,18 @@ message TransferConnection {
...
 
+// Request that is sent for getting a transfer by ID
+message GetTransferByIDRequest {
+  // Transfer ID to request data for.
+  string transfer_id = 1 [(google.api.field_behavior) = REQUIRED];
+}
+
+// Response that is received when querying transfers by ID
+message GetTransferByIDResponse {
+  // page containing the transfer requested by ID.
```
gordsport, line 1631: should this be a capital 'P' for consistency as per the other similar strings above

EVODelavega: This is actually an incorrect comment, must've deleted the wrong line

datanode/gateway/graphql/schema.graphql
```diff
@@ -858,6 +858,9 @@ type Query {
     pagination: Pagination
   ): TransferConnection
 
+  "Find a transfer using its ID"
```
gordsport, line 861: bit of a nit pick but should this be GET as per: https://github.com/vegaprotocol/vega/blob/develop/DOCUMENTATION_STYLE.md#naming-terminology

I know its FIND has been used below on line 864

jeremyletang (approved):

PR #9407, opened 2023-09-14
fix: Ethereum oracle validation failing when using go 1.19

### LINKED ISSUE #9406: [Bug]: Partially copied 1.20 errors library code allowed vega to compile under 1.19 but would result in incorrect eth oracle validation
### Problem encountered

Part of the errors library from go 1.20 to 'join' errors was copied into vega library to allow the the errors join function to be used in go 1.19, however, the corresponding errors.Is code was not copied across meaning that when compiled under go 1.19 the errors logic for validation of the eth oracle specs was broken as pre 1.20 the errors.Is logic does not know how to deal with joined errors.

### Observed behaviour

eth oracle validation systm tests would all pass on go 1.20, tests would fail when using 1.19

### Expected behaviour

"

### Steps to reproduce

```bash
"
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ettec:
closes #9406 
Will require that all envs are using go 1.20+ going forward, if for any reason this is a problem we should not merge this PR.


jeremyletang (dismissed):

jeremyletang (commented):
you will need to change the github workflow to use 1.20 too.

PR #9429, opened 2023-09-15
feat: add additional filters for the referral sets APIs

### LINKED ISSUE #9417: ListReferralSet and ListReferralSetReferees should also allow filtering by Referrer and Referee as well as set id
Currently only referral set id can be provided, but we should allow for filtering by the referrer or the referee.

guoguojin:
Closes #9417


jeremyletang (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3654,9 +3658,13 @@ message ReferralSetRefereeConnection {
 // Request to retrieve information about a referral set's referees
 message ListReferralSetRefereesRequest {
   // Referral set ID to retrieve information for
-  string referral_set_id = 1;
+  optional string referral_set_id = 1;
   // Pagination information to limit the data that is returned.
   optional Pagination pagination = 2;
+  // Referrer to filter by. If referrer set id is provided, this field is ignored.
```
gordsport, line 3664: according to the docs guidelines we should be using `ID` not `id`

https://github.com/vegaprotocol/vega/blob/develop/DOCUMENTATION_STYLE.md#what-not-to-use-and-what-to-use-instead

jeremyletang (approved):

gordsport (approved):

jeremyletang (approved):

PR #9448, opened 2023-09-18
fix: eth oracles broadcasting to every spec with matching data keys

### LINKED ISSUE #9332: [Bug]: Incorrect market is settled when using same settlement key with same filters BUT different contracts
### Problem encountered

Spec Ref:
https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0082-ETHD-ethereum-data-source.md

AC Ref:

Different contracts on different markets - Create 2 markets with ethereum oracle settlement data sources containing different contract addresses and with the same settlement keys, but with all conditions and filters the same. Confirm that sending settlement value that passes the market spec filter only settles one market. ([0082-ETHD-051](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0082-ETHD-ethereum-data-source.md#0082-ETHD-051))

Tried to test the above AC and looks like it isn't working correctly.

When testing the above , noticed that all markets using the same settlement key and filters are being settled although they are using different settlement contracts.

### Observed behaviour

Scenario:

Create 2 markets settling with different ethereum contracts BUT with the same settlement key and filter values.
Settle market 1 by passing the correct settlement value to the correct contract.

Both markets are being settled even though the markets are using different ethereum contracts.

### Expected behaviour

Only the market using the ethereum contract for which the settlement price is set should be settled.

### Steps to reproduce

```bash
Run the below system test:

System Test Name - test_eth_oracle_different_events_for_different_contracts
System Test Branch - feat_eth_oracle_acs_50_51
```

### Software version

vega - develop

### Failing test

test_eth_oracle_different_events_for_different_contracts

### Jenkins run

https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/6161/testReport/junit/tests.ethereum_oracles/ethereum_call_test/test_eth_oracle_different_events_for_different_contracts_True_/

### Configuration used

_No response_

### Relevant log output

```shell
All the relevant events for market and oracle spec creation , update and oracle data are here.

https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/6161/artifact/test_logs/ethereum_call_test/test_eth_oracle_different_events_for_different_contracts%5BTrue%5D/
```

pscott31:
closes #9332

### DIFF
--- a/core/datasource/spec/spec.go
+++ b/core/datasource/spec/spec.go
@@ -165,6 +165,12 @@ func (s *Spec) MatchData(data common.Data) (bool, error) {
 		return false, nil
 	}
 
+	// Don't broadcast ethcall data based unless it's 'EthKey' matches
+	// (which is currently the SpecID - see comment on the datasource.common.Data struct)
+	if data.EthKey != "" && data.EthKey != string(s.id) {
+		return false, nil
+	}
+
 	// if it is internal time data and we have a time-trigger check that we're past it
 	if ok, tt := isInternalTimeTrigger(data); ok && s.triggers[0] != nil {
 		if !s.triggers.IsTriggered(tt) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,7 @@
 - [9385](https://github.com/vegaprotocol/vega/issues/9385) - Support deprecated liquidity auction type for compatibility
 - [9398](https://github.com/vegaprotocol/vega/issues/9398) - Fix division by zero panic in market liquidity
 - [9413](https://github.com/vegaprotocol/vega/issues/9413) - Fix range validation for SLA parameters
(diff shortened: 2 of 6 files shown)


core/datasource/spec/spec.go
```diff
@@ -165,6 +165,12 @@ func (s *Spec) MatchData(data common.Data) (bool, error) {
 		return false, nil
 	}
 
+	// Don't broadcast ethcall data based unless it's 'EthKey' matches
+	// (which is currently the SpecID - see comment on the datasource.common.Data struct)
+	if data.EthKey != "" && data.EthKey != string(s.id) {
```
wwestgarth, line 170: could you add a little unit-test in `spec_test.go`?

wwestgarth (dismissed):

core/datasource/spec/spec.go
pscott31: sure

wwestgarth (approved):

PR #9459, opened 2023-09-18
Add steps for volume discount program

### LINKED ISSUE #9445: [Feature]:  Add steps for volume discount program
# Feature Overview

**In order to** test the volume discount program in feature tests 
**We will** require new steps
**So that** we can propose programs, and check discount factors.

#### Step: "the volume discount tiers 'name':"
A step which can define volume discount tiers to be passed in a volume discount program governance proposal.
- mandatory columns
  - min running volume
  - discount factor

- [x] Done
 
#### Step: “the volume discount program:”
A step for proposing a volume discount program.
- mandatory columns
  - discount tiers (should be the name of either default or previously defined discount tiers)
  - closing timestamp
  - window length
- optional columns
  - err

- [x] Done

#### Step: “the volume discount stats at epoch "N":”
A step for checking a parties volume discount program stats at a specific epoch.
(Can be flexible in this step, I have just specced it in a fashion similar to the equivalent referral program step).
- mandatory fields
  - epoch
- mandatory columns
  - party
  - running volume
  - discount factor

- [x] Done

#### Step: "the following trades should be executed:"
Needs to be enhanced to optionally report applied volume discounts (this is already done but including for completeness).

- [x] Done

#### Step: "the following network parameters are set:"
Needs to be enhanced to cover the following:
- `volumeDiscountProgram.maxBenefitTiers`
- `volumeDiscountProgram.maxVolumeDiscountFactor`

- [x] Done

## Tasks
- [ ] implement cucumber steps for the volume discount program

ValentinTrinque:
Close #9445


datanode/entities/volume_discount.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"code.vegaprotocol.io/vega/libs/num"
+	eventspb "code.vegaprotocol.io/vega/protos/vega/events/v1"
+)
+
+type (
+	VolumeDiscountStatsUpdated struct {
+		AtEpoch                  uint64
+		PartyVolumeDiscountStats []*PartyVolumeDiscountStats
```
guoguojin, line 14: If you don't need to, I think you can just use the proto message here instead of having to map it out.

ValentinTrinque: happy to use it !

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,24 @@ import "vega/vega.proto";
 
 option go_package = "code.vegaprotocol.io/vega/protos/vega/events/v1";
 
+// Stats of all parties eligible to volume discount.
```
candida-d, line 16:
```suggestion
// Stats of all parties eligible for volume discount.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,24 @@ import "vega/vega.proto";
...
+message VolumeDiscountStatsUpdated {
+  // Epoch at which the volume discount statistics are updated.
+  uint64 at_epoch = 1;
+  // All parties stats.
```
candida-d, line 20:
```suggestion
  // All parties' stats.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,24 @@ import "vega/vega.proto";
...
+  repeated PartyVolumeDiscountStats stats = 2;
+}
+
+// The volume discount stats for a given party.
+message PartyVolumeDiscountStats {
+  // The party.
```
candida-d, line 26:
```suggestion
// Volume discount stats for a given party.
message PartyVolumeDiscountStats {
  // Party ID.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,24 @@ import "vega/vega.proto";
...
+  string party_id = 1;
+  // Discount factor applied to the volume.
+  string discount_factor = 2;
+  // Running volume of the party.
```
candida-d, line 30:
```suggestion
  // The party's running volume.
```

guoguojin (dismissed):

jeremyletang (dismissed):

core/integration/steps/the_volume_discount_stats_should_be.go
```diff
@@ -0,0 +1,142 @@
...
+			PartyId:        partyID,
+			DiscountFactor: specificRow.DiscountFactor(),
+			RunningVolume:  specificRow.RunningVolume(),
+		}
+	}
+
+	return stats, nil
+}
+
+func compareVolumeDiscountStats(expectedVolumeDiscountStats map[string]eventspb.PartyVolumeDiscountStats, foundVolumeDiscountStats map[string]eventspb.PartyVolumeDiscountStats) error {
```
EVODelavega, line 63: nit-pick:
```go
func compareVolumeDiscountStats(expectedVolumeDiscountStats, foundVolumeDiscountStats map[string]eventspb.PartyVolumeDiscountStats) error {
```

So as to not repeat the rather long type name

core/integration/stubs/broker_stub.go
```diff
@@ -936,6 +936,21 @@ func (b *BrokerStub) VestingStats() []eventspb.VestingStatsUpdated {
...
 }
 
+func (b *BrokerStub) VolumeDiscountStats() []eventspb.VolumeDiscountStatsUpdated {
+	batch := b.GetBatch(events.VolumeDiscountStatsUpdatedEvent)
+
+	stats := make([]eventspb.VolumeDiscountStatsUpdated, 0, len(batch))
+	for _, event := range batch {
+		switch et := event.(type) {
+		case *events.VolumeDiscountStatsUpdated:
+			stats = append(stats, et.Proto())
```
EVODelavega, line 946: Probably out of scope of this PR, but the `events.Proto()` functions are inconsistent WRT their return values. `Party.Proto()` returns `*proto.Party`, whereas most other types like `RiskFactor.Proto()` return `proto.RiskFactor`. There may be a good reason for this behaviour, but if there isn't, we probably want to make this consistent at some point.

ValentinTrinque: Created the issue.

core/integration/steps/the_volume_discount_stats_should_be.go
```diff
@@ -0,0 +1,142 @@
...
+			PartyId:        partyID,
+			DiscountFactor: specificRow.DiscountFactor(),
+			RunningVolume:  specificRow.RunningVolume(),
+		}
+	}
+
+	return stats, nil
+}
+
+func compareVolumeDiscountStats(expectedStats map[string]eventspb.PartyVolumeDiscountStats, foundStats map[string]eventspb.PartyVolumeDiscountStats) error {
```
EVODelavega, line 63: repeated arg types

jeremyletang (approved):

PR #9467, opened 2023-09-19
refactor: include funding payment in liq. price

### LINKED ISSUE #8764: [Feature]: Perps - Datanode API: EstimatePosition
# Feature Overview

**In order to** Have the `EstimatePosition` end-point reflect the margin calculations for a perps market
**We will** extend it to use the `margin_funding_factor` where appropriate
**So that** FE can give better estimates for required margins in the UI

This may or may not be possible depending what data is available in the data node. If we want to add on `margin_funding_factor * G(t_k-1, t)` we could use the funding payment from the last complete funding period which would be easiest. Or if we want to use the funding factor of the in progress funding period we would have to calculate the twap-difference in the data node from the data points. So it depends how accurate we want to be.

There is also the `EstimateMargin` which I believe is deprecated in favour of `EstimatePosition` so we may or may not want to add something there.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

witgaw:
* Include funding payment in margin and liquidation price estimates for perps.
* Additional tests to assure consistency between internal margin calculations and the static method used by the `EstimatePosition` endpoint.
* Optional argument to return liquidation price estimate scaled to market decimal places.

Close https://github.com/vegaprotocol/vega/issues/8764


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3510,7 +3511,10 @@ message EstimatePositionRequest {
   // Open and/or hypothetical orders.
   repeated OrderInfo orders = 3;
   // Optional argument specifying collateral available for the position, if provided then response will contain the liquidation price estimate.
+  // Collateral available needs to be provided scaled to asset decimal places. Liquidation price returned scaled to asset decimal places by default.
```
candida-d, line 3514:
```suggestion
  // Collateral available needs to be provided scaled to asset decimal places. Liquidation price returned is scaled to asset decimal places by default.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3510,7 +3511,10 @@ message EstimatePositionRequest {
...
   optional string collateral_available = 4;
+  // Optional argument specifying if the liqudiation price estimates should be scaled to market decimal places (scaled to asset decimal places by default).
```
candida-d, line 3516:
```suggestion
  // Argument specifying if the liquidation price estimates should be scaled to market decimal places, by default they are scaled to asset decimal places.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -574,8 +574,9 @@ service TradingDataService {
 
   // Estimate position
   //
-  // Estimate the margin that would be required for maintaining the specified position.
+  // Estimate the margin that would be required for maintaining the specified position. Margin estimates are scaled to asset decimal places.
   // If the optional collateral available argument is supplied, the response also contains the estimate of the liquidation price.
+  // Liquidation price estimates are scaled to asset decimal places by default, unless a flag to scale to market decimal places is specified in the request.
```
candida-d, line 579:
```suggestion
  // Liquidation price estimates are scaled to asset decimal places by default, unless an argument to scale to market decimal places is specified in the request.
```

candida-d (changes requested):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3510,7 +3511,10 @@ message EstimatePositionRequest {
   // Open and/or hypothetical orders.
   repeated OrderInfo orders = 3;
   // Optional argument specifying collateral available for the position, if provided then response will contain the liquidation price estimate.
+  // Collateral available needs to be provided scaled to asset decimal places. Liquidation price returned scaled to asset decimal places by default.
   optional string collateral_available = 4;
+  // Optional argument specifying if the liqudiation price estimates should be scaled to market decimal places (scaled to asset decimal places by default).
+  optional bool scale_liquidation_to_market_decimals = 5;
```
candida-d, line 3517: Should this also be added to GraphQL?

jeremyletang (commented):
Could you had the changes to the graphql API as well please?

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Yep, added now, thanks

candida-d (dismissed):

jeremyletang (approved):

PR #9469, opened 2023-09-19
Feature test vdpr-0013

peterbarrow:
Check each tier level is used as taker notional changes.


core/integration/features/0084-VDPR-013.feature
```diff
@@ -0,0 +1,121 @@
...
+    Then the parties place the following orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     |
+      | party1 | ETH/MAR24 | buy  | 10     | 900   | 0                | TYPE_LIMIT | TIF_GTC |
+      | party1 | ETH/MAR24 | buy  | 1      | 1000  | 0                | TYPE_LIMIT | TIF_GTC |
+      | lp1    | ETH/MAR24 | buy  | 100    | 990   | 0                | TYPE_LIMIT | TIF_GTC |
+      | lp1    | ETH/MAR24 | sell | 100    | 1010  | 0                | TYPE_LIMIT | TIF_GTC |
+      | party2 | ETH/MAR24 | sell | 10     | 1100  | 0                | TYPE_LIMIT | TIF_GTC |
+      | party2 | ETH/MAR24 | sell | 1      | 1000  | 0                | TYPE_LIMIT | TIF_GTC |
+
+    Then the opening auction period ends for market "ETH/MAR24"
```
EVODelavega, line 86: This is fine here, but should be noted that in scenarios with more than 1 market, this will cause problems. It's probably best to use `When the network moves ahead "1" blocks` so it's clear that this step progresses time and therefore affects all markets.

EVODelavega (approved):

PR #9473, opened 2023-09-19
fix: list teams API returns an error when filtering

### LINKED ISSUE #9472: [Bug]: grpc query of ListTeams returns rpc error
### Problem encountered

After submitting a signed `CreateReferralSet` with teams

querying `ListTeams` by referrer returns the error

<_InactiveRpcError of RPC that terminated with:
	status = StatusCode.INTERNAL
	details = "Internal error"
	debug_error_string = "UNKNOWN:Error received from peer ipv4:127.0.0.1:3027 {created_time:"2023-09-19T13:44:19.829577+01:00", grpc_status:13, grpc_message:"Internal error"}"
>

### Observed behaviour

rpc error

### Expected behaviour

returns a suitable response when no team exists or a valid team response

### Steps to reproduce

```bash
1. run the test `test_referee_submit_signed_referral_code` from branch `0083-RFPR-025`
2. Query `ListTeams` API
3. Observe the error
```

### Software version

dev branch

### Failing test

test_referee_submit_signed_referral_code

### Jenkins run

https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/11526/testReport/junit/tests.teams/test_team_administration/test_referee_submit_signed_referral_code/

### Configuration used

_No response_

### Relevant log output

_No response_

guoguojin:
Closes #9472 

The ListTeams API leverages the GetTeam API call when filtering for a specific team ID or party ID as there should only be 1 result. When calling with an ID that doesn't exist, the GetTeam API correct returns a not found error, but the ListTeams API did not take this into account and should actually return an empty slice instead of the error.

### DIFF
--- a/datanode/api/trading_data_v2.go
+++ b/datanode/api/trading_data_v2.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/georgysavva/scany/pgxscan"
+
 	"code.vegaprotocol.io/vega/datanode/sqlstore"
 
 	"golang.org/x/sync/errgroup"
@@ -4227,10 +4229,22 @@ func (t *TradingDataServiceV2) ListTeams(ctx context.Context, req *v2.ListTeamsR
 func listTeam(ctx context.Context, teamsService *service.Teams, teamID entities.TeamID, partyID entities.PartyID) (*v2.ListTeamsResponse, error) {
 	team, err := teamsService.GetTeam(ctx, teamID, partyID)
 	if err != nil {
-		return nil, formatE(ErrTeamNotFound, err)
+		if pgxscan.NotFound(err) {
+			return &v2.ListTeamsResponse{
+				Teams: &v2.TeamConnection{
+					Edges: []*v2.TeamEdge{},
+				},
+			}, nil
+		}
+
+		return nil, formatE(ErrListTeams, err)
 	}
 
-	edges, err := makeEdges[*v2.TeamEdge]([]entities.Team{team})
+	if team == nil {
+		return nil, nil
+	}
+
+	edges, err := makeEdges[*v2.TeamEdge]([]entities.Team{*team})
 	if err != nil {
 		return nil, formatE(err)
 	}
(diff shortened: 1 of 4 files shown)


datanode/sqlstore/teams.go
```diff
@@ -121,13 +121,13 @@ func (t *Teams) RefereeSwitchedTeam(ctx context.Context, referee *entities.Refer
 	return err
 }
 
-func (t *Teams) GetTeam(ctx context.Context, teamID entities.TeamID, partyID entities.PartyID) (entities.Team, error) {
+func (t *Teams) GetTeam(ctx context.Context, teamID entities.TeamID, partyID entities.PartyID) (*entities.Team, error) {
 	defer metrics.StartSQLQuery("Teams", "GetTeam")()
 
 	var team entities.Team
 
 	if teamID == "" && partyID == "" {
```
wwestgarth, line 129: Does this useful user-input error ever make its way out atm? I think current it'll end up being hitting the `return nil, formatE(ErrTeamNotFound, err)`. We should probably validate it earlier in `func (t *TradingDataServiceV2) ListTeams()` itself.

guoguojin: So the GetTeam method on the service was because originally I had intended a GetTeam API for the gRPC, but later it just made more sense to add the Filter to the ListTeams API and use this method on the service to retrieve the team if the team ID or party ID filter is provided.

The ListTeams gRPC API can accept an empty Team ID and Party ID, in which case it will list all teams, so it doesn't make sense to have this check in the gRPC method itself.

wwestgarth (dismissed):

datanode/api/trading_data_v2.go
```diff
@@ -4227,10 +4229,22 @@ func (t *TradingDataServiceV2) ListTeams(ctx context.Context, req *v2.ListTeamsR
...
 	if err != nil {
+		if pgxscan.NotFound(err) {
+			return &v2.ListTeamsResponse{
+				Teams: &v2.TeamConnection{
+					Edges: []*v2.TeamEdge{},
+				},
+			}, nil
+		}
+
 		return nil, formatE(ErrTeamNotFound, err)
```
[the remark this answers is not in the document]
guoguojin: Yes, you're right, I'll update that, thanks.

jeremyletang (approved):

PR #9500, opened 2023-09-20
fix: improve error response for CSV exports

### LINKED ISSUE #9074: [Bug]: Internal error in CVS export should be a User input error
### Problem encountered

The [CSV export](https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-export-network-history) data node API requires the user to supply to/from blocks that lie on history segment boundaries.

If you don't you get an error like this:
```
{
  "error": {
    "code": 13,
    "message": "Internal error",
    "details": [
      {
        "code": 10000,
        "message": "failed to get contiguous history segments",
        "inner": "no contiguous segment range found with height 10817000 to 10817445"
      }
    ]
  }
}
```

There are two issues here. The first is that it is not an internal error, and we should instead return some other more helpful error code impling the input args are wrong.

Secondly this error text should be reworded to sound less surprised that it cannot be found i.e:
```
"no contiguous segment range found with height 10817000 to 10817445"
->
"heights 10817000 to 10817445 do not lie within a continuous segment range"
```
maybe it could also say the next nearest boundaries to the ones supplied if we are feeling fancy?

We may also want to give the doc strings of the API another look given it has caused some confusion:
https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1692096019292219

### Observed behaviour

Internal error for incorrect input data.

### Expected behaviour

Invalid args error for a user-input error.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

83406455de113020c12696a9ab459c6a4911b306

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #9074

### DIFF
--- a/datanode/api/errors.go
+++ b/datanode/api/errors.go
@@ -214,7 +214,7 @@ var (
 	ErrCopyHistorySegmentToFile            = errors.New("failed to copy history segment to file")
 	ErrGetIpfsAddress                      = errors.New("failed to get node's ipfs address")
 	ErrNetworkHistoryNoTableName           = errors.New("no table name for network history supplied")
-	ErrNetworkHistoryGetContiguousSegments = newInvalidArgumentError("failed to get contiguous history segments")
+	ErrNetworkHistoryGetContiguousSegments = newInvalidArgumentError("could not to get contiguous history segments")
 	ErrNetworkHistoryOpeningSegment        = errors.New("failed to open network history segment file")
 	ErrNetworkHistoryExtractingSegment     = errors.New("failed to extract data from network history segment file")
 	ErrNetworkHistoryCreatingZipFile       = errors.New("failed to create zip file writer for network history segment")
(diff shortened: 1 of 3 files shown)


datanode/api/errors.go
```diff
@@ -214,7 +214,7 @@ var (
 	ErrCopyHistorySegmentToFile            = errors.New("failed to copy history segment to file")
 	ErrGetIpfsAddress                      = errors.New("failed to get node's ipfs address")
 	ErrNetworkHistoryNoTableName           = errors.New("no table name for network history supplied")
-	ErrNetworkHistoryGetContiguousSegments = errors.New("failed to get contiguous history segments")
+	ErrNetworkHistoryGetContiguousSegments = newInvalidArgumentError("failed to get contiguous history segments")
```
wwestgarth, line 217: Minor: can we just change the text from `failed` to `could not` just so that it sounds less like its core's fault?

jeremyletang (approved):

PR #9506, opened 2023-09-20
fix: verify eth time on submitted chain events

### LINKED ISSUE #9344: [Bug]: Missing check on `eth-block-time` in `ethverifier` allows manipulation of funding-payments
### Problem encountered

As a malicious validator I can manually send in a `ContractCall` chain-event with an ethereum block time that does not correspond to the given ethereum block. This allows me to submit false data-points to a perpetual market and pull the external TWAP in a direction that suits me.

For example, if I know of a Ethereum block height `h` which provides a data-point with price `p` I can submit a lot  data-points with false times for height `h` within the funding period such that the TWAP calculations at the end of the funding period will be close to `p`.

This is because the `eth-block-time` is included in the `ContractCall`'s hash but is not verified in the `ethverifier`.

### Observed behaviour

Contract call events with a false time stamp are accepted and propagated to the product.

### Expected behaviour

Contract call events with a false time stamp are verified and rejected.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

pscott31:
closes #9344 

Very well spotted @wwestgarth that was quite a potentially quite nasty one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -197,6 +197,7 @@
 - [9449](https://github.com/vegaprotocol/vega/issues/9449) - if expiration is empty, never expire a discount/reward program
 - [9263](https://github.com/vegaprotocol/vega/issues/9263) - save dispatch strategy details in the database and allow for its retrieval.
 - [9374](https://github.com/vegaprotocol/vega/issues/9374) - `ListGovernanceData` returns an error instead of results.
+- [9344](https://github.com/vegaprotocol/vega/issues/9344) - Verify `EthTime` in submitted `ChainEvent` transactions matches the chain.
 - [9461](https://github.com/vegaprotocol/vega/issues/9461) - Do not make SLA related transfers for 0 amount.
 - [9462](https://github.com/vegaprotocol/vega/issues/9462) - Add missing proposal error `enum` values to the database.
 - [9466](https://github.com/vegaprotocol/vega/issues/9466) - `ListReferralSets` returns error when there are no stats available.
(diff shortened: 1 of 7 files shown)


EVODelavega (dismissed):

EVODelavega:
LGTM, just one thought I had considering why we want to check eth time here: it might be useful to memoise this data (possibly clearing it once per N epochs or something, or using an array and tracking the most recent N blocks or something like that, considering we're trying to prevent using the same price/height data with different time values

pscott31:
> LGTM, just one thought I had considering why we want to check eth time here: it might be useful to memoise this data (possibly clearing it once per N epochs or something, or using an array and tracking the most recent N blocks or something like that, considering we're trying to prevent using the same price/height data with different time values

Ha thanks - just realised I had done that but not pushed.

```
(system-tests-py3.11) ➜  vega git:(verify_eth_time) ✗ gpf
To https://github.com/vegaprotocol/vega.git
 ! [rejected]            verify_eth_time -> verify_eth_time (remote ref updated since checkout)
error: failed to push some refs to 'https://github.com/vegaprotocol/vega.git'
hint: Updates were rejected because the tip of the remote-tracking
```

Yeah, memoising  might be a good idea, but we don't want something growing without bound I can add a LRU cache in with say the last 50 attempts?

EVODelavega (approved):

PR #9512, opened 2023-09-20
Allow hysteresis period to be 0

### LINKED ISSUE #9426: Allow for PerformanceHysteresisEpochs to be set to 0
Allow for PerformanceHysteresisEpochs to be set to `0` so that only the performance of the LP in the current (just ended) epoch matters when fees are divvied up.

karlem:
Close #9426


ze97286:
would the liquidity engine still work with it? like the slice ring etc.?

karlem:
@ze97286 yes. Look at this test: https://github.com/vegaprotocol/vega/blob/5bdc1aee3cfcbc38df0f35eac5b65971c2af3293/core/liquidity/v2/sla_test.go#L191

ze97286 (approved):

PR #9515, opened 2023-09-21
feat: add trusted proxy support to XFF check

### LINKED ISSUE #8979: [GLS-064] DoS: Rate limits can be circumvented by providing a fictitious IP address in the X-Forwarded-For header.
During a recent pen test audit it has been noted that it may be possible to bypass the data node rate limiting:

Rate-limiting can be circumvented by utilising the X-Forwarded-For header containing a fictitious IP address

Example:
`wfuzz -v -w wordlist/stage_routes.txt -w wordlist/loop_stage.txt -H "X-Forwarded-For: $randomIPaddress" -X GET  http://n05.stagnet1.vega.xyz:3008/api/v2/FUZZ?x=FUZ2Z`

Adding a customised X-Forwarded-For header to the request makes it possible to utilise a new token bucket on every request, thereby circumventing rate limiting.

This needs to be investigated to ensure that a DDOS attack cannot be taken on data nodes

EVODelavega:
Closes #8979

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,7 @@
 - [8764](https://github.com/vegaprotocol/vega/issues/8764) - Include funding payment in margin and liquidation price estimates for `PERPS`.
 - [9519](https://github.com/vegaprotocol/vega/issues/9519) - Fix `oracle_specs` data in the `database` that was inadvertently removed during an earlier database migration
 - [9478](https://github.com/vegaprotocol/vega/issues/8764) - Add SLA statistics to market data and liquidity provision APIs.
+- [8979](https://github.com/vegaprotocol/vega/issues/8979) - Add trusted proxy config and verification for `XFF` header.
 
 ### 🐛 Fixes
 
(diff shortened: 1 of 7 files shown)


pscott31 (dismissed):

pscott31:
Do you think this warrants some sort of 'breaking change' warning because I think this will break everyone who is using a proxy until they add it to the trusted proxy list?

EVODelavega:
@pscott31 It wouldn't cause any and all requests to fail, as the default config just bypasses the forwarded for stuff), but it probably should be communicated clearly. Just going ask what makes the most sense according to @gordsport @jeremyletang @candida-d here

jeremyletang:
> @pscott31 It wouldn't cause any and all requests to fail, as the default config just bypasses the forwarded for stuff), but it probably should be communicated clearly. Just going ask what makes the most sense according to @gordsport @jeremyletang @candida-d here

I think a breaking change entry makes sense, at the end of the day, all request coming from a proxy which has not been setup as the trusted proxy will be considered as coming from a single user and apply rate limiting to it.

We should also probably do what ops has suggested and set 127.0.0.1 in the default trusted proxy list.

EVODelavega:
```
{
"system-tests": "8979-trusted-proxies"
}
```

jeremyletang (approved):

PR #9522, opened 2023-09-22
fix/remove market.liquidity.targetstake.triggering.ratio

### LINKED ISSUE #9484: Remove network param market.liquidity.targetstake.triggering.ratio
This is no longer referred to in specs and no longer provides a default for market proposal trigger ratio (which is now mandatory between 0 and 1)

See https://github.com/vegaprotocol/specs/pull/1964

davidsiska-vega:
Closes https://github.com/vegaprotocol/vega/issues/9484


core/governance/market.go
```diff
@@ -232,20 +232,19 @@ func buildMarketFromProposal(
 		tsTimeWindow, _ := netp.GetDuration(netparams.MarketTargetStakeTimeWindow)
 		tsScalingFactor, _ := netp.GetDecimal(netparams.MarketTargetStakeScalingFactor)
 		// get triggering ratio
-		triggeringRatio, _ := netp.GetDecimal(netparams.MarketLiquidityTargetStakeTriggeringRatio)
-
-		params := &types.TargetStakeParameters{
+		lmParams := &proto.LiquidityMonitoringParameters{}
```
witgaw, line 235: This seems redundant, instantiating an empty struct will result in default values at best or errors when trying to parse decimals form empty strings (not sure what the behaviour is).

I'd remove this entire section, these aren't marked as default parameters, so I believe the proto validation will assure these are populated.

core/netparams/keys.go
```diff
@@ -34,7 +34,6 @@ const (
 	MarketSuccessorLaunchWindow              = "market.liquidity.successorLaunchWindowLength"
 
 	MarketLiquidityProvisionShapesMaxSize            = "market.liquidityProvision.shapes.maxSize"
-	MarketLiquidityTargetStakeTriggeringRatio        = "market.liquidity.targetstake.triggering.ratio"
```
wwestgarth, line 37: Instead of removing it entirely, which counts as a breaking change and means we'll have to go around and remove the nework parameter from all the system tests, and all the example configs in the vegacapsule repo etc etc, can you instead just add the parameter to the deprecated list?
```
var Deprecated = map[string]struct{}{
	StakingAndDelegationRewardPayoutFraction:    {},
	StakingAndDelegationRewardPayoutDelay:       {},
	StakingAndDelegationRewardMaxPayoutPerEpoch: {},
}
```

this will mean that core won't add it to any newly generated genesis files, you can't propose to change the value, and then we can fully remove it peacefully after the release?

davidsiska-vega: Thank you, I'll look into doing that.

wwestgarth (dismissed):

EVODelavega:
Any specific reason for the repeated `lqm-params` table in the features? If so, it would perhaps be worth while adding that to the defaults

EVODelavega:
@vegaprotocol/qa Can we get a nightly run against this change? We want to make sure this isn't a breaking change WRT snapshots

EVODelavega (changes requested):
Not so much a change request, just want to prevent accidental merging until we've confirmed this doesn't cause issues with snapshots

EVODelavega (approved):

Jiajia-Cui (approved):
LGTM

PR #9568, opened 2023-09-26
markets cancel if the opening auction exceeds the max auction

### LINKED ISSUE #8421: A market that should get cancelled for sitting in an opening auction too long
A bit of spec that got ommitted:
```
Enactment date is passed by more than the maximum opening auction extension duration Network Parameter → Cancelled (the market ceases to exist, auction orders are cancelled, and no uncrossing occurs)
```
https://github.com/vegaprotocol/specs/blob/master/protocol/0043-MKTL-market_lifecycle.md#pending 

See also https://github.com/vegaprotocol/specs/issues/1728 for an incoming spec AC.

EVODelavega:
Closes #8421

### DIFF
--- a/core/execution/future/auction.go
+++ b/core/execution/future/auction.go
@@ -47,6 +47,8 @@ func (m *Market) checkAuction(ctx context.Context, now time.Time, idgen common.I
 		err    error
 	)
 
+	checkExceeded := m.mkt.State == types.MarketStatePending
+
 	// as soon as we have an indicative uncrossing price in opening auction it needs to be passed into the price monitoring engine so statevar calculation can start
 	isOpening := m.as.IsOpeningAuction()
 	if isOpening && !m.pMonitor.Initialised() {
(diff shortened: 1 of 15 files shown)


core/monitor/auction.go
```diff
@@ -30,6 +30,7 @@ type AuctionState struct {
 	m                  *types.Market           // keep market definition handy, useful to end auctions when default is FBA
 	extension          *types.AuctionTrigger   // Set if the current auction was extended, reset after the event was created
 	extensionEventSent bool
+	maxDuration        *time.Duration
```
ze97286, line 33: maybe I've missed it but I don't see how this is handled in the snapshot, i.e. suppose this has been updated, a node loading from snapshot will not see this value.

EVODelavega: I don't think we have to do anything in the snapshots. When we restore the markets themselves, we propagate the initial network parameter, which will allow us to check whether or not the opening auction has gone on for too long. There's no real state being added here

EVODelavega:
Full run: https://jenkins.vega.rocks/job/common/job/system-tests/3019/

core/execution/future/auction.go
```diff
@@ -61,6 +63,23 @@ func (m *Market) checkAuction(ctx context.Context, now time.Time, idgen common.I
 	}
 
 	if endTS := m.as.ExpiresAt(); endTS == nil || !endTS.Before(now) {
+		if isOpening && checkExceeded && m.as.ExceededMaxOpening(now) {
+			// cancel the market, exceeded opening auction
+			m.tradableInstrument.Instrument.Product.UnsubscribeTradingTerminated(ctx)
```
ze97286, line 68: don't we already have a close() that does all of these things correctly?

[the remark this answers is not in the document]
EVODelavega: settlement is unsubscribed in the call that actually cancels the market (does some other things like clearing the LPs, too)

EVODelavega: You're right, I didn't use it because it takes `settlementDataInAsset *num.Uint` as an argument, but that isn't used if the market is being cancelled.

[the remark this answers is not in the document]
EVODelavega: Same as comment below: good point, removed messy code-duplication, replaced with a single call to `terminateTrading` (had to split that up into 2 functions because `checkAuction` already acquires a mutex lock)

EVODelavega:
Started nightly run again: https://jenkins.vega.rocks/job/common/job/system-tests/3084/

jeremyletang (approved):

PR #9569, opened 2023-09-26
chore: added feature tests for return volatility metric and reward

### LINKED ISSUE #9559: [Feature]: Feature tests for return volatility metric and rewards
Cover ACs:
0056-REWA-088
0056-REWA-089
0056-REWA-090

ze97286:
Closes #9559 
close https://github.com/vegaprotocol/OctoberACs/issues/353
close https://github.com/vegaprotocol/OctoberACs/issues/354
close https://github.com/vegaprotocol/OctoberACs/issues/355

### DIFF
--- a/core/execution/common/market_activity_tracker.go
+++ b/core/execution/common/market_activity_tracker.go
@@ -706,7 +706,7 @@ func (mat *MarketActivityTracker) calculateMetricForParty(asset, party string, m
 	case vega.DispatchMetric_DISPATCH_METRIC_RETURN_VOLATILITY:
 		filteredReturns := []num.Decimal{}
 		for _, d := range returns {
-			if !d.IsZero() {
+			if d.IsPositive() {
 				filteredReturns = append(filteredReturns, d)
 			}
 		}
@@ -920,11 +920,7 @@ func (mt *marketTracker) getReturns(party string, windowSize int) ([]num.Decimal
 			returns = append(returns, num.DecimalZero())
 		}
 		epochData := mt.epochPartyM2M[len(mt.epochPartyM2M)-i-1]
-		if v, ok := epochData[party]; !ok {
-			returns = append(returns, num.DecimalZero())
-		} else {
-			returns = append(returns, num.MaxD(num.DecimalZero(), v))
-		}
+		returns = append(returns, epochData[party])
 	}
 	return returns, true
 }
--- a/core/integration/features/rewards/return_volatility_rewards.feature
+++ b/core/integration/features/rewards/return_volatility_rewards.feature
@@ -1,0 +1,405 @@
+Feature: Return volatility rewards
+
+  Background:
+    Given the following network parameters are set:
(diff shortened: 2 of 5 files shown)


jeremyletang (dismissed):

core/execution/common/market_activity_tracker.go
```diff
@@ -923,7 +923,7 @@ func (mt *marketTracker) getReturns(party string, windowSize int) ([]num.Decimal
 		if v, ok := epochData[party]; !ok {
 			returns = append(returns, num.DecimalZero())
 		} else {
-			returns = append(returns, num.MaxD(num.DecimalZero(), v))
+			returns = append(returns, v)
```
EVODelavega, line 926: nit-pick, I think we don't even need this `if-else` anymore, if `ok == false`, v should be `num.Decimal{}`, which is a nil/zero value

core/integration/features/rewards/return_volatility_rewards.feature
```diff
@@ -0,0 +1,405 @@
...
+    # setup recurring transfer to the reward account - this will start at the end of this epoch
+    Given the parties submit the following recurring transfers:
+      | id | from                                                             | from_account_type    | to                                                               | to_account_type                       | asset | amount | start_epoch | end_epoch | factor | metric                            | metric_asset | markets | lock_period | window_length | distribution_strategy | entity_scope | individual_scope | staking_requirement | notional_requirement |
+      | 1  | a3c024b4e23230c89884a54a813b1ecb4cb0f827a38641c66eeca466da6b2ddf | ACCOUNT_TYPE_GENERAL | 0000000000000000000000000000000000000000000000000000000000000000 | ACCOUNT_TYPE_REWARD_RETURN_VOLATILITY | VEGA  | 10000  | 1           |           | 1      | DISPATCH_METRIC_RETURN_VOLATILITY | ETH          |         | 2           | 2             | PRO_RATA              | INDIVIDUALS  | ALL              | 0                   | 10000                |
+
+    Then the network moves ahead "1" epochs
+
+    Given time is updated to "2023-09-23T00:00:30Z"
```
EVODelavega, line 180: Is there a reason why we're manually setting time to a specific value here? It's preferable to avoid this...

ze97286: yes there's a very good reason we do. we want position to be recorded at this time, which is with respect to the epoch start time the mid point of the epoch.

core/execution/common/market_activity_tracker.go
ze97286: good point. done.

core/integration/features/rewards/return_volatility_rewards.feature
EVODelavega: makes sense, alrighty

EVODelavega (dismissed):

EVODelavega (approved):

PR #9572, opened 2023-09-26
fix: ledger export end date + performance

### LINKED ISSUE #9570: [Bug]: Ledger entry export doesn't work if `dateRange.End` is specified
### Problem encountered

broken

### Observed behaviour

```
"https://api.n00.testnet.vega.rocks/api/v2/ledgerentry/export?partyId=099c59484e4a0db12d06dc52cb57f4c0e0b861a90d70d7afdfe6cd8e88d94612&dateRange.startTimestamp=1695726164455000000&dateRange.endTimestamp=1695738834409000000"
```

returns nothing

### Expected behaviour

should return stuff

### Steps to reproduce

```bash
as above
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #9571: [Bug]: Ledger entry CSV export slow even when looking at narrow time window
### Problem encountered

Even querying between in a short time window the query is very slow, because the table is partitioned on the synthetic 'ledger_entry_time' but we are filtering on vega_time so we don't get to make use of any of timescale's wonderfulness.

I actually don't think we need `ledger_entry_time`, which was I think added so we had a unique PK for ledger entries; but I am not sure we actually need one at all.

### Observed behaviour

It is slow

### Expected behaviour

It should be faster..

pscott31:
closes #9570  

close #9571

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,8 @@
 - [9530](https://github.com/vegaprotocol/vega/issues/9530) - Referral program end timestamp not correctly displaying in `GraphQL API`.
 - [9532](https://github.com/vegaprotocol/vega/issues/9532) - Data node crashes if referral program starts and ends in the same block.
 - [9540](https://github.com/vegaprotocol/vega/issues/9540) - Proposals connection errors for `UpdateReferralProgram`.
+- [9570](https://github.com/vegaprotocol/vega/issues/9570) - Ledger entry export doesn't work if `dateRange.End` is specified
+- [9571](https://github.com/vegaprotocol/vega/issues/9571) - Ledger entry `CSV` export slow even when looking at narrow time window
 - [8934](https://github.com/vegaprotocol/vega/issues/8934) - Check for special characters in headers.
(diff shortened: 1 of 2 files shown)


jeremyletang (dismissed):

CHANGELOG.md
```diff
@@ -226,6 +226,8 @@
 - [9530](https://github.com/vegaprotocol/vega/issues/9530) - Referral program end timestamp not correctly displaying in `GraphQL API`.
 - [9532](https://github.com/vegaprotocol/vega/issues/9532) - Data node crashes if referral program starts and ends in the same block.
 - [9540](https://github.com/vegaprotocol/vega/issues/9540) - Proposals connection errors for `UpdateReferralProgram`.
+- [9570](https://github.com/vegaprotocol/vega/issues/9570) - Ledger entry export doesn't work if dateRange.End is specified
```
gordsport, line 229:
```suggestion
- [9570](https://github.com/vegaprotocol/vega/issues/9570) - Ledger entry export doesn't work if `dateRange.End` is specified
```

jeremyletang (approved):

PR #9604, opened 2023-09-28
feat: referral spam protection

### LINKED ISSUE #9578: [Feature]:  implement spam protection for the referrals feature
# Feature Overview

**In order to** ensure the referrals feature is protected again spam attacks 
**We will** implement spam protection
**So to** keep the network safe to use

## Specs
- [Referrals spam specs PR](https://github.com/vegaprotocol/specs/pull/1978)

ze97286:
Closes #9578


candida-d (dismissed):

cdummett (dismissed):

EVODelavega (dismissed):

core/collateral/snapshot.go
```diff
@@ -106,6 +107,8 @@ func (e *Engine) restoreAccounts(ctx context.Context, accs *types.CollateralAcco
 	var err error
 	e.state.serialisedAccounts, err = proto.Marshal(p.IntoProto())
 	e.getOrCreateNetTreasuryAndGlobalInsForAssets(ctx, assets)
+	e.nextBalancesSnapshot = accs.NextBalanceSnapshot
+	e.state.updateBalanceSnapshotTime(e.nextBalancesSnapshot)
```
wwestgarth, line 111: Do we not need to call `snapshotBalances()` or something here to populate the cache after restore?

ze97286: we could. I don't think it's critical. As it is, a joining node will have a maximum of 5 minutes (or whatever the param will be on) during which it will return 0 for everything. If we do call here, it will return different results than other nodes. If the interval is longer though, it might be better to call it anyway. so in short, idk.

wwestgarth: Surely it also applies after a protocol upgrade though too, where all nodes will have cached balances of 0 for everyone for a bit of time.

But you're right, its not that much of a problem.

jeremyletang (approved):

PR #9614, opened 2023-09-29
fix: fix trigger times in GQL

AnExsomnis:


candida-d (dismissed):

datanode/gateway/graphql/oracles_resolver.go
```diff
@@ -76,9 +77,9 @@ func resolveTrigger(obj any) (trigger TriggerKind) {
 		switch trig := obj.(type) {
 		case *vegapb.EthCallTrigger_TimeTrigger:
 			if trig.TimeTrigger != nil {
-				init := int64(trig.TimeTrigger.GetInitial())
+				init := strconv.FormatUint(trig.TimeTrigger.GetInitial(), 10)
```
guoguojin, line 80: RFC3339 encoded should be:

```suggestion
				init := time.Unix(trig.TimeTrigger.GetInitial(), 0).Format(time.RFC3339)
```

datanode/gateway/graphql/oracles_resolver.go
```diff
@@ -76,9 +77,9 @@ func resolveTrigger(obj any) (trigger TriggerKind) {
...
 				every := int(trig.TimeTrigger.GetEvery())
-				until := int64(trig.TimeTrigger.GetUntil())
+				until := strconv.FormatUint(trig.TimeTrigger.GetUntil(), 10)
```
guoguojin, line 82: As above

guoguojin (dismissed):

wwestgarth:
All the timestamps across all of of GQL schema are `Timestamp` type, I'm not sure why this one needs to be different?

AnExsomnis:
> All the timestamps across all of of GQL schema are `Timestamp` type, I'm not sure why this one needs to be different? I think probably all we need to do here is to scale it because `initial` and `until` are in seconds but the timestamp resolver will expect them in nanoseconds.

Actually there are timestamps as trings - like CandleConnection, ERC20MultiSigSignerAddedBundle and ERC20MultiSigSignerRemovedBundle.

We were giving Timestamp fields in the GQL definition time in Unix seconds, but Timestamp data type is expecting Unixnanos and if they are being given as seconds, then you get the badly formatted time.
The idea here is change the GQL schema def to String, and then format the Unix seconds correctly in the GQL resolver.

AnExsomnis:
> > All the timestamps across all of of GQL schema are `Timestamp` type, I'm not sure why this one needs to be different? I think probably all we need to do here is to scale it because `initial` and `until` are in seconds but the timestamp resolver will expect them in nanoseconds.
> 
> Actually there are timestamps as strings - like CandleConnection, ERC20MultiSigSignerAddedBundle and ERC20MultiSigSignerRemovedBundle.
> 
> We were giving Timestamp fields in the GQL definition time in Unix seconds, but Timestamp data type is expecting Unixnanos and if they are being given as seconds, then you get the badly formatted time. The idea here is change the GQL schema def to String, and then format the Unix seconds correctly in the GQL resolver.

Following other conversation with William, i am going to check the Proposals timestamps before proceeding, and see if we really to change this.

wwestgarth (dismissed):

wwestgarth (approved):

PR #9621, opened 2023-09-29
Orders divergence map

ettec:
Experiment to gauge the impact of using an a generics based ordered map instead of the built in map to try and reduce the number of occurrences of non-deterministic event ordering.   This may not solve all ordering issues, but maybe its the braces of a belt and braces approach, the belt being the current sorts peppered in the code.

Syntactically it works well, theres is little to no inconvenience (arguagly iteration is a bit more verbose).

Performance wise (see the `temp_ordered_map_benchmarks_test`), it performs surprisingly well,  iteration, of all things seems to be quicker(!) - unless there is some issue with the benchmark test, I can't see it.

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	orderedmap "github.com/wk8/go-ordered-map/v2"
 	"sort"
 	"sync"
 	"time"
@@ -105,7 +106,7 @@ type Market struct {
 	closed               bool
 	finalFeesDistributed bool
 
-	parties map[string]struct{}
+	parties *orderedmap.OrderedMap[string, struct{}]
 
 	pMonitor common.PriceMonitor
 	lMonitor LiquidityMonitor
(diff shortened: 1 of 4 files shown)


EVODelavega:
This really isn't worth it IMO. Sorting a slice of keys whenever we need to deterministically iterate over a map may be slightly, or even significantly less efficient, but what matters is which operations we do more of: read/access the map, deleting entries. Using a simple map for that is O(1). Deterministic iterations are 3 lines of code. It's fine as it is, let's just get on with the AC's rather than debating the pro's and cons of nice-to-haves like this

ettec:
> This really isn't worth it IMO. Sorting a slice of keys whenever we need to deterministically iterate over a map may be slightly, or even significantly less efficient, but what matters is which operations we do more of: read/access the map, deleting entries. Using a simple map for that is O(1). Deterministic iterations are 3 lines of code. It's fine as it is, let's just get on with the AC's rather than debating the pro's and cons of nice-to-haves like this

So the point of the perf tests was not to see if the OrderedMap is more performant, just to see if it is in the same ball park as a map, which it is, ~15% more for get and puts, iteration I assumed would be slower, so was initially surprised to see it was quicker, but given it internally maintains a linked list, that's not so unexpected on reflection.  But, the point was to gauge the impact both in terms of performance and of the code changes required to use such a map that which will give deterministic iteration to avoid all the event ordering bugs, given the amount of time wasted on hunting down these bugs, and the random nature with which they occur, I'm not convinced its a 'nice to have' - but can certainly wait till post next release.

PR #9632, opened 2023-10-02
Rework referral set stats

### LINKED ISSUE #9258: [Feature]: Party reward & referral factors
# Feature Overview

**In order to** support front end
**We will** build a NEW party reward & referral factors API 
**So that** front end can build:
- The team member details on page 6 of the wireframes
- The profile information on page 11 of the wireframes
- The console deal ticket and fills table discounts applied 

## Specs
- 0083, 0084, 0085, 0086

## Tasks
- [x] Create new API to expose game metrics & discount factors for parties
- [x] Publish at the end of each epoch 

API to provide the following data:
- [x] Team ID
- [x] Number of epochs in team
- [x] Volume_discount_factor 
- [ ] Referral_discount_factor 
- [x] Referral_reward_factor (Not sure this is needed, only applies to referrer, and it would be on the referral set API right?)
- [x] Notional time weighted position (per asset)
- [x] Trade volume (party_epoch_notional_taker_volume)

Filtering requirements:
  - [x] Party ID
  - [x] Team ID

ValentinTrinque:
Close #9258


asiaznik (dismissed):

datanode/gateway/graphql/referral_set_resolver.go
```diff
@@ -22,33 +22,12 @@ func (r *referralSetRefereeResolver) RefereeID(ctx context.Context, obj *v2.Refe
...
-	return res.Stats, nil
-}
-
 type referralSetStatsResolver VegaResolverRoot
 
-func (r *referralSetStatsResolver) AtEpoch(_ context.Context, obj *v2.ReferralSetStats) (*int, error) {
-	if obj == nil {
-		return nil, nil
+func (r *referralSetStatsResolver) AtEpoch(_ context.Context, obj *v2.ReferralSetStats) (int, error) {
+	if obj.AtEpoch > math.MaxInt {
```
EVODelavega, line 28: I'm not really sure why we'd take a `uint64` and restrict it to the max value of an int here... why is a `uint64` not possible, or why not use a supported/larger type that can hold all _valid_ values?

datanode/sqlstore/referral_sets.go
```diff
@@ -117,73 +122,55 @@ func (rs *ReferralSets) AddReferralSetStats(ctx context.Context, stats *entities
...
-		stats []*entities.ReferralSetStats
-		query string
-		args  []interface{}
+		query    string
+		args     []interface{}
+		pageInfo entities.PageInfo
+	)
+
+	query = fmt.Sprintf(
+		`SELECT at_epoch,
```
EVODelavega, line 134: This looks very much like the view we dropped in the migrations, why should we not keep the view? (referral_set_stats)

datanode/gateway/graphql/referral_set_resolver.go
ValentinTrinque: AFAIK, GraphQL only support integer out of the box. For ref:
>The GraphQL schema language supports the scalar types of String, Int, Float, Boolean, and ID, so you can use these directly in the schema you pass to buildSchema.
> \- [GraphQL documentation](https://graphql.org/graphql-js/basic-types/)

About _holding all valid values_, I think we only have strings, out of the box. Otherwise, it down to creating a specific GraphQL object with a specific resolver.

datanode/sqlstore/referral_sets.go
ValentinTrinque: Two reasons:
1. There is no need for the view if we can get all from one table.
2. The view was missing some data, and to accommodate the query I reworked, working out with a join was harder than having this one.

datanode/gateway/graphql/referral_set_resolver.go
jeremyletang: yea  this is fine, not uint in gql, only Ints, and in this case it will be good enough.

datanode/entities/referral_set_stats.go
```diff
@@ -46,48 +46,56 @@ type (
...
-func (ref *ReferralSetRefereeStats) ToProto() *v2.ReferralSetStats {
-	stats := eventspb.RefereeStats{
-		PartyId:                  ref.PartyID,
-		DiscountFactor:           ref.DiscountFactor,
-		RewardFactor:             ref.RewardFactor,
-		EpochNotionalTakerVolume: ref.ReferralSetRunningNotionalTakerVolume.String(),
+func (c *ReferralSetStatsCursor) ToString() string {
+	bs, err := json.Marshal(c)
+	if err != nil {
+		panic(fmt.Errorf("could not marshal referral set stats cursor: %v", err))
```
EVODelavega, line 79: Really don't like seeing a `panic` in a `String()` function

datanode/gateway/graphql/resolvers.go
```diff
@@ -1654,6 +1650,26 @@ func (r *myQueryResolver) ReferralSetReferees(ctx context.Context, id, referrer,
...
+	}
+
+	resp, err := r.tradingDataClientV2.GetReferralSetStats(ctx, &v2.GetReferralSetStatsRequest{
+		ReferralSetId: setID,
+		AtEpoch:       epochU64Ptr,
+		Referee:       partyID,
+		Pagination:    pagination,
+	})
+	if err != nil {
+		return &v2.ReferralSetStatsConnection{}, err
```
EVODelavega, line 1667: Why return a pointer to a valid, but uninitialised type instead of `return nil, err`?

ValentinTrinque: A mistake.

jeremyletang (approved):

gordsport (approved):

asiaznik (approved):

jeremyletang (approved):

PR #9639, opened 2023-10-02
Proposal for organisation

ValentinTrinque:


jeremyletang (approved):
I'm gonna approve in place of Justin, but not sure we should merge this PR.

EVODelavega (changes requested):
I must object to merging without a prenup

PR #9649, opened 2023-10-03
feat: emit zero-margin event when clearing margin balance

### LINKED ISSUE #8475: [Feature]: Emit an event with 0 margin levels when party exits the market entirely
When party leaves the market (position becomes 0 and there are no open orders) either as a result of own action or a closeout a final margin event with 0 values should get emitted and ingested by the datanode to assure consistency of any downstream calculations relying on those values.

EVODelavega:
Closes #8475

### DIFF
--- a/core/integration/features/margin/1920-trader-with-no-pos-and-stopped-fok-should-have-0-margin.feature
+++ b/core/integration/features/margin/1920-trader-with-no-pos-and-stopped-fok-should-have-0-margin.feature
@@ -40,9 +40,11 @@ Feature: test for issue 1920
       | party  | market id | side | volume | price | resulting trades | type       | tif     | reference |
       | party1 | ETH/DEC19 | sell | 1      | 1000  | 0                | TYPE_LIMIT | TIF_FOK | ref-1     |
 
+    # If the order isn't filled, the margin levels are calculated, but a zero-event is sent once the order is stopped
     Then the parties should have the following margin levels:
       | party  | market id | maintenance | search | initial | release |
-      | party1 | ETH/DEC19 | 100         | 110    | 120     | 140     |
+      | party1 | ETH/DEC19 | 0           | 0      | 0       | 0       |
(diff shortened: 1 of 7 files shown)


jeremyletang (dismissed):

jeremyletang (dismissed):

core/integration/features/margin/1920-trader-with-no-pos-and-stopped-fok-should-have-0-margin.feature
```diff
@@ -40,9 +40,11 @@ Feature: test for issue 1920
       | party  | market id | side | volume | price | resulting trades | type       | tif     | reference |
       | party1 | ETH/DEC19 | sell | 1      | 1000  | 0                | TYPE_LIMIT | TIF_FOK | ref-1     |
 
+    # If the order isn't filled, the margin levels are calculated, but a zero-event is sent once the order is stopped
     Then the parties should have the following margin levels:
       | party  | market id | maintenance | search | initial | release |
-      | party1 | ETH/DEC19 | 100         | 110    | 120     | 140     |
+      | party1 | ETH/DEC19 | 0           | 0      | 0       | 0       |
+      #| party1 | ETH/DEC19 | 100         | 110    | 120     | 140     |
```
witgaw, line 47: I think the old values can just be deleted

witgaw (approved):

jeremyletang (approved):

core/integration/features/margin/1920-trader-with-no-pos-and-stopped-fok-should-have-0-margin.feature
EVODelavega: They can be, I just kept them here in case I got this comment, because there is an event that'll set the margin levels to these values, it's just that this `parties should have the following margin levels`  step takes the last event, which is 0 margin (as the order was stopped)

PR #9687, opened 2023-10-06
feat: cache and forward referral rewards multiplier and factor multplier

### LINKED ISSUE #9678: Add `RewardsFactorMultiplier` and `RewardsMultiplier` to referees stats
We need to add the RewardsMultiplier to the referees stats sent to the data-node.

This issue targets the ReferralEngine.

This is currently computing on demand in the `(*Engine).RewardsMultiplierForParty()` method.
Since the stats is computed ahead of time is `(*Engine).computeFactorsByReferee()` method, the computation done in the `RewardsMultiplierForParty()` should be moved inside the `computeFactorsByReferee()`, and the `RewardsMultiplierForParty()` while have to look for the data inside the stats (just like in the method `ReferralDiscountFactorForParty()`)

Same for `RewardsFactorMultiplier`

Bonus:
- Method `(*Engine).VolumeDiscountFactorForParty()` is not used anywhere, it should be removed from this engine.

ettec:
closes #9678


ValentinTrinque (changes requested):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3821,6 +3821,10 @@ message ReferralSetStats {
   string reward_factor = 5;
   // Current referee notional taker volume
   string epoch_notional_taker_volume = 6;
+  // Rewards multiplier applied to the party.
+  string rewards_multiplier = 7;
+  // Rewards factor multiplier for the party.
```
candida-d, line 3826: Could you go into more detail about what the difference is between these two?

jeremyletang:
@ettec could you fix the commit message and address @candida-d's comment please?

ValentinTrinque (dismissed):

jeremyletang (approved):

PR #9696, opened 2023-10-09
fix: distribute rewards amongst teams pro-rata by multipliers

cdummett:
Spec [0056](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0056-REWA-rewards_overview.md#0056-REWA-107) states rewards allocated to a team should be distributed pro-rata by each team members reward multiplier.

The parties performance (providing they have a non-zero score) should have no impact on the share of the rewards allocated to their team.

### DIFF
--- a/core/teams/engine.go
+++ b/core/teams/engine.go
@@ -197,6 +197,9 @@ func (e *Engine) GetAllPartiesInTeams(minEpochsInTeam uint64) []string {
 
 func (e *Engine) GetTeamMembers(team string, minEpochsInTeam uint64) []string {
 	t := e.teams[(types.TeamID(team))]
+	if t == nil {
+		return []string{}
+	}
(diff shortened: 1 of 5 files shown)


core/teams/engine.go
```diff
@@ -194,6 +194,9 @@ func (e *Engine) GetAllPartiesInTeams(minEpochsInTeam uint64) []string {
 
 func (e *Engine) GetTeamMembers(team string, minEpochsInTeam uint64) []string {
 	t := e.teams[(types.TeamID(team))]
+	if t == nil {
```
EVODelavega, line 197: assuming `e.teams` is  a map, checking for nil is redundant:

```go
t, ok := e.teams[types.TeamID(team)]
if !ok {
    return nil
}
```

EVODelavega (approved):

PR #9700, opened 2023-10-09
feat: add snapshot version to appstate payload

### LINKED ISSUE #9593: [Feature]: Add to snapshot the core version that created it, and whether it was for a upgrade
# Feature Overview

**In order to** make the migration between snapshot versions more robust
**We will** and additional data to each snapshot
**So that** each engine can make an informed decision on why/how to migrate snapshot data

We are finding ourselves quite often having to write code in particular engines to handle the migration of data. Up until now we've been doing this _usually_ by some logic of the form:
```
if [new field isn't set in snapshot] {
    // do some migration stuff to set it
}
```

This works enough right now, but eventually we will hit a case where this logic may hold a false assumption.

We should add to the app-state payload/snapshot metadata:
- the core version that created the snapshot
- whether the snapshot was taken for the purpose of a protocol upgrade

So that we can have logic in an engine that looks like:
```
meta := ctx.get_snapshot_metadata()
if meta.isUpgrade && meta.version == [known version to migrate from] {
    // do some migration stuff
}
```
This will ensure it is always a one-off migration, and will also mean leaving old migration code hanging around is harmless.

As a bonus: it will also allow us to add code that will let us fail more gracefully is we try to load a snapshot with the wrong version, instead of what could be crazy panics. We could even rejected snapshots offered with the wrong version during statesync. We can do this as a separate ticket though.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
close #9593 

Adds the core version that took the snapshot, and whether it was for a protocol upgrade to the app-state payload in the snapshot. Its not used at the moment, but its certainly useful that its there:

I've kicked off a few choice system tests, and have stared deeply at the logs to confirm full its working as intended

basic snapshot system test that state-syncs:
https://jenkins.vega.rocks/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/21472/pipeline

basic protocol upgrade:
https://jenkins.vega.rocks/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/21481/tests

more involved protocol upgrade:https://jenkins.vega.rocks/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/21486/tests


EVODelavega (dismissed):

core/snapshot/engine.go
```diff
@@ -642,49 +650,31 @@ func (e *Engine) snapshotNow(ctx context.Context, saveAsync bool) ([]byte, DoneC
...
+	e.appState.ChainID = cid
 
-	if appUpdate {
-		if err = e.updateAppState(); err != nil {
-			e.snapshotTreeLock.Unlock()
-			return nil, nil, err
-		}
-		updated = true
-	}
+	// TODO rename saveAsync
```
jeremyletang, line 671: todo as part of this PR?

wwestgarth: whoops, decided against it in the end.

EVODelavega (approved):

candida-d (approved):

PR #9701, opened 2023-10-09
fix: refactor referral engine snapshot

### LINKED ISSUE #9691: [Bug]: Referral engine reward factor snapshot handling is incorrect
### Problem encountered

* Fairground snapshot pipeline is falling out of consensus periodically
* It is observed that the difference is in account balances
* The root cause is the following: when loading referral engine from snapshot we're calling computeFactorsByReferee for the current epoch:
image
* As opposed to in a running network, where we only call it for the current epoch when it ends:
image (1)
* As a result, a node loaded from snapshot would think that a party is eligible for referral reward whereas the running nodes not loaded from snapshot would disagree until the end of the epoch.
* So if during the run there is a trade with a party who's eligible, then the node loaded from snapshot would pay fee rewards to the referrer and the fairground nodes would not, leading to balances discrepancy and falling out of consensus.

### Observed behaviour

Fairground snapshot pipeline sometimes falling out of consensus.

### Expected behaviour

Fairground snapshot pipeline should not be falling out of consensus.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.73.0-preview.11

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
Closes #9691


core/referral/snapshot.go
```diff
@@ -155,56 +165,10 @@ func (e *SnapshottedEngine) serialiseReferralSets() ([]byte, error) {
...
-func (e *SnapshottedEngine) serialiseReferralMisc() ([]byte, error) {
-	payload := &snapshotpb.Payload{
-		Data: &snapshotpb.Payload_ReferralMisc{
-			ReferralMisc: &snapshotpb.ReferralMisc{
-				LastProgramVersion: e.latestProgramVersion,
-				ProgramHasEnded:    e.programHasEnded,
-			},
-		},
-	}
+	println(payload.String())
```
jeremyletang, line 171: debug

ValentinTrinque (dismissed):

jeremyletang (dismissed):

candida-d (dismissed):

jeremyletang (approved):

PR #9721, opened 2023-10-11
feat: add Referral feature test

Jiajia-Cui:
closes: https://github.com/vegaprotocol/OctoberACs/issues/398


Jiajia-Cui:
close as the AC should be covered in system-test rather than feature test

PR #9732, opened 2023-10-11
fix: allow empty team scope value

### LINKED ISSUE #9731: [Bug]: rewards with entity scope TEAMS and an empty team scope apply to all teams
### Problem encountered

the team scope for team rewards is strictly applied. Only teams that are explicitly set are eligible for rewards, whereas the [AC 0056-REWA-103](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0056-REWA-rewards_overview.md#0056-REWA-103) explicitly states that:

> If the entity scope is ENTITY_SCOPE_TEAMS and the teams scope is not-set, then all teams are eligible for rewards

### Observed behaviour

An empty slice for teams translates to no teams being eligible for rewards

### Expected behaviour

All teams should be eligible

### Steps to reproduce

```bash
Covered by integration test in rewards/0056-REWA-103
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #9731

### DIFF
--- a/core/integration/steps/transfers.go
+++ b/core/integration/steps/transfers.go
@@ -213,7 +213,7 @@ func rowToRecurringTransfer(r RowWrapper) *types.RecurringTransfer {
 			if r.HasColumn("teams") {
 				teams = strings.Split(r.MustStr("teams"), ",")
 				if len(teams) == 1 && teams[0] == "" {
-					mkts = []string{}
+					teams = []string{}
 				}
 			}
 			ntop = r.MustStr("ntop")
--- a/core/integration/steps/transfers.go
+++ b/core/integration/steps/transfers.go
@@ -210,7 +210,9 @@ func rowToRecurringTransfer(r RowWrapper) *types.RecurringTransfer {
 		teams := []string{}
 		ntop := ""
 		if entityScope == proto.EntityScope_ENTITY_SCOPE_TEAMS {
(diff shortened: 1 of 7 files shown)


core/integration/steps/transfers.go
```diff
@@ -210,7 +210,9 @@ func rowToRecurringTransfer(r RowWrapper) *types.RecurringTransfer {
 		teams := []string{}
 		ntop := ""
 		if entityScope == proto.EntityScope_ENTITY_SCOPE_TEAMS {
-			teams = strings.Split(r.MustStr("teams"), ",")
+			if r.HasColumn("teams") {
+				teams = strings.Split(r.MustStr("teams"), ",")
+			}
 			if len(teams) == 1 && teams[0] == "" {
 				mkts = []string{}
```
ze97286, line 217: only half related but I think this should be teams = []string{}

EVODelavega: This was already there, and `teams = []string{}` is set a couple of lines above. I'll check if `mkts` is being used (I imagine it is - still, I'll check)

EVODelavega: removed

ze97286 (approved):

PR #9756, opened 2023-10-12
fix: apply funding payment increment correctly

EVODelavega:
Closes https://github.com/vegaprotocol/system-tests/issues/2971

### DIFF
--- a/core/risk/margins_calculation.go
+++ b/core/risk/margins_calculation.go
@@ -82,11 +82,8 @@ func (e *Engine) calculateMargins(m events.Margin, markPrice *num.Uint, rf types
 			CollateralReleaseLevel: num.UintZero(),
 		}
 	}
-	// long funding payment -> requires extra margin, if false, same holds for short
-	longFP := inc.IsNegative()
-	if longFP {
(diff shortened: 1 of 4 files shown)


jeremyletang (dismissed):

core/risk/margins_calculation.go
```diff
@@ -242,9 +239,8 @@ func (e *Engine) calculateMargins(m events.Margin, markPrice *num.Uint, rf types
 	}
 
 	if !inc.IsZero() && !openVolume.IsZero() {
-		// calculate margin increase based on position
-		// incD = max(0, inc * open volume)
-		incD := num.MaxD(num.DecimalZero(), inc.Mul(openVolume))
+		// openVolume and inc are signed, but this is fine, we only apply the positive values
+		incD := inc.Mul(openVolume)
```
wwestgarth, line 243: I don't get why we don't need the `num.MaxD`.

If the funding payment is `-1` and openVolume is `1` then incD will be negative and we'll "add" `-1` to `marginMaintenanceSht`. But really we there should be no change to the margin at all.

EVODelavega: Fixed, added integration test, too

pscott31 (approved):

PR #9758, opened 2023-10-12
feat: add referral set spam stats to core API

### LINKED ISSUE #9755: [Feature]: `GetSpamStatisitcs` needs information about the new referral set policies
# Feature Overview

The `GetSpamStatisitcs` API in core returns all the information about each parties spam status for each transaction type. New spam policies were recently added for the referral set feature but have not been added to this end-point.

So we should add them.

wwestgarth:
closes #9755 

I am also changing the nullchain-spam flag slightly to not enable the PoW engine. This is because the vega-market-sim code is not set up for calculating PoW's when sending in direct transactions, so keeping it disabled will help speed up QA's testing.


jeremyletang (dismissed):

protos/sources/vega/api/v1/core.proto
```diff
@@ -434,22 +434,28 @@ message PoWStatistic {
...
   PoWStatistic pow = 6;
-  // Statistics for multisig signatures issued for the party
+  // Statistics for multisig signatures issued for the party.
   SpamStatistic issue_signatures = 7;
-  // Epoch in which these statistics apply to
+  // Epoch in which these statistics apply to.
   uint64 epoch_seq = 8;
+  // Statistics for create referral set transactions made by the party.
+  SpamStatistic create_referral_set = 9;
+  // Statistics for update referral set transactions made by the party.
```
candida-d, line 455:
```suggestion
  // Statistics for transactions made by the party to update referral sets.
```

protos/sources/vega/api/v1/core.proto
```diff
@@ -434,22 +434,28 @@ message PoWStatistic {
...
+  SpamStatistic update_referral_set = 10;
+  // Statistics for apply referral code transactions made by the party.
```
candida-d, line 457:
```suggestion
  // Statistics for transactions made by the party to apply referral codes.
```

PR #9759, opened 2023-10-12
test: add integration test covering a failing system test

EVODelavega:
Closes https://github.com/vegaprotocol/system-tests/issues/2948

### DIFF
--- a/core/integration/features/auctions/cancel-gfn-auction.feature
+++ b/core/integration/features/auctions/cancel-gfn-auction.feature
@@ -66,10 +66,10 @@ Feature: When moving into auction, all GFN orders are cancelled
 
     # place GFN orders outside of bounds so we can trigger auction without those order uncrossing
     When the parties place the following orders:
-      | party  | market id | side | volume | price | resulting trades | type       | tif     |
-      | party1 | ETH/DEC20 | buy  | 1      | 990   | 0                | TYPE_LIMIT | TIF_GFN |
-      | party2 | ETH/DEC20 | sell | 1      | 1010  | 0                | TYPE_LIMIT | TIF_GFN |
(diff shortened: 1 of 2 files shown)


core/integration/features/auctions/cancel-gfn-auction.feature
```diff
@@ -0,0 +1,104 @@
...
+      | 1000       | TRADING_MODE_CONTINUOUS | 60      | 995       | 1005      | 7434         | 90000000       | 10            |
+      | 1000       | TRADING_MODE_CONTINUOUS | 600     | 978       | 1022      | 7434         | 90000000       | 10            |
+
+    When the parties place the following orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     |
+      | party3 | ETH/DEC20 | buy  | 1      | 1010  | 0                | TYPE_LIMIT | TIF_GTC |
+    Then the trading mode should be "TRADING_MODE_MONITORING_AUCTION" for the market "ETH/DEC20"
+    ## The GFN orders will should be removed from the book (buy at 990, sell at 1010)
+    ## the 2 GTC orders that triggered the price auction, however, remain
+    And the order book should have the following volumes for market "ETH/DEC20":
```
wwestgarth, line 94: This test doesn't match the system-test. 

The system test check's explicitly that the GFN order is cancelled, this integration test checks that _some_ order at price level at `1110` is no longer on the book.

For parity, you also need to check that it was the GFN order that is no longer on the book, and not the GTC one. And also that its no longer on the book because it was cancelled, and not for any other reason.

EVODelavega: Added references to the orders, and verified the order status after entering auction

AnExsomnis (approved):

PR #9767, opened 2023-10-12
Rename endpoints from refferralFeeStats to feesStats

### LINKED ISSUE #9766: Rename endpoints from refferralFeeStats to feesStats
... and anything relevent which do not imply that this endpoint is specific to the referral fees.

ValentinTrinque:
Close #9766


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -799,12 +799,12 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Teams"};
   }
 
-  // Get referral fee statistics
+  // Get fees statistics
```
candida-d, line 802: Is the description for this one still correct? I'd also expect stats about fees to tell me about the fees paid (which isn't mentioned) and discounts, but not necessarily about rewards.

jeremyletang (dismissed):

jeremyletang (approved):

protos/sources/data-node/api/v2/trading_data.proto
candida-d: It seems then it should say “rewards based on fees paid or earned” since it sounds like it won’t show all rewards then, right?

PR #9798, opened 2023-10-16
Fix/9797 default pagination not always set

guoguojin:
Closes #9797

### DIFF
--- a/datanode/entities/pagination.go
+++ b/datanode/entities/pagination.go
@@ -107,7 +107,10 @@ func NewCursorPagination(first *int32, after *string, last *int32, before *strin
 }
 
 func CursorPaginationFromProto(cp *v2.Pagination) (CursorPagination, error) {
-	if cp == nil {
+	if cp == nil || (cp != nil && cp.First == nil && cp.Last == nil) {
+		if cp != nil && cp.NewestFirst != nil {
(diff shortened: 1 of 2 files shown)


datanode/entities/pagination.go
```diff
@@ -107,7 +107,7 @@ func NewCursorPagination(first *int32, after *string, last *int32, before *strin
 }
 
 func CursorPaginationFromProto(cp *v2.Pagination) (CursorPagination, error) {
-	if cp == nil {
+	if cp == nil || (cp != nil && cp.First == nil && cp.After == nil) {
```
jeremyletang, line 110: is that enough? shouldn't we check if all fields are nil instead?
```suggestion
	if cp == nil || (cp != nil && cp.First == nil && cp.After == nil && cp.Last == nil && cp.Before == nil) {
```

guoguojin: The After and Before can be nil, the Last and After specify how many records to return, from the start and end. If After or Before are provided, then those cursors are used to determine where the position of the cursor in the data set.

jeremyletang: hum, alright, I suppose.

EVODelavega (approved):

PR #9809, opened 2023-10-17
feat: add created at to blockexplorer

### LINKED ISSUE #8919: [Feature]: Add timestamp to Transactions returned by getTxList
# Feature Overview

**In order to** address https://github.com/vegaprotocol/frontend-monorepo/issues/4236
**We will** add a single new field to each Transaction result from the Block Explorer API
**So that** block explorer does not have to fetch block data from Tendermint

## Specs
- https://github.com/vegaprotocol/frontend-monorepo/issues/4236 is what we want to achieve

## Tasks
- [x] Ensure internal type has `createdAt` [`TxResultRow`](https://github.com/vegaprotocol/vega/blob/develop/blockexplorer/entities/transaction.go#L31-L40)
- [ ] Verify that `createdAt` makes sense - i.e. it is when the TX was executed, or was in a block¹
- [ ] Ensure that is returned for all Transactions returned by [`ListTransactions`](https://docs.vega.xyz/testnet/api/rest/explorer/block-explorer-list-transactions) and [`GetTransaction`](https://docs.vega.xyz/testnet/api/rest/explorer/block-explorer-get-transaction)

# Notes
- This data will be used by the Block Explorer to show when a transaction 'happened'
- ¹ It is up to the implementer to decide what makes the most sense for this data - whether it's when the TX first hit the mempool of some node, or when it was in a block, or when it was executed. I just want a date.

EVODelavega:
Closes #8919


pscott31 (dismissed):

edd (dismissed):
👍 Thanks

protos/sources/blockexplorer/api/v1/blockexplorer.proto
```diff
@@ -109,4 +109,6 @@ message Transaction {
   // Optional error happening when processing / checking the transaction
   // This should be set if error code is not 0
   optional string error = 10;
+  // Timestamp when the transaction happened.
```
candida-d, line 112: Could you add the timestamp format?

EVODelavega: updated

candida-d (dismissed):

jeremyletang (dismissed):

edd (dismissed):

candida-d (approved):

edd (approved):
Really Can't Fathom how we missed that

jeremyletang (approved):

PR #9810, opened 2023-10-17
fix: show pending LP separately when an liquidity provision is updated

### LINKED ISSUE #9757: SLA API endpoints must report both current and pending amounts and fees
If a party has an active liquidity provision and then it makes an update to that it will now appear as PENDING on the API until the end of the current epoch when that update will be applied.

This is misleading because you can't tell the difference between a new provision which is contributing ZERO liquidity until the end of the epoch and one that is currently providing liquidity but will have some sort of change applied at end of epoch.

Update the SLA API endpoint so that it has both "current" part (amount, committed, fee proposed) and "pending part" (new amount and fee which will be activated at epoch boundary). 

For a party that submitted LP provision for the 1st time the "current" part can return `0` and `0` or `null` and `null` - whatever is more in keeping with the API style.

guoguojin:
Closes #9757 

Please be aware that in order to support the additional pending LP data, has required some breaking changes.


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
   LiquidityProvisionsConnection liquidity_provisions = 1;
 }
 
+// Liquidity provider commitment. For API purposes this can report a pending liquidity provision,
+// the current live provision or both in the event an update has been accepted on the network, but has yet
+// to go live and there is still a
```
candida-d, line 2614: Leaving me in suspense here

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+message LiquidityProvision {
+  // Unique ID for the liquidity provision.
+  string id = 1;
+  // Unique party ID for the creator of the provision.
```
candida-d, line 2618:
```suggestion
  // ID for the liquidity provision.
  string id = 1;
  // Party ID for the creator of the provision.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  string party_id = 2;
+  // Timestamp in Unix nanoseconds for when the order was created.
+  int64 created_at = 3;
+  // Timestamp in Unix nanoseconds for when the order was updated.
+  int64 updated_at = 4;
+  // Market ID for the order.
```
candida-d, line 2624:
```suggestion
  // Market ID for the market the liquidity provision was made on.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  string market_id = 5;
+  // Specified as a unitless number that represents the amount of settlement asset of the market.
```
candida-d, line 2626:
```suggestion
  // Specified as a unitless number that represents the amount of the market's settlement asset.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  // This field is an unsigned integer scaled to the asset's decimal places.
+  string commitment_amount = 6;
+  // Nominated liquidity fee factor, which is an input to the calculation of taker fees on the market, as per setting fees and rewarding liquidity providers.
```
candida-d, line 2629:
```suggestion
  // Provider's nominated liquidity fee factor, which is an input to the calculation of taker fees on the market, as per setting fees and rewarding liquidity providers.
```

CHANGELOG.md
```diff
@@ -322,6 +322,7 @@
 - [9775](https://github.com/vegaprotocol/vega/issues/9775) - Do not pay discount if set is not eligible
 - [9788](https://github.com/vegaprotocol/vega/issues/9788) - Fix transfer account validation.
 - [9797](https://github.com/vegaprotocol/vega/issues/9797) - Default pagination limits are not always correctly set.
+- [9757](https://github.com/vegaprotocol/vega/issues/9757) - Liquidity provisions `API` shows the pending `LP` instead of the current when an update is accepted by the network.
```
wwestgarth, line 325: Can you plop this in the "breaking change" section of the changelog please?

datanode/gateway/graphql/schema.graphql
```diff
@@ -4912,6 +4912,36 @@ type LiquidityProvision {
   status: LiquidityProvisionStatus!
   "A reference for the orders created out of this liquidity provision"
   reference: String
+  "Liquidity provision that has been updated by the liquidity provider, and has been accepted by the network, but will not be active until the next epoch."
+  pending: PendingLiquidityProvision
+}
+
+"Liquidity provision that has been accepted by the network, but is not yet active."
+type PendingLiquidityProvision {
+  "Unique identifier for the order (set by the system after consensus)"
```
candida-d, line 4921:
```suggestion
  "Unique identifier for the provision, set by the system after consensus"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -4912,6 +4912,36 @@ type LiquidityProvision {
...
+  id: ID!
+  "The party making this commitment"
+  party: Party!
+  "RFC3339Nano time when the liquidity provision was initially created"
+  createdAt: Timestamp!
+  "RFC3339Nano time of when the liquidity provision was updated"
```
candida-d, line 4927:
```suggestion
  "RFC3339Nano time when the liquidity provision was updated"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -4912,6 +4912,36 @@ type LiquidityProvision {
...
+  updatedAt: Timestamp
+  "Market for the order"
+  market: Market!
+  "Specified as a unit-less number that represents the amount of settlement asset of the market."
```
candida-d, line 4931:
```suggestion
  "Specified as a unitless number that represents the amount of the market's settlement asset for the commitment."
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -4912,6 +4912,36 @@ type LiquidityProvision {
...
+  commitmentAmount: String!
+  "Nominated liquidity fee factor, which is an input to the calculation of liquidity fees on the market, as per setting fees and rewarding liquidity providers."
```
candida-d, line 4933:
```suggestion
  "Provider's nominated liquidity fee factor, which is an input to the calculation of liquidity fees on the market, as per setting fees and rewarding liquidity providers."
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
 
+// Liquidity provider commitment. For API purposes this can report a pending liquidity provision,
+// the current live provision or both in the event an update has been accepted on the network, but has yet
+// to go live and there is still a
+message LiquidityProvision {
+  // Unique ID for the liquidity provision.
+  string id = 1;
+  // Unique party ID for the creator of the provision.
+  string party_id = 2;
+  // Timestamp in Unix nanoseconds for when the order was created.
```
candida-d, line 2620:
```suggestion
  // Timestamp in Unix nanoseconds for when the liquidity provision was created.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  int64 created_at = 3;
+  // Timestamp in Unix nanoseconds for when the order was updated.
```
candida-d, line 2622:
```suggestion
  // Timestamp in Unix nanoseconds for when the liquidity provision was updated.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  // Specified as a unitless number that represents the amount of settlement asset of the market.
+  // This field is an unsigned integer scaled to the asset's decimal places.
+  string commitment_amount = 6;
+  // Nominated liquidity fee factor, which is an input to the calculation of taker fees on the market, as per setting fees and rewarding liquidity providers.
+  string fee = 7;
+  // Set of liquidity sell orders to meet the liquidity provision obligation.
+  repeated vega.LiquidityOrderReference sells = 8;
+  // Set of liquidity buy orders to meet the liquidity provision obligation.
+  repeated vega.LiquidityOrderReference buys = 9;
+  // Version of this liquidity provision order.
```
candida-d, line 2635:
```suggestion
  // Version of this liquidity provision.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  uint64 version = 10;
+  // Status of this liquidity provision order.
```
candida-d, line 2637:
```suggestion
  // Status of this liquidity provision.
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -4912,6 +4912,36 @@ type LiquidityProvision {
...
+type PendingLiquidityProvision {
+  "Unique identifier for the order (set by the system after consensus)"
+  id: ID!
+  "The party making this commitment"
+  party: Party!
+  "RFC3339Nano time when the liquidity provision was initially created"
+  createdAt: Timestamp!
+  "RFC3339Nano time of when the liquidity provision was updated"
+  updatedAt: Timestamp
+  "Market for the order"
```
candida-d, line 4929:
```suggestion
  "Market ID for the liquidity provision"
```
assuming this does display as an ID

datanode/gateway/graphql/schema.graphql
```diff
@@ -4912,6 +4912,36 @@ type LiquidityProvision {
...
+  fee: String!
+  "A set of liquidity sell orders to meet the liquidity provision obligation."
+  sells: [LiquidityOrderReference!]!
+  "A set of liquidity buy orders to meet the liquidity provision obligation."
+  buys: [LiquidityOrderReference!]!
+  "The version of this liquidity provision"
+  version: String!
+  "The current status of this liquidity provision"
+  status: LiquidityProvisionStatus!
+  "A reference for the orders created out of this liquidity provision"
```
candida-d, line 4943:
```suggestion
  "A reference for the orders created to support this liquidity provision"
```
I think this is also a holdover from the shapes era, so it's worth checking these order ones are the same functionality

datanode/sqlstore/liquidity_provision.go
```diff
@@ -335,7 +335,7 @@ where liquidity_provider_sla != 'null' and liquidity_provider_sla is not null
 
 func (lp *LiquidityProvision) getWithCursorPagination(ctx context.Context, partyID entities.PartyID, marketID entities.MarketID,
 	reference string, live bool, pagination entities.CursorPagination,
-) ([]entities.LiquidityProvision, entities.PageInfo, error) {
+) ([]entities.CurrentAndPreviousLiquidityProvision, entities.PageInfo, error) {
```
candida-d, line 338: should this (and everywhere it's used) be plural?

candida-d (dismissed):

ValentinTrinque (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2609,10 +2609,43 @@ message ListLiquidityProvisionsResponse {
...
+  repeated vega.LiquidityOrderReference buys = 9;
+  // Version of this liquidity provision.
+  uint64 version = 10;
+  // Status of this liquidity provision.
+  vega.LiquidityProvision.Status status = 11;
+  // Reference shared between this liquidity provision and all its orders.
+  string reference = 12;
+  // Liquidity provision that is currently pending and will go live in the next epoch.
+  optional vega.LiquidityProvision pending = 13;
+}
```
jeremyletang, line 2643: Why not do something like this instead:
```
message LiquidityProvision {
   optional vega.LiquidityProvision active = 1 ;
   optional vega.LiquidityProvision pending = 2;
}
```

karlem: I agree. There is probably no need for all this extra information on the top level

guoguojin: Because I wanted to make the minimal breaking change possible. This additional field with the pending will only affect the gRPC API, the REST API should just have an additional field and wouldn't break I think.

karlem (dismissed):
Overall it looks good. I just wonder whether we want to address the Jeremy's comment. But it is a non blocker

protos/sources/data-node/api/v2/trading_data.proto
guoguojin: I don't think it would be a huge amount of work, no, up to you really.

guoguojin: After quick discussion, we're going to go with the deprecation route.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -475,9 +475,22 @@ service TradingDataService {
...
+  // provision is returned by the API instead of the provision that is currently active.
+  // Use ListLiquidityProvisionsWithPending instead.
   rpc ListLiquidityProvisions(ListLiquidityProvisionsRequest) returns (ListLiquidityProvisionsResponse) {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Liquidity"};
+    option deprecated = true;
+  }
+
+  // List liquidity provisions
+  //
+  // Get a list of liquidity provisions for a given market. This is similar to the list liquidity provisions API
```
wwestgarth, line 489: If the old API is being deprecated then we should treat it such that it no longer exists anymore, so we shouldn't really reference in the doc string here.

Also the `WithPending` bit feels strange? If the old API is deprecated and should no longer be used (and maybe even eventually removed completely) we're left with this one API called `ListLiquidityProvisionsWithPending` which just reads weird given there isn't a `ListLiquidityProvisionsWithoutPending`.

Can we not call is `ListAllLiquidityProvisions`? It will then at least keep in line with the name we used when we deprecated `ListPositions` in favour of `ListAllPositions`?

guoguojin: Yeah, I couldn't think of a better name, I didn't like the ListAll... for the API name but you make a good point in that we have done this in the past and @candida-d also suggested that so let's go with it.

I mention the deprecated API in the doc string because it's only deprecated and it still exists, and if/when we remove the old API, we can update the documentation, but I see your point and will remove it.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -475,9 +475,21 @@ service TradingDataService {
...
+  // Use ListAllLiquidityProvisions instead.
   rpc ListLiquidityProvisions(ListLiquidityProvisionsRequest) returns (ListLiquidityProvisionsResponse) {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Liquidity"};
+    option deprecated = true;
+  }
+
+  // List liquidity provisions
+  //
+  // Get a list of liquidity provisions for a given market. This API returns a current and pending liquidity provision
+  // in the event a provision has been updated by the provider but the updated provision will not be active until the next epoch.
```
candida-d, line 490:
```suggestion
  // Get a list of liquidity provisions for a given market. This API returns a current and pending liquidity provision
  // in the event that a provision has been updated by the provider but the updated provision will not be active until the next epoch.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2607,12 +2619,42 @@ message ListLiquidityProvisionsRequest {
...
 
+// Response from listing liquidity provisions
+message ListAllLiquidityProvisionsResponse {
+  // Page of liquidity provisions data and corresponding page information.
+  LiquidityProvisionsWithPendingConnection liquidity_provisions = 1;
+}
+
+// Liquidity provider commitment. For API purposes this can report a pending liquidity provision,
+// the current live provision or both in the event an update has been accepted on the network, but has yet
+// to go live and there is still a pending provision that will become active in the next epoch.
```
candida-d, line 2650:
```suggestion
// the current live provision or both in the event that an update has been accepted on the network, but has yet
// to go live and thus there is a pending provision that will become active in the next epoch.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2629,6 +2679,14 @@ message LiquidityProvisionsConnection {
   PageInfo page_info = 2;
 }
 
+// Page of liquidity provisions data and corresponding page information
+message LiquidityProvisionsWithPendingConnection {
+  // Page of liquidity provisions data and their corresponding cursors.
+  repeated LiquidityProvisionsWithPendingEdge edges = 1;
+  // Page information that is used for fetching further pages.
```
candida-d, line 2686:
```suggestion
  // Page information that is used to fetch further pages.
```

wwestgarth (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -2607,12 +2619,42 @@ message ListLiquidityProvisionsRequest {
...
 // Response from listing liquidity provisions
 message ListLiquidityProvisionsResponse {
   // Page of liquidity provisions data and corresponding page information.
   LiquidityProvisionsConnection liquidity_provisions = 1;
 }
 
+// Response from listing liquidity provisions
+message ListAllLiquidityProvisionsResponse {
+  // Page of liquidity provisions data and corresponding page information.
+  LiquidityProvisionsWithPendingConnection liquidity_provisions = 1;
```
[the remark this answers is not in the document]
guoguojin: I thought about it and I don't really like the idea of having the connection name starting with "All", I think the name correctly describes what type of messages the connection has

datanode/gateway/graphql/schema.graphql
```diff
@@ -5462,6 +5490,18 @@ type LiquidityProvisionsConnection {
   pageInfo: PageInfo!
 }
 
+"Edge type containing the liquidity provision and cursor information returned by a LiquidityProvisionsWithPendingConnection"
```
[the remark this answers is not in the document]
guoguojin: I renamed the edge type because an edge has a singular Liquidity Provision, while the Connection I left as Provisions (plural) because the connection can return a list.

karlem (approved):

PR #9812, opened 2023-10-17
fix: increase max page size to 5k

EVODelavega:


jeremyletang (dismissed):

datanode/entities/pagination.go
```diff
@@ -123,7 +126,7 @@ func CursorPaginationFromProto(cp *v2.Pagination) (CursorPagination, error) {
 	}
 
 	if cp.First != nil {
-		if *cp.First < 0 || *cp.First > defaultPageSize {
+		if *cp.First < 0 || *cp.First > maxPageSize {
 			return CursorPagination{}, ErrCursorOverflow
```
wwestgarth, line 131: Can we change the error to not be `negative pagination value` and say something useful about it no being in range `1-5000`?

EVODelavega: Changed that to be the message for `ErrCursorOverflow`

ValentinTrinque (approved):

jeremyletang (approved):

PR #9829, opened 2023-10-17
feat: upgrade to cometbft 0.38

### LINKED ISSUE #8051: Upgrade to cometbft 38
cometbft 0.38 has an implementation of abci++ which means we need to implement properly:
1. prepareProposal - to assemble a block
2. processProposal - validate a block proposal and reject it if it contains invalid transactions
3. finalizeBlock - beginBlock - deliverTx - endBlock
4. rewrite spam and pow engines to get rid of bans

ze97286:
Closes #8051


core/processor/abci.go
```diff
@@ -641,98 +656,93 @@ func (app *App) Info(_ tmtypes.RequestInfo) tmtypes.ResponseInfo {
...
+	return &tmtypes.ResponseLoadSnapshotChunk{
+		Chunk: raw.Data,
+	}, nil
 }
 
-func (app *App) OnInitChain(req tmtypes.RequestInitChain) tmtypes.ResponseInitChain {
+func (app *App) OnInitChain(req *tmtypes.RequestInitChain) (*tmtypes.ResponseInitChain, error) {
 	app.log.Debug("ABCI service InitChain start")
-	hash := hex.EncodeToString(vgcrypto.Hash([]byte(req.ChainId)))
+	hash := hex.EncodeToString(vgcrypto.Hash(req.AppStateBytes))
```
wwestgarth, line 737: The app-state bytes is not deterministic because the app-state contains maps, so this needs to stay as `vgcrypto.Hash([]byte(req.ChainId)` otherwise the event file diff will complain.

core/spam/engine_test.go
```diff
@@ -1,17 +1,14 @@
-// Copyright (C) 2023 Gobalsky Labs Limited
+// Copyright (c) 2022 Gobalsky Labs Limited
```
EVODelavega, line 1: 2023

EVODelavega (dismissed):

jeremyletang (approved):

PR #9858, opened 2023-10-20
fix: fix where in query

### LINKED ISSUE #9839: [Bug]: Specifically referral-set-referee pagination broken on fairground
### Problem encountered

I believe this to be a different issue to: https://github.com/vegaprotocol/vega/issues/9836 where something recent has broken pagniation for _all_ endpoints on stagnet.

What I see on fairground is that I can paginated through API's and it works except the one to get referees for a referral set. This means there is something specific to the referral-set-referee cursor generation that isn't working.

Here is a snippet of python to test it:
```
BASE_URL = "https://api.n00.testnet.vega.rocks/api/v2/"

def get_general(url, dict_key, n_in):
    print()
    total = 0
    n = str(n_in)
    r = requests.get(BASE_URL + url +"?pagination.first=" + n)
    rr = r.json()
    print("nrefs", len(rr[dict_key]["edges"]))
    total += len(rr[dict_key]["edges"])
    cursor = rr[dict_key]["pageInfo"]["endCursor"]
    print(cursor)
    while rr[dict_key]["pageInfo"]["hasNextPage"]:
        print()
        full_url = BASE_URL + url + "?pagination.first="+n+"&pagination.after="+ cursor
        print(full_url)
        r = requests.get(full_url)
        rr = r.json()
        print("nrefs", len(rr[dict_key]["edges"]))
        total += len(rr[dict_key]["edges"])
        cursor = rr[dict_key]["pageInfo"]["endCursor"]
        print(cursor)

    print("found this many things:", total)

def main():

    # pagniating through accounts works just fine
    get_general("accounts", "accounts", 1000)

    # works because there are 156 entries and so there is nothing to pagniation
    get_general("referral-sets/referees", "referralSetReferees", 1000)
    # does not work, loops forever
    get_general("referral-sets/referees", "referralSetReferees", 100)
    return

if __name__ == "__main__":
    main()
```

### Observed behaviour

Pagination of referal set referees does not work.

### Expected behaviour

Pagination of referal set referees works.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.73.0-preview.16

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

karlem:
Close #9839

### DIFF
--- a/datanode/sqlstore/referral_sets.go
+++ b/datanode/sqlstore/referral_sets.go
@@ -231,16 +231,25 @@ func (rs *ReferralSets) ListReferralSetReferees(ctx context.Context, referralSet
 
 	query := getSelectQuery(aggregationDays)
 
+	var hasWhere bool
 	// we only allow one of the following to be used as the filter
 	if referralSetID != nil {
 		query = fmt.Sprintf("%s where rf.referral_set_id = %s", query, nextBindVar(&args, referralSetID))
+		hasWhere = true
 	} else if referrer != nil {
 		query = fmt.Sprintf("%s where rs.referrer = %s", query, nextBindVar(&args, referrer))
+		hasWhere = true
 	} else if referee != nil {
 		query = fmt.Sprintf("%s where rf.referee = %s", query, nextBindVar(&args, referee))
+		hasWhere = true
 	}
 
-	query, args, err = PaginateQuery[entities.ReferralSetRefereeCursor](query, args, referralSetRefereeOrdering, pagination)
+	paginate := PaginateQueryWithWhere[entities.ReferralSetRefereeCursor]
+	if hasWhere {
+		paginate = PaginateQuery[entities.ReferralSetRefereeCursor]
+	}
+
+	query, args, err = paginate(query, args, referralSetRefereeOrdering, pagination)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


datanode/sqlstore/referral_sets.go
```diff
@@ -231,16 +231,25 @@ func (rs *ReferralSets) ListReferralSetReferees(ctx context.Context, referralSet
...
 	} else if referee != nil {
 		query = fmt.Sprintf("%s where rf.referee = %s", query, nextBindVar(&args, referee))
+		hasWhere = true
 	}
 
-	query, args, err = PaginateQuery[entities.ReferralSetRefereeCursor](query, args, referralSetRefereeOrdering, pagination)
+	paginate := PaginateQueryWithWhere[entities.ReferralSetRefereeCursor]
+	if hasWhere {
+		paginate = PaginateQuery[entities.ReferralSetRefereeCursor]
+	}
```
jeremyletang, line 250: For some reasons I'd expect PaginateQueryWithWhere to be inside the if, not the inverse?

jeremyletang (approved):

PR #9865, opened 2023-10-20
8056 transfer fees

### LINKED ISSUE #8056: [API]: Add `fee` to transfers objects in GraphQL & REST
>[!WARNING]
> Reopened as it is not possible in GraphQL to get the fee for an individual transfer

https://github.com/vegaprotocol/vega/blob/develop/datanode/gateway/graphql/schema.graphql#L718-L755

In GraphQL and REST, it is not possible to know what they fee paid for a transfer is. You can *guess* by looking at a network parameter, but that is unreliable for transfers in the past as the factor could have changed.

# Changes
- A new field should represent the fee paid for a transfer
- The fee should be the value in the same asset as the transfer
- The fee should not be the fee factor, but the actual amount paid for the transfer

# Tasks
- [ ] https://github.com/vegaprotocol/vega/issues/8056#issuecomment-1853673347
- [x] Expose `fee` in GraphQL
- [x] Expose `fee` in REST
- [ ] Either:
    - [x] Recurring transfers should contain an array of fee objects containing the individual fees paid, and the date and time of each
    - [ ] Recurring transfers should contain a total of the fees paid to date

EVODelavega:
Closes #8056


protos/sources/vega/events/v1/events.proto
```diff
@@ -445,6 +445,15 @@ message MarketEvent {
   string payload = 2;
 }
 
+message TransferFees {
+  // the transfer that triggered the colleciton of fees.
+  string transfer_id = 1;
+  // the amount of fees transferred.
+  string amount = 2;
+  // the epoch when the transfer was dispatched, and fees were paid (zero for one-off transfers).
```
candida-d, line 453:
```suggestion
  // Transfer that triggered the collection of fees.
  string transfer_id = 1;
  // Amount of fees paid.
  string amount = 2;
  // Epoch when the transfer was dispatched, and fees were paid. Zero for one-off transfers.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -1383,6 +1395,8 @@ message BusEvent {
     PaidLiquidityFeesStats paid_liquidity_fees_stats = 179;
     // Event notifying of an update to the vesting and locked balances.
     VestingBalancesSummary vesting_balances_summary = 180;
+    // Event notifying of fees being transferred related to a transfer.
```
candida-d, line 1398:
```suggestion
    // Event notifying of fees related to a transfer being paid.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1698,10 +1698,17 @@ message ListTransfersResponse {
...
+  vega.events.v1.Transfer transfer = 1;
+  // Fees paid for this transfer.
+  repeated vega.events.v1.TransferFees fees = 2;
+}
+
 // Transfers data with the corresponding cursor.
 message TransferEdge {
   // Data relating to a transfer that has been made.
-  vega.events.v1.Transfer node = 1;
+  TransferNode node = 1;
```
wwestgarth, line 1711: I think this makes it a breaking change, which annoyingly means we might have to have a whole new API unless theres is another way.

EVODelavega: The motivation for this change is that the current API is incomplete (some would even argue defective/broken), so a breaking change to a broken part of the API arguably isn't a bad thing. There's one alternative that _might_ reduce the number of new types here, and that would be to overload the transfers event type to contain a slice of fee payments, but that causes issues further down the line (and elsewhere in the code)

EVODelavega:
```
{
"system-tests": "8056-transfer-api"
}
```

mattrussell36 (approved):

ValentinTrinque (approved):

candida-d (approved):

PR #9873, opened 2023-10-23
feat: support timestamp for open oracle, fallback to vega time

### LINKED ISSUE #8769: [Feature]: Perps -  timestamp selection in settlement data-point data-sourcing

EVODelavega:
Closes #8769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -348,7 +348,7 @@
 - [9858](https://github.com/vegaprotocol/vega/issues/9858) - Fix missing where statement in list referral set referees.
 - [9868](https://github.com/vegaprotocol/vega/issues/9868) - Avoid empty ledger entries from governance transfers for metric when there is no activity.
 - [9871](https://github.com/vegaprotocol/vega/issues/9871) - Fix panic on snapshot after market termination following protocol upgrade. 
-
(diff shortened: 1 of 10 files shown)


wwestgarth:
Looks like the normal JSON oracle code path is still missing having vega-time added to its meta-data?

EVODelavega:
> Looks like the normal JSON oracle code path is still missing having vega-time added to its meta-data?

Yes, but there's no pre-set property that contains the data. All `common.Data` types that get broadcast will have the vega-time metadata set, so that should be covered by that.

wwestgarth (dismissed):

wwestgarth (approved):

PR #9885, opened 2023-10-24
feat: add reward filter to transfer API

### LINKED ISSUE #9883: Extend filters of transfer API
Add a new filter to the transfer API.

### todo
- [ ] add an onlyRewards boolean filter to the transfersConnection API
- [ ] if enabled returns only recurring and govRecurring transfers
- [ ] if enabled returns only transfers with a dispatch metric

EVODelavega:
Closes #9883


datanode/sqlstore/transfers.go
```diff
@@ -217,3 +217,37 @@ func (t *Transfers) getTransfers(ctx context.Context, pagination entities.Cursor
 
 	return transfers, pageInfo, nil
 }
+
+func (t *Transfers) GetAllRewards(ctx context.Context, pagination entities.CursorPagination) ([]entities.Transfer, entities.PageInfo, error) {
```
wwestgarth, line 221: Can you add some unittests please?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1687,9 +1687,12 @@ message ListTransfersRequest {
   // Restrict transfer to those where the given public key is a sender or receiver.
   optional string pubkey = 1;
   // Restrict transfers to those in the given direction from the supplied public key.
+  // When is_reward is true, and pubkey is set, then direction MUST be TRANSFER_DIRECTION_DIRECTION_FROM.
   TransferDirection direction = 2;
   // Optional pagination control
   optional Pagination pagination = 3;
+  // Optional filter to return only reward transfers.
```
wwestgarth, line 1694:
```suggestion
  // Restrict transfers to those related to reward payments.
```

datanode/sqlstore/transfers.go
EVODelavega: Added - thought I opened this as a draft for that reason

protos/sources/data-node/api/v2/trading_data.proto
candida-d: Is "payments" right here if this also includes transfers that add assets to a reward pool?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1687,9 +1687,12 @@ message ListTransfersRequest {
   // Restrict transfer to those where the given public key is a sender or receiver.
   optional string pubkey = 1;
   // Restrict transfers to those in the given direction from the supplied public key.
+  // When is_reward is true, and pubkey is set, then direction MUST be TRANSFER_DIRECTION_DIRECTION_FROM.
```
candida-d, line 1690:
```suggestion
  // When is_reward is true, and pubkey is set, then direction MUST be 'from'.
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -2602,6 +2604,8 @@ type Party {
     direction: TransferDirection
     "Pagination information"
     pagination: Pagination
+    "Filter reward transfers only. Direction must be From"
```
candida-d, line 2607:
```suggestion
    "Filter for reward transfers only. Direction must be 'From'"
```
It needs to be 'from' in order to filter?

datanode/gateway/graphql/schema.graphql
```diff
@@ -988,6 +988,8 @@ type Query {
     direction: TransferDirection
     "Pagination information"
     pagination: Pagination
+    "Filter reward transfers only. Direction must be From if a public key is used"
```
candida-d, line 991:
```suggestion
    "Filter for reward transfers only. Direction must be 'from' if a public key is used"
```

protos/sources/data-node/api/v2/trading_data.proto
EVODelavega: I'll keep it as transfers. I'd argue a payment is when funds move between accounts, this is just a standing order to conditionally/periodically make such payments.

protos/sources/data-node/api/v2/trading_data.proto
EVODelavega: Doesn't it make more sense to use the values that make sense within the types in the proto files? `'from'` is not a valid `TransferDirection` value after all

guoguojin (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
candida-d: Are they exactly the same for REST and gRPC on the user side?

EVODelavega: Yes, looking at the `trading_data.swagger.json` file:

```json
          {
            "name": "direction",
            "description": "Restrict transfers to those in the given direction from the supplied public key.",
            "in": "query",
            "required": false,
            "type": "string",
            "enum": [
              "TRANSFER_DIRECTION_UNSPECIFIED",
              "TRANSFER_DIRECTION_TRANSFER_FROM",
              "TRANSFER_DIRECTION_TRANSFER_TO",
              "TRANSFER_DIRECTION_TRANSFER_TO_OR_FROM"
            ],
            "default": "TRANSFER_DIRECTION_UNSPECIFIED"
          }
```

protos/sources/data-node/api/v2/trading_data.proto
candida-d: It could be "Restrict results to only reward transfers", or "Filter to return only reward transfers"

candida-d: I have a slight preference for the first one. But consistency is best so whichever style we use more frequently should be the winner.

EVODelavega: Done

candida-d (approved):

guoguojin (approved):

EVODelavega:
```
{
"system-tests": "8056-transfer-api"
}
```

PR #9888, opened 2023-10-25
Do not wipe the snapshots when rolling back from 0.73

ValentinTrinque:


core/snapshot/config_test.go
```diff
@@ -35,6 +36,8 @@ func testEngineConfigDefaultConfigIsValid(t *testing.T) {
 }
 
 func testEngineConfigInvalidConfigFails(t *testing.T) {
+	t.Skip()
```
EVODelavega, line 39: Do we have any tests for invalid data with this one skipped?

jeremyletang (approved):

PR #9905, opened 2023-10-26
fix: apply all fee discounts on the full fee amount

### LINKED ISSUE #9904: [Bug]: fee discounts are applied incorrectly
### Problem encountered

volume and referrer discounts are both expected to be applied on the full fee amount and applied together rather than one applied and the other applied on the remaining amount. 

### Observed behaviour

fee discounts are applied incrementally i.e. the volume discount is applied on the remaining fee after the referrer discount is applied. 

### Expected behaviour

https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0029-FEES-fees.md#applying-benefit-factors

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.73.0-rc.2

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
Closes #9904


ze97286:
closing for now as the decision is to change the spec.

PR #9917, opened 2023-10-27
fix: fix untilEpoch in activity streak even

### LINKED ISSUE #9916: [Bug]: The vesting summary seems to show the epoch as being out by 1
### Problem encountered

Running this query gives the wrong result 

```
query vestingDetails {
  party(id: "89fd87b86727fa4d9ab06b9f4e64b51c9b07055e8a0c8f0bc40f0a249e321525") {
    vestingStats{
      rewardBonusMultiplier
      epochSeq
      }
    vestingBalancesSummary {
      epoch
      vestingBalances {
        asset {
          id
          symbol
          quantum
          decimals
        }
        balance
      }
      lockedBalances {
        asset {
          id
          symbol
          quantum
          decimals
        }
        balance
        untilEpoch
      }
    }
  }
}
```

### Observed behaviour

I get back this result

So it is trying to tell me that I have rewards locked until epoch 9789 but:

1. We are through epoch 9789 and are in 9890
2. The [transfer](https://explorer.fairground.wtf/txs/EE30F7C4C3F8FA05D43BD7C7BECD2D47CD6B9670CFBECC2DB98C9C5B5F1FA1E9) started in 9788 and is locked for 2 epochs so the first reward SHOULD be in 9890

```
{
  "data": {
    "party": {
      "vestingStats": {
        "rewardBonusMultiplier": "1.05",
        "epochSeq": 9789
      },
      "vestingBalancesSummary": {
        "epoch": 9789,
        "vestingBalances": null,
        "lockedBalances": [
          {
            "asset": {
              "id": "c9fe6fc24fce121b2cc72680543a886055abb560043fda394ba5376203b7527d",
              "symbol": "tUSDC",
              "quantum": "1",
              "decimals": 5
            },
            "balance": "81599",
            "untilEpoch": 9789
          },
          {
            "asset": {
              "id": "c9fe6fc24fce121b2cc72680543a886055abb560043fda394ba5376203b7527d",
              "symbol": "tUSDC",
              "quantum": "1",
              "decimals": 5
            },
            "balance": "94236",
            "untilEpoch": 9790
          }
        ]
      }
    }
  }
}
```

### Expected behaviour

It should be locked until 9890 (I think it is) and should show that in the API (it is not)

### Steps to reproduce

```bash
See above
```

### Software version

73 rc4

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

jeremyletang:
close #9916

### DIFF
--- a/core/vesting/vesting.go
+++ b/core/vesting/vesting.go
@@ -372,7 +372,7 @@ func (e *Engine) broadcastSummary(ctx context.Context, seq uint64) {
 					&eventspb.PartyLockedBalance{
 						Asset:      asset,
 						Balance:    balance.String(),
-						UntilEpoch: seq + remainingEpochs,
+						UntilEpoch: seq + remainingEpochs + 1, // we add one here because the remainingEpochs can be 0, meaning the funds are released next epoch
 					},
 				)
 			}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -360,6 +360,7 @@
 - [9900](https://github.com/vegaprotocol/vega/issues/9900) - Fix for duplicate liquidity provisions in the `API`.
 - [9911](https://github.com/vegaprotocol/vega/issues/9911) - Bind correct property in `GraphQL` resolver.
 - [9913](https://github.com/vegaprotocol/vega/issues/9913) - Add missing properties in schema.
+- [9916](https://github.com/vegaprotocol/vega/issues/9916) - Fix activity streak locked account until epoch.
 
 ## 0.72.1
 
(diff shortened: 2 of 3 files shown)


core/vesting/vesting.go
```diff
@@ -372,7 +372,7 @@ func (e *Engine) broadcastSummary(ctx context.Context, seq uint64) {
 					&eventspb.PartyLockedBalance{
 						Asset:      asset,
 						Balance:    balance.String(),
-						UntilEpoch: seq + remainingEpochs,
+						UntilEpoch: seq + remainingEpochs + 1, // we add one here because the remainingEpochs can be 0, meaning the funds are released next epoch
```
ze97286, line 375: does it mean that? doesn't 0 mean they are not parked in the vesting account but move in the same epoch to the vested?

jeremyletang: well, it depends. when calling AddReward: https://github.com/vegaprotocol/vega/blob/develop/core/vesting/vesting.go#L151 then yes.

But in here: https://github.com/vegaprotocol/vega/blob/develop/core/vesting/vesting.go#L238-L246
We can have some with 0, which means it get distributed next epoch.

jeremyletang: in any case if it's 0 when sent to AddRewad, it never get locked, and go through this process.

ze97286: ok cool, so it's just the internal accounting of the vesting engine

ze97286 (approved):

core/vesting/vesting.go
jeremyletang: yes.

PR #9925, opened 2023-10-27
fix: referral set referees api should aggregate by number of epochs

### LINKED ISSUE #9924: [Feature]:
# Feature overview

I dont think the aggregationDays argument is doing anything.

There are only 3 parties in that set (2 in the previous epoch) and I did no volume in the last epoch but I still get the same value back whether I do 1 day, 2 days ... 30 days etc.

It sounds like the problem may be that we have actually done this API with DAY when we should be going in epochs so it aligns with all the rest of the referral program design.  As it stands I cant make sense of the numbers in here relative to the rest of the referral APIs .... we need **aggegrationDays** to be **aggregationEpochs** so that we can make it all add up please.

```
query refereesBySet {
  referralSetReferees(
    id:"876be7c89c40d4a898b886a17f0039f86063cb35223833e4d585cef18dc452b3",
    aggregationDays:2)
  {
    edges{
      node{
        referralSetId
        refereeId
        joinedAt
        atEpoch
        totalRefereeGeneratedRewards
        totalRefereeNotionalTakerVolume
      }
    }
  }
}

```

guoguojin:
Closes #9924


datanode/api/trading_data_v2.go
```diff
@@ -4410,10 +4410,10 @@ func (t *TradingDataServiceV2) ListReferralSetReferees(ctx context.Context, req
...
 
 	var (
-		id              *entities.ReferralSetID
-		referrer        *entities.PartyID
-		referee         *entities.PartyID
-		daysToAggregate uint32 = 30 // default is 30 days
+		id                *entities.ReferralSetID
+		referrer          *entities.PartyID
+		referee           *entities.PartyID
+		epochsToAggregate uint32 = 30 // default is 30 days
```
candida-d, line 4416:
```suggestion
		epochsToAggregate uint32 = 30 // default is 30 epochs
```

datanode/gateway/graphql/resolvers.go
```diff
@@ -1673,19 +1673,19 @@ func (r *myQueryResolver) ReferralSets(ctx context.Context, id, referrer, refere
 	return resp.ReferralSets, nil
 }
 
-func (r *myQueryResolver) ReferralSetReferees(ctx context.Context, id, referrer, referee *string, pagination *v2.Pagination, daysToAggregate *int) (*v2.ReferralSetRefereeConnection, error) {
-	var aggregationDays uint32 = 30 // default to 30 days
+func (r *myQueryResolver) ReferralSetReferees(ctx context.Context, id, referrer, referee *string, pagination *v2.Pagination, epochsToAggregate *int) (*v2.ReferralSetRefereeConnection, error) {
+	var aggregationEpochs uint32 = 30 // default to 30 days
```
candida-d, line 1677:
```suggestion
	var aggregationEpochs uint32 = 30 // default to 30 epochs
```

asiaznik (dismissed):

candida-d (commented):
doc strings look good, just highlighting 2 places where days were missed in annotations

candida-d (approved):

jeremyletang (approved):

PR #9926, opened 2023-10-27
9881 rm liq monitoring

### LINKED ISSUE #9881: [Chore]: remove liquidity monitoring auctions
In order to keep the code base clean we should remove liquidity monitoring auctions now we have the new liquidity mechanism.

As per:

- [Feat: remove liquidity monitoring auctions](https://github.com/vegaprotocol/specs/pull/1991)

EVODelavega:
Closes #9881


core/execution/future/market_test.go
```diff
@@ -5648,7 +5605,7 @@ func Test2963EnsureEquityShareAreInMarketData(t *testing.T) {
 	tm.market.OnTick(ctx, tm.now)
 
 	require.NoError(t, tm.market.SubmitLiquidityProvision(ctx, lp, "party-2", vgcrypto.RandomHash()))
-	assert.Equal(t, 1, tm.market.GetLPSCount())
+	assert.Equal(t, 0, tm.market.GetLPSCount())
```
witgaw, line 5608: Do you understand what happened here? Seems a bit puzzling that the LP would not be active now.

EVODelavega: Because we've removed liquidity monitoring, the market has left opening auction before we submit the LP, so in the liquidity engine, we return `false` for `submittedImmediately`, and so the equity shares aren't updated (which is where we get the LP count from). I could've moved around the LP submission and have `tm.market.GetLPSCount()` return the old value of 1, but I've left some comments on the spec PR, and wanted someone from research to verify this behaviour because I'm not sure what the desired behaviour here would be.  Technically, the LP submission now happens after the market leaves opening auction, but then the market is able to enter continuous trading without any provided liquidity. I'm not sure if that's entirely correct

EVODelavega:
Full run: https://jenkins.vega.rocks/job/common/job/system-tests/5455/

core/monitor/auction.go
```diff
@@ -252,26 +239,10 @@ func (a AuctionState) IsOpeningAuction() bool {
 	return a.trigger == types.AuctionTriggerOpening
 }
 
-func (a AuctionState) IsLiquidityAuction() bool {
-	// FIXME(jeremy): the second part of the condition is to support
-	// the compatibility on 72 > 73 snapshots.
-
-	return a.trigger == types.AuctionTriggerLiquidityTargetNotMet ||
-		a.trigger == types.AuctionTriggerUnableToDeployLPOrders
-}
```
jeremyletang, line 261: I'm pretty sure we still need at least the first part to make sure we are compatible with market being in liquidity auction before an upgrade?

Or we could also just make sure that at the time of the migration, the liquidity auction type is migrated to something else.

EVODelavega:
Another full run: https://jenkins.vega.rocks/job/common/job/system-tests/5473/

core/monitor/auction.go
EVODelavega: Updated, this should be fine now

wwestgarth (dismissed):

core/execution/future/market.go
```diff
@@ -1417,7 +1403,7 @@ func (m *Market) leaveAuction(ctx context.Context, now time.Time) {
 
 	m.checkForReferenceMoves(ctx, updatedOrders, true)
 
-	m.checkLiquidity(ctx, nil, true)
+	m.checkLiquidity(ctx)
```
ze97286, line 1406: could just call here m.checkBondBalance(ctx) I suppose and get rid of checkliquidity?

ze97286 (dismissed):

ze97286 (approved):

Jiajia-Cui (approved):
integration tests look good to me

PR #9929, opened 2023-10-30
feat: expose quanum balance in fee stats event

jeremyletang:
Referral set referree endpoint exposes aggregate data of fee from different assets. this cannot work, This exposes the quantum value of the fees in the core event, and uses it in the postgres query instead


core/execution/future/market.go
```diff
@@ -240,7 +240,8 @@ func NewMarket(
 	)
 	positionEngine := positions.NewSnapshotEngine(log, positionConfig, mkt.ID, broker)
 
-	feeEngine, err := fee.New(log, feeConfig, *mkt.Fees, asset, positionFactor)
+	quoteAssetQuantum, _ := collateralEngine.GetAssetQuantum(asset)
```
ze97286, line 243: what if asset quantum changes? is this being propagated somewhere?

jeremyletang: that's a good point, that I ignored.
I can pass around the collateral engine instead and just GetAssetQuantum everytime?

EVODelavega (approved):

ze97286 (approved):

candida-d (approved):

PR #9979, opened 2023-11-03
Support amending the order size by defining the target size

### LINKED ISSUE #9943: Order amendment improvements
This issue is to implement improvements to the order amendments based on user feedback.

- See feedback below
- Also requested an "amend ID"

Image

The details can be found in the spec change:

- https://github.com/vegaprotocol/specs/pull/2020
- https://github.com/vegaprotocol/specs/pull/2031

### LINKED ISSUE #9960: [0004-AMND-058] - test coverage for Palazzo
Add test coverage for [0004-AMND-058](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0004-AMND-amends.md#0004-AMND-058)

### LINKED ISSUE #9961: [0004-AMND-056] - test coverage for Palazzo
Add test coverage for [0004-AMND-056](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0004-AMND-amends.md#0004-AMND-056)

### LINKED ISSUE #9962: [0004-AMND-057] - test coverage for Palazzo
Add test coverage for [0004-AMND-057](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0004-AMND-amends.md#0004-AMND-057)

ValentinTrinque:
Close #9943 

close https://github.com/vegaprotocol/vega/issues/9960 
close https://github.com/vegaprotocol/vega/issues/9961 
close https://github.com/vegaprotocol/vega/issues/9962


core/integration/steps/errors.go
```diff
@@ -20,6 +20,8 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/exp/maps"
+
```
EVODelavega, line 24: RMBL (third party packages should be added at the bottom)

core/integration/steps/table_wrapper.go
```diff
@@ -240,6 +241,13 @@ func (r RowWrapper) MaybeUint(name string) *num.Uint {
 	return u
 }
 
+func (r RowWrapper) MaybeU64(name string) *uint64 {
```
EVODelavega, line 244: Don't like this very much... At the very least, I´d want this renamed to `U64Ptr` or something.

core/integration/steps/the_orders_should_have_the_following_states.go
```diff
@@ -18,36 +18,50 @@ package steps
 import (
 	"fmt"
 
-	"code.vegaprotocol.io/vega/core/integration/stubs"
 	"github.com/cucumber/godog"
+
```
EVODelavega, line 22: Other way around

core/integration/steps/the_orders_should_have_the_following_states.go
```diff
@@ -67,6 +81,9 @@ func parseOrdersStatesTable(table *godog.Table) []RowWrapper {
 		"side",
 		"volume",
 		"price",
+		"remaining",
```
EVODelavega, line 84: Why would `remaining` be a required field? It's far easier to add it as an optional one. Tests where we submit an order and check for resulting trades == 0 will automatically have a remaining == size, but this doesn't make a whole lot of sense to include in FOK/IOC or any order that has been filled. Checking the final status is more than enough.

libs/stringer/stringer.go
```diff
@@ -46,3 +46,10 @@ func Int64PointerToString(n *int64) string {
 	}
 	return strconv.FormatInt(*n, 10)
 }
+
+func Uint64PointerToString(n *uint64) string {
```
EVODelavega, line 50: Why not just:

```go
func PtrToString[T any](v *T) string {
    if v == nil {
        return "nil"
    }
    return fmt.Sprintf("%v", *v)
}
```

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -140,6 +141,14 @@ message OrderAmendment {
   string pegged_offset = 7;
   // New pegged reference for the order.
   vega.PeggedReference pegged_reference = 8;
+  // New size for the order.
+  // Amending the size causes the the size and remaining part of the order to be amended by the difference between the original and amended size.
+  // - Specifying a size lower than the current size leave the order in its current spot.
+  // - Specifying a size greater than the current size remove and reinsert the order at the back of the price level.
+  // - Specifying a size that result in the remaining part of the order being reduced below zero cancels the order.
+  // This field is an unsigned integer scaled to the market's decimal places.
+  // If specified, size_delta must be set to 0.
```
candida-d, line 150:
```suggestion
  // Amending the size causes the size and remaining part of the order to be changed by the difference between the original and amended size.
  // - Specifying a size smaller than the current size leaves the order at its current order book position.
  // - Specifying a size larger than the current size removes and reinsert the order at the back of the price level.
  // - Specifying a size that results in the remaining part of the order being reduced to zero cancels the order.
  // This field is an unsigned integer scaled to the market's decimal places.
  // If specified, size_delta must be set to 0.
```

core/integration/steps/table_wrapper.go
ValentinTrinque: What about the consistency with the rest of the functions alike ? Probably another PR, so we don't get side-tracked.

core/integration/steps/the_orders_should_have_the_following_states.go
ValentinTrinque: Being very explicit in the order state verification brings more confidence in the tests. As for verifying the trading core feature and state, I don't think we are _saving_ anything from not being extremely assertive. That might look redundant but these tests are also here to make us as confident as possible.

libs/stringer/stringer.go
ValentinTrinque: I can refactor this in another PR.

candida-d (dismissed):

jeremyletang (approved):

PR #9984, opened 2023-11-04
feat: remove fee and minimal amount on transfer from vested account

### LINKED ISSUE #9982: Improvements to transfers made from a vested account
## Issue Description

The following spec change specifies changes such that transfers from a vested account to general account of the same key should

- incur no transfer fee
- not be subject to minimum transfer amount if the full amount is transferred

This issue is to implement the changes as specified

spec:

- https://github.com/vegaprotocol/specs/issues/2032

## Workaround 

Users are able to redeem their rewards at any time using the transfer function in Console. 

If they do not wish to incur fees for this they should wait until this patch is applied.

jeremyletang:
close #9982

### DIFF
--- a/core/banking/transfers_common.go
+++ b/core/banking/transfers_common.go
@@ -66,24 +66,33 @@ func (e *Engine) CheckTransfer(t *types.TransferBase) error {
 		return fmt.Errorf("could not transfer funds, %w", err)
 	}
 
-	if err := e.ensureMinimalTransferAmount(a, t.Amount); err != nil {
+	if err := e.ensureMinimalTransferAmount(a, t.Amount, t.FromAccountType, t.From); err != nil {
 		return err
 	}
 
-	_, err = e.ensureFeeForTransferFunds(t.Amount, t.From, t.Asset, t.FromAccountType)
+	_, err = e.ensureFeeForTransferFunds(t.Amount, t.From, t.Asset, t.FromAccountType, t.To)
 	if err != nil {
 		return fmt.Errorf("could not transfer funds, %w", err)
 	}
 	return nil
 }
 
-func (e *Engine) ensureMinimalTransferAmount(a *assets.Asset, amount *num.Uint) error {
+func (e *Engine) ensureMinimalTransferAmount(
+	a *assets.Asset,
+	amount *num.Uint,
+	fromAccType types.AccountType,
+	from string,
+) error {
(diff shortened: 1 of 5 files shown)


core/banking/transfers_common.go
```diff
@@ -77,7 +77,11 @@ func (e *Engine) CheckTransfer(t *types.TransferBase) error {
 	return nil
 }
 
-func (e *Engine) ensureMinimalTransferAmount(a *assets.Asset, amount *num.Uint) error {
+func (e *Engine) ensureMinimalTransferAmount(a *assets.Asset, amount *num.Uint, fromAccType types.AccountType) error {
+	if fromAccType == types.AccountTypeVestedRewards {
```
ze97286, line 81: I think the requirement is that only if it's the full amount then there's no minimum.

core/banking/transfers_common.go
```diff
@@ -180,8 +184,13 @@ func (e *Engine) makeFeeTransferForTransferFunds(
 	from, asset string,
 	fromAccountType types.AccountType,
 ) *types.Transfer {
+	// no fee for Vested account
+	feeAmount := num.UintZero()
+
 	// first we calculate the fee
-	feeAmount, _ := num.UintFromDecimal(amount.ToDecimal().Mul(e.transferFeeFactor))
+	if fromAccountType != types.AccountTypeVestedRewards {
```
ze97286, line 191: I think the requirement is that this applies only if you're transferring between your own account, so if you transfer from vested to some other general account (not yours) there should still be fee (according to the ticket at least).

jeremyletang: Fair point.

core/banking/transfers_common.go
jeremyletang: I'm not sure that requirement makes all the sense, because you may want to take of some from the vested, but not all so you keep increase / conserve your multiplier.

ze97286: it just means that if the amount you want to transfer is so small (<min_amount), then we let you move it only if it's the full amount in the account. I don't know if the requirement makes sense or not, but it's what the ticket says :)

jeremyletang: hum right, I understood this as, all the balance, all the time not just if < min_amount. that makes more sense

ze97286 (dismissed):

EVODelavega (approved):

PR #10005, opened 2023-11-07
feat: track average entry price in position engine

### LINKED ISSUE #10004: [Feature]: Add average entry price to position engine
# Feature Overview

**In order to** support the requirements for margin isolation 
**We will** add average entry price to the position engine
**So that** so we can calculate the amount of funds that need to be moved to the margin account when switching between modes.

ze97286:
Closes #10004


core/positions/engine.go
```diff
@@ -612,3 +617,17 @@ func (e *Engine) remove(p *MarketPosition) {
...
+	if volume+addVolume == 0 || addPrice == nil {
+		return num.DecimalZero()
+	}
+
+	volumeDec := num.DecimalFromInt64(volume)
+	addVolumeDec := num.DecimalFromInt64(addVolume)
+	addPriceDec := num.DecimalFromUint(addPrice)
+
+	//	return ((vwap * float64(volume)) + (float64(addPrice) * float64(addVolume))) / (float64(volume) + float64(addVolume))
+	return vwap.Mul(volumeDec).Add(addPriceDec.Mul(addVolumeDec)).Div(volumeDec.Add(addVolumeDec))
```
jeremyletang, line 632: Any chance we could use Uints here instead?

ze97286: maybe lets see how it goes first performance wise and decide later?

jeremyletang: Alright.

ze97286: no, not alright. I changed my mind again.

jeremyletang: alright then.

ze97286: not so much because of data type, more because all of this is wrong

candida-d (approved):

EVODelavega (approved):

PR #10006, opened 2023-11-07
Liquidation strategy

### LINKED ISSUE #9945: [Liquidation improvements] Completely decouple MTM, margin checks and closeout trades
### Issue:

This issue is to implement liquidation improvements noted having seen markets working in mainnet since Alpha mainnet was deployed by the validators.

The current design around margins means that:

- closeout trades happen (or are attempted) when a party is in breach of margin requirement and when there is enough volume and when we're not in an auction.
- we cannot handle (future) frequent batch auctions
- we may reward a party that has a big position (MTM move against them, we want to close out but can't, MTM move for them they're OK again - the network was at risk but party kept profit).
- the logic in core is tightly coupled in places making moves towards WASM based smart products harder.
The aim here is:

Separate margin checks, mark-to-market and closeout trades into three separate streams that can run at arbitrary frequencies,
1. Accommodate closeout trades during auctions,
2. Avoid allowing parties with huge position breaching margin level hanging around.
3. Do this without changing any existing margin calculations.
4. Do this without changing how loss socialisation works.

### Specs:

- https://github.com/vegaprotocol/specs/issues/1641

EVODelavega:
First part of issue #9945: Update market framework


jeremyletang (changes requested):
Could you add the commands package validation as well please?

protos/sources/vega/markets.proto
```diff
@@ -389,3 +391,15 @@ message MarketTimestamps {
...
+
+// The liquidation strategy used when the network holds a position resulting from position resolution.
+message LiquidationStrategy {
+  // The period of time over which the network will attempt to close its position.
+  int64 disposal_time_step = 1;
+  // The fraction of the open position the market will attempt to close in a single attempt; range 0 through 1.
+  string disposal_fraction = 2;
+  // The size of the position that the network will attempt to close in a single attempt.
+  uint64 full_disposal_size = 3;
+  // The max fraction of the total volume of the orderbook (within liquidity bounds) the network can use to close its position; range 0 through 1.
```
candida-d, line 403:
```suggestion
// Liquidation strategy used when the network holds a position resulting from position resolution.
message LiquidationStrategy {
  // Period of time over which the network will attempt to close its position.
  int64 disposal_time_step = 1;
  // Fraction of the open position the market will attempt to close in a single attempt; range 0 through 1.
  string disposal_fraction = 2;
  // Size of the position that the network will attempt to close in a single attempt.
  uint64 full_disposal_size = 3;
  // Max fraction of the total volume of the orderbook, within liquidity bounds, that the network can use to close its position; range 0 through 1.
```

candida-d (dismissed):

candida-d (dismissed):

commands/proposal_submission.go
```diff
@@ -975,6 +977,27 @@ func checkPriceMonitoring(parameters *protoTypes.PriceMonitoringParameters, pare
 	return errs
 }
 
+func checkLiquidationStrategy(params *protoTypes.LiquidationStrategy, parent string) Errors {
```
jeremyletang, line 980: Is there no validation required fro:   `uint64 full_disposal_size = 3;`

core/types/market.go
```diff
@@ -999,6 +1000,12 @@ func MarketFromProto(mkt *vegapb.Market) (*Market, error) {
 	if mkt.ParentMarketId != nil {
 		parent = *mkt.ParentMarketId
 	}
+	var ls *LiquidationStrategy
+	if mkt.LiquidationStrategy != nil {
+		if ls, err = LiquidationStrategyFromProto(mkt.LiquidationStrategy); err != nil {
+			return nil, err
+		}
+	}
```
jeremyletang, line 1008: Will nil be good on a protocol upgrade? Should have default here?

core/types/governance_new_market.go
```diff
@@ -339,6 +348,12 @@ func NewMarketConfigurationFromProto(p *vegapb.NewMarketConfiguration) (*NewMark
 	if err != nil {
 		return nil, fmt.Errorf("error getting new market configuration from proto: %w", err)
 	}
+	var liqStrat *LiquidationStrategy
+	if p.LiquidationStrategy != nil {
+		if liqStrat, err = LiquidationStrategyFromProto(p.LiquidationStrategy); err != nil {
+			return nil, fmt.Errorf("error getting the liquidation strategy from proto: %w", err)
+		}
+	}
```
jeremyletang, line 356: Same thing re protocol upgrade.

core/types/market.go
EVODelavega: I'm ensuring nil will be acceptable. It kind of has to be until the actual functionality is implemented anyways. This is just part of the feature in this PR (to prevent a massive PR that will be a pain to maintain/review/merge)

core/types/governance_new_market.go
EVODelavega: same answer, too

commands/proposal_submission.go
EVODelavega: No, the spec says that the range of this parameter is 0-max, so no validation required.

datanode/gateway/graphql/schema.graphql
```diff
@@ -4583,6 +4590,17 @@ type LiquiditySLAParameters {
   slaCompetitionFactor: String!
 }
 
+type LiquidationStrategy {
+  "Specifies the time in which the network will try to unload its position"
```
candida-d, line 4594:
```suggestion
  "Specifies the time at which the network will try to unload its position"
```

candida-d: Is this in a specific timestamp format?

datanode/gateway/graphql/schema.graphql
```diff
@@ -4583,6 +4590,17 @@ type LiquiditySLAParameters {
...
+  disposalTimeStep: Int!
+  "Specifies the fraction of its position the network will try to reduce its position by in a single disposal attempt."
+  disposalFraction: String!
+  "Specifies the size of the position held by the network that it will attempt to dispose of in one go."
```
candida-d, line 4598:
```suggestion
  "Specifies the size of the position held by the network that it will try to dispose of in one attempt."
```

protos/sources/vega/markets.proto
```diff
@@ -389,3 +391,15 @@ message MarketTimestamps {
   // Time when the market closed.
   int64 close = 4;
 }
+
+// Liquidation strategy used when the network holds a position resulting from position resolution.
+message LiquidationStrategy {
+  // Period of time over which the network will attempt to close its position.
+  int64 disposal_time_step = 1;
+  // Fraction of the open position the market will attempt to close in a single attempt; range 0 through 1.
```
candida-d, line 399:
```suggestion
  // Fraction of the open position the market will try to close in a single attempt; range 0 through 1.
```

protos/sources/vega/markets.proto
```diff
@@ -389,3 +391,15 @@ message MarketTimestamps {
...
+  string disposal_fraction = 2;
+  // Size of the position that the network will attempt to close in a single attempt.
```
candida-d, line 401:
```suggestion
  // Size of the position that the network will try to close in a single attempt.
```

datanode/gateway/graphql/schema.graphql
EVODelavega: It's an int, number of seconds. It's not really _when_ something happens, but rather the period over which something happens. Combining all the parameters in this type, every block over the next N seconds, the network, if it has a non-zero position, will attempt to reduce its position by `total position * disposal_fraction` or `full_disposal_size` (if the position <= full disposal size) _or_ `order_book_volume * max_fraction_consumed`

So with a block time of 1s, a disposal time step of 20s, you could see the network place orders for the next 20 blocks.

protos/sources/vega/markets.proto
```diff
@@ -389,3 +391,15 @@ message MarketTimestamps {
   // Time when the market closed.
   int64 close = 4;
 }
+
+// Liquidation strategy used when the network holds a position resulting from position resolution.
+message LiquidationStrategy {
+  // Period of time over which the network will attempt to close its position.
```
candida-d, line 397:
```suggestion
  // Period of time over which the network will attempt to close its position, in seconds.
```

commands/proposal_submission.go
```diff
@@ -975,6 +977,29 @@ func checkPriceMonitoring(parameters *protoTypes.PriceMonitoringParameters, pare
...
+	errs := NewErrors()
+	if params == nil {
+		// @TODO these will be required, in that case the check for nil should be removed
+		// or return an error.
+		return errs
+	}
+	dispFrac, err := num.DecimalFromString(params.DisposalFraction)
+	if err != nil || dispFrac.IsNegative() || dispFrac.IsZero() || dispFrac.GreaterThan(num.DecimalZero()) {
```
wwestgarth, line 988: surely?
```suggestion
	if err != nil || dispFrac.IsNegative() || dispFrac.IsZero() || dispFrac.GreaterThan(num.DecimalOne()) {
```

could you add some unittests for this validation please?

commands/proposal_submission.go
```diff
@@ -975,6 +977,29 @@ func checkPriceMonitoring(parameters *protoTypes.PriceMonitoringParameters, pare
...
+		errs.AddForProperty(fmt.Sprintf("%s.disposal_fraction", parent), ErrMustBeBetween01)
+	}
+	maxFrac, err := num.DecimalFromString(params.MaxFractionConsumed)
+	if err != nil || maxFrac.IsNegative() || maxFrac.IsZero() || maxFrac.GreaterThan(num.DecimalZero()) {
+		errs.AddForProperty(fmt.Sprintf("%s.max_fraction_consumed", parent), ErrMustBeBetween01)
+	}
+	if params.DisposalTimeStep < 1 {
+		errs.AddForProperty(fmt.Sprintf("%s.disposal_time_step", parent), ErrMustBePositiveOrZero)
```
wwestgarth, line 996: Is `ErrMustBePositiveOrZero` correct here?

commands/proposal_submission.go
EVODelavega: Yup, silly me

commands/proposal_submission.go
EVODelavega: Not anymore, fixed

candida-d (dismissed):

gordsport (approved):

jeremyletang (approved):

candida-d (approved):

PR #10028, opened 2023-11-08
Stop orders linked to order or position

### LINKED ISSUE #9948: Position/order linked stop orders
This issue is to implement the changes to stop orders and laid out in the below spec changes:

- Attach stop to order (volume “activates” as order trades, unactivated is cancelled if order no longer on book or parked)
- Attach stop to position (volume grows/shrinks with position, cancelled if position 0 or changes direction)

Specs:

- https://github.com/vegaprotocol/specs/issues/2025

peterbarrow:
Closes #9948


core/execution/future/market.go
```diff
@@ -3581,10 +3581,32 @@ func (m *Market) submitStopOrders(
...
+							logging.Error(err))
+					}
+					v.OrderSubmission.Size = order.Size - order.TrueRemaining()
+
+					// Need to cancel the referenced order
+					toDelete = append(toDelete, order)
+				} else {
+					// Update the order size to match that of the party's position
+					pos, _ := m.position.GetPositionByPartyID(v.Party)
+					v.OrderSubmission.Size = uint64(pos.Size())
```
jeremyletang, line 3606: I realise that this is not finished, but pos.Size() can be negative here if the party is short which is not gonna play nice with the orderSubmission. Also, aren't we suppose to ensure that we are trying to reduce the position in this case too?

core/execution/future/market.go
```diff
@@ -1821,11 +1821,13 @@ func (m *Market) triggerStopOrders(
 	if m.lastTradedPrice == nil {
 		return nil
 	}
-
 	lastTradedPrice := m.priceToMarketPrecision(m.getLastTradedPrice())
 	triggered, cancelled := m.stopOrders.PriceUpdated(lastTradedPrice)
 
-	if len(triggered) <= 0 {
+	// See if there are any linked orders that are the wrong direction
+	cancelled = append(cancelled, m.stopOrders.CheckDirection(m.position)...)
```
jeremyletang, line 1828: Why are we doing this here?

core/execution/future/market.go
```diff
@@ -3364,10 +3370,54 @@ func (m *Market) submitStopOrders(
...
+				order, _, err := m.getOrderByID(v.SizeOverrideValue.OrderID)
+				if err != nil {
+					m.log.Error("could not get order to override size with",
+						logging.StopOrderSubmission(v),
+						logging.Error(err))
+				}
+				v.OrderSubmission.Size = order.Size - order.TrueRemaining()
+
+				// Need to cancel the referenced order
+				toDelete = append(toDelete, order)
```
jeremyletang, line 3390: We ignore the parameter returned which says if the order was in the book or not, however I don't think we should be submitting the order if it's not traded right? Also by that point, if the order has traded fully and available neither from the book or the pegged list, this will return nothing, we need for every order which is linked to a stop order to keep track of the actual volume which has been traded, and execute against this instead.

core/execution/future/market.go
```diff
@@ -3382,10 +3432,14 @@ func (m *Market) submitStopOrders(
...
 			}
 		}
-
 		evts = append(evts, events.NewStopOrderEvent(ctx, v))
 	}
 
+	// Remove any referenced orders
+	for _, order := range toDelete {
+		m.CancelOrder(ctx, order.Party, order.ID, order.ID)
+	}
```
jeremyletang, line 3441: Do we really need to do this?

core/execution/future/market.go
peterbarrow: I'm checking to see if the position has flipped sign every time we evaluate if we have a stop order to process. If felt right to process them both at the same place. Let me know if you have a better place for it.

core/execution/stoporders/stop_orders.go
```diff
@@ -107,13 +108,52 @@ func (p *Pool) Settled() []*types.StopOrder {
...
 
+func (p *Pool) CheckDirection(positions *positions.SnapshotEngine) []*types.StopOrder {
+	toCancel := []*types.StopOrder{}
+	for partyID, orders := range p.orders {
+		pos, ok := positions.GetPositionByPartyID(partyID)
+		if !ok {
+			continue
+		}
+		for _, order := range orders {
+			if order.SizeOverrideSetting == types.StopOrderSizeOverrideSettingOrder ||
```
jeremyletang, line 119: Shouldn't this apply only for the setting Postion?

core/execution/future/market.go
peterbarrow: That's what the spec says:
"If set to `Order` (which should be an object containing specification for a particular order ID), triggering should override the contained order's size with however much of the referenced order has traded. Additionally, the referenced order should have any remaining unmatched portion cancelled."

core/execution/future/market.go
peterbarrow: I think if the referenced order is filled then the triggered order should do nothing as the purpose of it was to unwind a non complete order. But I'd like @TomMcL to answer this one fully.

core/execution/future/market.go
jeremyletang: alright, didn't remembered that,

core/execution/future/market.go
peterbarrow: Linked orders no longer supported, closing.

protos/sources/vega/vega.proto
```diff
@@ -57,6 +70,8 @@ message StopOrder {
     REJECTION_REASON_STOP_ORDER_NOT_ALLOWED_WITHOUT_A_POSITION = 5;
     // This stop order does not close the position
     REJECTION_REASON_STOP_ORDER_NOT_CLOSING_THE_POSITION = 6;
+    // The percentage value for the linked order is invalid
```
candida-d, line 73:
```suggestion
    // The percentage value for the linked stop order is invalid
```

candida-d (commented):
Very minor vega.proto docstring change

candida-d (dismissed):

wwestgarth:
Can you add some validation here please: https://github.com/vegaprotocol/vega/blob/develop/commands/stop_orders_submission.go#L31 to make sure that the new `string` is a valid decimal and within the expected ranges?

protos/sources/vega/vega.proto
```diff
@@ -7,6 +7,19 @@ import "vega/markets.proto";
 option go_package = "code.vegaprotocol.io/vega/protos/vega";
 
 message StopOrder {
+  message SizeOverrideValue {
+    optional string percentage = 1;
+  }
```
jeremyletang, line 12: Why is that wrapped into another objet? Not a big deal but we end up with an optional message containing an optional value? How about use a oneOff instead?

jeremyletang (commented):
There's no datanode support for this for now, is that coming in a separate PR?

PR #10047, opened 2023-11-10
Support filtering on epoch range on transfers

ValentinTrinque:


candida-d (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1727,6 +1727,10 @@ message ListTransfersRequest {
   optional Pagination pagination = 3;
   // Restrict transfers to those related to reward transfers.
   optional bool is_reward = 4;
+  // Epoch to filter from (included). If omitted, `to epoch` must not be set. If both omitted, the most recent epoch's data is returned.
```
wwestgarth, line 1730: Any reason why this doesn't work like any of the "range" filters in other API, and allow a `to` to be set without a `from` and you just get everything from `0` up to the `to`?

ValentinTrinque: We made this choice because we couldn't find any reason to restrict this.

gordsport (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1730,8 +1730,12 @@ message ListTransfersRequest {
   TransferDirection direction = 2;
   // Pagination controls.
   optional Pagination pagination = 3;
-  // Restrict transfers to those related to reward transfers.
+  // Restrict transfers to athose related to reward transfers.
```
candida-d, line 1733:
```suggestion
  // Restrict transfers to those related to reward transfers.
```

candida-d: Tried to sneak this one by... ;)

guoguojin (dismissed):

MadalinaRaicu (dismissed):

candida-d (approved):

guoguojin (approved):

PR #10066, opened 2023-11-13
9945 liquidation strat

### LINKED ISSUE #9945: [Liquidation improvements] Completely decouple MTM, margin checks and closeout trades

EVODelavega:
Closes #9945


EVODelavega:
```
{
    "system_tests":"support-liquidation-strategy"
}
```

EVODelavega:
ST run: https://jenkins.vega.rocks/job/common/job/system-tests/5863/

candida-d (dismissed):

candida-d (dismissed):

ValentinTrinque (dismissed):

Jiajia-Cui (dismissed):
feature tests look good

candida-d (dismissed):

TomMcL (commented):
Working my way through, tests look reasonable so far, could we kill all the `debug` lines though?

EVODelavega:
> Working my way through, tests look reasonable so far, could we kill all the `debug` lines though?

They will be removed, along with most of the tags I've added. The reason why I'm keeping them is because this isn't the full implementation yet, and they're acting as a kind of bookmark around the things I'm going to be working on next

TomMcL (dismissed):

core/collateral/engine.go
```diff
@@ -2046,15 +2052,26 @@ func (e *Engine) getFeeTransferRequest(
...
-		generalID, err := e.CreatePartyGeneralAccount(ctx, t.Owner, assetID)
+	var (
+		general *types.Account
+		err     error
+	)
+	if t.Owner == types.NetworkParty {
+		general, err = e.GetMarketInsurancePoolAccount(marketID, assetID)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("no insurance pool for the market %w", err)
```
jeremyletang, line 2062: in what condition is that possible?

EVODelavega: It shouldn't be, the error was being returned already, I just made the error message more explicit in case we ever encounter it.

core/execution/future/liquidation.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil
+	}
+	m.broker.Send(events.NewOrderEvent(ctx, order))
+	m.handleConfirmationPassiveOrders(ctx, conf)
+
+	// no trades...
+	if len(conf.Trades) == 0 {
+		m.checkForReferenceMoves(ctx, conf.PassiveOrdersAffected, false)
+		return nil
+	}
```
jeremyletang, line 39: how come the references prices could have moved if there was not trade? I'm not sure what's the point of that? Also no passive order would have been affected?

core/execution/future/liquidation.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		m.log.Panic("Failure after submitting order to matching engine",
+			logging.Order(order),
+			logging.Error(err),
+		)
+	}
+	// no order was submitted
+	if conf == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
EVODelavega: remnants from the time where I submitted the order in the liquidation engine I guess

core/execution/future/liquidation.go
```diff
@@ -0,0 +1,99 @@
...
+	m.broker.Send(events.NewOrderEvent(ctx, order))
+	m.handleConfirmationPassiveOrders(ctx, conf)
+
+	// no trades...
+	if len(conf.Trades) == 0 {
+		m.checkForReferenceMoves(ctx, conf.PassiveOrdersAffected, false)
+		return nil
+	}
+	// transfer fees to the good party -> fees are now taken from the insurance pool
+	fees, _ := m.fee.GetFeeForPositionResolution(conf.Trades)
```
jeremyletang, line 41: No errors to check fdor here?

core/execution/future/liquidation.go
```diff
@@ -0,0 +1,99 @@
...
+	tresps, err := m.collateral.TransferFees(ctx, m.GetID(), m.settlementAsset, fees)
+	if err != nil {
+		// this is the current behaviour, even if this fails, we need to check for reference moves
+		// FIXME(): we may figure a better error handling in here
+		m.checkForReferenceMoves(ctx, conf.PassiveOrdersAffected, false)
+		m.log.Error("unable to transfer fees for positions resolutions",
+			logging.Error(err),
+			logging.String("market-id", m.GetID()))
+		return err
+	}
```
jeremyletang, line 51: Should we actually return here without doing anything else? that looks wrong, seeing that previously we already matched with the orderbook, and produce trades, skipping all that's after if we were just not able to produce transfer the fees seems wrong.

core/execution/future/liquidation.go
EVODelavega: it doesn't return an error, it returns the fee object itself. Right now we don't need it, but I'm keeping it in case it proves handy for the next PR (PnL for network position)

core/execution/liquidation/engine.go
```diff
@@ -0,0 +1,307 @@
...
+// once that's happened, this code can be removed.
+func GetDefaultStrategy() *types.LiquidationStrategy {
+	return defaultStrat.DeepClone()
+}
+
+// GetLegacyStrat is exported, same as defaul. This can be used for protocol upgrade
+// it most closely resebles the old behaviour (network attempts to close out fully, in one go)
+// this can be removed once protocol upgrade has completed.
+func GetLegacyStrat() *types.LiquidationStrategy {
+	return legacyStrat.DeepClone()
```
jeremyletang, line 89: GetDefaultStrategy, GetLegacyStrat, gotta love consistency.

core/execution/future/liquidation.go
EVODelavega: The only reason `TransferFees` would return an error is if we are missing accounts, or were otherwise unable to update the balances, if that's the case, we will run in to the same error (most likely) when performing the MTM settlement after handling the trades. I suppose it's better to replace this with a panic though

core/execution/liquidation/engine.go
EVODelavega: These are gone next PR

core/execution/liquidation/engine.go
```diff
@@ -0,0 +1,307 @@
...
+		mps = append(mps, cp)
+		parties = append(parties, cp.Party())
+	}
+	// send order events
+	e.broker.SendBatch(orders)
+	// send trade events
+	e.broker.SendBatch(trades)
+	// the network has no (more) remaining open position -> no need for the e.nextStep to be set
+	if e.pos.open == 0 {
+		e.nextStep = time.Time{}
```
jeremyletang, line 213: Don't know if this goes to the snapshot, but it doesn't marshal well if I remember properly.

core/execution/liquidation/engine.go
```diff
@@ -0,0 +1,307 @@
...
+	order := types.Order{
+		ID:            idgen.NextID(),
+		MarketID:      e.mID,
+		Status:        types.OrderStatusFilled,
+		Party:         types.NetworkParty,
+		Price:         price,
+		OriginalPrice: dpPrice,
+		CreatedAt:     now.UnixNano(),
+		Reference:     "close-out distressed",
+		TimeInForce:   types.OrderTimeInForceFOK, // this is an all-or-nothing order, so TIME_IN_FORCE == FOK
```
jeremyletang, line 261: Isn't the network order supposed to be IOC now?

jeremyletang: eh I'm stupid that's the order with the distressed party.

EVODelavega: The order the network places to close its own position, yes, and that's an IOC order. These are the "fake" orders that populate the order/trade events for the closeout of the distressed parties.

core/fee/engine_test.go
```diff
@@ -77,7 +76,6 @@ func TestFeeEngine(t *testing.T) {
 	t.Run("calculate batch auction trading fee empty trade", testCalcBatchAuctionTradingErrorEmptyTrade)
 	t.Run("calculate batch auction trading fee same batch", testCalcBatchAuctionTradingSameBatch)
 	t.Run("calculate batch auction trading fee different batches", testCalcBatchAuctionTradingDifferentBatches)
-	t.Run("calculate position resolution", testCalcPositionResolution)
```
jeremyletang, line 80: was there no chances to port that tests / write a new one instead of remove it completely?

core/execution/liquidation/engine.go
EVODelavega: Haven't encountered any issues TBH, but I believe they're converted to `UnixNano()` for the marshalling

core/fee/engine_test.go
EVODelavega: Not really. The old fee calculation calculated the amounts per distressed party (pro-rated WRT how much of the fees they actually have to end up paying). After checking with research, we're now paying the whole fee from the insurance pool, so close to none of the logic/calculations we were testing here remain

jeremyletang: Doesn't mean it's not worth a test for the whole bunch of logic added / moved around.

EVODelavega: Tests are already in the pipeline on next PR branch

TomMcL (dismissed):

TomMcL:
Approved but with a follow-up query. How would this play with current mainnet if we have to deploy it? Will current markets act similarly to how they do now or is it going to start using this network owned position stuff?

candida-d (dismissed):

TomMcL (approved):

Jiajia-Cui (approved):
feature tests look good

candida-d (approved):

core/collateral/engine.go
```diff
@@ -2176,11 +2181,16 @@ func (e *Engine) getFeeTransferRequest(
...
 		treq.ToAccount = []*types.Account{insurancePool}
-		return treq, nil
-
 	default:
 		return nil, ErrInvalidTransferTypeForFeeRequest
 	}
+	// we may be moving funds from the insurance pool, we cannot have more than 1 from account in that case
+	// because once the insurance pool is drained, and a copy of the same account without the updated balance
+	// sits in the FromAccount slice, we are magically doubling the available insurance pool funds.
+	if len(treq.FromAccount) > 0 && treq.FromAccount[0].Type == types.AccountTypeInsurance {
```
wwestgarth, line 2190: I don't like the sound of magically doubling funds. Should this not just be a panic if a configuration that shouldn't happen, instead of just fudging and removing the other accounts from the slice?

Feel like there should also be a unittest for this case to make sure we *definitely do not* double anything.

core/execution/future/liquidation.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"code.vegaprotocol.io/vega/core/events"
+	"code.vegaprotocol.io/vega/core/types"
+	"code.vegaprotocol.io/vega/libs/num"
+	"code.vegaprotocol.io/vega/logging"
+)
+
+func (m *Market) checkNetwork(ctx context.Context, now time.Time) error {
+	// debug
+	// this only returns an error if we couldn't get the price range, incidating no orders on book
```
wwestgarth, line 15: Feels a bit fragile that we're relying on the code path all the way into the matching engine to always only return an error when there are no orders on the book. Can we not catch the specific error, or change `OnTick()` to return `nil, nil` in that case so at least _here_ we can handle errors properly?

core/execution/liquidation/engine.go
```diff
@@ -0,0 +1,307 @@
...
+		Party:       types.NetworkParty,
+		Side:        side,
+		Price:       price,
+		Size:        size,
+		Remaining:   size,
+		TimeInForce: types.OrderTimeInForceIOC,
+		Type:        types.OrderTypeLimit,
+		CreatedAt:   now.UnixNano(),
+		Status:      types.OrderStatusActive,
+		Reference:   "LS", // Liquidity sourcing
```
wwestgarth, line 172: Minor: given this will be visible in the data-node we sould just give it a reference thats less cryptic like "liquidity-sourcing".

core/execution/future/liquidation.go
EVODelavega: This is fine here, in part because this isn't yet the final form of the liquidation implementation. The liquidation engine will do more heavy lifting come the next PR. I agree that this has the whiff of fragile code, but it's not intended to survive the next PR

core/execution/liquidation/engine.go
EVODelavega: I think I copied this from the existing code, comment and all, so I was reluctant to change it, given that this reference is already in use for position resolution

core/execution/liquidation/snapshot.go
```diff
@@ -0,0 +1,91 @@
...
+
+	switch d := pl.Data.(type) {
+	case *types.LiquidationNode:
+		e.mID = d.MarketID
+		e.pos.open = d.NetworkPos
+		e.nextStep = d.NextStep
+		if d.Config != nil {
+			e.cfg = d.Config.DeepClone()
+		} else {
+			// @NOTE this can be removed after protocol upgrade has completed
```
wwestgarth, line 60: Minor: I don't think we'd every hit this because they'll be no payload in the protocol-upgrade snapshot for us to load from. I think we're cover by the bit in `NewMarketFromSnapshot()` though, so thats good.

EVODelavega: yes, the places where nil values are being replaced with the legacy strat is overkill ATM. Part of the upcoming PR is requiring the liquidation strategy to be set in the proposals, and then I'm planning to go through all these bits and bobs, and remove dead code.

core/governance/engine.go
```diff
@@ -1240,6 +1241,10 @@ func (e *Engine) updatedMarketFromProposal(p *proposal) (*types.Market, types.Pr
 	default:
 		return nil, types.ProposalErrorUnsupportedProduct, ErrUnsupportedProduct
 	}
+	// assign the current liquidation strategy if none was set on the update proposal
```
ze97286, line 1244: how come we don't have this in new market (this is update market I think)?

EVODelavega: For this version, the liquidation strategy is still optional, the next PR will change the governance validation and all more, and liquidation strategy will no longer be optional then

ze97286: it will still be optional for the migration release, i.e. for proposals for new markets/update markets that will exist at the time of the migration.

EVODelavega: I've talked about this with David, he's leaning towards setting the liquidation strategy to some default value on migration.

wwestgarth (approved):

ze97286 (approved):

PR #10096, opened 2023-11-14
feat: add isolated margin support

### LINKED ISSUE #8295: [Feature]: Segregating/separating margin
This issue is to review the spec update for Segregating/separating margin covered in the following spec task and either break the change up into deliverable issues or implement under this issue

- [Margin Isolation](https://github.com/vegaprotocol/specs/issues/1627)

### LINKED ISSUE #9986: [Feature]: Add support for update margin mode in batch transaction
# Feature Overview

The transaction to update/change margin mode can be included in a batch transaction in order to allow updates when placing an order.

## Specs
- [[Spec](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0019-MCAL-margin_calculator.md#setting-margin-mode)]

### LINKED ISSUE #9987: [Feature]: Implement transition between margin mode and margin calculations for isolated margin
# Feature Overview
Add support for isolated margin and for handling the swapping between the modes in the futures market.

## Specs

## Tasks
- [Implement the logic for switching between margin modes  ]
- [Implement the order and position handling for isolated margin mode in future market]
- [For all margin feature tests add corresponding FTs for isolated margin and for switching between modes]

### LINKED ISSUE #10049: [Feature]: Add isolated margin support to feature tests
# Feature Overview

**In order to** be able to switch to/from isolated margin
**We will** add a step for the transaction
**So that** party can toggle between margin states

There may be other steps required for full support of isolated margin in integration tests.

### LINKED ISSUE #10095: [Feature]: Implement isolated margin
See spec [[0019-MCAL-margin_calculator.md](https://github.com/vegaprotocol/specs/pull/2055/files#diff-f1a21254aefda4db070ab08a262571555ce4bf07745077dee278de11fb7db32b)](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0019-MCAL-margin_calculator.md)

ze97286:
development:
Closes #10095 
close #10049 
close #8295 
close #9986 
close #9987

Test:
close https://github.com/vegaprotocol/core-test-coverage/issues/429
close https://github.com/vegaprotocol/core-test-coverage/issues/430
close https://github.com/vegaprotocol/core-test-coverage/issues/431
close https://github.com/vegaprotocol/core-test-coverage/issues/477
close https://github.com/vegaprotocol/core-test-coverage/issues/435
close https://github.com/vegaprotocol/core-test-coverage/issues/436
close https://github.com/vegaprotocol/core-test-coverage/issues/432
close https://github.com/vegaprotocol/core-test-coverage/issues/478
close https://github.com/vegaprotocol/core-test-coverage/issues/479
close https://github.com/vegaprotocol/core-test-coverage/issues/483
close https://github.com/vegaprotocol/core-test-coverage/issues/480
close https://github.com/vegaprotocol/core-test-coverage/issues/481
close https://github.com/vegaprotocol/core-test-coverage/issues/482
close https://github.com/vegaprotocol/core-test-coverage/issues/484
close https://github.com/vegaprotocol/core-test-coverage/issues/500
close https://github.com/vegaprotocol/core-test-coverage/issues/447
close https://github.com/vegaprotocol/core-test-coverage/issues/449
close https://github.com/vegaprotocol/core-test-coverage/issues/418
close https://github.com/vegaprotocol/core-test-coverage/issues/448
close https://github.com/vegaprotocol/core-test-coverage/issues/501
close https://github.com/vegaprotocol/core-test-coverage/issues/502
close https://github.com/vegaprotocol/core-test-coverage/issues/503
close https://github.com/vegaprotocol/core-test-coverage/issues/508
close https://github.com/vegaprotocol/core-test-coverage/issues/509
close https://github.com/vegaprotocol/core-test-coverage/issues/510
close https://github.com/vegaprotocol/core-test-coverage/issues/511
close https://github.com/vegaprotocol/core-test-coverage/issues/512
close https://github.com/vegaprotocol/core-test-coverage/issues/513


core/processor/batch_market_instructions_processor.go
```diff
@@ -124,10 +131,28 @@ func (p *BMIProcessor) ProcessBatch(
...
 
+	failedMarkets := map[string]error{}
+	for _, umm := range batch.UpdateMarginMode {
+		err := p.validator.CheckUpdateMarginMode(umm)
+		if err == nil {
+			var marginFactor num.Decimal
+			if umm.MarginFactor == nil || len(*umm.MarginFactor) == 0 {
+				marginFactor = num.DecimalZero()
+			} else {
+				marginFactor = num.MustDecimalFromString(*umm.MarginFactor)
```
jeremyletang, line 142: MustDecimal here will panic with a malformed string isn't it?
```suggestion
				marginFactor = num.MustDecimalFromString(*umm.MarginFactor)
```

core/processor/batch_market_instructions_processor.go
```diff
@@ -124,10 +131,28 @@ func (p *BMIProcessor) ProcessBatch(
 	// these need to be determinitistic
 	idgen := idgeneration.New(determinitisticID)
 
+	failedMarkets := map[string]error{}
+	for _, umm := range batch.UpdateMarginMode {
```
jeremyletang, line 135: Should we add some sort of validation to ensure that at most 1 update margin mode is submitted, per market?

ze97286: idk, is there a requirement for that?

core/execution/future/isolated_margin.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, err
+	}
+	price := m.getMarketObservable(order.Price.Clone())
+	increment := m.tradableInstrument.Instrument.Product.GetMarginIncrease(m.timeService.GetTimeNow().UnixNano())
+	return m.risk.UpdateIsolatedMarginsOnPositionChange(ctx, pos, price, increment, m.matching.GetOrdersPerParty(order.Party), []*types.Trade{trade}, order.Side, m.getMarginFactor(order.Party))
+}
+
+func (m *Market) getIsolatedMarginContext(mpos *positions.MarketPosition, order *types.Order) (*num.Uint, events.Margin, num.Decimal, *num.Uint, num.Decimal, []*types.Order, error) {
+	var orderPrice *num.Uint
+	if order != nil {
```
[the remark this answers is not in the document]
ze97286: multiple, but essentially when it is not called in the context of any order. E.g. at the end of uncrossing an auction when we iterate all the parties in isolated margin and release excess margin for orders they placed during the auction

core/execution/future/market.go
```diff
@@ -3986,5 +4308,113 @@ func (m *Market) GetRiskFactors() *types.RiskFactor {
...
+// order margin account. Add this amount to the difference calculated in step 1. If this amount is less than or equal to the
+// amount in the general account, perform the transfers (first move funds into/out of margin account, then move funds into
+// the order margin account). If there are insufficient funds, reject the transaction.
+// 3. Move account to isolated margin mode on this market
+//
+// When switching from isolated margin mode to cross margin mode, the following steps will be taken:
+// 1. Any funds in the order margin account will be moved to the margin account.
+// 2. At this point trading can continue with the account switched to the cross margining account type.
+// If there are excess funds in the margin account they will be freed at the next margin release cycle.
+func (m *Market) switchMarginMode(ctx context.Context, party string, marginMode types.MarginMode, marginFactor num.Decimal) error {
```
jeremyletang, line 4349: would be useful for the frontend here to add some sort of event, indicating that the party is changing marging mode.

Also to store in the datanode: the party, margin mode, and eventual factor. that'll be useful for UIs to know what's the user selected for margin mode. I realise that it's also set in the margin level, but if the user has never traded, the frontend won't be able to know using that.

core/execution/future/market_snapshot.go
```diff
@@ -294,6 +301,16 @@ func (m *Market) GetState() *types.ExecMarket {
 	sort.Strings(parties)
 	assetQuantum, _ := m.collateral.GetAssetQuantum(m.settlementAsset)
 
+	partyMarginFactors := make([]*snapshot.PartyMarginFactor, 0, len(m.partyMarginFactor))
+	for k, d := range m.partyMarginFactor {
+		if !d.IsZero() {
```
jeremyletang, line 306: In what situation could this be 0?

core/execution/future/market.go
ze97286: I don't understand. it's already there, in margin levels.

core/execution/future/market_snapshot.go
ze97286: in some situation I forgot to remove it from the map.

core/positions/market_position.go
```diff
@@ -71,6 +71,22 @@ func (p *MarketPosition) Closed() bool {
 	return p.size == 0 && p.buy+p.sell == 0
 }
 
+func (p *MarketPosition) UpdateInPlaceOnTrades(log *logging.Logger, traderSide types.Side, trades []*types.Trade, order *types.Order) *MarketPosition {
```
jeremyletang, line 74: In what context is that used, the naming suggest we are updating something, but it returns a clone, is that expected, could lead to some mistake if for the naming only in the future.

Sohill-Patel:
```{"system_tests":"0019-MCAL-092"}```

File: protos/sources/vega/vega.proto [hunk 7]
```diff
@@ -478,6 +478,10 @@ enum OrderError {
   ORDER_ERROR_POST_ONLY_ORDER_WOULD_TRADE = 49;
   // Post order would trade
   ORDER_ERROR_REDUCE_ONLY_ORDER_WOULD_NOT_REDUCE_POSITION = 50;
+  // Isolated margin check failed
+  ORDER_ERROR_ISOLATED_MARGIN_CHECK_FAILED = 51;
```
wwestgarth, line 482: I think you might need to add these to the `OrderRejectionReason` enum in the graphQL schema.

protos/sources/vega/vega.proto
```diff
@@ -1154,6 +1166,12 @@ message MarginLevels {
...
   int64 timestamp = 8;
+  // Margin required to cover orders in isolated margin mode.
+  string order_margin = 9;
+  // Margin mode for the party, cross margin or isolated margin.
+  MarginMode margin_mode = 10;
+  // Margin factor, relevant only for isolated margin, empty otherwise.
+  string margin_factor = 11;
 }
 
 // Represents market data specific to a perpetual market.
```
wwestgarth, line 1177: Not part of this PR, but I've spotted it so I'll comment:
the new account type `ACCOUNT_TYPE_ORDER_MARGIN` needs to be added to the big list in the API that exports ledger entries `func (ls *Ledger) Export()`.

core/collateral/engine.go
```diff
@@ -4390,6 +4519,28 @@ func (e *Engine) TransferSpot(ctx context.Context, partyID, toPartyID, asset str
 	return res, nil
 }
 
+func (e *Engine) GetOrCreateOrderPartyMarginAccount(ctx context.Context, partyID, marketID, asset string) (string, error) {
```
jeremyletang, line 4522: GetOrCreatePartyOrderMarginAccount

core/execution/future/isolated_margin.go
```diff
@@ -0,0 +1,76 @@
...
+	orders := m.matching.GetOrdersPerParty(order.Party)
+	marginFactor := m.getMarginFactor(order.Party)
+	r, err := m.risk.UpdateIsolatedMarginsOnPositionChange(ctx, pos, price, increment, orders, []*types.Trade{trade}, order.Side, marginFactor)
+	if err == nil {
+		for _, rr := range r {
+			m.transferMargins(ctx, []events.Risk{rr}, nil)
+		}
+	} else {
+		return err
+	}
```
jeremyletang, line 43:
```suggestion
	r, err := m.risk.UpdateIsolatedMarginsOnPositionChange(ctx, pos, price, increment, orders, []*types.Trade{trade}, order.Side, marginFactor)
	if err != nil {
		return err
	}
	for _, rr := range r {
		m.transferMargins(ctx, []events.Risk{rr}, nil)
	}
```

core/execution/future/market.go
```diff
@@ -1355,6 +1360,24 @@ func (m *Market) uncrossOnLeaveAuction(ctx context.Context) ([]*types.OrderConfi
...
+
+	// after auction uncrossing we can relax the price requirement and release some excess order margin if any was placed during an auction.
+	for k, d := range m.partyMarginFactor {
+		if !d.IsZero() {
+			partyPos, _ := m.position.GetPositionByPartyID(k)
+			if partyPos != nil && (partyPos.Buy() != 0 || partyPos.Sell() != 0) {
+				marketObservable, mpos, increment, _, _, orders, err := m.getIsolatedMarginContext(partyPos, nil)
+				if err != nil {
+					continue
+				}
```
jeremyletang, line 1372: Could anything happen here if we are just ignoring this error?

core/execution/future/market.go
```diff
@@ -2650,12 +2839,32 @@ func (m *Market) collateralAndRisk(ctx context.Context, settle []events.Transfer
...
+		mrgns, err := m.risk.CheckMarginInvariants(ctx, evt, m.getMarketObservable(nil), increment, m.matching.GetOrdersPerParty(evt.Party()), m.getMarginFactor(evt.Party()))
+		if err == risk.ErrInsufficientFundsForMaintenanceMargin {
+			m.log.Debug("party in isolated margin mode has insufficient margin", logging.String("party", evt.Party()))
+			isolatedMarginPartiesToClose = append(isolatedMarginPartiesToClose, mrgns)
+		}
+	}
+
+	if len(crossRiskUpdates) == 0 {
+		return nil, isolatedMarginPartiesToClose
+	}
```
jeremyletang, line 2866: doesn't matter but not sure why we add this check? an empty slice == nil too

core/execution/future/market.go
```diff
@@ -2870,7 +3086,14 @@ func (m *Market) parkOrder(ctx context.Context, orderID string) *types.Order {
...
 	}
 
-	_ = m.position.UnregisterOrder(ctx, order)
+	pos := m.position.UnregisterOrder(ctx, order)
+	if m.getMarginMode(order.Party) == types.MarginModeIsolatedMargin {
+		pos, _ := m.position.GetPositionByPartyID(pos.Party())
+		if err := m.updateIsolatedMarginOnOrder(ctx, pos, order); err != nil {
+			m.log.Panic("failed to update order margin after order park", logging.Order(order), logging.String("party", pos.Party()))
+		}
+	}
```
jeremyletang, line 3095: Is that necessary? seeing that we do not support pegged order with isolated margin anymore, that should not be necessary, or be a panic as we should never end up in that position.

core/execution/future/market.go
```diff
@@ -2889,6 +3112,23 @@ func (m *Market) AmendOrder(
...
 }
 
+// handleIsolatedMarginInsufficientOrderMargin stops all party orders
+// Upon position/order update if there are insufficient funds in the order margin, all open orders are stopped and margin re-evaluated.
+func (m *Market) handleIsolatedMarginInsufficientOrderMargin(ctx context.Context, party string) {
+	orders := m.matching.GetOrdersPerParty(party)
+	for _, o := range orders {
+		if !o.IsFinished() {
+			m.unregisterAndReject(ctx, o, types.OrderErrorIsolatedMarginCheckFailed)
+			m.matching.RemoveOrder(o.ID)
```
jeremyletang, line 3122: Seems like RemoveOrder set the order status to parked, which is unlikely to be what we want here?
Also we may want to send order events?

core/execution/future/market.go
```diff
@@ -2911,6 +3151,9 @@ func (m *Market) AmendOrderWithIDGenerator(
 
 	conf, updatedOrders, err := m.amendOrder(ctx, orderAmendment, party)
 	if err != nil {
+		if m.getMarginMode(party) == types.MarginModeIsolatedMargin && err == common.ErrMarginCheckFailed {
```
jeremyletang, line 3154: So, is when trying to amend 1 order, we now fail the isolated margin check, we are cancelling ALL the orders of the party? that doesn't seounds right to me is it?

core/execution/future/market.go
```diff
@@ -3199,9 +3445,19 @@ func (m *Market) amendOrder(
...
-			// ensure we release excess if party reduced the size of their order
-			m.recheckMargin(ctx, m.position.GetPositionsByParty(amendedOrder.Party))
+			if m.getMarginMode(party) == types.MarginModeCrossMargin {
+				// ensure we release excess if party reduced the size of their order
+				m.recheckMargin(ctx, m.position.GetPositionsByParty(amendedOrder.Party))
+			}
+		}
+		if m.getMarginMode(party) == types.MarginModeIsolatedMargin {
+			pos, _ := m.position.GetPositionByPartyID(amendedOrder.Party)
+			if err := m.updateIsolatedMarginOnOrder(ctx, pos, amendedOrder); err == risk.ErrInsufficientFundsForMarginInGeneralAccount {
```
jeremyletang, line 3455: Same thing here, that sounds weird to me, but this might be per spec, I haven't checked for this specifically.

core/execution/future/market.go
jeremyletang: it's there in margin level, but in the case the party do not have a position or open orders yet, or there's no market observable etc, then it's never sent to the data node and there's no way to expose that to the API. for the frontends we'll need most likely something exposing: market_id, party_id, margin_mode, margin_factor.

We'll also need to send an event for that on the protocol upgrade to default this to cross most likely.

core/execution/future/market_snapshot.go
jeremyletang: lol ok.

core/processor/batch_market_instructions_processor.go
jeremyletang: I don't know, but it would make sense to me.

datanode/gateway/graphql/resolvers.go
```diff
@@ -2404,6 +2404,21 @@ func (r *myMarginLevelsUpdateResolver) MaintenanceLevel(_ context.Context, m *ve
...
+	return m.OrderMargin, nil
+}
+
+func (r *myMarginLevelsUpdateResolver) MarginFactor(_ context.Context, m *vegapb.MarginLevels) (string, error) {
+	return m.MarginFactor, nil
+}
+
+func (r *myMarginLevelsUpdateResolver) MarginMode(_ context.Context, m *vegapb.MarginLevels) (MarginMode, error) {
+	if m.MarginMode == vegapb.MarginMode_MARGIN_MODE_ISOLATED_MARGIN {
+		return MarginModeMarginModeIsolatedMargin, nil
```
jeremyletang, line 2417: We usually map enums with in the gqlgen.yml file instead and add the marshaller in the marshallers package.

core/execution/future/market.go
ze97286: well then if you did check the spec you'd see that:

image

core/execution/future/market.go
ze97286: interesting, because the line above is setting the order to stopped and feature tests are validating that the orders are in state stopped. I suppose swapping the lines may work as desired.

core/execution/future/market.go
ze97286: it's not, I forgot to remove it when I disabled pegged

core/collateral/engine.go
ze97286: indeed

core/execution/future/market.go
ze97286: I don't think it's really possible for an error to occur here so the continue is definitely not a problem. panic here would be a bit too much in my opinion.

protos/sources/vega/vega.proto
ze97286: added

commands/update_margin_mode.go
```diff
@@ -33,7 +33,7 @@ func checkUpdateMarginMode(cmd *commandspb.UpdateMarginMode) Errors {
 		return errs.FinalAddForProperty("update_margin_mode", ErrIsRequired)
 	}
 
-	if cmd.Mode == commandspb.UpdateMarginMode_MODE_CROSS_UNSPECIFIED {
+	if cmd.Mode == commandspb.UpdateMarginMode_MODE_UNSPECIFIED {
 		errs.AddForProperty("update_margin_mode.margin_mode", ErrIsNotValid)
 	}
```
wwestgarth, line 38: Can you add a check that the integer within `cmd.Mode` is within in enum bounds? like:
```
if _, ok := commandspb.UpdateMarginMode_Mode_name[int32(cmd.Mode)]; !ok {
	errs.AddForProperty("update_margin_mode.margin_mode", ErrIsNotValid)
}
```

core/execution/future/market.go
```diff
@@ -3986,5 +4300,131 @@ func (m *Market) GetRiskFactors() *types.RiskFactor {
...
+				m.cancelOrder(ctx, o.Party, o.ID)
+			}
+		}
+		return nil
+	} else {
+		// switching from isolated margin to cross margin
+		// 1. Any funds in the order margin account will be moved to the margin account.
+		// 2. At this point trading can continue with the account switched to the cross margining account type. If there are excess funds in the margin account they will be freed at the next margin release cycle.
+		risk := m.risk.SwitchFromIsolatedMargin(ctx, margins, marketObservable, increment)
+		if err != nil {
```
wwestgarth, line 4409: I don't know what error this is checking, because `SwitchFromIsolatedMargin` isn't returning one.

ze97286: it used to though...

core/execution/future/market_snapshot.go
wwestgarth: This doesn't make sense though, forgetting to remove it from the map doesn't make the value become zero? 

Forgetting to remove it from the map will just leave someone in cross-margin mode with a non-zero margin factor, left as a non-zero margin factor that was set originally which is we get into that state will cause loops like this to go wonky:
```
for k, d := range m.partyMarginFactor {
		if !d.IsZero() {
			partyPos, _ := m.position.GetPositionByPartyID(k)
			if partyPos != nil && (partyPos.Buy() != 0 || partyPos.Sell() != 0) {
				marketObservable, mpos, increment, _, _, orders, err := m.getIsolatedMarginContext(partyPos, nil)
				if err != nil {
					continue
				}
				r := m.risk.ReleaseExcessMarginAfterAuctionUncrossing(ctx, mpos, marketObservable, increment, d, orders)
				if r != nil && r.Transfer() != nil {
					m.transferMargins(ctx, []events.Risk{r}, nil)
				}
			}
		}
	}
```

So I don't think the `if !d.IsZero()` guards are doing anything at all.

ze97286: no. that's not what I meant. what I meant is setting it to zero and still not removing it from the map. Hope it makes sense.

ze97286: but given that it seems to be impossible to set it to 0 in the map (considering it is only set when the margin mode passed to the switch is isolated margin, in which case the validation won't ket it get there with a margin factor of 0), then it seems safe to remove the checks.

wwestgarth (dismissed):

protos/sources/vega/vega.proto
```diff
@@ -1154,6 +1166,12 @@ message MarginLevels {
   string asset = 7;
   // Timestamp in Unix nanoseconds for when the ledger entry was created.
   int64 timestamp = 8;
+  // Margin required to cover orders in isolated margin mode.
+  string order_margin = 9;
+  // Margin mode for the party, cross margin or isolated margin.
+  MarginMode margin_mode = 10;
+  // Margin factor, relevant only for isolated margin, empty otherwise.
```
candida-d, line 1173:
```suggestion
  // Margin factor, relevant only for isolated margin, 0 otherwise.
```

candida-d (approved):

peterbarrow (approved):

wwestgarth (approved):

PR #10101, opened 2023-11-15
Remove constraint on epoch range to list fees stats

ValentinTrinque:


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4180,7 +4180,7 @@ message GetFeesStatsForPartyRequest {
   string party_id = 1;
   // Restrict fees statistics to those related to the given asset.
   optional string asset_id = 2;
-  // Epoch to filter from (included). If omitted, `to epoch` must not be set. If both omitted, the most recent epoch's data is returned.
+  // Epoch to filter from (included). If omitted, the range goes from from the oldest epoch to the `to epoch`.
```
candida-d, line 4183:
```suggestion
  // Epoch to filter from (included). If omitted, the range goes from the oldest epoch to the `to epoch`.
```

candida-d, line 4183: Does this also need to be done in GraphQL?

wwestgarth (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
ValentinTrinque: Good catch.

wwestgarth (approved):

candida-d (approved):

mattrussell36 (approved):

PR #10134, opened 2023-11-17
refactor: fix `CalculateLiquidationPriceWithSlippageFactors` signature

witgaw:
Fix signature of `CalculateLiquidationPriceWithSlippageFactors` the downstream code assumed the return ordering correctly, it's just the signature that was incorrect. Currently it would only be a problem if reutrn value was used despite an error.

### DIFF
--- a/core/risk/liquidation_calculation.go
+++ b/core/risk/liquidation_calculation.go
@@ -28,7 +28,7 @@ type OrderInfo struct {
 	IsMarketOrder bool
 }
 
-func CalculateLiquidationPriceWithSlippageFactors(sizePosition int64, buyOrders, sellOrders []*OrderInfo, currentPrice, collateralAvailable num.Decimal, positionFactor, linearSlippageFactor, quadraticSlippageFactor, riskFactorLong, riskFactorShort, fundingPaymentPerUnitPosition num.Decimal) (liquidationPriceForOpenVolume, liquidationPriceWithSellOrders, liquidationPriceWithBuyOrders num.Decimal, err error) {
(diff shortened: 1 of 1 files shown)


core/risk/liquidation_calculation.go
```diff
@@ -28,7 +28,7 @@ type OrderInfo struct {
 	IsMarketOrder bool
 }
 
-func CalculateLiquidationPriceWithSlippageFactors(sizePosition int64, buyOrders, sellOrders []*OrderInfo, currentPrice, collateralAvailable num.Decimal, positionFactor, linearSlippageFactor, quadraticSlippageFactor, riskFactorLong, riskFactorShort, fundingPaymentPerUnitPosition num.Decimal) (liquidationPriceForOpenVolume, liquidationPriceWithSellOrders, liquidationPriceWithBuyOrders num.Decimal, err error) {
+func CalculateLiquidationPriceWithSlippageFactors(sizePosition int64, buyOrders, sellOrders []*OrderInfo, currentPrice, collateralAvailable num.Decimal, positionFactor, linearSlippageFactor, quadraticSlippageFactor, riskFactorLong, riskFactorShort, fundingPaymentPerUnitPosition num.Decimal) (liquidationPriceForOpenVolume, liquidationPriceWithBuyOrders, liquidationPriceWithSellOrders num.Decimal, err error) {
```
EVODelavega, line 31: why are the args switched?

wwestgarth: Looks like the callers are assuming its buys then sells, but this was returning sells then buys.

wwestgarth (approved):

EVODelavega (approved):

PR #10174, opened 2023-11-24
fix: restore the position resolution event

### LINKED ISSUE #10173: [Bug]: restore the position resolution event
### Problem encountered

When parties are distressed, we used to emit a position resolution event containing the mark price, number of distressed parties, and the number of parties that were closed out (the difference between the two being the parties that were able to keep their position after their orders were closed)

### Observed behaviour

No position resolution event is emitted

### Expected behaviour

The event should be emitted if parties are distressed

### Steps to reproduce

```bash
Run any test that results in parties getting closed out, and capture the events. The event is not there
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #10173


wwestgarth (dismissed):

jeremyletang (approved):

EVODelavega:
Not needed

PR #10176, opened 2023-11-27
0053-PERP-028

Sohill-Patel:
close https://github.com/vegaprotocol/core-test-coverage/issues/470

also renamed `0053-PERP-027_twap_calculation.feature` for consistency


core/integration/features/settlement/0053-PERP-028_twap_calculation_with_auction.feature
```diff
@@ -0,0 +1,156 @@
...
+        Given time is updated to "2020-10-16T00:20:00Z"
+        When the parties place the following orders:
+            | party  | market id | side | volume | price | resulting trades | type       | tif     |
+            | party1 | ETH/DEC19 | buy  | 1      | 30    | 0                | TYPE_LIMIT | TIF_GTC |
+            | party2 | ETH/DEC19 | sell | 1      | 30    | 1                | TYPE_LIMIT | TIF_GTC |
+        # in theory internal TWAP = 9.625 external TWAP = 10.25, if the auction period is excluded
+        # but these are type int so the decimal is truncated
+        Then the product data for the market "ETH/DEC19" should be:
+            | internal twap | external twap |
+            | 9             | 10            |
```
wwestgarth, line 153: Given that these numbers end up being the same as the test where there isn't any auction periods, I think we need to scale up the prices/set-different-decimal-places-in-the-markets so that the TWAP calculations do not lose all the precision.

Sohill-Patel: change the mark/spot price during auction so that if the auction was included in the calculation then the twap values are off by a whole number

wwestgarth (approved):

PR #10178, opened 2023-11-27
fix: validate against large order sizes that may caught internal over…

### LINKED ISSUE #10177: [Bug]: Panic on fairground
### Problem encountered

The below panic occurred on fairgraound:
```
Nov 26 12:13:45 n01.testnet.vega.rocks visor[3758517]: 2023-11-26T12:13:45.470Z        ERROR        tendermint        consensus/state.go:732        CONSENSUS FAILURE!!!        {"err": "cannot unregister order with potential sell + size change < 0", "stack": "goroutine 490 [running]:
runtime/debug.Stack()
    /opt/hostedtoolcache/go/1.20.11/x64/src/runtime/debug/stack.go:24 +0x65
github.com/tendermint/tendermint/consensus.(*State).receiveRoutine.func2()
    /home/runner/go/pkg/mod/github.com/vegaprotocol/cometbft@v0.34.29-patch.1/consensus/state.go:732 +0x4c
panic({0x31776e0, 0xc01e972ef0})
    /opt/hostedtoolcache/go/1.20.11/x64/src/runtime/panic.go:884 +0x213
go.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x2?, 0x2?, {0x0?, 0x0?, 0xc011debc40?})
    /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/zapcore/entry.go:198 +0x65
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc01987a270, {0xc014644f80, 0x2, 0x2})
    /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/zapcore/entry.go:264 +0x3ec
go.uber.org/zap.(*Logger).Panic(0x2?, {0x3b55544?, 0x2?}, {0xc014644f80, 0x2, 0x2})
    /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/logger.go:258 +0x59
code.vegaprotocol.io/vega/core/positions.(*MarketPosition).UpdateOnOrderChange(0xc0122f1090, 0xc00dba7860, 0x8881798?, 0xc034f3c360, 0x7ffffffffff0bdbf, 0x0)
    /home/runner/work/vega/vega/core/positions/market_position.go:106 +0x897
code.vegaprotocol.io/vega/core/positions.(*MarketPosition).UnregisterOrder(0x32a6bc0?, 0xc00dba7890?, 0xc021351200?)
    /home/runner/work/vega/vega/core/positions/market_position.go:75 +0x45
code.vegaprotocol.io/vega/core/positions.(*Engine).UnregisterOrder(0xc00db8af00, {0x41cfe50, 0xc0341c42d0}, 0xc01ced7ee0)
    /home/runner/work/vega/vega/core/positions/engine.go:239 +0x23c
code.vegaprotocol.io/vega/core/execution/future.(*Market).submitValidatedOrder(0xc00dbdc300, {0x41cfe50, 0xc0341c42d0}, 0xc01ced7ee0)
    /home/runner/work/vega/vega/core/execution/future/market.go:2074 +0xd5c
code.vegaprotocol.io/vega/core/execution/future.(*Market).submitOrder(0xc00dbdc300, {0x41cfe50, 0xc0341c42d0}, 0xc01ced7ee0)
    /home/runner/work/vega/vega/core/execution/future/market.go:1943 +0x24e
code.vegaprotocol.io/vega/core/execution/future.(*Market).SubmitOrderWithIDGeneratorAndOrderID(0xc00dbdc300, {0x41cfe50, 0xc0341c42d0}, 0xc012ccf500?, {0xc021351200, 0x40}, {0x419bec0?, 0xc01e30b128?}, {0xc014336d80, 0x40}, ...)
    /home/runner/work/vega/vega/core/execution/future/market.go:1906 +0x525
code.vegaprotocol.io/vega/core/execution.(*Engine).SubmitOrder(0xc0014c02c0, {0x41cfe50, 0xc0341c42d0}, 0xc01e670a20, {0xc021351200, 0x40}, {0x419bec0, 0xc01e30b128}, {0xc014336d80, 0x40})
```

### Observed behaviour

Network panic when trying to close entire position.

### Expected behaviour

Not network panic when trying to close entire position.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.73.6

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #10177 

Adds validation that the order size that we might be regsitering with the position engine isn't so large that it'll cause us to overflow. We do this both in command validation, and order valiation in the execution engine.

If we do _somehow_ have an order that makes its way through, we now panic ealier with more helpful error-text.

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -1940,6 +1940,10 @@ func (m *Market) submitOrder(ctx context.Context, order *types.Order) (*types.Or
 		return nil, nil, err
 	}
 
+	if err := m.position.ValidateOrder(order); err != nil {
+		return nil, nil, err
+	}
+
 	// Now that validation is handled, call the code to place the order
 	orderConf, orderUpdates, err := m.submitValidatedOrder(ctx, order)
 	if err == nil {
@@ -2990,6 +2994,10 @@ func (m *Market) amendOrder(
 
 	amendedOrder, err := existingOrder.ApplyOrderAmendment(orderAmendment, m.timeService.GetTimeNow().UnixNano(), m.priceFactor)
 	if err != nil {
+		return nil, nil, err
+	}
+
+	if err := m.position.ValidateAmendOrder(existingOrder, amendedOrder); err != nil {
 		return nil, nil, err
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 - [10123](https://github.com/vegaprotocol/vega/issues/10123) - Ledger exports contain account types of "UNKNOWN" type
 - [10132](https://github.com/vegaprotocol/vega/issues/10132) - Add mapping in `GraphQL` for update perps market proposal.
 - [10125](https://github.com/vegaprotocol/vega/issues/10125) - Wire the `JoinTeam` command in the wallet.
+- [10177](https://github.com/vegaprotocol/vega/issues/10177) - Add validation that order sizes are not strangely large.
 - [10166](https://github.com/vegaprotocol/vega/issues/10166) - Closed markets should not be subscribed to data sources when restored from a snapshot.
 - [10127](https://github.com/vegaprotocol/vega/issues/10127) - Untangle `ApplyReferralCode` and `JoinTeam` command verification.
 - [10153](https://github.com/vegaprotocol/vega/issues/10153) - Add metrics and reduce amount of request sent to the Ethereum `RPC`.
--- a/commands/errors.go
+++ b/commands/errors.go
@@ -81,6 +81,7 @@ var (
 	ErrMustBeSetTo0IfSizeSet                           = errors.New("must be set to 0 if the property \"order_amendment.size\" is set")
 	ErrMustBeAtMost3600                                = errors.New("must be at most 3600")
 	ErrMustBeWithinRangeGT0LT20                        = errors.New("price range must be strictly greater than 0 and less than or equal to 20")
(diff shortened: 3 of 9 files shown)


core/execution/future/market.go
```diff
@@ -1940,6 +1940,10 @@ func (m *Market) submitOrder(ctx context.Context, order *types.Order) (*types.Or
 		return nil, nil, err
 	}
 
+	if err := m.position.ValidateOrder(ctx, order); err != nil {
```
ze97286, line 1943: shouldn't we also validate in amend?
also, what happens if we have now (in the protocol upgrade snapshot) such an order already?

wwestgarth: True about amend, I'll have a look although that code path is much more complicated.

I've not checked mainnet snapshots yet to see if we have a bad one. I imagine the chances are slim because its only in the case where we reduce-only and the order `size` is massive, and so we'd have already seen a panic. In all other cases I think we would just fail at the margin check before we even got that far. Which is probably also true when amending.

jeremyletang: It should be quite unlikely, as these are the one used to close position and are reduce only. Because of this we don’t check margin for themS orders woth such a size which would have to go through margin check would surely not go through isn’t it?

EVODelavega (approved):

PR #10182, opened 2023-11-29
feat: ft for closeout bug 1

Jiajia-Cui:
This feature test is to replicate a bug during closeout 

we're transferring the entirety of the margin account to the insurance account:
if !marginAcc.Balance.IsZero() {
            resp.Entries = append(resp.Entries, &types.LedgerEntry{
                FromAccount: marginAcc.ToDetails(),
                ToAccount:   ins.ToDetails(),
                Amount:      marginAcc.Balance.Clone(),
                Type:        types.TransferTypeMarginConfiscated,
                // Reference:   "position-resolution",
                Timestamp:          now,
                FromAccountBalance: num.UintZero(),
                **ToAccountBalance:   num.Sum(ins.Balance, marginAcc.Balance),**
            })
            if err := e.IncrementBalance(ctx, ins.ID, marginAcc.Balance); err != nil {
                return nil, err
            }
            if err := e.UpdateBalance(ctx, marginAcc.ID, marginAcc.Balance.SetUint64(0)); err != nil {
                return nil, err
            }
        }
The important bit is the line in bold which is basically setting the balance of the insurance account after topping it up to what it had before (0) + the balance of the marginAcc BEFORE we topped it up from the general account! because  marginAcc.Balance is only a static copy of the actual margin account balance, so it doesn't get updated when we do the increment above from the general account

### DIFF
--- a/core/integration/features/margin/closeout-bug-1.feature
+++ b/core/integration/features/margin/closeout-bug-1.feature
@@ -1,0 +1,163 @@
+Bug
+
+Feature: replicate a closeout bug, when party is distressed, party's order gets cancelled, and then MTM, party gets closed out
+  Background:
+    Given the following network parameters are set:
+      | name                                    | value |
+      | network.markPriceUpdateMaximumFrequency | 0s    |
+    And the liquidity monitoring parameters:
+      | name       | triggering ratio | time window | scaling factor |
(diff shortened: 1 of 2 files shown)


Jiajia-Cui:
close it as the bug has been fixed

PR #10183, opened 2023-11-29
Fix trading fees register

karlem:
To fix bugs:

1. The api `GetTotalTransferFeeDiscount` always returned 0 discount
2. If more than one party on the same asset accrued discount then the decay for a party would trigger for each additional party for the same asset

### DIFF
--- a/datanode/gateway/graphql/schema.graphql
+++ b/datanode/gateway/graphql/schema.graphql
@@ -1054,11 +1054,11 @@ type Query {
   totalTransferFeeDiscount(
     "ID of party eligible for the discount."
     partyId: String!
-    "ID of asset to associated with the discount."
+    "ID of asset to be associated with the discount."
     assetId: String!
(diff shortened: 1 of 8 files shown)


peterbarrow (approved):

candida-d (approved):

datanode/gateway/graphql/schema.graphql
```diff
@@ -1054,11 +1054,11 @@ type Query {
   totalTransferFeeDiscount(
     "ID of party eligible for the discount."
     partyId: String!
-    "ID of asset to associated with the discount."
+    "ID of asset to be associated with the discount."
```
candida-d, line 1057: Actually this has been rephrased in another PR that Tan has open

candida-d, line 1057:
```suggestion
    "ID of the asset that the discount relates to."
```

PR #10212, opened 2023-12-04
fix: ensure infra fees reward are not counted for vesting

### LINKED ISSUE #10211: [Bug]: Vesting balance shown in API does not match the actual account balance
### Problem encountered

Querying the vesting balance summary using the API vestingBalancesSummary returns incorrect balances. 

### Observed behaviour

When querying the vestingBalancesSummary on testnet with the query:

```
query vestingDetails {
  party(id: "89fd87b86727fa4d9ab06b9f4e64b51c9b07055e8a0c8f0bc40f0a249e321525") {
    vestingStats{
      rewardBonusMultiplier
      quantumBalance
      epochSeq
      }
    vestingBalancesSummary {
      epoch
      vestingBalances {
        asset {
          id
          symbol
          quantum
          decimals
        }
        balance
      }
      lockedBalances {
        asset {
          id
          symbol
          quantum
          decimals
        }
        balance
        untilEpoch
      }
    }
  }
}
```

It returns non zero balances, however when looking at the database, the actual collateral account is 0 and has never seen funds being transferred to it. This seems to be the results of a change for not transferring infra fees reward to the vesting account but directly to the general account. While as far as actual funds transfers, the infra fees are moved to the general account, they are still reported to the vesting account where they are counted as being vested. 

### Expected behaviour

the vesting summary report should match what's actually being vested/locked. 

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.73.6

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
Closes #10211

### DIFF
--- a/core/vesting/vesting.go
+++ b/core/vesting/vesting.go
@@ -18,6 +18,7 @@ package vesting
 import (
 	"context"
 	"sort"
+	"time"
 
 	"code.vegaprotocol.io/vega/core/assets"
 	"code.vegaprotocol.io/vega/core/events"
@@ -39,6 +40,7 @@ type Collateral interface {
 	) ([]*types.LedgerMovement, error)
 	GetVestingRecovery() map[string]map[string]*num.Uint
 	GetAllVestingQuantumBalance(party string) *num.Uint
+	GetVestingAccounts() []*types.Account
 }
 
 type ActivityStreakVestingMultiplier interface {
@@ -76,7 +78,9 @@ type Engine struct {
 	baseRate     num.Decimal
 	benefitTiers []*types.VestingBenefitTier
 
-	state map[string]*PartyRewards
+	state                map[string]*PartyRewards
+	epochSeq             uint64
+	upgradeHackActivated bool
 }
 
 func New(
@@ -146,7 +150,9 @@ func (e *Engine) OnEpochEvent(ctx context.Context, epoch types.Epoch) {
 	}
 }
 
-func (e *Engine) OnEpochRestore(ctx context.Context, epoch types.Epoch) {}
+func (e *Engine) OnEpochRestore(ctx context.Context, epoch types.Epoch) {
+	e.epochSeq = epoch.Seq
+}
 
 func (e *Engine) AddReward(
 	party, asset string,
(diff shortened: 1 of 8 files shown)


core/vesting/vesting_snapshot.go
```diff
@@ -64,21 +65,34 @@ func (e *SnapshotEngine) GetState(k string) ([]byte, []types.StateProvider, erro
...
-		e.loadStateFromSnapshot(data.Vesting)
+		e.loadStateFromSnapshot(ctx, data.Vesting)
 		return nil, nil
 	default:
 		return nil, types.ErrUnknownSnapshotType
 	}
 }
 
-func (e *SnapshotEngine) loadStateFromSnapshot(state *snapshotpb.Vesting) {
+func (e *SnapshotEngine) recoverVesting736(ctx context.Context) {
```
peterbarrow, line 82: What does the 736 bit mean?

peterbarrow: Ah it's code for a specific version of the chain, would it be good to add a comment above this to explain that?

peterbarrow (dismissed):

core/vesting/vesting_snapshot.go
davidsiska-vega: I'd hazard a guess that's it's the v0.73.6 tag.

core/vesting/vesting.go
```diff
@@ -343,7 +348,14 @@ func (e *Engine) broadcastSummary(ctx context.Context, seq uint64) {
 		PartiesVestingSummary: []*eventspb.PartyVestingSummary{},
 	}
 
-	for p, pRewards := range e.state {
+	parties := make([]string, 0, len(e.state))
+	for k := range e.state {
+		parties = append(parties, k)
+	}
+	sort.Strings(parties)
```
jeremyletang, line 355: Is this necessary? Things are sorted at the end of the loop?

core/vesting/vesting_snapshot.go
```diff
@@ -64,21 +65,34 @@ func (e *SnapshotEngine) GetState(k string) ([]byte, []types.StateProvider, erro
...
+	accs := e.c.GetVestingAccounts()
+	for _, a := range accs {
+		e.increaseVestingBalance(a.Owner, a.Asset, a.Balance)
+	}
+	e.broadcastSummary(ctx, e.epochSeq)
```
jeremyletang, line 87: This will need to be broadcast at the start of the first block instead.

core/vesting/vesting.go
ze97286: not necessary, no

jeremyletang (approved):

PR #10225, opened 2023-12-05
Introduce batched proposals

### LINKED ISSUE #9880: [Feature]: Batch proposals
# Feature Overview

**In order to** allow users to submit a batch of proposals at the same time
**We will** implement the following changes as specified

## Specs
- [WIP](https://github.com/vegaprotocol/specs/issues/1994)

## Tasks
- [ ]
- [ ]

karlem:
Close #9880


core/governance/engine.go
```diff
@@ -332,16 +336,121 @@ func (e *Engine) removeActiveProposalByID(ctx context.Context, id string) {
...
+				}
+			}
+
+			// do not check parent market, the market was either rejected when the parent was succeeded
+			// or, if the parent market state is gone (ie succession window has expired), the proposal simply
+			// loses its parent market reference
+			if proposal.ShouldClose(now) {
+				proposal.Close(e.accs, e.markets)
+				if proposal.IsPassed() {
+					e.log.Debug("Proposal passed", logging.ProposalID(proposal.ID))
```
jeremyletang, line 379: maybe add the batch ID in this log, might be useful.

wwestgarth:
I think it might be worth clarifying with the protocol-design team whats supposed to happen when batch proposing network parameters with dependent values.

I think as it stands if I have a batch proposal which increases the below two at the same time to the same value:
```
network.validators.multisig.numberOfSigners 
network.validators.tendermint.number
```
the first one will get rejected in `e.preEnactProposal(ctx, proposal)`. My intuition is that if they are both in the batch with the same enactment time then they'd be no problems. But saying that I can't see anything in the spec about it.

karlem:
@davidsiska-vega can you clarify please

davidsiska-vega:
That said, if, as another ticket, we could look at whether we need to enforce "multisig signers <= consensus validators" at netparam level, that would be good. 

I can imagine a setup where the netparam says: multisig signers `20`, consensus validators `5` and we simply say okay, the number we'll use in core is `max(netparam multisig signers, consensus validators = 5`.

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -226,6 +226,17 @@ message ProposalSubmission {
   vega.ProposalRationale rationale = 3;
 }
 
+// Command that allows a token holder to submit a batch governance proposal that can be voted on by any other token holders, and eventually enacted on the Vega network.
+// For example this command can be used to propose a new market and a network paramater change with it.
```
candida-d, line 230:
```suggestion
// For example this command can be used to propose a new market and a network parameter change with it.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
   }
 }
 
+// Terms change for a batch governance proposal on Vega
```
candida-d, line 355:
```suggestion
// Terms change for a batch governance proposal
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+message BatchProposalTermsChange {
+  // Timestamp as Unix time in seconds when a proposal term gets enacted if proposal passed,
```
candida-d, line 357:
```suggestion
  // Timestamp as Unix time in seconds when proposal terms gets enacted if proposal passed the vote,
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+  // constrained by `minEnact` and `maxEnact` network parameters.
+  int64 enactment_timestamp = 1;
+  // Change being proposed.
```
candida-d, line 360:
```suggestion
  // Changes being proposed.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+  oneof change {
+    // Proposal change for modifying an existing futures market on Vega.
```
candida-d, line 362:
```suggestion
    // Proposal change for modifying an existing futures market.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+    UpdateMarket update_market = 101;
+    // Proposal change for creating new futures market on Vega.
```
candida-d, line 364:
```suggestion
    // Proposal change for creating new futures market.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+    NewMarket new_market = 102;
+    // Proposal change for updating Vega network parameters.
```
candida-d, line 366:
```suggestion
    // Proposal change for updating network parameters.
```

core/governance/engine.go
```diff
@@ -332,16 +334,16 @@ func (e *Engine) removeActiveProposalByID(ctx context.Context, id string) {
 	}
 }
 
-// OnTick triggers time bound state changes.
 func (e *Engine) OnTick(ctx context.Context, t time.Time) ([]*ToEnact, []*VoteClosed) {
+	now := t.Unix()
+
+	voteClosed, addToActiveProposals := e.evaluateBatchProposals(ctx, now)
```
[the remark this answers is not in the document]
karlem: Given that they have just passed, I thought it might be useless to add them as they should not really need to be evaluated by all the code below (they were just closed). So it's ok to evaluate them in the next cycle - next ontick.

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+    UpdateNetworkParameter update_network_parameter = 103;
+    // Proposal change for a freeform request, which can be voted on but does not change the behaviour of the system,
+    // and can be used to gauge community sentiment.
+    NewFreeform new_freeform = 104;
+    // Proposal change for updating an asset.
+    UpdateAsset update_asset = 105;
+    // Proposal change for creating new spot market on Vega.
+    NewSpotMarket new_spot_market = 106;
+    // Proposal change for modifying an existing spot market on Vega.
```
candida-d, line 375:
```suggestion
    // Proposal change for creating new spot market.
    NewSpotMarket new_spot_market = 106;
    // Proposal change for modifying an existing spot market.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+    UpdateSpotMarket update_spot_market = 107;
+    // Proposal change for a governance transfer.
+    NewTransfer new_transfer = 108;
+    // Cancel a governance transfer.
```
candida-d, line 379:
```suggestion
    // Proposal change for a governance initiated transfer.
    NewTransfer new_transfer = 108;
    // Proposal change to cancel a governance initiated transfer.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+    UpdateReferralProgram update_referral_program = 111;
+    // Proposal change for updating the volume discount program.
+    UpdateVolumeDiscountProgram update_volume_discount_program = 112;
+  }
+}
+
+message ProposalParameters {
+  // Represents the minimum time before a proposal can be closed.
+  int64 min_close = 1;
+  // Represents the maximum time before a proposal must be closed.
```
candida-d, line 393:
```suggestion
  // Represents the maximum time that a proposal can be open for voting.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+    UpdateMarketState update_market_state = 110;
+    // Proposal change for updating the referral program.
+    UpdateReferralProgram update_referral_program = 111;
+    // Proposal change for updating the volume discount program.
+    UpdateVolumeDiscountProgram update_volume_discount_program = 112;
+  }
+}
+
+message ProposalParameters {
+  // Represents the minimum time before a proposal can be closed.
```
candida-d, line 391:
```suggestion
  // Represents the minimum time before a proposal can be closed for voting.
```

davidsiska-vega:
> as another ticket,

https://github.com/vegaprotocol/vega/issues/10231

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,78 @@ message ProposalTerms {
...
+  int64 max_close = 2;
+  // Represents the minimum time before an enacted proposal takes effect.
+  int64 min_enact = 3;
+  // Represents the maximum time before an enacted proposal must take effect.
+  int64 max_enact = 4;
+  // Specifies the required percentage of eligible participants that must partake in the proposal for it to be valid.
+  string required_participation = 5;
+  // Specifies the required percentage of votes a proposal needs to be approved.
+  string required_majority = 6;
+  // Specifies the minimum balance a proposer must hold to initiate a proposal.
```
candida-d, line 403:
```suggestion
  // Specifies the minimum governance token balance a proposer must hold to initiate a proposal.
```

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,76 @@ message ProposalTerms {
   }
 }
 
+// Terms change for a batch governance proposal
```
candida-d, line 355: Shouldn't these governance.proto changes also be reflected in GraphQL?

karlem: There is another ticket for the API. It will be reflected there

core/processor/abci.go
```diff
@@ -1728,6 +1737,39 @@ func (app *App) CheckPropose(_ context.Context, tx abci.Tx) error {
...
+func (app *App) CheckBatchPropose(_ context.Context, tx abci.Tx) error {
+	p := &commandspb.BatchProposalSubmission{}
+	if err := tx.Unmarshal(p); err != nil {
+		return err
+	}
+
+	ids := make([]string, 0, len(p.Terms.Changes))
+	for i := 0; i < len(p.Terms.Changes); i++ {
+		ids = append(ids, fmt.Sprintf("prop-%d", i))
+	}
```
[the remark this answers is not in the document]
jeremyletang: I mean, not that it's not deterministic this way here, but would be nicer to do like elsewhere and derive the IDs as usual.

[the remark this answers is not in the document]
karlem: Yeah, I think I forgot to change it. Will use the generator like I do in deliver propose

core/governance/engine_batch.go
```diff
@@ -0,0 +1,356 @@
...
+		if len(closedProposals) != len(batchProposal.Proposals) {
+			e.log.Panic("Failed to close all proposals in batch proposal",
+				logging.ProposalBatchID(batchID),
+			)
+		}
+
+		proposalEvents := make([]events.Event, 0, len(closedProposals))
+		for _, proposal := range closedProposals {
+			if proposal.IsPassed() && batchHasDeclinedProposal {
+				proposal.Decline(types.ProposalErrorProposalInBatchDeclined)
```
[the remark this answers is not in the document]
karlem: Yeah, that's what I thought, but the fact that it doesn't have the error but just says something else has failed notes that this has not failed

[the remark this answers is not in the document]
karlem: So I think now it's just on UI how they display these with this error.

ValentinTrinque (dismissed):

protos/sources/vega/governance.proto
```diff
@@ -352,6 +352,76 @@ message ProposalTerms {
...
+  string min_proposer_balance = 7;
+  // Specifies the minimum governance token balance a voter must hold to participate in voting.
+  string min_voter_balance = 8;
+  // Specifies the required percentage of participation from liquidity providers for the proposal to be valid.
+  string required_participation_lp = 9;
+  // Specifies the required majority percentage from liquidity providers for a proposal to be approved.
+  string required_majority_lp = 10;
+  // Specifies the minimum share that is akin to equity, which a participant must hold for their vote to be eligible.
+  string min_equity_like_share = 11;
+}
```
jeremyletang, line 413: Do we need to add this? Assuming that the most restrictive parameters are to be used accross the batch, not sure we need to add these things here?

karlem: @jeremyletang well because of snapshot. But give what @davidsiska-vega said I have not implemented this correctly. So will need to remove it.

jeremyletang: @davidsiska-vega true, my point is that, as all proposal needs to pass, the only parameters relevant are the most restrictives accross all proposals somehow.

core/governance/engine_batch.go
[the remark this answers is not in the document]
jeremyletang: I'm not sure I understand you comment.

core/processor/abci.go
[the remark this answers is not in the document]
karlem: Hmm actually, does it really matter since this is just for check propose? These IDs will never be used anywhere.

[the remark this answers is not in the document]
jeremyletang: I think it does. I'd expect to be able on the API, to fetch a proposal which is part of a batch, from its ID.

protos/sources/vega/governance.proto
davidsiska-vega: @jeremyletang You seem to be right. We still need to evaluate each proposal individually because of the LP votes. But we only need to store the most restrictive numbers on participation and majority for token holders as they all need to pass.

core/governance/engine_batch.go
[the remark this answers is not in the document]
karlem: What I am saying is - this error I use there right now has been added exactly for the reason you have described above. The one that has passed get `ProposalErrorProposalInBatchDeclined` error. The one that did not pass they get the error reason of why they did not pass.

core/spam/engine.go
```diff
@@ -130,6 +130,7 @@ func New(log *logging.Logger, config Config, epochEngine EpochEngine, accounting
 	}
 
 	e.transactionTypeToPolicy[txn.ProposeCommand] = proposalPolicy
+	e.transactionTypeToPolicy[txn.BatchProposeCommand] = proposalPolicy
```
wwestgarth, line 133: Is this correct to split the policies like this? Because now we have two spam policies for proposals, one for batch and one for normal proposals, so I can actually submit `2 * SpamProtectionMaxProposals` per epoch if I split them across the two transaction types.

karlem: Yeah true. Probably not correct

protos/sources/vega/governance.proto
```diff
@@ -441,6 +512,9 @@ message Proposal {
   optional string required_liquidity_provider_participation = 12;
   // Required majority from liquidity providers, optional but is required for market update proposal.
   optional string required_liquidity_provider_majority = 13;
+  // Batch proposal terms.
+  optional BatchProposalTerms batch_terms = 14;
+  optional string batch_id = 15;
```
wwestgarth, line 517: Can you add a doc-string for `batch_id` please?

wwestgarth:
Can you add some command validation like we have for other transactions: https://github.com/vegaprotocol/vega/tree/develop/commands

Also I can't see anywhere where we reject a batch proposal if it contains a `NewAsset` proposal?

karlem:
> Can you add some command validation like we have for other transactions: https://github.com/vegaprotocol/vega/tree/develop/commands
> 
> 
> 
> Also I can't see anywhere where we reject a batch proposal if it contains a `NewAsset` proposal?

Because newAsset proposal is not even part of the batch proposal submission.

wwestgarth:
> > Can you add some command validation like we have for other transactions: https://github.com/vegaprotocol/vega/tree/develop/commands
> > Also I can't see anywhere where we reject a batch proposal if it contains a `NewAsset` proposal?
> 
> Because newAsset proposal is not even part of the batch proposal submission.

Ah makes sense! I see `UpdateAsset` is in the list of allowed proposals which I would have thought should be no-supported as well?

karlem:
> > > Can you add some command validation like we have for other transactions: https://github.com/vegaprotocol/vega/tree/develop/commands
> 
> > > Also I can't see anywhere where we reject a batch proposal if it contains a `NewAsset` proposal?
> 
> > 
> 
> > Because newAsset proposal is not even part of the batch proposal submission.
> 
> 
> 
> Ah makes sense! I see `UpdateAsset` is in the list of allowed proposals which I would have thought should be no-supported as well?

It is supported because that apparently doesn't need Eth validation :)

wwestgarth (approved):

PR #10250, opened 2023-12-13
Filter ledger by transfer ID

### LINKED ISSUE #10102: Update `LedgerEntries` filter to accept `transferId`
# API Overview

**In order to** show the movement of money caused by [Governance Initiated Transfers](https://github.com/vegaprotocol/specs/blob/master/protocol/0028-GOVE-governance.md#5-transfers-initiated-by-governance)
**We will update the Ledger Entries filtesr to** accept a `transferId`
**So that** there is a reliable way to fetch account movements caused by a specific governance proposal

There is no current way to list the money movements triggered by an individual or one off transfer:
- `LedgerEntries` accepts `PartyID` and `AccountType` as filters, but Governance transfers can be between the `network` party and a blank party ID. Filtering only by account type doesn't guarantee that the transfers were caused by a specific proposal and not any other action
- The transfers API does not contain multiple entries in the case of recurring transfer. A `Transfer` is not the actual movement of money.
- The `Entities` API does not work. The only `txHash` we have is for the Proposal, and the only `entity` created by that `txHash` is a `Proposal`, not the `Transfer` that is created when the proposal is enacted

## Specs
[Spec name](URL) to spec or section within a spec

### Filtering requirements (inputs) (GraphQL, REST, GRPC)
  - [ ] `LedgerEntries` must accept a new optional filter, `transferId`.
    - [ ] If specified, `LedgerEntries` that are triggered by that proposal should be returned

### Sample API calls
```shell
curl -L -X GET 'https://api.n00.testnet.vega.rocks/api/v2/ledgerentry/history?filter.transferId=764b6520f369f1094b2e28e40876ea4adec1562069ede9631b2d8f2bf88690df&pagination.first=1' \
-H 'Accept: application/json'
```

```graphql
{
  ledgerEntries(filter: { transferId: "764b6520f369f1094b2e28e40876ea4adec1562069ede9631b2d8f2bf88690df" } ) {
    edges {
      node {
        vegaTime
      }
    }
  }
}
```

## API test scenarios

**GIVEN** I submit a Governance Proposal for a one off transaction (e.g. [example one off proposal](https://governance.fairground.wtf/proposals/764b6520f369f1094b2e28e40876ea4adec1562069ede9631b2d8f2bf88690df))
**AND** that proposal passes and has the ID `764b6520f369f1094b2e28e40876ea4adec1562069ede9631b2d8f2bf88690df`
**WHEN** I query `LedgerEntries` with a `transferId` parameter set to `764b6520f369f1094b2e28e40876ea4adec1562069ede9631b2d8f2bf88690df`
**THEN** `LedgerEntries` returns 1 ledger entry

**GIVEN** I submit a Governance Proposal for a recurring transaction (e.g. [example recurring proposal](https://governance.fairground.wtf/proposals/6655aa91eace8ac36ff59c84026d7982985d3b96a9728ab8ad0093892a0c7ca1)) 
**AND** that proposal passes and has the ID `6655aa91eace8ac36ff59c84026d7982985d3b96a9728ab8ad0093892a0c7ca1`
**WHEN** I query `LedgerEntries` with a `transferId` parameter set to `6655aa91eace8ac36ff59c84026d7982985d3b96a9728ab8ad0093892a0c7ca1`
**AND** enough time has passed that 3 transfers should have been triggered
**THEN** `LedgerEntries` returns 3 ledger entries

ValentinTrinque:
- Made the ledger movement type a postgresql enum
- Clean up unnecessary complexity while building SQL query
- Remove intermediary object reading the rows as deemed unnecessary

Close #10102

### DIFF
--- a/datanode/entities/aggregated_ledgerentries.go
+++ b/datanode/entities/aggregated_ledgerentries.go
@@ -33,17 +33,18 @@ import (
 type AggregatedLedgerEntry struct {
 	VegaTime     time.Time
 	Quantity     decimal.Decimal
-	TransferType *LedgerMovementType
-	AssetID      *AssetID
+	TransferType LedgerMovementType
+	AssetID      AssetID
 
-	FromAccountPartyID  *PartyID
-	ToAccountPartyID    *PartyID
-	FromAccountMarketID *MarketID
-	ToAccountMarketID   *MarketID
-	FromAccountType     *types.AccountType
-	ToAccountType       *types.AccountType
+	FromAccountPartyID  PartyID
+	ToAccountPartyID    PartyID
+	FromAccountMarketID MarketID
+	ToAccountMarketID   MarketID
+	FromAccountType     types.AccountType
+	ToAccountType       types.AccountType
 	FromAccountBalance  decimal.Decimal
 	ToAccountBalance    decimal.Decimal
+	TransferID          TransferID
 }
 
 func (ledgerEntries *AggregatedLedgerEntry) ToProto() *v2.AggregatedLedgerEntry {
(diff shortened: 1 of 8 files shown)


guoguojin (dismissed):

guoguojin (dismissed):

datanode/sqlstore/migrations/0063_ledger_types.sql
```diff
@@ -0,0 +1,39 @@
+-- +goose Up
+CREATE TYPE ledger_movement_type AS ENUM (
```
wwestgarth, line 2: Is there are particular reason for changing this to an enum? I'm asking because its a little bit less error prone when its not an enum, as we often have situations where a new one is added to the protos but not added to the schema and it causes the data-node to panic.

ValentinTrinque: Since we are usually using enum for this kind of field, I made that part consistent as part of the code cleanup. I don't think we are earning anything regarding performance or anything, except getting a strong data model, so I am happy to revert that part if we are prone to panic.

jeremyletang (approved):

PR #10251, opened 2023-12-13
Fix batch proposal submission

### LINKED ISSUE #10069: [API]: Batch Governance Proposal APIs
# API Overview

**In order to** build support for [Batch Governance Proposals](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0028-GOVE-governance.md#batch-proposals) in to the Governance site
**We will extend the existing API to** include batch governance proposals
**So that** the Governance site can show meaningful data (e.g. pass/fail likelihood) for a batch proposal

tl;dr a batch proposal is an object like any other proposal, except its `terms` field changes to be an array of terms fields.

## Specs
[0028-GOVE-Governance#batch-proposals](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0028-GOVE-governance.md#batch-proposals) 

## API request details
- [x] Updates to [ProposalSubmission](https://docs.vega.xyz/testnet/api/grpc/vega/commands/v1/commands.proto#proposalsubmission) _(GRPC)_
  - [x] [https://docs.vega.xyz/testnet/api/grpc/vega/governance.proto#proposalterms](ProposalTerms) *must* accept a new named proposal type, `batch`, which can contain any of the existing proposal types
     - [x] The closing timestamp of all submissions must match {AC: [0028-GOVE-148](https://github.com/vegaprotocol/specs/blob/palazzo/protocol/0028-GOVE-governance.md#0028-GOVE-148)}
     - [x] A batch proposal cannot contain the following types, and will be rejected if it does:
       - [x] Another batch proposal
       - [x] A new asset proposal
- [ ] Update to [Get governance data](https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-get-governance-data) _(Data Node: GRPC, REST, GraphQL)_
  - [ ] GraphQL: Update `ProposalChange` union to include a batch type
  - [ ] GraphQL: A Proposal of type `ProposalChange` contains an array of `ProposalTerms` as its `terms` field
  - [x] REST: `terms` field of a `batch` proposal is an array of `terms`
- [ ] Updates to [List governance data](https://docs.vega.xyz/testnet/api/rest/data-v2/trading-data-service-list-governance-data) _(Data Node: GRPC, REST, GraphQL. Block Explorer REST)_
  - [ ] Update `proposalType` filter
    - [ ] Add Batch Governance Proposal as an optional `proposalType` filter
    - [ ] When filtering for a `proposalType`, if a batch proposal includes a proposal of that type, it must be included in the result.
- [ ] Block Explorer
  - [ ] Ensure block explorer properly decodes batch proposals 
### Sample API output (fetching a batch by ID)

```shell
curl -L -X GET 'https://api.n00.testnet.vega.rocks/api/v2/governance?proposalId=cafecafed432035ae7e623470e6627e89f4913bcb79dc9bb57c409b9b7649cf5' \
-H 'Accept: application/json'
```

This returns a batch that contains two proposals, a transfer and a freeform governance proposal. One person (the proposer) has voted `yes`.

```json
{
  "data": {
    "proposal": {
      "id": "cafecafed432035ae7e623470e6627e89f4913bcb79dc9bb57c409b9b7649cf5",
      "reference": "",
      "partyId": "69464e35bcb8e8a2900ca0f87acaf252d50cf2ab2fc73694845a16b7cafecafe",
      "state": "STATE_ENACTED",
      "timestamp": "1699610240697284000",
      "terms": {
        "closingTimestamp": "1699610488",
        "enactmentTimestamp": "1699610548",
        "validationTimestamp": "0",
        "batch": [
          {
            "closingTimestamp": "1699015142",
            "enactmentTimestamp": "1699015142",
            "validationTimestamp": "0",
            "newTransfer": {
              "changes": {
                "sourceType": "ACCOUNT_TYPE_NETWORK_TREASURY",
                "source": "",
                "transferType": "GOVERNANCE_TRANSFER_TYPE_BEST_EFFORT",
                "amount": "3333000000",
                "asset": "8ba0b10971f0c4747746cd01ff05a53ae75ca91eba1d4d050b527910c983e27e",
                "fractionOfBalance": "0.1",
                "destinationType": "ACCOUNT_TYPE_REWARD_MAKER_PAID_FEES",
                "destination": "",
                "recurring": {
                  "startEpoch": "9800",
                  "endEpoch": "9839",
                  "dispatchStrategy": {
                    "assetForMetric": "8ba0b10971f0c4747746cd01ff05a53ae75ca91eba1d4d050b527910c983e27e",
                    "metric": "DISPATCH_METRIC_MAKER_FEES_PAID",
                    "markets": [],
                    "entityScope": "ENTITY_SCOPE_INDIVIDUALS",
                    "individualScope": "INDIVIDUAL_SCOPE_ALL",
                    "teamScope": [],
                    "nTopPerformers": "",
                    "stakingRequirement": "",
                    "notionalTimeWeightedAveragePositionRequirement": "",
                    "windowLength": "1",
                    "lockPeriod": "0",
                    "distributionStrategy": "DISTRIBUTION_STRATEGY_PRO_RATA",
                    "rankTable": []
                  }
                }
              }
            }
          },
          {
            "closingTimestamp": "1684342983",
            "enactmentTimestamp": "0",
            "validationTimestamp": "0",
            "newFreeform": {}
          }
        ]
      },
      "rationale": {
        "description": "(The freeform proposal here is pointless)",
        "title": "Do a transfer, and make it snappy"
      },
      "requiredParticipation": "0.07",
      "requiredMajority": "0.66"
    },
    "yes": [
      {
        "partyId": "69464e35bcb8e8a2900ca0f87acaf252d50cf2ab2fc73694845a16b7cafecafe",
        "value": "VALUE_YES",
        "proposalId": "cafecafed432035ae7e623470e6627e89f4913bcb79dc9bb57c409b9b7649cf5",
        "timestamp": "1699610284537647000",
        "totalGovernanceTokenBalance": "50000001000000000000000000",
        "totalGovernanceTokenWeight": "1",
        "totalEquityLikeShareWeight": "0"
      }
    ],
    "no": [],
    "yesParty": {},
    "noParty": {}
  }
}
```

### Submitting a proposal
- See linked spec for acceptance criteria

### On the list API batches count as both their own type and the type of each of the proposals they contain

**GIVEN** there are two proposals
- And the first proposal proposal is a batch, and that batch contains a New Market proposal and a Network Parameter Update proposal
- And the second proposal is a New Market proposal

**WHEN** I use GraphQL to list all proposals, filtered by `TYPE_NEW_MARKET` proposals
**THEN** the list of results will include two proposals: the batch, and the new market proposal

**GIVEN** there are two proposals
- And the first proposal proposal is a batch, and that batch contains a New Market proposal and a Network Parameter Update proposal
- And the second proposal is a New Market proposal

**WHEN** I use GraphQL to list all proposals, filtered by `TYPE_BATCH` proposals
**THEN** the list of results will include one proposals: the batch,

### Block explorer can decode a batch proposal transaction 
**GIVEN** there is a batch proposal transaction in block 5
**WHEN** I use Block Explorer's REST API to fetch block 5
**THEN** I can see the full batch proposal including all of its contained proposals

## Decision log 
- If List Governance Proposals is filtered by (for example) New Market Proposal, it could return an 'unwrapped' market proposal with a new field that indicated it was part of a batch proposal. This would require fewer changes on clients, but add unneccesary complexity on the API side.
- Filtering by type will allow filtering for just batch proposals. This will be useful for block explorer
- Filtering by other types will return proposals of that type *and batches that contain one or more of the given proposal type*. This may be a breaking change for clients (getting new proposal types in queries that used to only return one type).

### LINKED ISSUE #10248: [Bug]: batch proposal error submission is not supported
### Problem encountered

submitting a batch proposal is rejected

`{"jsonrpc":"2.0","error":{"code":1070,"message":"Network error","data":"the network could not process the transaction: command *v1.InputData_BatchProposalSubmission is not supported (ABCI code 70)"},"id":"request"}`

for the request

`request{'batchProposalSubmission':{'reference':'2C6EEE2108','terms':{'closingTimestamp':'1702417536','changes':[{'enactmentTimestamp':'1702417537','newMarket':{'changes':{'instrument':{'name':'Futures System-Tests Market','code':'CRYPTO:VEGA CORE SYS TEST','future':{'settlementAsset':'9acf5cec6d24a35b1f417ada51eff5f3d46ae15ed3df877ff2539447bfe83de4','quoteName':'USD','dataSourceSpecForSettlementData':{'external':{'oracle':{'signers':[{'pubKey':{'key':'5f4ce567db148298a70334fcb67b3c2595a747acef6ea39bb0f6f4ebbd767cb0'}}],'filters':[{'key':{'name':'external.test.settlement_6B1C4A3826','type':'TYPE_INTEGER','numberDecimalPlaces':'3'},'conditions':[{'operator':'OPERATOR_GREATER_THAN_OR_EQUAL','value':'0'}]}]}}},'dataSourceSpecForTradingTermination':{'external':{'oracle':{'signers':[{'pubKey':{'key':'5f4ce567db148298a70334fcb67b3c2595a747acef6ea39bb0f6f4ebbd767cb0'}}],'filters':[{'key':{'name':'external.test.termination_CEC2300159','type':'TYPE_BOOLEAN'},'conditions':[{'operator':'OPERATOR_EQUALS','value':'True'}]}]}}},'dataSourceSpecBinding':{'settlementDataProperty':'external.test.settlement_6B1C4A3826','tradingTerminationProperty':'external.test.termination_CEC2300159'}}},'decimalPlaces':'5','metadata':['asset_class:fx/crypto','product:futures'],'liquidityMonitoringParameters':{'targetStakeParameters':{'timeWindow':'10','scalingFactor':5.0},'triggeringRatio':'0.01','auctionExtension':'1'},'positionDecimalPlaces':'5','linearSlippageFactor':'0.01','liquiditySlaParameters':{'priceRange':'1.0','commitmentMinTimeFraction':'0','performanceHysteresisEpochs':'1','slaCompetitionFactor':'1'},'liquidationStrategy':{'disposalTimeStep':'1','disposalFraction':'1','fullDisposalSize':'18446744073709551615','maxFractionConsumed':'1'},'logNormal':{'riskAversionParameter':0.001,'tau':0.00011407711613050422,'params':{'r':0.016,'sigma':1.5,'mu':0.0}},'quadraticSlippageFactor':''}}}]},'rationale':{'description':'Batch proposal submitted','title':'Batch proposal'}},'pubKey':'5f4ce567db148298a70334fcb67b3c2595a747acef6ea39bb0f6f4ebbd767cb0','propagate':True}`

### Observed behaviour

batch proposal is rejected

### Expected behaviour

batch proposal is accepted

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

dev branch

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

karlem:
Close #10248 
Close #10069


datanode/sqlstore/migrations/0063_batch_proposals.sql
```diff
@@ -0,0 +1,32 @@
...
+);
+
+-- +goose Down
+
+DROP INDEX IF EXISTS proposals_idx_batch_id;
+
+ALTER TABLE proposals
+      DROP COLUMN IF EXISTS batch_id,
+      DROP COLUMN IF EXISTS batch_terms;
+
```
guoguojin, line 29: You need to regenerate the proposals_current view here.

guoguojin (changes requested):

guoguojin (dismissed):

protos/sources/vega/governance.proto
```diff
@@ -437,6 +437,11 @@ message ProposalRationale {
 
 // Governance data
 message GovernanceData {
+  enum Type {
+    TYPE_SINGLE = 0;
+    TYPE_BATCH = 1;
+  }
```
candida-d, line 443: docs here please!

protos/sources/vega/governance.proto
```diff
@@ -451,6 +456,9 @@ message GovernanceData {
   // where key (string) is the party ID i.e. public key and
   // value (Vote) is the vote cast by the given party.
   map<string, Vote> no_party = 5;
+
+  Type proposal_type = 6;
+  repeated Proposal proposals = 7;
```
candida-d, line 461: This could use some docs too

candida-d (commented):
2 requests for more docs, otherwise looks good

datanode/entities/proposal.go
```diff
@@ -127,14 +138,18 @@ func (p Proposal) ToProto() *vega.Proposal {
 		errDetails = ptr.From(p.ErrorDetails)
 	}
 
+	batchID := ptr.From(p.BatchID.String())
```
wwestgarth, line 141: Does this need to be the below, so for a non-batch-proposal the `batchID` will be `nil` and not a pointer to an empty string? 
```suggestion
        var batchID *string
        if len(p.BatchId) != 0 {
            batchID := ptr.From(p.BatchID.String())
        }
```

datanode/sqlstore/votes.go
```diff
@@ -54,14 +54,16 @@ func (vs *Votes) Add(ctx context.Context, v entities.Vote) error {
...
+		 VALUES ($1,  $2,  $3,  $4,  $5, $6, $7, $8, $9, $10)
 		 ON CONFLICT (proposal_id, party_id, vega_time) DO UPDATE SET
 			value = EXCLUDED.value,
 			total_governance_token_balance =EXCLUDED.total_governance_token_balance,
 			total_governance_token_weight = EXCLUDED.total_governance_token_weight,
 			total_equity_like_share_weight = EXCLUDED.total_equity_like_share_weight,
+			per_market_equity_like_share_weight = EXCLUDED.per_market_equity_like_share_weight,
 			tx_hash = EXCLUDED.tx_hash;
 		`,
 		v.ProposalID, v.PartyID, v.Value, v.TxHash, v.VegaTime, v.InitialTime,
```
wwestgarth, line 69: `v.PerMarketEquityLikeShareWeight` is missing being passed in here. But also I don't think we can use a map because the non-determinism might cause history segments to diverge.

core/governance/engine_batch.go
```diff
@@ -331,13 +341,14 @@ func (e *Engine) addBatchVote(ctx context.Context, batchProposal *batchProposal,
...
-		TotalGovernanceTokenBalance: getTokensBalance(e.accs, party),
-		TotalGovernanceTokenWeight:  num.DecimalZero(),
-		TotalEquityLikeShareWeight:  num.DecimalZero(),
+		PartyID:                        party,
+		ProposalID:                     cmd.ProposalID,
+		Value:                          cmd.Value,
+		Timestamp:                      e.timeService.GetTimeNow().UnixNano(),
+		TotalGovernanceTokenBalance:    getTokensBalance(e.accs, party),
+		TotalGovernanceTokenWeight:     num.DecimalZero(),
+		TotalEquityLikeShareWeight:     totalELSWeight,
```
wwestgarth, line 350: I don't think `totalELSWeight` being the sum of the per-market one makes much sense. Can we just leave it as zero? and in `governance.proto` say how that `TotalEquityLikeShareWeight` is only populated if it was a vote on a single-proposal and `PerMarketEquityLikeShareWeight` is only populated if its a batch proposal?

wwestgarth:
Its not in this PR, but theres this bit of code in `DeliverBatchPropose` which isn't right:
```
idgen := idgeneration.New(deterministicBatchID)
ids := make([]string, 0, len(prop.Terms.Changes))

for i := 0; i < len(prop.Terms.Changes); i++ {
    ids = append(ids, idgen.NextID())
}
```
the first call to `idgen.NextID() == deterministicBatchID` so the first proposal in the batch will have the same ID as the batch-proposal which, imagine this is going to cause a lot of issue. So we'll need to throw away the first ID you get from `idgen.NextID()`.

datanode/sqlstore/proposals.go
```diff
@@ -311,19 +386,40 @@ func (ps *Proposals) Get(ctx context.Context,
...
+				batchID = p.BatchID
+			} else if p.IsBatch() {
+				batchID = p.ID
+			}
+
+			if _, ok := fetchedBatches[batchID]; ok {
+				continue
+			}
+
+			p, err := ps.extendOrGetBatchProposal(ctx, p)
```
wwestgarth, line 415: The way this works seems very unintuitive to me. 

If I ask to list all proposals, we get given the batch proposals which contains a list of the batch terms, but the sub-proposals themselves are not returned. So I can't get to the ID of the sub-proposals.

If I _somehow_ did manage to work out the sub-proposal ID and called the `GetGovernanceData` with the sub-proposal ID, it then returns the batch-proposal (which consequently has a different ID) so it doesn't return what I asked.

The result is that I can't seem to find a way to get a sub-proposal from the governance data end-points. Which means I can't see the `state` field of any of them and thats important because it is possible for a proposal to fail between it passing and enacting.

I feel like @edd might have some opinions on what the API should look like.

wwestgarth: Ignore the above: I didn't expand the JSON enough and missed that for a batch proposal it also has a slice of all the sub-proposals, so you can see everything. I still think its a bit weird to ask for a proposal by ID and get a different proposal back, but its no the end of the world.

wwestgarth:
This function never populates `Proposals` so when listing all proposals from the `List` end-points the sub-proposals for a batch will not be populated:
```
func (p Proposal) ToProtoEdge(_ ...any) (*v2.GovernanceDataEdge, error) {
	return &v2.GovernanceDataEdge{
		Node: &vega.GovernanceData{
			Proposal: p.ToProto(),
		},
		Cursor: p.Cursor().Encode(),
	}, nil
}
```

candida-d (dismissed):

datanode/sqlstore/votes.go
ValentinTrinque: Seems to be addressed.

core/governance/engine_batch.go
ValentinTrinque: Seems to be addressed.

PR #10269, opened 2023-12-18
feat: allow changing market name in update market proposal

### LINKED ISSUE #10143: Governance proposal to change market name
With the introduction of perps there is a requirement for the ability to change a market name via the governance process.

This has been updated in the following specs PR and thus needs implementation:

- https://github.com/vegaprotocol/specs/pull/2062

ze97286:
Closes #10143 

```
{"system-tests":"update_market"}
```


datanode/gateway/graphql/schema.graphql
```diff
@@ -4508,6 +4508,7 @@ type UpdateMarketConfiguration {
 
 type UpdateInstrumentConfiguration {
   code: String!
+  name: String!
```
jeremyletang, line 4511: Might need to make it optional here to be compatible with previous versions of the payload?

protos/sources/vega/governance.proto
```diff
@@ -238,6 +238,8 @@ message UpdateSpotMarketConfiguration {
 message UpdateInstrumentConfiguration {
   // Instrument code, human-readable shortcode used to describe the instrument.
   string code = 1;
+  // Instrument name
+  string name = 2;
```
jeremyletang, line 242: Isn't this used as well for the API? in which case it needs to be set optional to be comaptible to previous protos where this fields didn't exists?

jeremyletang (approved):

PR #10279, opened 2023-12-20
update required go version to 1.21

jeremyletang:


EVODelavega (approved):

daniel1302 (approved):
Let's merge it because other repositories are already merged and now we will see some red marks on jenkins

PR #10283, opened 2023-12-20
Introduce `cometbtf` command to replace `tendermint`

### LINKED ISSUE #10000: Alias `tendermint` sub-command as `cometbft`

ValentinTrinque:
Close #10000


cmd/vega/commands/root.go
```diff
@@ -48,7 +48,7 @@ func Main(ctx context.Context) error {
 	if len(os.Args) >= 2 {
 		switch os.Args[1] {
 		case "tendermint", "tm":
```
jeremyletang, line 50: can you add as well "cometbft" in here.

ValentinTrinque: nice catch.

ValentinTrinque: I simply ran `vega cometbft`.

cmd/vega/commands/cometbft.go
```diff
@@ -0,0 +1,94 @@
...
+		cmtcmd.ShowNodeIDCmd,
+		cmtcmd.GenNodeKeyCmd,
+		cmtcmd.VersionCmd,
+		cmtcmd.RollbackStateCmd,
+		cmtcmd.CompactGoLevelDBCmd,
+		cmtdebug.DebugCmd,
+		cmtcli.NewCompletionCmd(rootCmd, true),
+	)
+
+	baseCmd := cmtcli.PrepareBaseCmd(rootCmd, "TM", os.ExpandEnv(filepath.Join("$HOME", cmtcfg.DefaultTendermintDir)))
```
jeremyletang, line 88: That env thing have changed with cometbft, I guess we need to support also the COMET dir or whatever they renamed it with comet

ValentinTrinque: Ah right ! I looked into it already, and I apparently should have not trusted the outdated comment inside the function that still talk about `TM`. Indeed, their main is using `CMT`.

jeremyletang (approved):

PR #10292, opened 2023-12-22
Send event when party updates margin mode

### LINKED ISSUE #10275: [API]: Margin mode API
# API Overview

**In order to** ... support isolated margin mode
**We will build an API to**... provide details of the mode the user is currently in
**So that** ... Console can communicate this to the user

## Specs
https://github.com/vegaprotocol/specs/issues/2075

## API request details

- [ ] Create an API that shows for a given party and market
  - [ ] The margin mode that you are in (cross vs isolated)
  - [ ] If the user has not yet set a margin mode we must be able to tell this, either by explicitly providing this unknown state or - as has been discussed previously - inferring this from the fact the API does not have a mode for that party/market
  - [ ] The margin factor for that market if the user is in isolated mode
  - [ ] The minimum theoretical margin factor for that market
  - [ ] The maximum theoretical leverage for that market

N.B. Both margin / leverage estimates should be provided EXCLUDING the theoretical additional leverage available due to slippage on the market.  We dont need this, and the slightly more conservative value will be fine for isolated mode.

  - [ ] Party ID
  - [ ] Market ID

```

```

## Questions 

## API test scenarios

GIVEN (setup/context) 
WHEN (action) 

Expected FE requirements this supports:
- Showing a user the margin mode a market is on 
- Showing a user that the margin mode has not been set yet so they know they need to
- Showing a user the max leverage available on the market so we can set the max on the "leverage slider"

https://miro.com/app/board/uXjVOs_PBKk=/

ValentinTrinque:
close https://github.com/vegaprotocol/vega/issues/10275


protos/sources/vega/events/v1/events.proto
```diff
@@ -1054,6 +1054,23 @@ message PaidLiquidityFeesStats {
   repeated PartyAmount fees_paid_per_party = 5;
 }
 
+message PartyMarginModeUpdated {
+  // Unique ID of the market where the update happened.
```
candida-d, line 1058:
```suggestion
  // Unique ID of the market in which the update happened.
```

jeremyletang (approved):

PR #10315, opened 2024-01-03
test: add coverage for AC 0012-POSR-012

EVODelavega:
Closes https://github.com/vegaprotocol/core-test-coverage/issues/459
Closes https://github.com/vegaprotocol/core-test-coverage/issues/460

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 - [9948](https://github.com/vegaprotocol/vega/issues/9948) - Add support for linked stop orders.
 - [9849](https://github.com/vegaprotocol/vega/issues/9849) - Add database support for `num.Uint`.
 - [10275](https://github.com/vegaprotocol/vega/issues/10275) - Add API to list party's margin mode.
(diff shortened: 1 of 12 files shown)


core/execution/liquidation/engine.go
```diff
@@ -167,6 +167,7 @@ func (e *Engine) OnTick(ctx context.Context, now time.Time) (*types.Order, error
 		return nil, nil
 	}
 	e.nextStep = now.Add(e.cfg.DisposalTimeStep)
+	fmt.Printf(">> NETWORK ORDER READY, network position %d, order %d", e.pos.open, size)
```
jeremyletang, line 170: debug

EVODelavega: remvoed

peterbarrow (dismissed):

davidsiska-vega (commented):
I believe the position size now looks correct (from market sim runs). 

The PnL in the feature test is correct. 

@cdummett is working on at least two more feature tests to test insurance pool balances and PnL. 

The PnL in data node is wrong and it should not be reported or report "None" or something to make it clear it's not to be relied on.

davidsiska-vega (approved):

jeremyletang (approved):

PR #10325, opened 2024-01-04
fix: market data API should only return the most recent when no dates…

### LINKED ISSUE #10324: [Bug]: GraphQL query getMarketDataHistoryConnectionByID does not return only the latest snapshot
### Problem encountered

The docs for this query state when both start and end date are not provided then the latest snapshot is returned.

image

However all snapshots are turned

### Observed behaviour

all snapshots are returned

### Expected behaviour

only latest snapshot returned

### Steps to reproduce

```bash
query MarketData{
	getMarketDataHistoryConnectionByID(id: "73486cfb14d9c62b1cf2520dc7e0391bfffe178d969e3ae996ba793aebaca9cf") {
    edges{
      node{       
        timestamp       
      }
    }
  }
}
```

### Software version

dev branch

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

guoguojin:
… are provided

Close #10324

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@
 - [10251](https://github.com/vegaprotocol/vega/issues/10251) - Add batch proposal API and fix batch proposal submission.
 - [10285](https://github.com/vegaprotocol/vega/issues/10285) - Fix `MTM` settlement panic.
 - [10321](https://github.com/vegaprotocol/vega/issues/10321) - Fix `ListPartyMarginModes` panic.
+- [10324](https://github.com/vegaprotocol/vega/issues/10324) - Fix `GetMarketHistoryByID` to only return the most recent market data information when no dates are provided.
 
 ## 0.73.0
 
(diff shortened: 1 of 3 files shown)


wwestgarth:
Should we not just update the API docs to say it returns all data (paginated)? That seems most consistent with our other API with start/end ranges. Also we have the `GetLatestMarketData()` API that returns just the latest market data.

jeremyletang (approved):

karlem (approved):

karlem:
I understand this is a quick fix, by I think that endpoint that has `get by id` in the name should not be paginated and should always returns a single item.

peterbarrow (approved):

guoguojin:
@karlem the API documentation specifies that if no start/end is provided we just returned the last market data record for the market. I remember updating the API documentation to reflect that, but the API wasn't doing that, and given how large the market data table can get, we probably had good reason for restricting it that way

PR #10329, opened 2024-01-05
feat: add test for network profit and loss

cdummett:
Close https://github.com/vegaprotocol/core-test-coverage/issues/593

close https://github.com/vegaprotocol/core-test-coverage/issues/594

### DIFF
--- a/core/integration/features/verified/network_pnl.feature
+++ b/core/integration/features/verified/network_pnl.feature
@@ -21,7 +21,7 @@ Feature: Profit and loss for network a running liquidation strategy
 
 
   @NoPerp @NetPNL
-  Scenario: Network liquidates distressed parties and has positive pnl
+  Scenario: Network long then liquidates short positions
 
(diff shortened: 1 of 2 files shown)


EVODelavega (dismissed):

davidsiska-vega (approved):

davidsiska-vega (changes requested):
Let's add some AC codes as well so we've got all this in the bag.

EVODelavega:
> Let's add some AC codes as well so we've got all this in the bag.

Given that the spec doesn't state that the PnL for the network should be available (and that we're not providing that data), these specific tests arguably all fall under `0012-POSR-014`, or maybe `0012-POSR-009` as it's checking the network position over time, I guess

davidsiska-vega (approved):
🚀

EVODelavega (approved):

ze97286 (approved):

PR #10340, opened 2024-01-08
Support updating profile for a party

### LINKED ISSUE #10307: Support setting metadata for keys
Updated requirements:

- Tx to add/edit key metadata
- Ability to associate an alias to a single key
- A check for uniqueness on that key in the Tx processing
- An API to expose metadata for a key (list the metadata for a given set of keys)

ValentinTrinque:
Close #10307


commands/update_party_profile.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if cmd == nil {
+		return errs.FinalAddForProperty("update_party_profile", ErrIsRequired)
+	}
+
+	if len(cmd.Alias) > 32 {
+		errs.AddForProperty("update_party_profile.alias", ErrIsLimitedTo32Characters)
+	}
+
+	cmd.Alias = vgstrings.TrimBlank(cmd.Alias)
```
jeremyletang, line 40: should that been done before validatingthe len?

commands/update_party_profile.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if len(cmd.Metadata) > 10 {
+		errs.AddForProperty("update_party_profile.metadata", ErrIsLimitedTo10Entries)
+	} else {
+		for i, m := range cmd.Metadata {
+			if len(m.Key) > 32 {
+				errs.AddForProperty(fmt.Sprintf("update_party_profile.metadata.%d.key", i), ErrIsLimitedTo32Characters)
+			}
+
+			m.Key = vgstrings.TrimBlank(m.Key)
```
jeremyletang, line 50: Same question here?  to be fair I don't think these function should changed the payload, as if the transaction is build with another library (e.g: webwallet) it's gonna be changed at runtime. I'd remove this completely

commands/update_party_profile.go
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) 2023 Gobalsky Labs Limited
```
EVODelavega, line 1: 2024?

commands/update_party_profile.go
ValentinTrinque: We could, I just do it after so we don't trigger a trim on something that is 100MB.

commands/update_party_profile.go
ValentinTrinque: so we accept blanks ? like:
- "  hello "
- "  hello  "
As different alias ?

commands/update_party_profile.go
ValentinTrinque: We have to update the linter then. Another PR I believe.

core/parties/engine.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("invalid profile update: %w", err)
+	}
+
+	profile, exists := e.profiles[partyID]
+	if !exists {
+		profile = &types.PartyProfile{}
+		e.profiles[partyID] = profile
+	}
+
+	profile.PartyID = partyID
```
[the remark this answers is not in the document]
ValentinTrinque: Sure

core/parties/helpers_for_test.go
```diff
@@ -0,0 +1,97 @@
...
+func expectPartyProfileUpdatedEvent(t *testing.T, engine *testEngine) {
+	t.Helper()
+
+	engine.broker.EXPECT().Send(gomock.Any()).Do(func(evt events.Event) {
+		_, ok := evt.(*events.PartyProfileUpdated)
+		assert.True(t, ok, "Event should be a PartyProfileUpdated, but is %T", evt)
+	}).Times(1)
+}
+
+func newSnapshotEngine(t *testing.T, vegaPath paths.Paths, now time.Time, engine *parties.SnapshottedEngine) *snapshot.Engine {
```
EVODelavega, line 61: This looks to me like a strange place for this helper... why, in the parties package, are we creating a snapshot engine? This feels to me like it belongs in the snapshot package. As far as unit tests go, we should test the API that the snapshot engine uses/expects with known data, and we can use the same known data in the snapshot engine unit test

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -415,3 +415,15 @@ message JoinTeam {
   // ID of the team to join, this is the same as the referral code used to generate the team.
   string id = 1;
 }
+
+// Command to associate metadata to a public key, known as a party ID.
+// Partial update is not supported, meaning previous values must be set to be kept,
+// otherwise, they are voided.m
```
candida-d, line 421:
```suggestion
// Partial update is not supported, meaning previous values must be set to be kept,
// otherwise, they are voided.
```

core/processor/abci.go
```diff
@@ -2635,3 +2647,32 @@ func (app *App) JoinTeam(ctx context.Context, tx abci.Tx) error {
...
+func (app *App) CheckUpdatePartyProfile(ctx context.Context, tx abci.Tx) error {
+	params := &commandspb.UpdatePartyProfile{}
+	if err := tx.Unmarshal(params); err != nil {
+		return fmt.Errorf("could not deserialize UpdatePartyProfile command: %w", err)
+	}
+
+	partyID := types.PartyID(tx.Party())
+	if err := app.partiesEngine.CheckProfile(ctx, partyID, params); err != nil {
+		return err
+	}
```
jeremyletang, line 2660: We can'd do that here, this being called by checkTx can be called concurrently, because of that, if I'm not mistaken, we could read from the map used in the parties engine whicle calling checkTx and processing a PartiesUpdates transaction. While that's very unlikely it happen in real life, I'd rather not risk it.

core/parties/helpers_for_test.go
ValentinTrinque: Only testing the API the snapshot engine is using prevent the test of catching a bunch of error cases. Consuming to API is not enough to prevent certain errors.

protos/sources/vega/commands/v1/commands.proto
candida-d: And a follow-up question about what actually gets voided here, it wasn't clear to me.

core/parties/dependencies.go
```diff
@@ -0,0 +1,27 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package parties
+
+import (
+	"code.vegaprotocol.io/vega/core/events"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -destination mocks/mocks.go -package mocks code.vegaprotocol.io/vega/core/parties Broker
```
EVODelavega, line 22: The broker package is the one exception to the rule of _"declare dependency interfaces and generate mocks in the package"_. Keeping the interface here is fine, but just import `core/broker/mocks`

EVODelavega (changes requested):

commands/update_party_profile.go
```diff
@@ -0,0 +1,65 @@
...
+		errs.AddForProperty("update_party_profile.alias", ErrIsLimitedTo32Characters)
+	}
+
+	cmd.Alias = vgstrings.TrimBlank(cmd.Alias)
+
+	if len(cmd.Metadata) > 10 {
+		errs.AddForProperty("update_party_profile.metadata", ErrIsLimitedTo10Entries)
+	} else {
+		for i, m := range cmd.Metadata {
+			if len(m.Key) > 32 {
```
wwestgarth, line 46: Do want want to check that they don't have two things in the meta-data slice with the same key? Because once it gets into core I guess whichever was first would just disappear.

commands/update_party_profile.go
jeremyletang: hum, I quite think so. My point is we should not have the node transform the transaction when doing validation. so we should either:
- accept those as different alias,
- simply reject the transaction with an appropriate error.
-

core/parties/helpers_for_test.go
EVODelavega: hence the second part of the test: using the values returned in this test in a unit test that sits in the snapshot package. If we call `GetState()` here, and use those return values in the snapshot test, what are we missing?

commands/update_party_profile.go
ValentinTrinque: I can do it if we think it's important.

core/parties/dependencies.go
ValentinTrinque: Ok

commands/update_party_profile.go
ValentinTrinque: I will accept them then. We will how it goes then.

protos/sources/vega/vega.proto
```diff
@@ -6,6 +6,25 @@ import "vega/markets.proto";
...
 
+// Holds metadata associated to a party.
+message PartyProfile {
+  // Party ID associated to the profile.
+  string party_id = 1;
+  // Alias given to the party.
+  string alias = 2;
+  // Metadata to associate to a party, in a key/value format where the key
+  // describes the type of metadata in the value field.
+  vega.Metadata metadata = 3;
```
wwestgarth, line 17: Should this be `repeated`?
```suggestion
  repeated vega.Metadata metadata = 3;
```

commands/update_party_profile.go
ValentinTrinque: I removed it entirely.

libs/strings/strings.go
```diff
@@ -0,0 +1,22 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package strings
+
+import "strings"
+
+func TrimBlank(s string) string {
```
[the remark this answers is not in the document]
ValentinTrinque: Removed it.

protos/sources/vega/vega.proto
ValentinTrinque: Nice catch.

commands/update_party_profile.go
ValentinTrinque: It is seen as necessary, so I will do it

EVODelavega (dismissed):

candida-d (approved):

EVODelavega (approved):

PR #10341, opened 2024-01-08
Adding AC for 0014-ORDT-130

peterbarrow:


core/integration/steps/parties_place_the_following_orders.go
```diff
@@ -437,8 +449,10 @@ func parseSubmitOrderTable(table *godog.Table) []RowWrapper {
 		"fb trailing",
 		"ra price trigger",
 		"ra trailing",
-		"so expires in",
-		"so expiry strategy",
+		"ra expires in",
```
EVODelavega, line 452: Not entirely sure why rise above/fall below require distinct fields here, if I'm honest, maybe it'll become clear as I get further down in this PR

peterbarrow: In the real world the user would be able to set each one independently so it felt right to do the same for the feature tests. Also because before it was only set once, it wasn't exercising all the code paths as only one side had an expiry set. Yes there is now a little more to type but it does mean better tests.

EVODelavega: Makes sense, all good

EVODelavega (dismissed):

EVODelavega (approved):

candida-d (approved):

PR #10364, opened 2024-01-10
fix: add retry logic for specific SQL status error

### LINKED ISSUE #10326: Fix flaky unit test
### Problem encountered

The `TestLoadingDataFetchedAsynchronously` test (datanode/networkhistory) fails fairly regularly, and should be made more stable. The test can fail both in the regular unit test run, and the run with race detection.

### Observed behaviour

Looking at the logs, we can see the test fails because of a deadlock on the DB side:

```
2024-01-04T19:23:13.037Z] 2024-01-04T19:21:02.404Z	INFO	revert-schema-to-version-0	v3@v3.6.1/migration.go:46	OK   0062_teams_stats.sql
[2024-01-04T19:23:13.037Z]     service_test.go:431: 
[2024-01-04T19:23:13.037Z]         	Error Trace:	/jenkins/workspace/vega_PR-10315/vega/datanode/networkhistory/service_test.go:431
[2024-01-04T19:23:13.037Z]         	Error:      	Received unexpected error:
[2024-01-04T19:23:13.037Z]         	            	failed to load snapshot data:failed to revert scheam to version zero: failed to goose down the schema to version 0: ERROR 0061_additional_candle_intervals.sql: failed to run SQL migration: failed to execute SQL query "SELECT remove_continuous_aggregate_policy('trades_candle_4_hours', true);\n": ERROR: deadlock detected (SQLSTATE 40P01)
[2024-01-04T19:23:13.037Z]         	Test:       	TestLoadingDataFetchedAsynchronously
[2024-01-04T19:23:13.037Z] --- FAIL: TestLoadingDataFetchedAsynchronously (4.91s)
```

Looking at the specific error message, and the migration itself (removing a continuous aggregate policy), the likely cause is a pgsql jobworker is using the underlying materialized view while we're trying to delete it, causing a deadlock. During the down step, we might be able to mitigate the issue by disabling the scheduler, although this is speculation at this point.

### Expected behaviour

The test passing.

### Steps to reproduce

```bash
run the datanode/network unit tests a number of times with `-count=1` to disable caching, the failure is intermittent.
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #10326 

Not the perfect fix, but as this is an intermittent failure, most likely caused by the migration and job workers accessing the same materialised view, as per PG docs: https://www.postgresql.org/docs/current/mvcc-serialization-failure-handling.html

Retrying the operation is a valid approach

### DIFF
--- a/datanode/networkhistory/service.go
+++ b/datanode/networkhistory/service.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -261,6 +262,9 @@ func (d *Service) LoadNetworkHistoryIntoDatanode(ctx context.Context, chunk segm
 func (d *Service) LoadNetworkHistoryIntoDatanodeWithLog(ctx context.Context, log snapshot.LoadLog, chunk segment.ContiguousHistory[segment.Full],
 	connConfig sqlstore.ConnectionConfig, withIndexesAndOrderTriggers, verbose bool,
 ) (snapshot.LoadResult, error) {
+	maxRetries := 2 // max 2 retries
+	// the deadlock error that should trigger a retry
+	status := "deadlock detected (SQLSTATE 40P01)"
 	datanodeBlockSpan, err := sqlstore.GetDatanodeBlockSpan(ctx, d.connPool)
 	if err != nil {
 		return snapshot.LoadResult{}, fmt.Errorf("failed to get data node block span: %w", err)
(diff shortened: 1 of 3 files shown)


datanode/networkhistory/service.go
```diff
@@ -274,14 +278,29 @@ func (d *Service) LoadNetworkHistoryIntoDatanodeWithLog(ctx context.Context, log
 
 	chunks := chunk.Slice(datanodeBlockSpan.ToHeight+1, chunk.HeightTo)
 	loadResult, err := d.snapshotService.LoadSnapshotData(ctx, log, chunks, connConfig, withIndexesAndOrderTriggers, verbose)
-	if err != nil {
+	if err == nil {
+		log.Info("loaded all available data into datanode", logging.String("result", fmt.Sprintf("%+v", loadResult)),
+			logging.Duration("time taken", time.Since(start)))
+
+		return loadResult, err
+	}
```
jeremyletang, line 286: Why not just put that in the for loop and give it 1 more retry?

EVODelavega: Because that would be the obvious thing to do :stuck_out_tongue:  fair enough

jeremyletang: hum, alright

guoguojin (dismissed):

datanode/networkhistory/service.go
```diff
@@ -261,6 +262,9 @@ func (d *Service) LoadNetworkHistoryIntoDatanode(ctx context.Context, chunk segm
 func (d *Service) LoadNetworkHistoryIntoDatanodeWithLog(ctx context.Context, log snapshot.LoadLog, chunk segment.ContiguousHistory[segment.Full],
 	connConfig sqlstore.ConnectionConfig, withIndexesAndOrderTriggers, verbose bool,
 ) (snapshot.LoadResult, error) {
+	maxRetries := 3 // max 2 retries
```
jeremyletang, line 265: lol

EVODelavega: removed

ValentinTrinque (dismissed):

jeremyletang (approved):

PR #10379, opened 2024-01-11
0014-ORDT-131 AC

peterbarrow:


core/execution/stoporders/stop_orders.go
```diff
@@ -319,7 +319,7 @@ func (p *Pool) RemoveExpired(orderIDs []string) []*types.StopOrder {
 				// is the OCO link already mapped
 				if _, ok := ordersM[order.OCOLinkID]; !ok {
 					ordersM[order.OCOLinkID] = p.orders[p.orderToParty[id]][order.OCOLinkID]
-					ordersM[order.OCOLinkID].Status = types.StopOrderStatusExpired
+					ordersM[order.OCOLinkID].Status = types.StopOrderStatusCancelled
```
ze97286, line 322: shouldn't this be stopped (=requested by the protocol) rather than cancelled (=requested by the user)

peterbarrow: Maybe, I was going from the spec:
"OCOs work exactly like two separate stop orders except that if one of the pair is triggered, cancelled, deleted, or rejected, the other one is automatically cancelled"

ze97286: understood, might be worth confirming with the owner of the spec, for consistency with other parts of the protocol.

ze97286 (approved):

PR #10387, opened 2024-01-12
feat: set the default chain ID on oracle spec

EVODelavega:


core/datasource/datasource.go
```diff
@@ -83,7 +84,8 @@ func SpecFromProto(specProto *vegapb.DataSourceSpec) *Spec {
 
 func (s *Spec) FromDefinition(d *definition.Definition) *Spec {
 	if d != nil {
-		bytes, _ := proto.Marshal(d.IntoProto())
+		// @TODO provide value somehow
+		bytes, _ := proto.Marshal(d.IntoProto(1))
```
jeremyletang, line 88: ?

core/datasource/external/ethcall/common/spec.go
```diff
@@ -46,14 +47,14 @@ type Spec struct {
 	L2ChainID             uint64
```
wwestgarth, line 47: I think instead of trying to pass a default chainID all the way through all these many, many, many, many, many, many layers of data-source types, we should just make this field `*uint64` and leave it as `nil` to implie its for the L1 chain. Then in `OnSpecActivated()` on the ethcall engine where we do this:
```
if e.chainID != d.L2ChainID {
    return nil
}
```

have something like

```
if d.L2ChainID == nil && e.iAmTheL1Engine {
    // this one is for me
   return
}

if e.chainID == d.L2ChainID {
	// I'm an L2 engine and this one is for me
        return
}

// this one is not for me
return nil
```

core/datasource/datasource.go
```diff
@@ -39,7 +39,8 @@ type Spec struct {
 func (s *Spec) IntoProto() *vegapb.DataSourceSpec {
 	config := &vegapb.DataSourceDefinition{}
 	if s.Data != nil {
-		config = s.Data.IntoProto()
+		// @TODO check value set on data
+		config = s.Data.IntoProto(1)
```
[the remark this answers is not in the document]
EVODelavega: I had to pass the chain ID somewhere. On these types, which are constantly being converted, and owing to the unique way in which the datasource package is written, this appears to be the only place

core/datasource/external/ethcall/common/spec.go
EVODelavega: Given that the L1 engine has a set ID, the check we have should have the same effect, surely

wwestgarth: It does, but has the advantage of not having to awkwardly pass a default chainID down all these functions that should really just be a basic type conversion.

There is also the second problem that if someone sets `nil` as the change ID when proposing a market for the field called `l2_chain_id`, when it comes out the other end in the event (and then through the API) the field called `l2_chain_id` will be populated with the Ethereum L1 chain ID, which doesn't make much sense.

PR #10391, opened 2024-01-15
chore: add oracle supports for composite price in feature tests

ze97286:


core/types/governance_new_market.go
```diff
@@ -200,14 +200,10 @@ func (mpc *CompositePriceConfiguration) DeepClone() *CompositePriceConfiguration
...
-		specDefinition := s.Data.DeepClone()
-		sources = append(sources, datasource.SpecFromDefinition(*definition.NewWith(specDefinition)))
-	}
+	sources = append(sources, mpc.DataSources...)
+
 	bindings := make([]*datasource.SpecBindingForCompositePrice, 0, len(mpc.SpecBindingForCompositePrice))
-	for _, sbfcp := range mpc.SpecBindingForCompositePrice {
-		bindings = append(bindings, sbfcp.DeepClone())
-	}
+	bindings = append(bindings, mpc.SpecBindingForCompositePrice...)
```
jeremyletang, line 206: why those changes?

ze97286: because the way this was deep cloned was incorrect and was yielding a mismatching sources.

karlem (dismissed):

core/types/governance_new_market.go
wwestgarth: What was incorrect about it? The way it is now its not really a deep-clone because the original and the new `CompositePriceConfiguration` will share pointers to the same thing.

ze97286: 1. the deep cloned version was not equal to the uncloned one - I don't think the way I was constructing it was correct (* I wrote the deep clone in the first place, I don't think it got a chance to be tested or used so I'd be careful of treating it as ground truth) - I'm sure there's a way to get it right - but it's not the way it was, at least I don't think it is.
2. that's true but I don't think it's an issue. It's treated as read only really, so while it may share a pointer and it's a bit unclean, at least at the time of writing I couldn't think of a potential problematic scenario for that, so I copy the arrays (rather than deep copy) to avoid reordering just in case, but I couldn't think of a risk worth the messy deep clone. I could be wrong.

karlem (approved):
@wwestgarth Can you please have a look and see if you are happy with Zohars comment?

ze97286:
> @wwestgarth Can you please have a look and see if you are happy with Zohars comment?

he doesn't need to be, he need's to be happy with the code :)

PR #10395, opened 2024-01-15
test: move netparams in snapshot higher up

EVODelavega:
DO NOT MERGE

Testing a change in snapshot order, moving netparams higher up


wwestgarth:
protocol upgrade pipeline step looks broken for dev-op-y reasons. Kicked off again without it: https://jenkins.vega.rocks/blue/organizations/jenkins/common%2Fsystem-tests/detail/system-tests/7592/pipeline

PR #10398, opened 2024-01-15
feat: set default chain ID

EVODelavega:
Bit of a mess, would've preferred having the stuff to set the chain ID field in the types package, but I believe that won't work due to circular imports between datasource and types... It's almost as though they should go in the same package


core/governance/snapshot.go
```diff
@@ -319,8 +323,12 @@ func (e *Engine) restoreBatchActiveProposals(ctx context.Context, active *types.
 	vevts := []events.Event{}
 	e.log.Debug("restoring active proposals snapshot", logging.Int("nproposals", len(active.BatchProposals)))
 	for _, bpp := range active.BatchProposals {
+		bpt := types.BatchProposalFromSnapshotProto(bpp.BatchProposal.Proposal, bpp.Proposals)
+		if vgcontext.InProgressUpgradeFrom(ctx, "v0.74.0") {
```
wwestgarth, line 327:
```suggestion
		if vgcontext.InProgressUpgradeFrom(ctx, "v0.73.12") {
```

wwestgarth (approved):

PR #10399, opened 2024-01-15
feat: improve validation of evm config netparam

wwestgarth:
Improves the validation of the L2 configuartions network parameters to ensure that:
- A network name/chain-id/network-id does not appear in two chain configs
- That a chain configuration cannot be removed

The second one appears in the spec to try and reduce scope, which makes sense. If it causes issues when writing system-tests by not being able to restore the network to a prior state, then we may have to remove the restriction.


protos/sources/vega/chain_events.proto
```diff
@@ -16,6 +16,8 @@ message EthContractCallEvent {
   bytes result = 4;
   // Error message if the call failed.
   optional string error = 5;
+  // the l2 chain for this chain event.
```
candida-d, line 19:
```suggestion
  // Layer 2 chain for this chain event.
```

protos/sources/vega/data_source.proto
```diff
@@ -103,6 +103,12 @@ message EthCallSpec {
   // in the second result returned from the contract for a structure with a key
   // called 'price' and use that if it exists.
   repeated Normaliser normalisers = 8;
+
+  // an optional ethereum L2 chain id, if not specified this EthCallSpec is
+  // to be used against ethereum mainnet, if set the matching L2 network
+  // for the chain_id is to be used, the network must have been registered via governance
+  // first.
```
candida-d, line 110:
```suggestion
  // Ethereum layer 2 chain ID. If not specified this EthCallSpec is
  // to be used against ethereum mainnet. If set, the matching L2 network
  // for the chain ID is to be used, and the network must have been registered via governance
  // first.
```

protos/sources/vega/vega.proto
```diff
@@ -1467,6 +1467,25 @@ message LiquidityProvision {
   string reference = 12;
 }
 
+message EthereumL2Config {
+  // Network ID of this Ethereum L2 network.
```
candida-d, line 1471:
```suggestion
  // Network ID of this Ethereum layer 2 network.
```

protos/sources/vega/vega.proto
```diff
@@ -1467,6 +1467,25 @@ message LiquidityProvision {
...
+  string network_id = 1;
+  // Chain ID of this Ethereum L2 network.
```
candida-d, line 1473:
```suggestion
  // Chain ID of this Ethereum layer 2 network.
```

protos/sources/vega/vega.proto
```diff
@@ -1467,6 +1467,25 @@ message LiquidityProvision {
...
+  string chain_id = 2;
+  // Number of block confirmations to wait to consider an Ethereum transaction trusted.
+  // An Ethereum block is trusted when there are at least "n" blocks confirmed by the
+  // network, "n" being the number of `confirmations` required. If `confirmations` was set to `3`,
+  // and the current block to be forged (or mined) on the L2 is block 14, block
```
candida-d, line 1478:
```suggestion
  // and the current block to be forged, or mined, on the L2 is block 14, block
```

wwestgarth:
Move to https://github.com/vegaprotocol/vega/pull/10406 because I got myself in a git-tangle.

PR #10426, opened 2024-01-17
Modify `EstimatePosition` API

### LINKED ISSUE #10247: [Feature]: Account for margin mode in the liquidation price estimate.

witgaw:
Modify `EstimatePosition` API to account for isolated margin mode.

Closes https://github.com/vegaprotocol/vega/issues/10247, https://github.com/vegaprotocol/vega/issues/10439


datanode/api/server.go
```diff
@@ -94,7 +139,7 @@ type GRPCServer struct {
 	networkLimitsService          *service.NetworkLimits
 	marketDataService             *service.MarketData
 	tradeService                  *service.Trade
-	assetService                  *service.Asset
+	AssetService                  AssetService
```
jeremyletang, line 142: Why has this been made public?

datanode/api/server.go
```diff
@@ -460,17 +505,18 @@ func (g *GRPCServer) Start(ctx context.Context, lis net.Listener) error {
 	vegaprotoapi.RegisterCoreServiceServer(g.srv, coreProxySvc)
 
 	tradingDataSvcV2 := &TradingDataServiceV2{
```
jeremyletang, line 507: Same question in this type, a few have been made public?

jeremyletang: ah right, most likely for the tests.

datanode/api/server.go
```diff
@@ -80,6 +81,50 @@ type NetworkHistoryService interface {
...
+// Asset service ...
+//
+//go:generate go run github.com/golang/mock/mockgen -destination mocks/asset_service_mock.go -package mocks code.vegaprotocol.io/vega/datanode/api AssetService
+type AssetService interface {
+	Add(ctx context.Context, a entities.Asset) error
+	GetByID(ctx context.Context, id string) (entities.Asset, error)
+	GetByTxHash(ctx context.Context, txHash entities.TxHash) ([]entities.Asset, error)
+	GetAll(ctx context.Context) ([]entities.Asset, error)
+	GetAllWithCursorPagination(ctx context.Context, pagination entities.CursorPagination) ([]entities.Asset, entities.PageInfo, error)
+}
```
jeremyletang, line 126: Can we only add the methods that are used in here in those mocks? Not a big deal per se, but first things I've seen here when looking at the file is that the interfaces include all the write methods, and I thoughts straight away -> there's something very wrong going on with this PR if we write from the API layer.

jeremyletang (commented):
Not sure why lots of proto generated files are updated while the relevant .proto files are untouched here? Are you using a different version of the protobuf compiler?

datanode/api/server.go
witgaw: Fixed now

witgaw:
> Not sure why lots of proto generated files are updated while the relevant .proto files are untouched here? Are you using a different version of the protobuf compiler?

All versions should be as per:
script/gettools.sh

Unless that's stale and there's some other way tooling should be installed now?

`go version go1.21.5 darwin/arm64` in case that impacts it.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3899,19 +3899,38 @@ message EstimatePositionRequest {
...
-  optional bool scale_liquidation_price_to_market_decimals = 5;
+  optional bool scale_liquidation_price_to_market_decimals = 10;
 }
 
 // Response for the estimate of the margin level and, if available collateral was provided in the request, liquidation price for the specified position.
 message EstimatePositionResponse {
   // Margin level range estimate for the specified position.
   MarginEstimate margin = 1;
+  // Margin account balance increase estimate.
+  CollateralIncreaseEstimate colateral_increase_estimate = 2;
```
candida-d, line 3923:
```suggestion
  CollateralIncreaseEstimate collateral_increase_estimate = 2;
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3899,19 +3899,38 @@ message EstimatePositionRequest {
...
   // Whether the liquidation price estimates should be scaled to market decimal places or by asset decimal places. If not set, asset decimal places are used.
-  optional bool scale_liquidation_price_to_market_decimals = 5;
+  optional bool scale_liquidation_price_to_market_decimals = 10;
 }
 
 // Response for the estimate of the margin level and, if available collateral was provided in the request, liquidation price for the specified position.
 message EstimatePositionResponse {
   // Margin level range estimate for the specified position.
   MarginEstimate margin = 1;
+  // Margin account balance increase estimate.
```
candida-d, line 3922:
```suggestion
  // Estimated margin account balance increase.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3890,7 +3890,7 @@ message OrderInfo {
   bool is_market_order = 4;
 }
 
-// Request for the estimate of the margin level and, if available collateral is provided, liquidation price for the specified position.
+// Request for the estimate of the margin level, margin account balance change estimate and liquidation price for the specified position.
```
candida-d, line 3893:
```suggestion
// Request for the estimated margin level, margin account balance change, and liquidation price for the specified position.
```

All 3 are estimates, right? It's clearer if either all 3 say "estimated" or there's just one use of it to cover all 3 values.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3899,19 +3899,38 @@ message EstimatePositionRequest {
...
+  string margin_account_balance = 4;
+  // General account balance. Needs to be provided scaled to asset decimal places.
+  string general_account_balance = 5;
+  // Order margin account balance. Needs to be provided scaled to asset decimal places.
+  string order_margin_account_balance = 6;
+  // Margin mode for the party, cross margin or isolated margin.
+  vega.MarginMode margin_mode = 7;
+  // Margin factor to be used along with isolated margin mode
+  optional string margin_factor = 8;
+  // Whether the estimated collateral increase should be assumed as available for liquidation price calculation in isolated margin mode.
```
candida-d, line 3912: I don't understand this description - could you rephrase, or describe what this does and we can work on it together?

candida-d (changes requested):

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Yes, thanks, I've applied all the suggestions.

protos/sources/data-node/api/v2/trading_data.proto
witgaw: Would that be clearer?
> Whether the estimated collateral increase should be included in available collateral for liquidation price calculation in isolated margin mode.

It's for deal ticket purposes: there the user is specifing an order but the extra collateral needed for that order is not yet in any of the margin accounts, however the user would most likely be interested in the liquidaiton price assuming the order gets posted and the appropriate additional collateral amount gets transfered.

When looking at the existing position and orders the liquidation price should be calculated only based on current level of margin accounts (main and order account) without any transfers even if the margin account balance is below the current initial margin level.

candida-d: That is clearer! Thank you for the explanation, too. Would you mind making that change for the GQL too?

witgaw: Sure, done now.

candida-d (dismissed):

datanode/api/__debug_bin3396181152
peterbarrow, line 1: Can we remove this file please, I think it was added by mistake.

core/risk/engine_test.go
```diff
@@ -482,24 +482,24 @@ func TestMarginWithNoOrdersOnBook(t *testing.T) {
 			positionSize:   9,
 			buyOrders: []*risk.OrderInfo{
 				{
-					Size:          3,
+					Remaining:     3,
```
EVODelavega, line 485: VERY much pedantry and nit-picky, but I've encountered it a couple times recently and it got on my nerves a bit: is this `Remaining` the visible remaining size on the book, or `TrueRemaining` (ie including the invisible part of iceberg orders)?

peterbarrow (dismissed):

EVODelavega (dismissed):

datanode/api/__debug_bin3396181152
EVODelavega: when removing, make sure to rebase/squash to remove it from history entirely, or use `git filter-branch`,

core/risk/engine_test.go
witgaw: Well it's whatever the caller of the endpoint chooses is to be, but I guess `TrueRemaining` might be better to encourage setting it as we'd anticipate.

witgaw: Renamed now

datanode/api/__debug_bin3396181152
witgaw: Thanks, good spot. Removed now.

EVODelavega: can you rebase the commits still? This just adds a commit removing the file, it's still there in the history. An empty file doesn't increase the size of the repo significantly, but it's good practice to remove files that should've never been tracked to begin with

witgaw: Sure, squashed it all together now.

datanode/api/trading_data_v2.go
```diff
@@ -3370,13 +3411,43 @@ func (t *TradingDataServiceV2) computeMarginRange(
...
+			}
+		}
+		orderMargin = risk.CalcOrderMarginIsolatedMode(openVolume, bNonMarketOrders, sNonMarketOrders, positionFactor, marginFactor, auctionPrice)
+	}
+
+	worst := risk.CalculateMaintenanceMarginWithSlippageFactors(openVolume, bOrders, sOrders, marketObservable, positionFactor, linearSlippageFactor, quadraticSlippageFactor, riskFactors.Long, riskFactors.Short, fundingPaymentPerUnitPosition, auction)
+	best := risk.CalculateMaintenanceMarginWithSlippageFactors(openVolume, bOrders, sOrders, marketObservable, positionFactor, num.DecimalZero(), num.DecimalZero(), riskFactors.Long, riskFactors.Short, fundingPaymentPerUnitPosition, auction)
+
+	// margins := e.calculateMargins(m, marketObservable, *e.factors, false, auction, inc)
+	// margins.OrderMargin = calcOrderMargins(m.Size(), orders, e.positionFactor, marginFactor, auctionPrice)
```
jeremyletang, line 3446: ?

witgaw:
Need to revisit a test before merging, should be able to do that tomorrow afternoon, marking as draft for now.

jeremyletang (dismissed):

candida-d (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3890,28 +3890,50 @@ message OrderInfo {
...
-  LiquidationEstimate liquidation = 2;
+  LiquidationEstimate liquidation = 3;
+}
+
+// Estimates the additional funds needed to support the specified position
+message CollateralIncreaseEstimate {
+  // Estimate assuming slippage cap is applied
+  string worst_case = 1;
+  // Estimate assuming no slippage
+  string best_case = 2;
```
candida-d, line 3936: Shouldn't this also be reflected in the GraphQL?

witgaw: It should, I'll add now

datanode/gateway/graphql/resolvers.go
```diff
@@ -1348,8 +1360,9 @@ func (r *myQueryResolver) EstimatePosition(
 	}
 
 	return &PositionEstimate{
-		Margin:      resp.Margin,
-		Liquidation: resp.Liquidation,
+		Margin:                    resp.Margin,
+		CollateralIncreasEstimate: resp.CollateralIncreaseEstimate,
```
candida-d, line 1364:
```suggestion
		CollateralIncreaseEstimate: resp.CollateralIncreaseEstimate,
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -3340,6 +3350,8 @@ type OrderUpdate {
 type PositionEstimate {
   "Margin level range estimate for the specified position"
   margin: MarginEstimate!
+  "Estimated margin account balance increase"
+  collateralIncreasEstimate: CollateralIncreaseEstimate!
```
candida-d, line 3354:
```suggestion
  collateralIncreaseEstimate: CollateralIncreaseEstimate!
```

candida-d (changes requested):

datanode/gateway/graphql/schema.graphql
witgaw: Fixed

jeremyletang (approved):

candida-d (approved):

bglownia (approved):

jeremyletang (approved):

PR #10427, opened 2024-01-17
feat: add upgrade to check

### LINKED ISSUE #10400: [Feature]: Reverse direction of snapshot-migration version check
# Feature Overview

A function exists in core that allows migrations paths to be executed if and only if the snapshot we were loading was for a protocol-upgrade, and from a specific version:
```
if vgcontext.InProgressUpgradeFrom(ctx, "v0.73.12") {
    // do migration code
}
```
this gives us more control over when to do the upgrade, and also means once the upgrade has occurred the code paths will no longer execute and are harmless (ideally being removed later).

The problem is that the logic checks that we are upgrading _from_ a particular version. This means whenever a patch is made on mainnet, while the next minor release is in development we have to remember to go through and update the version string in all of these calls.

**What would be better** is if the check was switched to be and upgrade _to_ a particular version. Then we could use:
```
if vgcontext.InProgressUpgradeTo(ctx, "v0.74.0") {
    // do migration code
}
```
and not needed to worry about updating it if there is a mainnet patch.

**The catch is** that in our testing enviroments we do sometimes want to test migration to not-quite-the-next-minor-version. For example:
- on stagnet1 we protocol-upgrade to version `v0.74.0-preview.1
- on the snapshot compatibility pipeline we protocol-upgrade to version `v0.74+dev-[git-hash]

So we will need the code in `InProgressUpgradeTo()` to be able to treat those effectively as `v0.74.0`.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

EVODelavega:
Closes #10400
And
Closes #10438 10435

### DIFF
--- a/libs/context/context.go
+++ b/libs/context/context.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"strings"
 
+	vgversion "code.vegaprotocol.io/vega/version"
+
 	uuid "github.com/satori/go.uuid"
 )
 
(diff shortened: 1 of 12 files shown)


libs/context/context.go
```diff
@@ -158,3 +166,17 @@ func InProgressUpgradeFrom(ctx context.Context, from string) bool {
 	}
 	return from == si.version && si.upgrade
 }
+
+// InProgressUpgradeTo returns whether the data in the contexts tells us that the
+// node is restoring from a snapshot as part of an upgrade to a specific version.
+// This ignores suffixes (preview/rc/dev/commit hash) are ignored, only the semver part is checked.
+func InProgressUpgradeTo(ctx context.Context, to string) bool {
```
wwestgarth, line 173: Can you add some unittests in `context_test.go` just to be sure?

EVODelavega: No problem, done

wwestgarth (dismissed):

PR #10438, opened 2024-01-18
fix: block explorer database migration is slow

### LINKED ISSUE #10419: [Bug]: Very slow migration for the block explorer
The migration#0005 for the block explorer: https://github.com/vegaprotocol/vega/blob/develop/blockexplorer/store/migrations/0005_add_block_height_to_tx_result.sql

Takes a very long time on the stagnet1. 

At the moment We have 9 MLN rows in stagnet1 (3x less than We have on the mainnet).
The migration takes 16 hours already and the database is very very slow.

```
postgres=# SELECT pid, age(clock_timestamp(), query_start), usename, query
FROM pg_stat_activity
WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%'
ORDER BY query_start desc;
   pid   |       age       |  usename   |                query
---------+-----------------+------------+--------------------------------------
 2200902 |                 | postgres   |
 2200899 |                 |            |
 2200901 |                 | postgres   |
 2200904 |                 | postgres   |
 2200903 |                 | postgres   |
 2200896 |                 |            |
 2200898 |                 |            |
 2200897 |                 |            |
 2202674 | 00:00:16.588972 | monitoring | SELECT * FROM pg_stat_bgwriter;
 3280199 | 16:13:54.483036 | postgres   | UPDATE tx_results                   +
         |                 |            | SET block_height=b.height           +
         |                 |            | FROM blocks b                       +
         |                 |            | WHERE b.rowid = tx_results.block_id;
```

# Expected behaviour

We should not update all the rows at the same time, We should do it in somehow in the batches. to let the service start and use users use the block explorer.

The long migration is not a problem with our testnets, but We cannot stop the service for 20 hours in the mainnet.

guoguojin:
Step one, this is the first step to solving the issue, we are just moving the existing tx_results in order to back it up, and allow tendermint to continue populating a new tx_results table.

Once everything is up and running, a second PR will add a background migration process that will move the existing data from the backup to the new tx_results table giving us a full history in block explorer.

### DIFF
--- a/blockexplorer/store/migrations/0005_add_block_height_to_tx_result.sql
+++ b/blockexplorer/store/migrations/0005_add_block_height_to_tx_result.sql
@@ -63,7 +63,7 @@ BEGIN
     WHERE b.rowid = NEW.block_id
       AND tx_results.rowid = NEW.rowid;
 
-    RETURN NULL;
+    RETURN NEW;
 END;
 $$;
 -- +goose StatementEnd
--- a/blockexplorer/store/migrations/0005_add_block_height_to_tx_result.sql
+++ b/blockexplorer/store/migrations/0005_add_block_height_to_tx_result.sql
@@ -15,56 +15,17 @@ ALTER INDEX tx_results_tx_hash_index RENAME TO tx_results_old_tx_hash_index;
 ALTER INDEX tx_results_submitter_block_id_index_idx RENAME TO tx_results_old_submitter_block_id_index_idx;
 ALTER INDEX tx_results_cmd_type_block_id_index RENAME TO tx_results_old_cmd_type_block_id_index;
 
--- We need to make sure the next value in the rowid serial for the new tx_results table
--- continues where the old one leaves off otherwise we will break foreign key constraints
--- in the events table which we have had to drop temporarily and will restore once all the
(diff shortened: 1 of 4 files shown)


EVODelavega (dismissed):

EVODelavega (dismissed):

blockexplorer/store/migrations/0005_add_block_height_to_tx_result.sql
```diff
@@ -5,38 +5,65 @@
...
+    UPDATE tx_results
+    SET block_height=b.height
+    FROM blocks b
+    WHERE b.rowid = NEW.block_id
+      AND tx_results.rowid = NEW.rowid;
 
+    RETURN NULL;
+END;
+$$;
+-- +goose StatementEnd
```
EVODelavega, line 69: So I don't see a down migration (I can understand why we wouldn't have one), but I do need to ask: with these updates to the tables/schema's, are we sure migrating down will not break other views and queries?

guoguojin:
Closing this PR in favour of https://github.com/vegaprotocol/vega/pull/10455 which contains the background migration handler too

PR #10455, opened 2024-01-19
Fix/10419 migrate old tx results

### LINKED ISSUE #10419: [Bug]: Very slow migration for the block explorer

guoguojin:
Closes #10419


blockexplorer/store/migrator.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return err
+	}
+
+	// loop through each date and migrate the data for the date
+	for _, d := range migrateDates {
+		// we're going to make use of temporary tables which are only visible to the session that created them
+		// we therefore have to use a single connection and make sure that the same connection is used for the
+		// migration process.
+		conn, err := m.pool.Acquire(context.Background())
```
EVODelavega, line 90: Actually, I'm not sure about the use of the background index here, shouldn't we pass in/use the same context with which we call `Migrate` in `blockexplorer.go`?

EVODelavega (dismissed):

EVODelavega (approved):

PR #10489, opened 2024-01-26
fix: get epoch by block query fixed

### LINKED ISSUE #10301: [Bug]: rest API: get epoch by block height is broken
### Problem encountered

According to the docs, [I should be able to pass a block height in to this API](https://docs.vega.xyz/mainnet/api/rest/data-v2/trading-data-service-get-epoch#get-epoch) and get back the relevant epoch.

### Observed behaviour

It always returns this, no matter the block height

```json
{
  "code": 5,
  "message": "NotFound error",
  "details": [
    {
      "code": 250001,
      "message": "failed to get epoch",
      "inner": "no resource corresponding to this id"
    }
  ]
}
```

### Expected behaviour

The epoch, e.g.:

```javascript
{
  "epoch": {
    "seq": "782",
    "timestamps": {
      "startTime": "1704131174989622000",
      "expiryTime": "1704217574989622000",
      "endTime": "0",
      "firstBlock": "29595975",
      "lastBlock": "0"
    },
    "validators": [
      // ...
    ]
  }
}

```

### Steps to reproduce

```bash
1. Submit an epoch request
```

```sh
curl -L -X GET 'https://api.n06.stagnet1.vega.rocks/api/v2/epoch?block=8998254' -H 'Accept: application/json'
```

```
{"code":5,"message":"NotFound error","details":[{"@type":"type.googleapis.com/vega.ErrorDetail","code":250001,"message":"failed to get epoch","inner":"no resource corresponding to this id"}]}
```

### Software version

0.74.0

### Failing test

I suspect there is no api test

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
Closes #10301


jeremyletang (dismissed):

datanode/sqlstore/epochs.go
```diff
@@ -80,10 +80,10 @@ func (es *Epochs) Get(ctx context.Context, ID uint64) (entities.Epoch, error) {
 
 func (es *Epochs) GetByBlock(ctx context.Context, height uint64) (entities.Epoch, error) {
 	defer metrics.StartSQLQuery("Epochs", "GetByBlock")()
-	query := `WITH epochs_current AS (SELECT DISTINCT ON (id) * FROM epochs ORDER BY id, vega_time DESC)
-		SELECT e.id, e.start_time, e.expire_time, e.end_time, e.tx_hash, e.vega_time, bs.height first_block, be.height last_block FROM epochs_current AS e
+	query := `SELECT e.id, e.start_time, e.expire_time, e.end_time, e.tx_hash, e.vega_time, bs.height first_block, be.height last_block FROM epochs AS e
```
wwestgarth, line 83: What was broken about it? Because there is a unittest for this call which was passing and is still passing. Also I think as it stands if you query a block from the current in-progress epoch it'll still return `no resource corresponding to this id`.

ze97286: that was an oversight. I tested it with some block on the database and incidentally it was a block for an epoch that has already finished. The unit test was making the same mistake so it didn't fail. Fixed both, thanks.

jeremyletang (approved):

PR #10521, opened 2024-01-31
feat: replicate perp termination bug on mainnet

Jiajia-Cui:


Jiajia-Cui:
close it, because of the "feature" in feature test, I am not able to reproduce the bug

PR #10580, opened 2024-02-07
feat: implement order matching for hybrid AMM and CLOB

### LINKED ISSUE #10538: [Feature]: AMM matching in continuous trading
# Feature Overview

Part of: https://github.com/vegaprotocol/vega/issues/8228

**In order for** orders to be placed against an AMM while in continuous trading
**We will** integrate the pool logic into the orderbook
**So that** when orders are submitted to Vega they are matched against both the orderbook and any AMM's with volume in range

## Specs
- [Link](https://github.com/vegaprotocol/specs/issues/1990) to spec or section within a spec

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
close #10538 

note: this is merging into the AMM feature branch


core/execution/spot/market.go
```diff
@@ -273,9 +273,18 @@ func NewMarket(
...
 }
 
+// some dummy offbook source to stop the spot tests falling over, but we should revisit when we revive spots.
+type offbook struct{}
+
+func (ob *offbook) SubmitOrder(_ *types.Order, _, _ *num.Uint) []*types.Order {
+	return nil
+}
+func (ob *offbook) NotifyFinished() {}
+
```
jeremyletang, line 287: Spot are being revived now, but without AMM support. let's make sure this is good enough to run in production.

core/matching/side.go
```diff
@@ -409,6 +416,12 @@ func (s *OrderBookSide) fakeUncross(agg *types.Order, checkWashTrades bool) ([]*
 			checkPrice = func(levelPrice *num.Uint) bool { return levelPrice.GTE(agg.Price) }
 		}
 
+		// first check for volume between the theoretical best price and the first price level
+		_, oo := s.uncrossOffbook(len(s.levels)-1, fake, idealPrice, true)
```
jeremyletang, line 420: these functions calls do not commit any state to the pools right?

core/execution/spot/market.go
wwestgarth: It think it will be fine, there was a feature test that went through it that is still passing. We'll want to rejected `create-amm` transactions to spot markets anyway, so I'll make a ticket for that and to make sure nothing else weird might happen when spots are about.

core/matching/side.go
wwestgarth: No state change, just returns some orders. The state will only change later down the line when we're in `handleConfirmation()` and the position engine gets updated when the trades are finalised.

PR #10592, opened 2024-02-08
fix: game api and team api amounts in quantum

### LINKED ISSUE #10567: Rewards / teams / games API disagree on numbers
## Issue Summary

We are seeing some discrepancy in testing on the APIs.

This image shows the rewards page which is based on accounts API and rewards API.  You can see iif shows that this key earned 37.19 tDAI.

image

Looking at the games view, which takes its numbers from the `games` API you can see it thinks the team earned 74 .. which is 37 x 2 so adds up.

image

However, looking at the members info and at the team profile at the top, you can see this view thinks the two kkeys both earned 39 tDAI, to give a total of 78 between them at the top.  These numbers are taken from the `teamStatistics` and `teamMemberStatstics` APIs.

## Fix

Given that the actual Vega reward pot (top left of first image) is taken from an account balance it seems that 37.19 is accurate,, which suggests the `games` API is accurare and there is an issue on the `teamStatistics` and `teamMemberStatistics` APIs.

### LINKED ISSUE #10578: Game API reward amount issue
## Issue

Currently we are showing a `rewardEarned` and I dont really know what the point of this is.

Per the docs it is "The rewards earned by the individual during the epoch"

But as you can see it doesnt match the value in the `totalRewardEarned`.

image

I wondered if this was because of a quantum conversion but quantum is one for these assets so its not that.  Also I dont know why it has more decimals....they look like it is in asset decimals but since the asset ID isnt given on this API you cant do anything with that.  The definition also implies maybe it is the reward earned iin the epoch across ALL GAMES?  If that is true it doesnt make sense.

## More Info

We have two sets of matching rewards open currently 

image

I am the only team that traded so I think I should get all rewards... 

image

So why does one team have totalRewardsEarned 4667 and one has 2000?
The second one just seems to be the 2000000000 adjusted for asset decimals
But the first one sounds like we got 2000000000 this game / epoch .... so I dont know what the 4667 is telling me?

You can see on the rewards API (which is what this rewards page uses) that I have indeed earned all those rewards .... 2000 of each asset. 

image

## Proposal

This actually could be quite useful, we should:

- Show the reward amount in terms of the actual asset in THIS game for THIS entity
- Give the asset ID of the reward so that we can do the conversion

guoguojin:
Closes #10567 
Closes #10578


jeremyletang (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4666,6 +4668,10 @@ message TeamGameEntity {
   string reward_earned = 5;
   // Total rewards earned by the individual for the team
   string total_rewards_earned = 6;
+  // Reward earned by the team in quantum value
```
candida-d, line 4671:
```suggestion
  // Rewards earned by the team in quantum value
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4692,6 +4698,10 @@ message IndividualGameEntity {
   string reward_earned = 5;
   // Total rewards earned by the individual for the game
   string total_rewards_earned = 6;
+  // Reward earned by the individual in quantum value
```
candida-d, line 4701:
```suggestion
  // Rewards earned by the individual in quantum value
```

candida-d (approved):

jeremyletang (approved):

PR #10626, opened 2024-02-12
Ensure chain event deduplication account for missing chain ID

### LINKED ISSUE #10624: Ensure chain event deduplication account for missing chain ID
# Algorithm

- Hash with Chain ID
   - if seen, deduplicate
   - if not seen, hash without chain ID
      - if seen, deduplicate
      - if not seen, track event hash with with chain ID.

ValentinTrinque:
Close #10624


candida-d (approved):

core/banking/deduplicate.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// First time we see this transaction, so we keep track of it.
+	e.seenAssetActions.Add(refKey)
+
+	return nil
+}
+
+func buildRefKey(ref snapshot.TxRef) (string, error) {
+	buf, err := vgproto.Marshal(&ref)
```
wwestgarth, line 72: Do we know for sure 100% that that the marshalling of `message TxRef {}` _before the chain-id field existed_ is the same as it existing but an empty string?

ValentinTrinque: Yes, I tested it locally, it does generate the same hash, even before the field existed.

There is no real way to test this sadly...

wwestgarth: Ok I tested it and I think it does, so thats good.

wwestgarth: Ok thats great! I just want to doubly doubly make sure. If we've both been able to see it come out the same locally then we are golden.

PR #10653, opened 2024-02-13
Integrate SLA and ELS to AMM

### LINKED ISSUE #10541: [Feature]: AMM SLA integration
# Feature Overview

**In order for** a party with an active AMM to meet their liquidity-provision commitment
**We will** extend the logical in the liquidity engine
**So that** the AMM's volume is considering in the SLA penalty calculations

Bit of a discussion:
https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1706880490090979

## Specs
- [Link](https://github.com/vegaprotocol/specs/issues/1990) to spec or section within a spec

### LINKED ISSUE #10543: [Feature]: AMM equity-like-shares and liquidity-fee settings
# Feature Overview

**In order for** for a party with an active AMM to gain ELS for the liquidity they have provided
**We will** integrate logic
**So that** the party is rewarded for being an early MM

Also wire up the "proposed_fee" field into wherever its supposed to end up:
https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1706880490090979

## Specs
- [Link](https://github.com/vegaprotocol/specs/issues/1990) to spec or section within a spec

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

karlem:
close https://github.com/vegaprotocol/vega/issues/10541
close https://github.com/vegaprotocol/vega/issues/10543
closes https://github.com/vegaprotocol/core-test-coverage/issues/971


core/execution/common/liquidity_provision_fees.go
```diff
@@ -84,6 +85,8 @@ func (m *MarketLiquidity) AllocateFees(ctx context.Context) error {
 	scoresPerLp := m.liquidityEngine.GetAverageLiquidityScores()
 	// Multiplies each equity like share with corresponding score.
 	updatedShares := m.updateSharesWithLiquidityScores(sharesPerLp, scoresPerLp)
+	// now adjust the score to account for the AMM share
+	updatedShares = m.mergeScores(updatedShares, ammScores)
```
wwestgarth, line 89: From my reading of the spec, this new network parameter isn't strictly related to AMM's. 

Say the network parameter is 0.5, it means half the fees are distrubuted calcuated with the current way by doing `ELS * LP-score`, and the other half are distributed using `LP-score` only. Theres no mention of amm's in that part of the spec.

EVODelavega: Indeed, this has changed a bit, but we are going to have to calculate the LP scores for the AMMs, and add them to the map (which I'm now doing).

core/execution/amm/pool.go
```diff
@@ -468,14 +521,10 @@ func (p *Pool) fairPrice() *num.Uint {
 // virtualBalances returns the pools x, y values where x is the balance in contracts and y is the balance in asset.
 func (p *Pool) virtualBalances(pos int64, ae *num.Uint, side types.Side) (num.Decimal, num.Decimal) {
 	switch {
-	case pos < 0:
-		return p.virtualBalancesShort(pos, ae)
-	case pos > 0:
-		return p.virtualBalancesLong(pos, ae)
-	case side == types.SideBuy:
+	case pos < 0, side == types.SideBuy:
```
wwestgarth, line 524: So it was like how it was before because only want to consider the side if position is `0`. If the position is non-zero we already know which curve to look at and we always want to use that one. When its `0` we use the incoming order side to work out which curve the AMM will move into, so we only want to do those cases after `pos > 0, pos <0` haven't been hit.

EVODelavega: I'm not sure if this is going to be functionally different, but will fix to be safe

wwestgarth (approved):

PR #10674, opened 2024-02-14
feat: add AMM pools events API

### LINKED ISSUE #10557: [Feature]: AMM events and API (pool configuration and sub accounts)
Feature Overview
In order to provide a view of pool configuration and sub accounts
We will emit events from core, and provide API
So that front-end and do their magic and provide a good user experience

Specs
https://github.com/vegaprotocol/specs/issues/1990 to spec or section within a spec

guoguojin:
Closes #10557


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4774,3 +4781,36 @@ message PartyMarginMode {
...
+message ListAMMPoolsRequest {
+  // Party ID to filter for.
+  optional string party_id = 1;
+  // Market ID to filter for. If party ID is provided, the market filter is ignored
+  optional string market_id = 2;
+  // Pool ID to filter for. If party ID or market ID is provided, the pool ID filter is ignored.
+  optional string pool_id = 3;
+  // Sub-account to filter for. Used in conjunction with the pool id filter only.
+  // If party ID is not provided, the sub-account filter is ignored.
+  optional string sub_account = 4;
```
wwestgarth, line 4794: I think I've been using the term "pool" wrong in all of this and it has been causing confusion (apologies). This reads like we're expecting one AMM, that multiple sub-accounts add to. How it actually works is that one person can create one AMM per market, that AMM gets a sub-account and thats it. If the pool is cancelled, then another is created by the same person, the AMM will get a new ID but will have the same sub-account as before.

So can we remove the restriction that you need to set a pool-id to be able to filter by sub-account? I was hoping this API could be use to answer the question "is this party a sub-account of any pools" (because sub-account is actually just a party-id the AMM acts as).

guoguojin: No worries, so:

  - the sub-account is tied to the actual party who creates the AMM 
  - the AMM sub-account is created after a party has created an AMM for the first time 
  - regardless of what AMM a party creates, they will all be linked to the same AMM sub-account

Does that sound right?

wwestgarth: yup that is all correct!

wwestgarth (approved):

peterbarrow (approved):

EVODelavega (approved):

PR #10679, opened 2024-02-15
fix: clone internal composite price before scaling for API

### LINKED ISSUE #10676: [Bug]: Snapshot failure on fairground pipeline
### Problem encountered

On the fairground snapshot pipeline this morning a run failed with:
```
CONSENSUS FAILURE!!!	{"err": "+2/3 committed an invalid block: wrong Block.Header.AppHash.  Expected 38EBE07AD5275AA7B2582F9223781EF879F8BE80E105C913835F0595FA46FB24, got A23CF05E9E06D8AED2792BB6335089421DF5BB4E09CF6CFA269CB599E8EF5BDD"
```

https://jenkins.vega.rocks/blue/organizations/jenkins/private%2FSnapshots%2FFairground/detail/Fairground/14681/pipeline/258

### Observed behaviour

Node fell out of consensus when starting from a snapshot.

### Expected behaviour

Node does not fall out of consensus when starting from a snapshot.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.74.1

### Failing test

_No response_

### Jenkins run

https://jenkins.vega.rocks/blue/organizations/jenkins/private%2FSnapshots%2FFairground/detail/Fairground/14681/pipeline/258

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
close #10676 

The internal-composite price is scaled before being added to the market data, but it wasn't cloned beforehand which meant it was changing the price in the actual market-price calculator too. So each block it would get chopped down until it eventually reached zero:
https://api.n00.testnet.vega.rocks/api/v2/market/data/6a6520fb81106f127411bf35cfc16210992d6e82eb66b6213f51fdccf3f8d353/latest

This causes issues with snapshot because, for whatever reason, we call GetMarketData() after loading so the snapshot version of node will have done an one extra scaling and it gets out of sync.

### DIFF
--- a/core/execution/common/mark_price.go
+++ b/core/execution/common/mark_price.go
@@ -255,6 +255,9 @@ func (mpc *CompositePriceCalculator) CalculateBookMarkPriceAtTimeT(initialScalin
 }
 
 func (mpc *CompositePriceCalculator) GetPrice() *num.Uint {
+	if mpc.price != nil {
+		return mpc.price.Clone()
+	}
 	return mpc.price
 }
 
--- a/core/integration/features/settlement/0019-MCAL_fundingMargin5.feature
+++ b/core/integration/features/settlement/0019-MCAL_fundingMargin5.feature
@@ -131,6 +131,6 @@ Feature: check when settlement data precision is different/equal to the settleme
       | market | party2 | ACCOUNT_TYPE_SETTLEMENT | ACCOUNT_TYPE_MARGIN     | ETH/DEC19 | 432741366 | USD   |
       | market | party3 | ACCOUNT_TYPE_SETTLEMENT | ACCOUNT_TYPE_MARGIN     | ETH/DEC19 | 432741368 | USD   |
       | market | aux2   | ACCOUNT_TYPE_SETTLEMENT | ACCOUNT_TYPE_MARGIN     | ETH/DEC19 | 432741366 | USD   |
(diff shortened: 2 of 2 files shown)


jeremyletang (dismissed):

core/execution/future/market.go
```diff
@@ -761,7 +761,7 @@ func (m *Market) GetMarketData() types.MarketData {
 			if internalCompositePrice == nil {
 				internalCompositePrice = num.UintZero()
 			} else {
-				internalCompositePrice = m.priceToMarketPrecision(internalCompositePrice)
+				internalCompositePrice = m.priceToMarketPrecision(internalCompositePrice.Clone())
```
EVODelavega, line 764: Wouldn't it make more sense to have `CompositePriceCalculator.GetPrice()` return a cloned value? Relying on the caller to clone the value feels like we're not addressing the root cause

wwestgarth: Good point. That method doesn't exist in `v0.74.1` which I was debugging in, I'll update.

EVODelavega (dismissed):
LGTM - although with `GetPrice` returning a cloned value, there's some stuff in `BlockEnd` (market.go) that probably can be optimised a bit, but that's beyond the scope of this fix

jeremyletang (approved):

PR #10697, opened 2024-02-16
chore: add block interval for ethcall Polling

### LINKED ISSUE #10698: Restrict amount of L2 RPC calls
Update the network parameter to support a block interval at which to call for new blocks

jeremyletang:
close #10698


protos/sources/vega/vega.proto
```diff
@@ -1524,6 +1524,9 @@ message EthereumL2Config {
   uint32 confirmations = 3;
   // Display name of this network
   string name = 4;
+  // Polling interval for the internal engine, need to be adapted
+  // to the source chain depending of it's block production rate.
```
candida-d, line 1528:
```suggestion
  // Polling interval for the internal engine, needs to be adapted
  // to the source chain depending on its block production rate.
```

candida-d (dismissed):

wwestgarth (dismissed):

guoguojin (approved):

PR #10706, opened 2024-02-19
Fixes amm

wwestgarth:
Three little niggles found while writing basic system tests for AMM.

### DIFF
--- a/core/collateral/engine.go
+++ b/core/collateral/engine.go
@@ -3652,7 +3652,7 @@ func (e *Engine) SubAccountRelease(
 	ctx context.Context,
 	party, subAccount, asset, market string,
 	pos events.MarketPosition,
-) (*types.LedgerMovement, events.Margin, error) {
+) ([]*types.LedgerMovement, events.Margin, error) {
 	ownerGeneral, subAccountGeneral, subAccountMargin, err := e.getSubAccounts(subAccount, party, asset, market)
 	if err != nil {
 		return nil, nil, err
@@ -3669,7 +3669,6 @@ func (e *Engine) SubAccountRelease(
 			marginShortFall: num.UintZero(),
 		}
 	}
-
 	// one transfer from general to general
 	treq := &types.TransferRequest{
 		Amount:      subAccountGeneral.Balance.Clone(),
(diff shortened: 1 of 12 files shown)


core/collateral/engine.go
```diff
@@ -3526,22 +3526,13 @@ func (e *Engine) CreatePartyAMMsSubAccounts(
...
 
-	_, err = e.CreatePartyGeneralAccount(ctx, subAccount, asset)
+	generalID, err := e.CreatePartyGeneralAccount(ctx, subAccount, asset)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	_, err = e.CreatePartyMarginAccount(ctx, subAccount, market, asset)
+	marginID, err := e.CreatePartyMarginAccount(ctx, subAccount, market, asset)
 	if err != nil {
```
jeremyletang, line 3536: Any chance that would be doing something weird in the case a party had created an amm, cancelled it, creating another one again?

core/collateral/engine.go
```diff
@@ -3526,22 +3526,13 @@ func (e *Engine) CreatePartyAMMsSubAccounts(
...
-		return nil, nil, errors.New("general sub account already exists")
-	}
-
-	marginID := e.accountID(market, subAccount, asset, types.AccountTypeMargin)
-	if _, ok := e.accs[marginID]; ok {
-		return nil, nil, errors.New("general sub account already exists")
-	}
 
-	_, err = e.CreatePartyGeneralAccount(ctx, subAccount, asset)
+	generalID, err := e.CreatePartyGeneralAccount(ctx, subAccount, asset)
```
EVODelavega, line 3530: Is it not possible for a party to submit an AMM, cancel it, and submit a second one later on? In that case, it feels a bit off to call `CreatePartyGeneralAccount` and `CreatePartyMarginAccount`

EVODelavega (approved):

core/collateral/engine.go
wwestgarth: I have a test that does exactly that and its all fine. Both `CreatePartyGeneralAccount()` and `CreatePartyMarginAccount()` just return the ID if the account already exists, doesn't try to re-create or anything

jeremyletang: sounds good then, my bad I didn't catch it in the test.

EVODelavega (approved):

PR #10742, opened 2024-02-23
fix: ledger entries api takes ages to return data when filtering by t…

### LINKED ISSUE #10374: Transfer API shows fees but why no payments?
> [!IMPORTANT]  
> Edit: see ticket history for the previous request

There is currently no way for the Explorer to fetch all of the ledger entries related to a recurring transfer. THis makes it impossible to show on the Network Treasury page all of the occurences of a transfer.

# Tasks
- [ ] Update `ledger` table to correctly store transfer ID for recurring transfers. Currently it is `null`.
- [ ] Update `LedgerEntries` API to accept a `TransferId` which will narrow the ledger entries by the txId field.
  - [ ] Ensure change is in GRPC API
  - [ ] Ensure change is in REST API
  - [ ] Ensure change is in GraphQL API

guoguojin:
…ransfer id

I have taken the route of fixing the ledger entries API rather than changing the transfers API. 

The issue with the ledger entries API was that if a transfer id was provided in the filter, only that would be used in order to filter the ledger entries. Unfortunately ledger entries is an incredibly large table with lots of rows, and the query must scan the entire table in order to locate a single entry.

Adding an index for the transfer id took approximately 8m9s on staging, however while this does improve performance, it still has to scan hypertable chunks in order to find a transfer id and the API still takes much too long to complete.

It also doesn't make sense for the API to ignore other filter inputs when a transfer id is provided. It would make more sense for a transfer id to be found within a given start/end time, or for a specific asset, or specific party.

Given these reasons, I've altered the filtering so that the transfer id is used along with all other filter inputs. I have tested the query on staging and the performance of the query is not impacted if there is no index on the transfer id when a date range, or party is provided.


guoguojin:
I haven't added an index for the transfer id column on the ledger table yet in this PR opting to get comments about it in the review process. The reason being in my testing, I didn't see a performance boost with the index as long as the other filters, i.e. party/asset/data range etc. are used in conjunction with the transfer id.

Happy to add it, but on staging adding the index took over 8m, so this is something to consider as it will have an impact when it is released.

guoguojin:
Also, I think having to use the transfer id to join the ledger to add the payments details to the transfer API would be detrimental to the transfers API, the index would definitely be required, and I'm not sure how the performance will be impacted when joining the transfers/transfer fees and ledger entries table.

edd:
This PR doesn't close #10374, but it does address my comment below the ticket. This is enough to unblock some aspects of #5841.

guoguojin:
I have also updated the ledger entries API to expose the transfer id on the aggregated ledger entries, the data is queried from the database, but for some reason the API was not updated so that the data was returned.

EVODelavega (dismissed):

candida-d (approved):

PR #10768, opened 2024-02-27
Prevent reduce only iceberg orders

peterbarrow:
Currently reduce only iceberg orders are blocked at the wallet level. This change blocks them in the core to prevent issues with people using their own wallet software.

Covers 0014-ORDT-090


peterbarrow:
This functionality is not needed, this is just a quirk of the feature test system. In a real system the validation is performed at CheckTx time and the order will be rejected correctly. Closing PR.

PR #10776, opened 2024-02-27
Add AMM scope for rewards

### LINKED ISSUE #10616: [Feature]: Add AMM entity-Scope in rewards metrics
# Feature Overview

**In order to** allow a reward metric that transfers to owners of AMM's
**We will** implement a new entity scope
**So that** a reward metric can be set up that is distributed only between AMMs

Spec change:
https://github.com/vegaprotocol/specs/pull/2019/commits/3ad22682f0d912053b50a56033757bce20111f84

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

karlem:
Close #10616


core/execution/common/market_activity_tracker.go
```diff
@@ -530,6 +555,18 @@ func (mat *MarketActivityTracker) getAllParties(asset string, mkts []string) map
 	return parties
 }
 
+func (mat *MarketActivityTracker) getAllParties(asset string, mkts []string) map[string]struct{} {
+	return mat.filterParties(asset, mkts, func(mt *marketTracker) map[string]struct{} {
```
ze97286, line 559: why pass a function that returns a map rather than just the map?

protos/sources/vega/vega.proto
```diff
@@ -1095,6 +1095,9 @@ enum IndividualScope {
   INDIVIDUAL_SCOPE_IN_TEAM = 2;
   // All parties that are not part of a team are within the scope of this reward.
   INDIVIDUAL_SCOPE_NOT_IN_TEAM = 3;
+  // All keys representing AMM parties, excluding those directly controlled by parties
+  // are within the scope of this reward.
```
candida-d, line 1099:
```suggestion
  // All keys representing AMM parties are within the scope of this reward.
```

core/execution/common/market_activity_tracker.go
karlem: Because it's nested inside right?

PR #10797, opened 2024-03-01
feat: cancel pegged orders when tick size changes and mismatches the …

### LINKED ISSUE #10778: [Feature]: cancel pegged orders when updating tick size
## Current

Currently the spec says for pegged orders.

- Round pegged buy orders **UP** to the nearest market tick.
- Round pegged sell orders **DOWN** to the nearest market tick.

This ensured the smallest possible spread (1 tick) can be achieved with pegged orders.

However as orders are not cancelled if the markets tick size is updated through governance and orders are only validated on submission. If the tick size is increased, we can enter a situation where pegged orders now cross.

## Change

When the tick size is updated, all existing pegged orders are checked. If the offset of an order is no longer an integer multiple of the tick size, the order is canceled.

# Additional Details

Spec PR
https://github.com/vegaprotocol/specs/pull/2191

ze97286:
…offset

Closes #10778

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -581,6 +581,8 @@ func (m *Market) GetSettlementAsset() string {
 }
 
 func (m *Market) Update(ctx context.Context, config *types.Market, oracleEngine products.OracleEngine) error {
+	tickSizeChanged := config.TickSize.NEQ(m.mkt.TickSize)
+
 	config.TradingMode = m.mkt.TradingMode
 	config.State = m.mkt.State
 	config.MarketTimestamps = m.mkt.MarketTimestamps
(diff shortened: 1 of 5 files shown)


core/execution/future/market.go
```diff
@@ -1326,9 +1340,15 @@ func (m *Market) getNewPeggedPrice(order *types.Order) (*num.Uint, error) {
 	offset := num.UintZero().Mul(order.PeggedOrder.Offset, m.priceFactor)
 	if order.Side == types.SideSell {
 		price = price.AddSum(offset)
+		// this can only happen when pegged to mid, in which case we want to round to the nearest *better* tick size
+		// but this can never cross the mid by construction as the the minimum offset is 1 tick size and all prices must be
+		// whole multiples of tick size.
 		if !num.UintZero().Mod(price, m.mkt.TickSize).IsZero() {
-			price = price.Div(price.AddSum(m.mkt.TickSize), m.mkt.TickSize)
-			price = price.Mul(price, m.mkt.TickSize)
+			price.Div(price, m.mkt.TickSize)
```
wwestgarth, line 1347: I think you can just subtract the result from `.Mod()` from the price instead of dividing and multiplying. Or expose `.DivMod()` and just use the quotient/remainder from that in a single call for the checks.

core/execution/future/market.go
```diff
@@ -638,7 +640,19 @@ func (m *Market) Update(ctx context.Context, config *types.Market, oracleEngine
 		// market is settled, unsubscribe all
 		m.tradableInstrument.Instrument.Unsubscribe(ctx)
 	}
-
+	if tickSizeChanged {
+		peggedOrders := m.matching.GetActivePeggedOrderIDs()
+		peggedOrders = append(peggedOrders, m.peggedOrders.GetParkedIDs()...)
```
jeremyletang, line 645: Why get the parked orders here if we just continue if the order is not in the orderbook after?

jeremyletang (approved):

PR #10809, opened 2024-03-04
fix: save history of seen ethereum verifier events

### LINKED ISSUE #10811: [Bug]: Snapshot is not saving history of seen ethereum verifier events
### Problem encountered

During a snapshot test is was noticed that some of the state in the staking engine and ETH verifier engine used to deduplicate events (from ethereum or L2s) was not being saved in the snapshot as they should.

### Observed behaviour

State is not being saved in the snapshot

### Expected behaviour

State should be saved in the snapshot

### Software version

0.74.7

### Jenkins:

https://jenkins.vega.rocks/blue/organizations/jenkins/private%2FSnapshots%2FFairground/detail/Fairground/15957/pipeline/254

wwestgarth:
close https://github.com/vegaprotocol/vega/issues/10811


core/datasource/external/ethverifier/verifier_snapshot.go
```diff
@@ -126,14 +156,22 @@ func (s *Verifier) LoadState(ctx context.Context, payload *types.Payload) ([]typ
...
+		s.restoreMisc(ctx, pl.Misc)
 		return nil, nil
 	default:
 		return nil, types.ErrUnknownSnapshotType
 	}
 }
 
 func (s *Verifier) OnStateLoaded(ctx context.Context) error {
+
+	if vgcontext.InProgressUpgradeFrom(ctx, "v0.74.8") {
```
jeremyletang, line 171: this will need to be 74.7, the version on mainnet is 74.7 and we will likely skip 74.8 in favour of a patch with this code

PR #10788, opened 2024-02-29
chore: add feature tests for 0011-MARA

guoguojin:
Closes https://github.com/vegaprotocol/core-test-coverage/issues/788
Closes https://github.com/vegaprotocol/core-test-coverage/issues/789
Closes https://github.com/vegaprotocol/core-test-coverage/issues/790
Closes https://github.com/vegaprotocol/core-test-coverage/issues/791
Closes https://github.com/vegaprotocol/core-test-coverage/issues/792
Closes https://github.com/vegaprotocol/core-test-coverage/issues/793
Closes https://github.com/vegaprotocol/core-test-coverage/issues/794
Closes https://github.com/vegaprotocol/core-test-coverage/issues/795
Closes https://github.com/vegaprotocol/core-test-coverage/issues/796
Closes https://github.com/vegaprotocol/core-test-coverage/issues/797
Closes https://github.com/vegaprotocol/core-test-coverage/issues/798
Closes https://github.com/vegaprotocol/core-test-coverage/issues/799

PR #10814, opened 2024-03-04
feat: create ft for spot market 0011

Jiajia-Cui:


Jiajia-Cui:
close as its duplicated with https://github.com/vegaprotocol/vega/pull/10788/files

PR #10856, opened 2024-03-09
feat: add API for time weighted notional position

### LINKED ISSUE #10831: [API]: Add notionalTimeWeightedAveragePosition to some API
# API Overview

**In order to** ... determine if a user is eligible for a game
**We will build an API to**... expose the `notionalTimeWeightedAveragePosition` of a party
**So that** ... Console can compare to the `notionalTimeWeightedAveragePositionRequirement` on the `transfers` API to see if they are eligible

## Specs
[Spec name](URL) to spec or section within a spec

## API request details

- [ ] Create or extend an API that... provides the `notionalTimeWeightedAveragePosition` of a party 
- [ ] It will need to be aggregated over whatever time window Core is aggregating to determine this eligibility for rewards

  - [ ] Party ID

```

```

I would assume that all this information is known inside the Core since they are doing this comparison in order to see if a party should be given rewards.  So hopefully it is not a question of having to work something new out, rather exposing something existing out on some API ... via party would seem logical ... but leave that to implementation.

guoguojin:
Closes #10831


core/execution/future/market.go
```diff
@@ -2661,7 +2661,9 @@ func (m *Market) handleConfirmation(ctx context.Context, conf *types.OrderConfir
 
 		tradeEvts = append(tradeEvts, events.NewTradeEvent(ctx, *trade))
 		for _, mp := range m.position.Update(ctx, trade, conf.PassiveOrdersAffected[idx], conf.Order) {
-			m.marketActivityTracker.RecordPosition(m.settlementAsset, mp.Party(), m.mkt.ID, mp.Size(), trade.Price, m.positionFactor, m.timeService.GetTimeNow())
+			if twNotionalPosition := m.marketActivityTracker.RecordPosition(m.settlementAsset, mp.Party(), m.mkt.ID, mp.Size(), trade.Price, m.positionFactor, m.timeService.GetTimeNow()); twNotionalPosition != nil {
```
ze97286, line 2664: this means that for an epoch with no position change there will be none and for epoch with many position changes there will be plenty. I think that's an overkill. Why not just emit an event at the end/beginning of the epoch with the tw notional position for all?

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4780,3 +4790,36 @@ message PartyMarginMode {
...
+  // Epoch at which the time weighted notional position is calculated.
+  uint64 at_epoch = 3;
+  // Time weighted notional position for the party and asset.
+  string time_weighted_notional_position = 4;
+  // Vega time of the block in which the last position update occurred.
+  int64 last_updated = 5;
+}
+
+// Request to retrieve the time weighted notional position for a party and asset.
+// If no epoch is given, the most current value will be returned.
```
candida-d, line 4810: Depending on changes to functionality, could you update this (and other instances) if it becomes “last epoch” rather than live current?

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,22 @@ import "vega/vega.proto";
 
 option go_package = "code.vegaprotocol.io/vega/protos/vega/events/v1";
 
+// Time weighted notional position update for the current epoch.
+// The time weighted notional position is used to determine whether
+// a party is eligible for receiving rewards at the end of an epoch.
+// These updates are provided to facilitate knowing whether or not a party
+// is on track to qualify.
+message TimeWeightedNotionalPositionUpdated {
+  // Epoch the time weighted notional position is eligible for
```
candida-d, line 22:
```suggestion
  // Epoch the time weighted notional position is relevant for
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,22 @@ import "vega/vega.proto";
...
+  uint64 epoch_seq = 1;
+  // Asset ID the position is for
```
candida-d, line 24:
```suggestion
  // Asset ID the position is denominated in
```

core/execution/future/market.go
guoguojin: Thanks, as advised I've added to the market activity tracker and published from there at the start of each epoch. The way I've done it, the rewards calculation will automatically calculate the time weighted notional positions anyway, so I've stored those values when they're calculated, and then at the start of the next epoch, I create the time weighted notional positions and set those for the epoch at the start of each epoch.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 - [10392](https://github.com/vegaprotocol/vega/issues/10392) - Integrate mark price with price monitoring.
 - [10845](https://github.com/vegaprotocol/vega/issues/10845) - Remove price monitoring trigger staleness check.
 - [10810](https://github.com/vegaprotocol/vega/issues/10810) - Do not update mark price from book during auctions.
+- [10831](https://github.com/vegaprotocol/vega/issues/10831) - Add API for notionalTimeWeightedAveragePosition.
```
gordsport, line 34:
```suggestion
- [10831](https://github.com/vegaprotocol/vega/issues/10831) - Add API for `notionalTimeWeightedAveragePosition`.
```

core/execution/common/market_activity_tracker.go
```diff
@@ -387,6 +394,14 @@ func (mat *MarketActivityTracker) OnEpochEvent(ctx context.Context, epoch types.
 		mat.clearDeletedMarkets()
 		mat.clearNotionalTakerVolume()
 		mat.takerFeesPaidInEpoch = map[string]map[string]map[string]*num.Uint{}
+		var twNotionalPositionEvents []events.Event
+		for asset := range mat.assetToPartyTWNotionalPositionCache {
```
ze97286, line 398: 1. why not iterate the key,value directly rather than look up every key?
2. this will generate events in a non deterministic order
3. why do this caching on end epoch and emit the event on new epoch rather than just emit the event on end epoch?

guoguojin: So to answer number 3 there's two reasons why I did the caching:

1. At the end of an epoch, the rewards OnEpochEvent already calls the calculations for the time weighted notional position so I saw no reason to re-run the calculations by two separate processes. It made more sense to use the data after it was created on the start of the next epoch.
2. The calculation aggregates across all markets of an asset for a given party, this information, i.e. the markets a party is involved in doesn't appear to be available in the market activity tracker.

Sending the previously cached calculated values on the next epoch I know the data will have been calculated at the end of the previous epoch when the rewards are calculated so I can guarantee they are the most recent values for the next epoch, if that makes sense.

I'll rework it so that the actual events to be sent are cached when the twNotionalPosition is calculated.

core/execution/common/market_activity_tracker.go
```diff
@@ -387,6 +400,12 @@ func (mat *MarketActivityTracker) OnEpochEvent(ctx context.Context, epoch types.
 		mat.clearDeletedMarkets()
 		mat.clearNotionalTakerVolume()
 		mat.takerFeesPaidInEpoch = map[string]map[string]map[string]*num.Uint{}
+		var twNotionalPositionEvents []events.Event
+		for _, update := range mat.twNotionalUpdateCache {
```
ze97286, line 404: you should reset mat.twNotionalUpdateCache after emitting the events batch.

ze97286 (commented):
I think you're missing conversion code for the new event in datanode/broker/convert.go

core/execution/common/market_activity_tracker.go
ze97286: I don't see that sorry. I'm not suggesting you rerun the calculation. You're only adding to the cache once, at which point I'm asking why not emit the event rather than add to the cache and process the cache later?

guoguojin: I see what you're saying now, yes. Sorry, I was being thick, I'll fix it.

datanode/gateway/graphql/schema.graphql
```diff
@@ -7259,3 +7269,16 @@ type PartyMarginMode {
   "Epoch at which the update happened."
   atEpoch: Int!
 }
+
+type TimeWeightedNotionalPosition {
+  "Asset in which the position is denominated in"
```
candida-d, line 7274:
```suggestion
  "Settlement asset for this position"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -1184,6 +1184,16 @@ type Query {
     pagination: Pagination
   ): TeamRefereeHistoryConnection
 
+  "Time weighted notional position is a metric used to determine if a reward should be paid to a party"
+  timeWeightedNotionalPosition(
+    "Asset the position is denominated in"
```
candida-d, line 1189:
```suggestion
    "Settlement asset for the position"
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -927,6 +927,16 @@ service TradingDataService {
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {tags: "Margin"};
   }
 
+  // Get time weighted notional position
+  //
+  // Get the time weighted notional position for a given party and asset. The time weighted notional position
+  // is used to check if a party qualifies for a reward.
+  // If no epoch is specified, the most current time weighted notional position is returned.
```
candida-d, line 934:
```suggestion
  // If no epoch is specified, the final time weighted notional position from the end of the most recent completed epoch is returned.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4780,3 +4790,36 @@ message PartyMarginMode {
   // Epoch at which the update happened.
   uint64 at_epoch = 7;
 }
+
+// Time weighted notional position for a party and asset at a given epoch. It is used as a metric to calculate whether
+// the party is eligible for receiving a reward.
+message TimeWeightedNotionalPosition {
+  // Asset ID for the asset associated with the position.
```
candida-d, line 4797:
```suggestion
  // Asset ID for the settlement asset associated with the position.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,22 @@ import "vega/vega.proto";
...
 
+// Time weighted notional position update for the current epoch.
+// The time weighted notional position is used to determine whether
+// a party is eligible for receiving rewards at the end of an epoch.
+// These updates are provided to facilitate knowing whether or not a party
+// is on track to qualify.
+message TimeWeightedNotionalPositionUpdated {
+  // Epoch the time weighted notional position is relevant for
+  uint64 epoch_seq = 1;
+  // Asset ID the position is denominated in.
```
candida-d, line 24:
```suggestion
  // Asset ID for the position’s settlement asset.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -13,6 +13,22 @@ import "vega/vega.proto";
...
+  string asset = 2;
+  // Party holding the position
+  string party = 3;
+  // Current time weighted notional position
```
candida-d, line 28:
```suggestion
  // Time weighted notional position from the end of the most recent completed epoch
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -1184,6 +1184,16 @@ type Query {
...
+    assetId: ID!
+    "Party holding the position"
+    partyId: ID!
+    "Optional epoch to filter for. If not provided, the latest time weighted notional position is returned."
```
candida-d, line 1193:
```suggestion
    "Optional epoch to filter for. If no epoch is specified, the time weighted notional position from the end of the most recent completed epoch is returned."
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -1184,6 +1184,16 @@ type Query {
     pagination: Pagination
   ): TeamRefereeHistoryConnection
 
+  "Time weighted notional position is a metric used to determine if a reward should be paid to a party"
+  timeWeightedNotionalPosition(
+    "Asset the position is denominated in"
+    assetId: ID!
+    "Party holding the position"
+    partyId: ID!
```
mattrussell36, line 1192: Its possibly scope creep for this PR.. but I'd like to get the queries that take partyId to be consistently able to accept multiple partyIds. We have a bit of a mixture in the graphql/rest APIs with regard to this so its fine if this isn't done now

EVODelavega: @mattrussell36 Could you create an issue for that, ideally listing as many queries as possible where this would need to be added, because as it stands I would agree that this is scope-creep :)

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4780,3 +4790,36 @@ message PartyMarginMode {
...
+  // Epoch at which the time weighted notional position is calculated.
+  uint64 at_epoch = 3;
+  // Time weighted notional position for the party and asset.
+  string time_weighted_notional_position = 4;
+  // Vega time of the block in which the last position update occurred.
+  int64 last_updated = 5;
+}
+
+// Request to retrieve the time weighted notional position for a party and asset.
+// If no epoch is given, the time weighted notional position at the start of the current epoch will be returned.
```
candida-d, line 4810:
```suggestion
// If no epoch is given, the time weighted notional position at the end of the most recently completed epoch will be returned.
```

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -4780,3 +4790,36 @@ message PartyMarginMode {
...
+// If an epoch is given, the time weighted notional position at the start of that epoch will be returned.
```
candida-d, line 4811:
```suggestion
// If an epoch is given, the time weighted notional position achieved by the end of the given epoch will be returned.
```

EVODelavega (dismissed):

candida-d (dismissed):

guoguojin:
> I think you're missing conversion code for the new event in datanode/broker/convert.go

Yes, you were right, I added the conversion.

EVODelavega (dismissed):

core/execution/common/market_activity_tracker.go
```diff
@@ -593,7 +596,9 @@ func (mat *MarketActivityTracker) isEligibleForReward(asset, party string, marke
 		}
 	}
 	if !notionalTimeWeightedAveragePositionRequired.IsZero() {
-		if mat.getTWNotionalPosition(asset, party, markets).LT(notionalTimeWeightedAveragePositionRequired) {
+		notional := mat.getTWNotionalPosition(asset, party, markets)
```
ze97286, line 599: ok, there's a few problem here (partly my fault):
1. so the isEligibleForReward will be called possibly a lot more times than you need an event for so maybe emitting the event as it is just now is not a good idea (that's on me)
2. note that the function is called with a set of markets - so what it means is that this is the tw notional for the party for the asset *in this set of markets* - now in the event you're ignoring this context so it gets lost and you will emit the same value regardless of the set of markets which I think is wrong - I think if you were including the markets in the 
3. I think we might want to emit an event with 0 for a party with 0 as its still informative.

maybe instead of or in addition to party and asset, just add the game_id (as hash(dispatch strategy) to the event - this would fix problem 2 and possible 1 as well.

mattrussell36 (dismissed):

EVODelavega (approved):

candida-d (approved):

core/execution/common/market_activity_tracker.go
guoguojin: Added the game id as an additional grouping for the time weighted notional position as suggested and updated the API accordingly.

core/execution/common/market_activity_tracker.go
```diff
@@ -582,28 +585,34 @@ func (mat *MarketActivityTracker) CalculateMetricForTeams(ds *vega.DispatchStrat
...
+func (mat *MarketActivityTracker) isEligibleForReward(asset, party string, markets []string, minStakingBalanceRequired *num.Uint, notionalTimeWeightedAveragePositionRequired *num.Uint, gameID string) bool {
 	if !minStakingBalanceRequired.IsZero() {
 		balance, err := mat.balanceChecker.GetAvailableBalance(party)
 		if err != nil || balance.LT(minStakingBalanceRequired) {
 			return false
 		}
 	}
 	if !notionalTimeWeightedAveragePositionRequired.IsZero() {
-		if mat.getTWNotionalPosition(asset, party, markets).LT(notionalTimeWeightedAveragePositionRequired) {
+		notional := mat.getTWNotionalPosition(asset, party, markets)
```
ze97286, line 603: what about the case it is zero?

ze97286 (approved):

core/execution/common/market_activity_tracker.go
guoguojin: As discussed, the is zero check on on the notional time weighted average position required config value. If the party's notional is zero, the event will still emit.

PR #10868, opened 2024-03-11
Add test and implementation for AMM fees

karlem:
Close https://github.com/vegaprotocol/core-test-coverage/issues/956


core/integration/steps/amm_submission.go
```diff
@@ -218,6 +222,13 @@ func (a ammRow) marketID() string {
 	return a.r.MustStr("market id")
 }
 
+func (a ammRow) proposedFee() num.Decimal {
+	if !a.r.HasColumn("proposed fee") {
+		return num.DecimalFromFloat(0.01)
```
EVODelavega, line 227: Not a fan of default values for what is a required parameter. This makes existing tests exhibit somewhat magic behaviour, and doesn't quite make sense when it comes to amending an AMM. I also think the fee parameter needs to be added to the event in the event proto, datanode, and integration tests

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -462,6 +465,9 @@ message AmendAMM {
   string slippage_tolerance = 3;
   // The concentrated liquidity parameters.
   optional ConcentratedLiquidityParameters concentrated_liquidity_parameters = 4;
+  // Nominated liquidity fee factor, which is an input to the calculation of taker fees on the market,
+  // as per setting fees and rewarding liquidity providers and AMM pools.
```
candida-d, line 469:
```suggestion
  // Nominated liquidity fee factor, which is an input to the calculation of taker fees on the market.
```

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -437,6 +437,9 @@ message SubmitAMM {
   string slippage_tolerance = 3;
   // The concentrated liquidity parameters.
   ConcentratedLiquidityParameters concentrated_liquidity_parameters = 4;
+  // Nominated liquidity fee factor, which is an input to the calculation of taker fees on the market,
+  // as per setting fees and rewarding liquidity providers and AMM pools.
```
candida-d, line 441:
```suggestion
  // Nominated liquidity fee factor, which is an input to the calculation of taker fees on the market.
```

wwestgarth (approved):

core/integration/steps/amm_submission.go
```diff
@@ -101,9 +102,10 @@ func parseSubmitAMMTable(table *godog.Table) []RowWrapper {
 
 func parseAmendAMMTable(table *godog.Table) []RowWrapper {
 	return StrictParseTable(table, []string{
-		"party",     // str
-		"market id", // str
-		"slippage",  // dec
+		"party",        // str
+		"market id",    // str
+		"slippage",     // dec
+		"proposed fee", // str
```
EVODelavega, line 108: is this going to be required for an amend? And shouldn't this be a decimal, rather than a string?

karlem: For the sake of test is better for it to be required.

PR #10872, opened 2024-03-12
fix: send a heartbeat ethcall event so that core's view of the latest…

### LINKED ISSUE #10841: [Bug]: High number of Ethereum calls in mainnet since upgrade to 0.74.9
### Problem encountered

It has been reported that validators have seen a massive increase in calls to ethereum since the upgrade to 0.74.9

### Observed behaviour

reported between 15k and 60k requests in 2h

image

image

### Expected behaviour

the usually seen number of calls

### Software version

0.74.9

wwestgarth:
closes #10841

The ethcall engine will now send to core a dummy chain event if it hasn't sent anything to core in the past hour. It will contain just the last eth-block-height checked. This allows core to store in the snapshot a more up to date  `last-seen` block height meaning that a node will "re-check" fewer ethereum blocks after a protocol upgrade.

For reference, the last protocol upgrade has `last-seen` set to 80,000 Ethereum blocks in the past and causes a spike in RPC calls.


core/datasource/external/ethcall/engine.go
```diff
@@ -36,6 +36,10 @@ import (
 	"github.com/ethereum/go-ethereum"
 )
 
+// if we haven't sent a contract call event in this many e.blockIntervals we'll send a dummy one so that
+// core knows of the latest eth block height we've checked.
+var heatbeatInterval = time.Hour
```
jeremyletang, line 41: Should we have some sort of configuration in the config file which a long weird explicit name (like some others) so we can set it easily to something small so we can see it in like the system tests quickly, but default it to 1hr for mainnet?

jeremyletang (dismissed):

core/datasource/external/ethcall/engine.go
wwestgarth: Can't hurt. I've been testing it locally with a lower time but it would make sense for the code to be flexed regularly in the system-tests.

jeremyletang (approved):

candida-d (approved):

EVODelavega (approved):

PR #10918, opened 2024-03-18
fix: add duplicate rejected status

### LINKED ISSUE #10915: [Bug]: STATUS not being updated when a duplicate deposit is seen
### Problem encountered

A duplicate deposit TX has been noticed and the status of this has remained as OPEN. We also need to check if this issue is alsol possible with staking.

### Observed behaviour

https://api1.vega.community/graphql/

```
query{
  deposits{
    edges{
      node{
        asset{symbol}
        party{id}
        amount
        status
        createdTimestamp,
        txHash
      }
    }
  }
}
```

"txHash": "0xc2e24585893c352728e9bfea4d4088593a209ddb7c9cc6136648d3e79eefae74"

### Additional information 

This duplicate deposit has been rejected by Vega and will never finalise and it’s only the status in the API that is showing as open forever until the fix can be deployed in the next release.

### Expected behaviour

The duplicated deposit should be marked as `REJECTED` or something similar

### Software version

0.74.10

EVODelavega:
Closes #10915


datanode/sqlstore/migrations/0096_add_deposit_rejected_state.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+
+ALTER TYPE deposit_status ADD VALUE 'STATUS_DUPLICATE_REJECTED';
```
guoguojin, line 3:
```suggestion
ALTER TYPE deposit_status ADD VALUE IF NOT EXISTS 'STATUS_DUPLICATE_REJECTED';
```

Otherwise it could cause issue with network history because there's no down migration.

guoguojin (dismissed):

candida-d (dismissed):

mattrussell36 (dismissed):

wwestgarth (dismissed):

peterbarrow (approved):

candida-d (approved):

PR #10963, opened 2024-03-21
fix: validate EVM chain ID in asset proposal earlier on in governance…

### LINKED ISSUE #10955: [Bug]: Non-validator nodes fall out of consensus when proposing asset with bogus chain-id
### Problem encountered

When proposing an ERC20 asset and _accidently_ setting the EVM chain-id to be one that does not correspond to any of Vega's bridge, non-validator nodes fall out of consensus. This is because the code that validates an asset against bridge chain-ids is a validator-only code path and so the asset proposal does not fail on non-validator nodes:
```
if s.isValidator {
	client, err := s.ethClientByChainID(assetDetails.GetERC20().ChainID)
	if err != nil {
		return nil, err
	}
	a, err := erc20.New(assetID, assetDetails, s.ethWallet, client)
	if err != nil {
		return nil, err
	}
	asset = a
} else {
	a, err := erc20.New(assetID, assetDetails, nil, nil)
	if err != nil {
		return nil, err
	}
	asset = a
}
```

### Observed behaviour

Non-validator nodes fall out of consensus

### Expected behaviour

Non-validator nodes remain in consensus

### Steps to reproduce

```bash
Checkout branch `0027-ASSP-032`
Run `test_propose_new_erc20_bogus_chain_id`
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #10955

### DIFF
--- a/core/assets/assets.go
+++ b/core/assets/assets.go
@@ -34,8 +34,10 @@ import (
 )
 
 var (
-	ErrAssetDoesNotExist  = errors.New("asset does not exist")
-	ErrUnknownAssetSource = errors.New("unknown asset source")
+	ErrAssetDoesNotExist        = errors.New("asset does not exist")
+	ErrUnknownAssetSource       = errors.New("unknown asset source")
+	ErrErc20AddressAlreadyInUse = errors.New("erc20 address already in use")
+	ErrUnknownChainID           = errors.New("erc20 chain-id does not correspond to a bridge")
 )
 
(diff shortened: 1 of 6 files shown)


wwestgarth:
```
{
  "system-tests": "epic/arbitrum",
  "vegacapsule": "epic/arbitrum"
}
```

peterbarrow (approved):

PR #10972, opened 2024-03-22
Fix teams stats

### LINKED ISSUE #10969: Incorrect rewards shhowing in mainnet teams leadeboard
Gnilrad has rewards ad has games but showuldnt 

image

I suspect these are the LP rewards which are individual which I'm digging into.

If I am right then the teamStats API is not filtering on entityScorpre = Team as it should.

ValentinTrinque:
Close #10969

### DIFF
--- a/datanode/sqlstore/accounts.go
+++ b/datanode/sqlstore/accounts.go
@@ -56,7 +56,7 @@ func (as *Accounts) Add(ctx context.Context, a *entities.Account) error {
 		`INSERT INTO accounts(id, party_id, asset_id, market_id, type, tx_hash, vega_time)
 		 VALUES ($1, $2, $3, $4, $5, $6, $7)
 		 RETURNING id`,
-		deterministicIDFromAccount(a),
+		DeterministicIDFromAccount(a),
 		a.PartyID,
 		a.AssetID,
 		a.MarketID,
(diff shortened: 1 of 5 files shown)


datanode/sqlstore/migrations/0076_teams_stats.sql
```diff
@@ -1,71 +1,83 @@
...
 CREATE TABLE IF NOT EXISTS teams_stats
 (
-  team_id              BYTEA             NOT NULL,
-  party_id             BYTEA             NOT NULL,
-  at_epoch             BIGINT            NOT NULL,
-  total_quantum_volume HUGEINT           NOT NULL,
-  total_quantum_reward NUMERIC(1000, 16) NOT NULL,
-  games_played         JSONB             NOT NULL,
-  PRIMARY KEY (team_id, party_id, at_epoch)
+    team_id              BYTEA             NOT NULL,
```
EVODelavega, line 4: You can't update a migration that has already been deployed

datanode/sqlstore/accounts.go
```diff
@@ -186,7 +186,7 @@ func (as *Accounts) getAccountFromCache(id entities.AccountID) (entities.Account
 	return entities.Account{}, false
 }
 
-func deterministicIDFromAccount(a *entities.Account) entities.AccountID {
+func DeterministicIDFromAccount(a *entities.Account) entities.AccountID {
```
EVODelavega, line 189: why export this func?

ValentinTrinque: For test purpose, so I can craft the account ID for the transfers foreign key.

datanode/sqlstore/migrations/0076_teams_stats.sql
ValentinTrinque: Indeed, that's a mistake.

EVODelavega (approved):

PR #10978, opened 2024-03-22
fix: marshaller for stop order rejection error implemented

### LINKED ISSUE #10975: [Bug]: stop order rejection reason is REJECTION_REASON_UNSPECIFIED
### Problem encountered

Stop order API returns REJECTION_REASON_UNSPECIFIED. This is not unexpected value which is not present in StopOrderRejectionReason enum.

It looks like this is only value that is returned, even if there is obvious different reason.

image

### Observed behaviour

Stop order API returns rejection reason not present in StopOrderRejectionReason enum

### Expected behaviour

Proper rejection reason should be returned

### Steps to reproduce

```bash
Submit stop order that will be rejected (ie on market where there is no position, or reach max stop orders per party).
Query GraphQL to check rejectionReason
```

### Software version

v0.76.0-preview.1

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
Closes #10975

### DIFF
--- a/datanode/gateway/graphql/marshallers/marshallers.go
+++ b/datanode/gateway/graphql/marshallers/marshallers.go
@@ -420,7 +420,17 @@ func MarshalStopOrderRejectionReason(s vega.StopOrder_RejectionReason) graphql.M
 }
 
 func UnmarshalStopOrderRejectionReason(v interface{}) (vega.StopOrder_RejectionReason, error) {
-	return vega.StopOrder_REJECTION_REASON_UNSPECIFIED, ErrUnimplemented
+	s, ok := v.(string)
+	if !ok {
+		return vega.StopOrder_REJECTION_REASON_UNSPECIFIED, fmt.Errorf("expected stop order rejection reason to be a string")
+	}
+
+	t, ok := vega.Order_TimeInForce_value[s]
+	if !ok {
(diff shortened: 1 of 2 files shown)


EVODelavega (dismissed):

datanode/gateway/graphql/marshallers/marshallers.go
```diff
@@ -420,7 +420,17 @@ func MarshalStopOrderRejectionReason(s vega.StopOrder_RejectionReason) graphql.M
 }
 
 func UnmarshalStopOrderRejectionReason(v interface{}) (vega.StopOrder_RejectionReason, error) {
-	return vega.StopOrder_REJECTION_REASON_UNSPECIFIED, ErrUnimplemented
+	s, ok := v.(string)
+	if !ok {
+		return vega.StopOrder_REJECTION_REASON_UNSPECIFIED, fmt.Errorf("expected stop order rejection reason to be a string")
+	}
+
+	t, ok := vega.Order_TimeInForce_value[s]
```
wwestgarth, line 428:
```suggestion
	t, ok := vega.StopOrder_RejectionReason_value[s]
```

peterbarrow (approved):

PR #11001, opened 2024-03-26
refactor: address the underlying cause of #10997

witgaw:
This PR proses to extend the existing behaviour (price history within price monitoring engine being continually reset whlist in opening auction) to the first mark price update after leaving opening auction (this is guaranteed to exist as we fall back on last traded price if no other sources exist at that stage).

Done as part of https://github.com/vegaprotocol/vega/issues/10997 investigation which was already addressed with an umbrella fix here: https://github.com/vegaprotocol/vega/pull/11007

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -1082,7 +1082,7 @@ func (m *Market) BlockEnd(ctx context.Context) {
 			t.UnixNano(),
 			m.matching,
 			m.mtmDelta,
-			m.tradableInstrument.MarginCalculator.ScalingFactors.InitialMargin, m.mkt.LinearSlippageFactor, m.risk.GetRiskFactors().Short, m.risk.GetRiskFactors().Long, false)
+			m.tradableInstrument.MarginCalculator.ScalingFactors.InitialMargin, m.mkt.LinearSlippageFactor, m.risk.GetRiskFactors().Short, m.risk.GetRiskFactors().Long, false, false)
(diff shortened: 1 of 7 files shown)


davidsiska-vega (approved):

core/execution/future/market.go
```diff
@@ -1654,6 +1655,7 @@ func (m *Market) leaveAuction(ctx context.Context, now time.Time) {
 
 	if wasOpeningAuction && !m.as.IsOpeningAuction() && m.getCurrentMarkPrice().IsZero() {
 		m.markPriceCalculator.OverridePrice(m.lastTradedPrice)
+		m.pMonitor.ResetPriceHistory([]*types.Trade{{Price: m.lastTradedPrice, Size: 1}})
```
EVODelavega, line 1658: Does the uncross volume matter in any way here? We would be discarding the external price data entirely, so the next time around, we could end up calculating a mark price that is perhaps more likely to breach a monitoring bound. Not really stating that as a fact, I just wonder if this reset call would essentially prime the price monitoring bounds as though the only price data available is the last traded price, but next time around the mark price calculator could end up providing a significantly different price

witgaw: The volume only matters within the same timestamp, but there all trades will be at the same price so doesn't matter. 

The reset only happens when mark price was 0 after opening auction, so better to have last traded price than 0.

peterbarrow (approved):

PR #11043, opened 2024-03-28
Migration fix

### LINKED ISSUE #11042: [Bug]: order rejection reason missing
### Problem encountered

the enum for the order reject reason is missing `REJECTION_REASON_STOP_ORDER_LINKED_PERCENTAGE_INVALID`

### Observed behaviour

datanode crashes

### Expected behaviour

datanode doesn't crash

### Steps to reproduce

```bash
submit a stop order that gets rejected for the above reason
```

### Software version

develop

EVODelavega:
Closes #11042


daniel1302 (commented):
Does it matter if your migration is 99 not 101?

datanode/sqlstore/migrations/0099_add_stop_order_rejection_reason.sql
```diff
@@ -0,0 +1,7 @@
+-- +goose Up
+
+ALTER TYPE stop_order_rejection_reason ADD VALUE IF NOT EXISTS 'REJECTION_REASON_STOP_ORDER_LINKED_PERCENTAGE_INVALID' AFTER 'REJECTION_REASON_STOP_ORDER_NOT_CLOSING_THE_POSITION';
```
ze97286, line 3: don't you want to be nice and also add it in schema.graphql? (only REJECTION_REASON_STOP_ORDER_LINKED_PERCENTAGE_INVALID is missing)

EVODelavega: didn't spot that

ze97286, line 3: REJECTION_REASON_STOP_ORDER_NOT_CLOSING_THE_POSITION does seem to exist in the original migration 0027

EVODelavega:
@daniel1302 The migration scripts sequence does matter, so I had to rename the ones that aren't part of 75.x

datanode/sqlstore/migrations/0099_add_stop_order_rejection_reason.sql
ze97286: the order seems correct.

EVODelavega: yes, I'm adding the `REJECTION_REASON_STOP_ORDER_LINKED_PERCENTAGE_INVALID` value after the one that already exists, before 2 more values that were added with migration 0085. The order of the enum values has to match the sequence in the protos, so I have to be specific about where to add the value

candida-d (dismissed):

candida-d (approved):

peterbarrow (approved):

PR #11068, opened 2024-04-04
fix: SLA param changes that push an LP out of meeting commitment rese…

### LINKED ISSUE #11065: [Bug]: `currentEpochFractionOfTimeOnBook` reporting value bigger than `1`
### Problem encountered

Since the proposal to change some SLA settings on some mainnet markets was enacted, the `currentEpochFractionOfTimeOnBook` for some LP's are being reported as being bigger than one:
[https://vega-data.nodes.guru:3008/api/v2/liquidity/providers?marketId=4e9081e20e9e81f3e74[…]27bd526b46f1db2db315df08637331beaffa8ca4fd2c7065a4d7e4a7](https://vega-data.nodes.guru:3008/api/v2/liquidity/providers?marketId=4e9081e20e9e81f3e747d42cb0c9b8826454df01899e6027a22e771e19cc79fc&partyId=6d4c7e9727bd526b46f1db2db315df08637331beaffa8ca4fd2c7065a4d7e4a7)

This makes no sense since how can they be meeting their commitment more than 100% of the time.

We need to work out why this has happened, and to understand the impact of rolling into the next epoch with a value of `currentEpochFractionOfTimeOnBook` bigger than one. Will it panic? Will it overflow and cause havoc?
```
// p = (1-[timeBookFraction-commitmentMinTimeFraction/1-commitmentMinTimeFraction]) * slaCompetitionFactor
	return one.Sub(
		timeBookFraction.Sub(e.slaParams.CommitmentMinTimeFraction).Div(one.Sub(e.slaParams.CommitmentMinTimeFraction)),
	).Mul(e.slaParams.SlaCompetitionFactor)
```

### Observed behaviour

SLA `currentEpochFractionOfTimeOnBook` seen as being bigger than 1.

### Expected behaviour

`currentEpochFractionOfTimeOnBook` should never be bigger than one.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.74.10

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #11065

### DIFF
--- a/core/liquidity/v2/sla.go
+++ b/core/liquidity/v2/sla.go
@@ -46,6 +46,7 @@ func (e *Engine) ResetSLAEpoch(
 	}
 
 	for party, commitment := range e.slaPerformance {
+		commitment.start = time.Time{}
 		if e.doesLPMeetsCommitment(party, markPrice, midPrice, positionFactor) {
 			commitment.start = now
 		}
@@ -87,8 +88,7 @@ func (e *Engine) calculateCurrentTimeBookFraction(now, start time.Time, s time.D
 	if lNano > 0 {
 		timeBookFraction = num.DecimalFromInt64(s.Nanoseconds()).Div(num.DecimalFromInt64(lNano))
 	}
-
-	return timeBookFraction
+	return num.MinD(num.DecimalOne(), timeBookFraction)
 }
 
 // CalculateSLAPenalties should be called at the and of epoch to calculate SLA penalties based on LP performance in the epoch.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 
 - [10946](https://github.com/vegaprotocol/vega/issues/10945) - Save dispatch strategy for recurring governance transfer in the database.
 - [11000](https://github.com/vegaprotocol/vega/issues/11000) - Vesting summary events are now sorted properly before sending.
+- [11065](https://github.com/vegaprotocol/vega/issues/11065) - `SLA` commitment state is reset properly at the end of an epoch
 - [10960](https://github.com/vegaprotocol/vega/issues/10960) - Only the owner of a referral set can update it.
 - [10970](https://github.com/vegaprotocol/vega/issues/10970) - Logic for whether a team member joined more than `rewards.team.minEpochsInTeam` is fixed.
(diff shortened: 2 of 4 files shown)


core/liquidity/v2/sla.go
```diff
@@ -88,7 +89,7 @@ func (e *Engine) calculateCurrentTimeBookFraction(now, start time.Time, s time.D
 		timeBookFraction = num.DecimalFromInt64(s.Nanoseconds()).Div(num.DecimalFromInt64(lNano))
 	}
 
-	return timeBookFraction
+	return num.MinD(num.DecimalOne(), timeBookFraction)
```
EVODelavega, line 92: are we sure we want to keep this? If the value > 1, wouldn't that imply that the party hasn't met the commitment last time around, and therefore the fraction would _have_ to be < 1

wwestgarth: I'm not sure I follow. It should always be that `timeBookFraction <= 1` and if its not something has gone wrong so the cap is just there for safety.

EVODelavega (dismissed):

guoguojin (approved):

core/liquidity/v2/sla.go
```diff
@@ -87,8 +88,7 @@ func (e *Engine) calculateCurrentTimeBookFraction(now, start time.Time, s time.D
 	if lNano > 0 {
 		timeBookFraction = num.DecimalFromInt64(s.Nanoseconds()).Div(num.DecimalFromInt64(lNano))
 	}
-
-	return timeBookFraction
+	return num.MinD(num.DecimalOne(), timeBookFraction)
```
davidsiska-vega, line 91: Can we also cap from below by `0.0`?

davidsiska-vega (dismissed):

EVODelavega:
One failed system test, looks like a rare/random failure, re-running the test itself: https://jenkins.vega.rocks/job/common/job/system-tests-wrapper/53752

core/liquidity/v2/sla.go
EVODelavega: negative values here necessitate time-travel, which is currently unsupported by vega. :)

PR #11074, opened 2024-04-08
10998 sparse rewards

### LINKED ISSUE #10998: [Bug]: totalRewardsEarned is not always showing the cumulative total as expected
### Problem encountered

The totalRewardsEarned and totalRewardsEarnedQuantum are meant to be cumulative over the lifetime of a game and they are just repeasting the other rewards fields

### Observed behaviour

See above

### Expected behaviour
The totalRewards and quantum should be summing the rewards over the lifetime of that game 

### Steps to reproduce

Look at tjis team 

https://vegaprotocol.eth.limo/#/competitions/teams/d332f78ccffe70c0e7e59c3b37bb93ef9247e1f46acbff807be928f647b18faf

YOu can see it has cumulative rewards building up over epochs.

Then look at this team 

https://vegaprotocol.eth.limo/#/competitions/teams/9cd31bf2edeebb145f81d6b5aaf9ee9c26ef66ac8a0d10f69497f9cd9159ed37

It is just saying the rewards and rewardsTotal are equal.

I suspect this is because that team earned no rwards in epoch 864 and the count has reset somehow.

Is the code maybe doing totalRewardsEarned(previous epoch) + RewardsEarned(current epoch) each time

### Software version

0.75.3

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #10998


guoguojin (dismissed):

datanode/sqlstore/migrations/0102_retro_fill_team_reward_totals.sql
```diff
@@ -0,0 +1,72 @@
...
+    AND NOT EXISTS (
+        SELECT 1
+        FROM game_reward_totals AS grtc
+        WHERE grtc.party_id = grto.party_id
+        AND grtc.asset_id = grto.asset_id
+        AND grtc.market_id = grto.market_id
+        AND grtc.game_id = grto.game_id
+        AND grtc.team_id = grto.team_id
+        AND grtc.epoch_id = grto.epoch_id + 1
+    );
```
guoguojin, line 25: I ran a quick test with this inside a transaction and I don't think this does what it's supposed to. I think it will only carry over missing transactions from one epoch, to the next, but not any further than that.

Also we would only want the stats carried over for the duration of the game, there's no point in carrying forward the stats if in the next epoch the game is over, or we will end up carrying this forward indefinitely.

datanode/sqlstore/migrations/0102_retro_fill_team_reward_totals.sql
```diff
@@ -0,0 +1,72 @@
...
+        AND NOT EXISTS (
+            SELECT 1
+            FROM game_reward_totals AS grtc
+            WHERE grtc.party_id =  grto.party_id
+            AND grtc.game_id = grto.game_id
+            AND grtc.asset_id = grto.asset_id
+            AND grtc.market_id = grto.market_id
+            AND grtc.team_id = grto.team_id
+            AND grtc.epoch_id = (NEW.id - 1)
+        );
```
guoguojin, line 55: I think this should exclude games which have expired in the last epoch.

jeremyletang (approved):

PR #11080, opened 2024-04-08
fix: network should not reduce its position during auction

EVODelavega:
Network orders shouldn't be placed when the market is in auction.

### DIFF
--- a/core/execution/future/liquidation.go
+++ b/core/execution/future/liquidation.go
@@ -26,6 +26,9 @@ import (
 )
 
 func (m *Market) checkNetwork(ctx context.Context, now time.Time) error {
+	if m.as.InAuction() {
+		return nil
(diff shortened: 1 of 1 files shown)


core/execution/future/liquidation.go
```diff
@@ -26,6 +26,9 @@ import (
 )
 
 func (m *Market) checkNetwork(ctx context.Context, now time.Time) error {
+	if m.as.InAuction() {
+		return nil
+	}
 	// debug
```
ze97286, line 32: not relevant but what is this?

EVODelavega: Not a clue, doesn't mean anything (anymore), so it's gone now

jeremyletang (approved):

PR #11108, opened 2024-04-12
Refactor price monitoring to register prices (with equal weights) rather than trades

### LINKED ISSUE #11127: Price moniotoring engine should record all observations with the same weight
Currently we still have legacy VWAP functionality within the price monitoring auction, but the only scenario in which it is employed is when we leave a price moniotring auction with an uncrossing trade and immediately update the mark price afterwards. This results in a reference price which is the following weighted average: (uncrossing volume * uncrossing price + mark price) / (uncrossing volume + 1). 

This should be modified so that it's a simple average.

witgaw:
- Modify price monitoring engine so that all observations recorded withing the same timestamp have equal weights.
- Simplify price monitoring engine (taking into account that we no longer have liquidity monitoring auctions or a need for VWAP calculations).

Closes https://github.com/vegaprotocol/vega/issues/11127


witgaw (commented):
Ready for review

candida-d (dismissed):

protos/sources/vega/snapshot/v1/snapshot.proto
```diff
@@ -446,7 +441,7 @@ message PriceMonitor {
   repeated PriceRangeCache price_range_cache = 9;
   int64 ref_price_cache_time = 10;
   repeated DecimalMap ref_price_cache = 11;
-  repeated CurrentPrice prices_now = 12;
+  repeated string prices_now = 12;
```
ze97286, line 444: both this and the other change above will not work. you can't change the types and/or names of existing fields or it won't be backward compatible. you can add a new field, migrate to it, then later we can mark old field as unused.

witgaw: Would it work with latest changes?

ze97286: it would but it adds a bit of confusion by the different naming between the proto layer and typed.

ze97286 (approved):

PR #11117, opened 2024-04-15
Order events

### LINKED ISSUE #11123: [Feature]: cut down on the number of order events especially when cancelling orders in batch
When orders are cancelled, core is to emit a simplified event that can cover mutliple orders being cancelled.

When all orders are cancelled, there's not really a point to releasing the margin bit by bit, so instead only make the call to release excess margin after all orders have been cancelled.

EVODelavega:
cancelled order event stuff

Closes https://github.com/vegaprotocol/vega/issues/11123


candida-d (dismissed):

EVODelavega:
```
{
    "system-tests": "order-events"
}
```

jeremyletang (approved):

PR #11122, opened 2024-04-16
Remove auciton trigger staleness functionality

### LINKED ISSUE #11121: Remove trigger staleness functionality
There's still a remaining bit of trigger staleness functionality: if the reference price within price monitoring engine hasn't been updated for longer than trigger's horizon then, once the auction is activated any such triggers which remain active will be switched off. This can lead to puzzling behaviour like auction uncrossing with a price violating a set of repeated triggers only after the extension of a single trigger elapses rather than sum of all repeated triggers which may be conunterinutitive and confusing for the users. 

Therefore the trigger stalness functionality should be removed altogether.

witgaw:
Closes https://github.com/vegaprotocol/vega/issues/11121

### DIFF
--- a/core/monitor/price/pricemonitoring_test.go
+++ b/core/monitor/price/pricemonitoring_test.go
@@ -137,7 +137,6 @@ func TestRecordPriceChange(t *testing.T) {
 
 	auctionStateMock.EXPECT().IsFBA().Return(false).Times(4)
 	auctionStateMock.EXPECT().InAuction().Return(false).Times(4)
-	auctionStateMock.EXPECT().IsPriceAuction().Return(false).Times(2)
 	statevar := mocks.NewMockStateVarEngine(ctrl)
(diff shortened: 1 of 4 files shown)


core/monitor/price/pricemonitoring_test.go
```diff
@@ -348,230 +343,6 @@ func TestCheckBoundViolationsWithinCurrentTimeWith2HorizonProbabilityPairs(t *te
 	require.False(t, b)
 }
 
-func TestCheckBoundViolationsAcrossTimeWith1HorizonProbabilityPair(t *testing.T) {
```
witgaw, line 351: This test effectively didn't test anything (price ranges withing the price monitoring engine were never updated) so I've removed it altogether.

witgaw (commented):
Ready for review

peterbarrow (approved):

PR #11137, opened 2024-04-17
chore: change the netparameter for evm chains config to be a list

### LINKED ISSUE #11138: [Feature]: Make external views on EVM bridge extensible to multiple bridges
# Feature Overview

**In order to** ensure we do not shoot ourselves in the foot
**We will** make all external types related to the second bridge lists of those types
**So that** if we implement another EVM bridge in the future there is less chance of having awkward incompatibilites

Currently the Arbitrum bridge feature has been written with the assumption we will have 1 new bridge and this is fine internally when we don't really know what the future holds. 

Despite this it would make sense for all _external_ types to be such they _could_ easily extend to multiple EVM bridges so that there are fewer breaking changes that we might need to do if we implement another bridge.

For example we should:

- [ ] Change the single network parameter from a single `EVMChainConfig`  -> `listof(EMVChainConfig)` like we have for L2 data sources
- [ ] Change the config file RPC endpoint for the second bridge to be a `listof(RPCEndpoint+ChainID)` like we have for L2 data sources
- [ ] Change the snapshot payloads for the new bridge states to be `listof(bridge-states)` so that there is less snapshot migratiion needed later

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

jeremyletang:
close #11138


wwestgarth:
```
{
    "system-tests": "arbitrum-updates",
    "vegacapsule": "arbitrum-updates",
}
```

candida-d (approved):

TomMcL (approved):

peterbarrow (approved):

PR #11152, opened 2024-04-18
10995 liquidation range

### LINKED ISSUE #10995: [Feature] Liquidation strategy to have its own range
Add a market parameter into the liquidation strategy section and wire it up as per: https://github.com/vegaprotocol/specs/pull/2218 

Basically, the new range parameter will replace the SLA range in all the liquidation behaviour.

EVODelavega:
Closes #10995 

```
{
    "system-tests": "liquidation_slippage"
}
```


core/execution/liquidation/engine.go
```diff
@@ -87,7 +83,11 @@ var (
 		DisposalFraction:    num.DecimalOne(),
 		FullDisposalSize:    math.MaxUint64,
 		MaxFractionConsumed: num.DecimalOne(),
+		DisposalSlippage:    num.DecimalFromFloat(10.0),
 	}
+
+	// just because we use it for min-max
+	dOne = num.DecimalFromFloat(1.0)
```
jeremyletang, line 90: don't we have num.DecimalOne()

core/governance/market.go
```diff
@@ -705,6 +705,9 @@ func validateLiquidationStrategy(ls *types.LiquidationStrategy) (types.ProposalE
 	} else if ls.DisposalTimeStep > time.Hour {
 		return types.ProposalErrorInvalidMarket, fmt.Errorf("liquidation strategy time step can't be more than 1h")
 	}
+	if ls.DisposalSlippage.IsZero() || ls.DisposalSlippage.IsNegative() {
+		return types.ProposalErrorInvalidMarket, fmt.Errorf("liquidation strategy must specify a disposal slippage range > 1")
```
jeremyletang, line 709: this wrong.

core/execution/liquidation/engine.go
```diff
@@ -132,14 +128,20 @@ func (e *Engine) Update(cfg *types.LiquidationStrategy) {
...
-		return nil, err
+
+	// get the min/max price from the range based on slippage parameter
+	mpDec := num.DecimalFromUint(midPrice)
+	minP := num.UintZero()
+	if e.cfg.DisposalSlippage.LessThan(num.DecimalZero()) {
+		minD := mpDec.Mul(num.DecimalZero().Sub(e.cfg.DisposalSlippage))
+		minP, _ = num.UintFromDecimal(minD)
 	}
+	maxD := mpDec.Mul(num.DecimalZero().Add(e.cfg.DisposalSlippage))
```
wwestgarth, line 143: Should this be?
```suggestion
	maxD := mpDec.Mul(num.DecimalOne().Add(e.cfg.DisposalSlippage))
```

Also we're only setting `minP` away from zero if `DisposalSlippage < 0` which doesn't make sense since I thought is must be a positive decimal.

core/execution/liquidation/engine.go
```diff
@@ -132,14 +128,20 @@ func (e *Engine) Update(cfg *types.LiquidationStrategy) {
...
 	}
-	minP, maxP, err := e.ml.ValidOrdersPriceRange()
-	if err != nil {
-		return nil, err
+
+	// get the min/max price from the range based on slippage parameter
+	mpDec := num.DecimalFromUint(midPrice)
+	minP := num.UintZero()
+	if e.cfg.DisposalSlippage.LessThan(num.DecimalZero()) {
+		minD := mpDec.Mul(num.DecimalZero().Sub(e.cfg.DisposalSlippage))
```
wwestgarth, line 140:
```suggestion
		minD := mpDec.Mul(num.DecimalOne().Sub(e.cfg.DisposalSlippage))
```
and also need to `max(,0, minD)` it.

core/execution/liquidation/engine.go
```diff
@@ -132,14 +128,20 @@ func (e *Engine) Update(cfg *types.LiquidationStrategy) {
 	e.cfg = cfg
 }
 
-func (e *Engine) OnTick(ctx context.Context, now time.Time) (*types.Order, error) {
+func (e *Engine) OnTick(ctx context.Context, now time.Time, midPrice *num.Uint) (*types.Order, error) {
 	if e.pos.open == 0 || e.as.InAuction() || e.nextStep.After(now) {
 		return nil, nil
 	}
-	minP, maxP, err := e.ml.ValidOrdersPriceRange()
-	if err != nil {
```
wwestgarth, line 140: Just like what `ValidOrdersPriceRange()` does, if there is no mid-price we should bail out and not do any liquidation bits.

core/execution/liquidation/snapshot.go
```diff
@@ -57,9 +57,13 @@ func (e *Engine) LoadState(ctx context.Context, pl *types.Payload) ([]types.Stat
 		if d.Config != nil {
 			e.cfg = d.Config.DeepClone()
 		} else {
-			// @NOTE this can be removed after protocol upgrade has completed
+			// this can probably be removed now
 			e.cfg = GetLegacyStrat()
 		}
+		// @NOTE this should have a protocol upgrade guard around it
+		if e.cfg.DisposalFraction.IsZero() {
```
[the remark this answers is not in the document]
EVODelavega: Had to read it again, it does say that... Not sure if that's really the way to go. If that's the case, then it definitely wouldn't make sense to also specify a default value of `0.1` for the slippage parameter (also in spec: https://github.com/vegaprotocol/specs/pull/2218/files#diff-3d81a39049a36bb18ecac5bd2f70835f6faabe4410a489eddc8f531411322bacR114)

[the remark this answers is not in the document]
wwestgarth: Probably worth clearing that up with protocol-design people then. Because also there being a "default" implies that the new field should be optional, whereas we have it here as a required parameter that must be set to something.

core/execution/liquidation/engine.go
EVODelavega: good point

core/execution/liquidation/engine.go
EVODelavega: yup, pretty sure I got carried away at some point replacing `dOne` with `num.DecimalZero()` instead of `num.DecimalOne()`

core/execution/future/market.go
```diff
@@ -709,6 +709,21 @@ func (m *Market) midPrice() *num.Uint {
 	return midPrice
 }
 
+func (m *Market) getMarketMidPrice() *num.Uint {
```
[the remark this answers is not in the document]
EVODelavega: Nothing... not a clue why I'm not using that one, must've overlooked it (probably assumed `getMidPrice` instead of `midPrice`.

EVODelavega:
```
{
    "system-tests": "liquidation_slippage"
}
```

core/execution/liquidation/snapshot.go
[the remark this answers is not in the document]
EVODelavega: https://github.com/vegaprotocol/specs/pull/2249

Default of 0.1 to be removed from spec, using SLA price range for proposals/existing markets now

core/execution/future/market.go
```diff
@@ -963,6 +963,11 @@ func (m *Market) PostRestore(ctx context.Context) error {
 		}
 	}
 
+	// Disposal slippage was set as part of this upgrade, send event to ensure datanode is updated.
+	if vegacontext.InProgressUpgradeFrom(ctx, "v0.75.8") {
```
wwestgarth, line 967: Ran into this with the second bridge migrations:
We can't use `v0.75.8` because thats the special mainnet-only patch and stagnet/fairground will migration from `v0.75.7`. Which is annoying, so we'll have to just use `vgcontext.InProgressUpgrade(ctx)` and remember to whip it out asap once the release is on mainnet.

core/execution/liquidation/engine.go
```diff
@@ -132,14 +128,21 @@ func (e *Engine) Update(cfg *types.LiquidationStrategy) {
...
-		return nil, err
+
+	one := num.DecimalOne()
+	// get the min/max price from the range based on slippage parameter
+	mpDec := num.DecimalFromUint(midPrice)
+	minP := num.UintZero()
+	if e.cfg.DisposalSlippage.LessThan(one) {
+		minD := mpDec.Mul(one.Sub(e.cfg.DisposalSlippage))
+		minP, _ = num.UintFromDecimal(minD)
```
wwestgarth, line 142: I still don't understand why we have the if when `DisposalSlippage` is _always_ bigger then zero. Shouldn't it just be the below?

```suggestion
minD := num.MaxD(num.DecimalZero(), mpDec.Mul(one.Sub(e.cfg.DisposalSlippage)))
minP, _ = num.UintFromDecimal(minD)
```

core/execution/future/market.go
EVODelavega: we could just `InProgressUpgradeFrom(ctx, "v0.75.8") || InProgressUpgradeFrom(ctx, "v0.75.7")` no?

core/execution/liquidation/engine.go
EVODelavega: because if disposal slippage > 1, the min price calculated here will _always_ be negative, and there's no point in calculating it, just set the min price to 0, and be done with it.

core/execution/future/market.go
EVODelavega: seeing as we've had this happen twice (fairground <=> mainnet tags), I added `InProgressUpgradeFromMultiple`, and now check for both tags

jeremyletang (approved):

PR #11176, opened 2024-04-23
chore: account for new slippage factor in protocol upgrade

EVODelavega:
Apply protocol-upgrade related fix to liquidation strategy snapshots


wwestgarth:
I was just looking at the API for the markets to see if they inherit the SLA-params on upgrade:
https://api.n00.stagnet1.vega.rocks/api/v2/markets

But it looks like for all markets apart from one `disposalSlippageRange` is `0` even when the SLA price-range is non-zero. So it would be worth looking into that, I imagine we just need to re-send out the market event so the data-node has a value for it.

jeremyletang (approved):

PR #11206, opened 2024-04-26
fix: apply transfer interval correctly and add transfer_interval opti…

### LINKED ISSUE #11205: [Bug]: The check for transfer interval is incorrect in evaluation of recurring/governance transfers
### Problem encountered

transfer is being made outside transfer interval schedule. 

### Observed behaviour

The check for transfer interval in recurring/governance transfer is incorrect and leading to transfers going through not only interval but continuously. 

### Expected behaviour

when transfer interval is defined, transfers should only go through on the defined interval. 

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.76.0-preview.6

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
…on to feature tests

Closes #11205


Jiajia-Cui (approved):
the feature test looks reasonable

jeremyletang (approved):

PR #11226, opened 2024-05-01
fix: reverse order of batch proposals

### LINKED ISSUE #11005: [Bug]: Sub proposals returned in reverse order in GraphQL
### Problem encountered

The subProposals are returned in the incorrect order. This makes it harder to determine market Ids in batch proposals

### Observed behaviour

The order appears to be reversed when using the `subProposals` field. The `batchTerms` are the correct order

### Expected behaviour

The order should be the same order as provided in the batch proposal

### Steps to reproduce

```bash
1. Submit a batch proposal with more than one batch
2. Use the query below
3. Notice that sub proposals are in the wrong order


{
  proposalsConnection {
    edges {
      proposalNode {
        ... on BatchProposal {
          id
          subProposals {
            terms {
              change {
                __typename
                ... on NewMarket {
                  instrument {
                    name
                    code
                  }
                }
              }
            }
          }
          batchTerms {
            changes {
              change {
                __typename
                ... on NewMarket {
                  instrument {
                    name
                    code
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```
```


### Software version

0.75.3

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

### PR DESCRIPTION
**Pull request by EVODelavega**
Closes #11005

### DIFF
--- a/datanode/entities/proposal.go
+++ b/datanode/entities/proposal.go
@@ -173,10 +173,13 @@ func (p Proposal) Cursor() *Cursor {
 }
 
 func (p Proposal) ToProtoEdge(_ ...any) (*v2.GovernanceDataEdge, error) {
-	proposalsProto := make([]*vega.Proposal, len(p.Proposals))
+	j := len(p.Proposals) - 1
+	proposalsProto := make([]*vega.Proposal, j+1)
 
 	for i, proposal := range p.Proposals {
-		proposalsProto[i] = proposal.ToProto()
+		// fill slice in reverse order. The records are inserted (and returned) in reverse order, so to match
+		// the order in which markets were created, the proposals should be returned in the original order.
+		proposalsProto[j-i] = proposal.ToProto()
 	}
 
 	return &v2.GovernanceDataEdge{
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: datanode/entities/proposal.go [hunk 1]
```diff
@@ -173,10 +173,13 @@ func (p Proposal) Cursor() *Cursor {
 }
 
 func (p Proposal) ToProtoEdge(_ ...any) (*v2.GovernanceDataEdge, error) {
-	proposalsProto := make([]*vega.Proposal, len(p.Proposals))
+	j := len(p.Proposals) - 1
+	proposalsProto := make([]*vega.Proposal, j+1)
 
 	for i, proposal := range p.Proposals {
-		proposalsProto[i] = proposal.ToProto()
+		// fill slice in reverse order. The records are inserted (and returned) in reverse order, so to match
```
wwestgarth, line 180: I don't think the issue is that they are backwards? The query returns them ordered by enactment time (see `getProposalsInBatch(ctx context.Context, batchID string)`), whereas FE need them ordered the same as they wre in the original proposal submission.

EVODelavega: I was wondering if it could really be this simple... clearly not. I'll have a rethink

PR #11258, opened 2024-05-09
AMM handle "near-zero-volume" price levels

### LINKED ISSUE #10784: [Feature]: AMM handle "near-zero-volume" price levels
# Feature Overview

For an AMM with wide bounds and a low commitment (and especially if the market is configured with 0 position DPs) it is possible for its curve to represent near-zero volume between price levels.

For example say we have an AMM whose lower curve is in the price range `(1800, 2000)` with low committment. If I query its volume between the prices `(1999, 2000)` say it comes out as `0.25`. Now an incoming sell order comes in at price `1999` how much do we trade?

Should we:
- reject AMM submissions with zero volume price-levels. I believe the curve is monotonic so we only have to check one end (super easy)
- round up, over trade with the AMM potentially leaving is crossed with the orderbook/other AMM's 
- round down and handle zero volume regions (very awkward)

There is a network parameter that governs the smallest possible commitment `market.amm.minCommitmentQuantum`, but this doesn't stop someone creating a pool which is very wide.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

karlem:
Close #10784

### DIFF
--- a/core/execution/amm/pool.go
+++ b/core/execution/amm/pool.go
@@ -401,7 +401,7 @@ func (p *Pool) setCurves(
 			true,
 		)
 
-		highPriceMinusOne := num.UintZero().Sub(p.lower.high, num.UintOne())
+		highPriceMinusOne := num.UintZero().Sub(p.lower.high, p.oneTick)
 		// verify that the lower curve maintains sufficient volume from highPrice - 1 to the end of the curve.
 		if p.lower.volumeBetweenPrices(p.sqrt, highPriceMinusOne, p.lower.high) < 1 {
 			return fmt.Errorf("insufficient volume in the lower curve from high price - 1 to the end")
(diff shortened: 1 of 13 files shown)


core/execution/amm/pool.go
```diff
@@ -411,7 +422,15 @@ func (p *Pool) setCurves(
 			p.positionFactor,
 			false,
 		)
+
+		highPriceMinusOne := num.UintZero().Sub(p.upper.high, num.UintOne())
```
wwestgarth, line 426:
```suggestion
		highPriceMinusOne := num.UintZero().Sub(p.upper.high, p.oneTick)
```

wwestgarth (approved):

PR #11260, opened 2024-05-10
fix: reset failed attempts counter when it works, but no upgrade is p…

### LINKED ISSUE #11242: [Bug]: Visor reports invalid value for the attempts left
### Problem encountered

This is low priority I guess

When We start visor from network history it decreases the `attemptLeft` from 10 and at some point it becomes negative. It is confusing why the left attempts may be negative.

Another thing is that even if we hit 0 and negative value the node will be able to start

```
2024-05-04T23:23:39.459+0200    DEBUG   visor   visor/visor.go:175      Failed to get upgrade status from API   {"error": "failed to get protocol update status: request failed: Post \"http://unix/rpc\": dial unix /home/daniel/www/snapshot-testing/workdir/vega.sock: connect: connection refused"}
2024-05-04T23:23:39.459+0200    INFO    visor   visor/visor.go:179      Still waiting for vega to start...      {"attemptLeft": -47}
```

### Observed behaviour

1. Visor reports `"attemptLeft":` as negative
2. attempts left is not working
3. attempts left is by default 10 which is not enough.

### Expected behaviour

1. attempts left should be reported correctly
2. attempts left should work
3. attempts left should be bigger by default

### Steps to reproduce

```bash
1. Start a node from the network history
2. Use bigger number for the network history segments blocks
```

### Software version

v0.76.0

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #11242

### DIFF
--- a/visor/visor.go
+++ b/visor/visor.go
@@ -29,7 +29,6 @@ import (
 
 const (
 	upgradeAPICallTickerDuration = time.Second * 2
-	maxUpgradeStatusErrs         = 10
 	namedLogger                  = "visor"
 )
 
@@ -114,8 +113,11 @@ func (v *Visor) Run(ctx context.Context) error {
 			runConf.Vega.RCP.HTTPPath,
 		)
 
+		// how many times to try and connect on the first start up of the binaries
 		maxNumberOfFirstConnectionRetries := v.conf.MaxNumberOfFirstConnectionRetries()
 
+		// how many times to try and connect in subsequent restarts of the binaries where it is expected to be much quicker
+		maxUpgradeStatusErrs := v.conf.MaxNumberOfRestartConnectionRetries()
 		numOfUpgradeStatusErrs := 0
 		maxNumRestarts := v.conf.MaxNumberOfRestarts()
 		restartsDelay := time.Second * time.Duration(v.conf.RestartsDelaySeconds())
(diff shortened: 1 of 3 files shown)


visor/visor.go
```diff
@@ -176,12 +178,17 @@ func (v *Visor) Run(ctx context.Context) error {
...
+					if numOfRestarts == 0 {
+						attemptsLeft = maxNumberOfFirstConnectionRetries - numOfUpgradeStatusErrs
+					}
+					v.log.Info("Still waiting for vega to start...", logging.Int("attemptLeft", attemptsLeft))
 
 					break
 				}
 
 				if !upStatus.ReadyToUpgrade {
+					numOfUpgradeStatusErrs = 0
```
karlem, line 191: Not sure why do you reset it here?

visor/visor.go
```diff
@@ -176,12 +178,17 @@ func (v *Visor) Run(ctx context.Context) error {
 
 					numOfUpgradeStatusErrs++
 
-					v.log.Info("Still waiting for vega to start...", logging.Int("attemptLeft", maxUpgradeStatusErrs-numOfUpgradeStatusErrs))
+					attemptsLeft := maxUpgradeStatusErrs - numOfUpgradeStatusErrs
+					if numOfRestarts == 0 {
```
karlem, line 182: This is a no show stopper but I would (since it's used in 2 places) store it in a variable called something like `isFirstStartup`. Just to make it very clear what we are looking for.

visor/visor.go
wwestgarth: No reason other than it feels like that the counter should be for the number of times of times we can't get the upgrade status in a row?

karlem: Yes I guess that's fair.

jeremyletang (approved):

daniel1302 (approved):

PR #11274, opened 2024-05-13
Vested account transfers from a pool sub account to the parent general account

### LINKED ISSUE #11263: [vAMM] allow transfers from a pool derived key to the parent key
Add an optional from field for the transaferFund command.
Requires the key to be:
- a valid subaccount for the main key
- from account to be REWARD_VESTED
- asset to be the same as the transfer

karlem:
Close #11263


protos/sources/vega/commands/v1/commands.proto
```diff
@@ -304,6 +304,12 @@ message Transfer {
   string amount = 5;
   // Reference to be attached to the transfer.
   string reference = 6;
+  // Optional derived subaccount key from which assets are transferred. 
+  // The submitter of the transaction must be the owner of this derived subaccount key.
+  // If provided, the 'from_account_type' must be REWARDS_VESTED, and the asset in this account 
+  // must match the asset specified in the transfer.
+  optional string from_sub_account = 7;
```
candida-d, line 311:
```suggestion
  optional string from = 7;
```
removing sub-account here we're tidying up the use of sub-account, which is used to mean several different things

protos/sources/vega/commands/v1/commands.proto
```diff
@@ -304,6 +304,12 @@ message Transfer {
   string amount = 5;
   // Reference to be attached to the transfer.
   string reference = 6;
+  // Optional derived subaccount key from which assets are transferred. 
+  // The submitter of the transaction must be the owner of this derived subaccount key.
```
candida-d, line 308:
```suggestion
  // The submitter of the transaction must be the owner of this derived key.
```

PR #11288, opened 2024-05-16
feat: add FT for 0016-PFUT-022

Jiajia-Cui:
closes 
https://github.com/vegaprotocol/core-test-coverage/issues/1318


Jiajia-Cui:
close it as its duplicated

PR #11303, opened 2024-05-21
feat: add test for AC 163 stop order in spot market

Jiajia-Cui:
closes https://github.com/vegaprotocol/specs/pull/2284

### DIFF
--- a/core/integration/features/spot/orders/0014-ORDT-163.feature
+++ b/core/integration/features/spot/orders/0014-ORDT-163.feature
@@ -1,0 +1,90 @@
+Feature: Spot market
+
+  Background:
+    Given time is updated to "2024-01-01T00:00:00Z"
+
+    Given the following network parameters are set:
+      | name                                    | value |
+      | network.markPriceUpdateMaximumFrequency | 1s    |
+      | market.value.windowLength               | 1h    |
+      | spam.protection.max.stopOrdersPerMarket | 5     |
+
+    Given the following assets are registered:
+      | id  | decimal places |
+      | ETH | 2              |
+      | BTC | 2              |
+
+    Given the fees configuration named "fees-config-1":
+      | maker fee | infrastructure fee |
+      | 0.01      | 0.03               |
+    Given the log normal risk model named "lognormal-risk-model-1":
+      | risk aversion | tau  | mu | r   | sigma |
+      | 0.001         | 0.01 | 0  | 0.0 | 1.2   |
+    And the price monitoring named "price-monitoring-1":
+      | horizon | probability | auction extension |
+      | 360000  | 0.999       | 1                 |
+
+    And the spot markets:
+      | id      | name    | base asset | quote asset | risk model             | auction duration | fees          | price monitoring   | decimal places | position decimal places | sla params    |
+      | BTC/ETH | BTC/ETH | BTC        | ETH         | lognormal-risk-model-1 | 1                | fees-config-1 | price-monitoring-1 | 2              | 2                       | default-basic |
+
+    # setup accounts
+    Given the parties deposit on asset's general account the following amount:
+      | party  | asset | amount |
+      | party1 | ETH   | 100    |
+      | party1 | BTC   | 11     |
+      | party2 | ETH   | 10000  |
+      | party2 | BTC   | 1000   |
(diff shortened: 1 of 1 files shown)


core/integration/features/spot/orders/0014-ORDT-163.feature
```diff
@@ -0,0 +1,108 @@
...
+    And the opening auction period ends for market "BTC/ETH"
+    When the network moves ahead "1" blocks
+    Then the trading mode should be "TRADING_MODE_CONTINUOUS" for the market "BTC/ETH"
+    And the mark price should be "1000" for the market "BTC/ETH"
+
+  Scenario:0014-ORDT-163, 0014-ORDT-164: A wrapped buy/sell order will be rejected when triggered if the party doesn't have enough of the required quote asset to cover the order.
+    Given the parties place the following orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     |
+      | party1 | BTC/ETH   | sell | 10     | 1000  | 0                | TYPE_LIMIT | TIF_GTC |
+      | party2 | BTC/ETH   | buy  | 10     | 1000  | 1                | TYPE_LIMIT | TIF_GTC |
```
cdummett, line 59: Could we maybe simplify the test a little by removing this step? There is no requirement for the parties to have made any trades to place a stop order on a spot market since there is no concept of a position.

core/integration/features/spot/orders/0014-ORDT-163.feature
```diff
@@ -0,0 +1,108 @@
...
+    # place an order to match with the limit order then check the stop is filled
+    And the parties place the following orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     | reference |
+      | party4 | BTC/ETH   | sell | 10     | 1010  | 0                | TYPE_LIMIT | TIF_GTC | p4-sell   |
+
+    # create party1 stop order
+    And the parties place the following orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     | only   | ra price trigger | error | reference |
+      | party1 | BTC/ETH   | buy  | 50     | 1010  | 0                | TYPE_LIMIT | TIF_GTC | reduce | 1005             |       | stop1     |
```
cdummett, line 68: Shouldn't need the reduce only here for a stop order on a spot market.

core/integration/features/spot/orders/0014-ORDT-163.feature
cdummett: If we do that then the balances at the end should match the deposited balances which I think is easier to reason - with this trade we should at least check the balances after the trade.

core/integration/features/spot/orders/0014-ORDT-163.feature
```diff
@@ -0,0 +1,108 @@
...
+      | party  | market id | status           | reference |
+      | party1 | BTC/ETH   | STATUS_TRIGGERED | stop1     |
+
+    Then "party1" should have general account balance of "200" for asset "ETH"
+    Then "party1" should have general account balance of "1" for asset "BTC"
+    Then "party1" should have holding account balance of "0" for asset "BTC"
+
+    Then the parties amend the following orders:
+      | party  | reference | price | size delta | tif     |
+      | party4 | p4-sell   | 1010  | -10        | TIF_GTC |
```
cdummett, line 87: This cancels the order right by removing all the volume? I don't think we have anything on the book after for party2's stop order to match against.

Could we make this a cancel so it's clearer and add a buy limit order at 1015 here as well so there would be something to match against?

core/integration/features/spot/orders/0014-ORDT-163.feature
```diff
@@ -0,0 +1,108 @@
...
+
+    # create party2 stop order
+    And the parties place the following orders:
+      | party  | market id | side | volume | price | resulting trades | type       | tif     | only   | ra price trigger | error | reference |
+      | party2 | BTC/ETH   | sell | 50     | 1015  | 0                | TYPE_LIMIT | TIF_GTC | reduce | 1020             |       | stop2     |
```
cdummett, line 92: Again no reduce only for spot market - don't think it matters but good to be clear.

cdummett (approved):

peterbarrow (approved):

PR #11305, opened 2024-05-21
fix: Correctly verify pegged order offset with respect to tick size i…

### LINKED ISSUE #11304: [Bug]: offset is allowed to be less than the tickSize
### Problem encountered

the offset validation against a tick size is not taking into account the priceFactor (market/asset decimals ratio)

### Observed behaviour

example:
pegged order pegged to mid price with offset 100 in market decimals.
market has one more decimals than asset so price factor = 0.1
tickSize is 100 (in asset!) 
this setting shouldn't be allowed as the effective offset is actually 10 which is lower than the market tick size.

### Expected behaviour

pegged offset should be validated against a tickSize in asset units. 

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

v0.76.4

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
…n the right units

closes #11304

### DIFF
--- a/core/integration/features/spot/orders/0037-OPEG-023.feature
+++ b/core/integration/features/spot/orders/0037-OPEG-023.feature
@@ -1,0 +1,91 @@
+
+Feature: 0001 tick size should be using market decimal, A pegged order specifying an offset which is not an integer multiple of the markets tick size should be rejected.
+
+    Background:
+        Given the following network parameters are set:
+            | name                                                | value |
+            | market.liquidity.bondPenaltyParameter               | 1     |
+            | network.markPriceUpdateMaximumFrequency             | 0s    |
+            | limits.markets.maxPeggedOrders                      | 6     |
+            | validators.epoch.length                             | 5s    |
+            | market.liquidity.earlyExitPenalty                   | 0.25  |
+            | market.liquidity.stakeToCcyVolume                   | 1.0   |
+            | market.liquidity.sla.nonPerformanceBondPenaltySlope | 0.19  |
+            | market.liquidity.sla.nonPerformanceBondPenaltyMax   | 1     |
+
+        And the liquidity monitoring parameters:
+            | name       | triggering ratio | time window | scaling factor |
+            | lqm-params | 0.1              | 24h         | 1              |
+
+        And the following assets are registered:
+            | id  | decimal places |
+            | ETH | 1              |
+            | BTC | 1              |
+
+        And the average block duration is "1"
+        And the simple risk model named "simple-risk-model-1":
+            | long | short | max move up | min move down | probability of trading |
+            | 0.1  | 0.1   | 60          | 50            | 0.2                    |
+        And the fees configuration named "fees-config-1":
+            | maker fee | infrastructure fee |
+            | 0         | 0                  |
+        And the price monitoring named "price-monitoring-1":
+            | horizon | probability | auction extension |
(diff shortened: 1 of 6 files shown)


jeremyletang (dismissed):

peterbarrow (dismissed):

core/integration/features/spot/orders/0037-OPEG-023.feature
```diff
@@ -0,0 +1,79 @@
...
+    #0037-OPEG-027:Given a market with non-zero market and asset decimals where the asset decimals are strictly more than the market decimals (yielding a negative price factor). A pegged order specifying an offset which is not an integer multiple of the markets tick size should be rejected.
+    And the spot markets:
+      | id        | name    | quote asset | base asset | liquidity monitoring | risk model          | auction duration | fees          | price monitoring   | sla params    | decimal places | tick size |
+      | ETH/DEC21 | BTC/ETH | BTC         | ETH        | lqm-params           | simple-risk-model-1 | 1                | fees-config-1 | price-monitoring-1 | default-basic | 2              | 10        |
+      | ETH/DEC22 | BTC/ETH | BTC         | ETH        | lqm-params           | simple-risk-model-1 | 1                | fees-config-1 | price-monitoring-1 | default-basic | 2              | 5         |
```
cdummett, line 45: Maybe this will become apparent once the test is updated to get out of opening auction, but these markets shouldn't be possible right?

We have...
- `asset decimals = 1`
- `market decimals = 2`
- `position decimals = 0`

Therefore violating the required `market decimal + position decimal < asset decimals`

core/integration/features/spot/orders/0037-OPEG-022.feature
```diff
@@ -0,0 +1,80 @@
...
+  #0037-OPEG-026:Given a market with non-zero market and asset decimals where the asset decimals are strictly more than the market decimals (yielding a negative price factor). A pegged order specifying an offset which is not an integer multiple of the markets tick size should be rejected.
+    And the markets:
+      | id        | quote name | asset | liquidity monitoring | risk model          | margin calculator         | auction duration | fees          | price monitoring   | data source config     | linear slippage factor | quadratic slippage factor | sla params | decimal places | tick size |
+      | ETH/DEC21 | ETH        | ETH   | lqm-params           | simple-risk-model-1 | default-margin-calculator | 1                | fees-config-1 | price-monitoring-1 | default-eth-for-future | 0.5                    | 0                         | SLA        | 2              | 10        |
+      | ETH/DEC22 | ETH        | ETH   | lqm-params           | simple-risk-model-1 | default-margin-calculator | 1                | fees-config-1 | price-monitoring-1 | default-eth-for-future | 0.5                    | 0                         | SLA        | 2              | 5         |
```
cdummett, line 45: Maybe this will become apparent once the test is updated to get out of opening auction, but these markets shouldn't be possible right?

We have...
- `asset decimals = 1`
- `market decimals = 2`
- `position decimals = 0`

Therefore violating the required `market decimal + position decimal < asset decimals`

cdummett, line 45: Could we also keep it to three markets each using some multiple of 10 as the tick size so we can line each market up with an AC easily?

PR #11312, opened 2024-05-23
feat: replicate a bug: when market is leaving opening auction, and ex…

Jiajia-Cui:
…it price is 0, the market is stuck


Jiajia-Cui:
fixed in other PR

PR #11343, opened 2024-06-03
feat: add decay rate to gov transfers and emit event with the initial…

### LINKED ISSUE #11336: [Bug]: Transfers & governance initiated transfers are presented inconsistently on the API
### Problem encountered

On a transfer on the transfers API the "amount" is the initial amount set up by the user. 

I would expect that the same behaviour would be observed on governance initiated transfers but it is not.

### Observed behaviour

On the governance transfer the below calculation is done.

```
transfer_amount = min(
    proposal.fraction_of_balance * source.balance,
    proposal.amount,
    NETWORK_MAX_AMOUNT,
    NETWORK_MAX_FRACTION * source.balance
)
where
NETWORK_MAX_AMOUNT is a network parameter specifying the maximum absolute amount that can be transferred by governance for the source account type
NETWORK_MAX_FRACTION is a network parameter specifying the maximum fraction of the balance that can be transferred by governance for the source account type (must be <= 1)
If type is "all or nothing" then the transfer will only proceed if:
    transfer_amount == min(proposal.fraction_of_balance * source.balance,proposal.amount).
```

### Expected behaviour

I would expect that both types of transfer are presented in exactly the same way ... however right now neither one is ideal to be honest.

- [ ] The "amount" should be the initial amount set by the user when setting up the transfer or governance transfer so should be taken from the transfer transaction or the governance proposal directly 

### Steps to reproduce

```bash
1. Submit `xyz` transaction
4. With this payload
5. Query `abc` API
6. Observe the error
```

### Software version

0.76

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
… amount rather than 0 on enactment and termination

Closes #11336


candida-d (dismissed):

Sohill-Patel:
```{"system-tests":"support-gov-tx-factor"}```

File: commands/proposal_submission.go [hunk 1]
```diff
@@ -697,6 +697,14 @@ func checkNewTransferConfiguration(changes *vegapb.NewTransferConfiguration) Err
 			return errs.FinalAddForProperty("new_transfer.changes.recurring.end_epoch", ErrIsNotValid)
 		}
 
+		if f, ok := big.NewFloat(0).SetString(recurring.Factor); !ok {
+			errs.AddForProperty("new_transfer.changes.recurring.factor", ErrNotAValidFloat)
+		} else {
+			if f.Cmp(big.NewFloat(0)) <= 0 {
```
EVODelavega, line 703: Any particular reason why we're not using `num.DecimalFromString(recurring.Factor)`  and subsequently `f.IsZero()`?

ze97286: just consistency with the way we check this elsewhere.

core/matching/side.go
```diff
@@ -220,10 +220,6 @@ func (s *OrderBookSide) amendOrder(orderAmend *types.Order) (int64, error) {
 // if removeOrders is set to True then the relevant orders also get removed.
 func (s *OrderBookSide) ExtractOrders(price *num.Uint, volume uint64, removeOrders bool) []*types.Order {
 	extractedOrders := []*types.Order{}
-	if volume == 0 {
```
wwestgarth, line 223: I'm assuming a bad merge has removed this. Same for the unit test below.

ze97286: super weird.

wwestgarth (approved):

candida-d (approved):

PR #11344, opened 2024-06-04
Update account API to include derivedParties

### LINKED ISSUE #11265: [vAMM] update account API to include derivedParties
update the account APIs to include derivedParties (see position API for reference)

karlem:
Close #11265


protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1181,6 +1187,9 @@ message ObserveAccountsRequest {
   string asset = 3;
   // Account type to subscribe to, required field.
   vega.AccountType type = 4;
+  // Whether to return all derived parties from AMMs for the given party.
+  // If used, party ID is required.
+  optional bool include_derived_parties = 5;
```
candida-d, line 1192: These new fields probably also need to be included in GraphQL

karlem: You are right. Just added it.

candida-d: @karlem Should the parent party ID field go in too, or is it already there? I didn't see it in the changes.

karlem: Maybe not. let me double check

karlem: Should be theere.

candida-d (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -1139,6 +1139,9 @@ message AccountBalance {
   string market_id = 5;
   // Account type of this account.
   vega.AccountType type = 6;
+  // Parent party ID of the account.
+  // This field is optional and is used in cases where the account is owned by a derived key.
```
wwestgarth, line 1143:
```suggestion
  // This field is optional and is used in cases where the account is derived from another party's account.
```

wwestgarth (dismissed):

wwestgarth (approved):

PR #11349, opened 2024-06-04
scale funding payments by fraction of time spent outside auction

### LINKED ISSUE #10360: Scale funding payment by fraction of period spent outside of auction
If the market has spent any time in auction mode during the funding period then the funding payment should be scaled down by the fraction of funding period spend outside auction mode. Please see https://github.com/vegaprotocol/specs/pull/2108 for details.

witgaw:
Closes https://github.com/vegaprotocol/vega/issues/10360


core/products/perpetual.go
```diff
@@ -164,18 +164,15 @@ type cachedTWAP struct {
 	end         int64        // time of the last calculated sub-product that was >= the last added data-point
 	sumProduct  *num.Uint    // the sum-product of all the intervals between the data-points from `start` -> `end`
 	points      []*dataPoint // the data-points used to calculate the twap
-
-	auctions *auctionIntervals
```
wwestgarth, line 168: Before you get too far I'd like to understand why this needs to be removed? As far as I understand we only need a few lines in `func (p *Perpetual) calculateFundingPayment(t int64)` to do:
```
factor := (t - p.startedAt) /  p.internalTWAP.auctions.timeSpent(p.startedAt, t)
```
and thats basically it?

witgaw: The formula is slightly different, but yes. 

I couldn't understand why there's a need to duplicate auction intervals withing `cachedTWAP`, thought it's just a leftover and since no tests failed after that initial refactoring I assumed I was correct. Is that not the case?

wwestgarth: They aren't copies its just a point, and its shared expressly to avoid having to pass it around everywhere and muddy up function signatures.

witgaw: Fair enough, I'll revert it then.

core/integration/features/settlement/0053-PERP-024_Closeout.feature
witgaw, line 1: 0053-PERP-024 is covered elsewhere and the behaviour which was described here wasn't in line with AC anyway.

core/integration/features/settlement/0019-MCAL_018.feature
```diff
@@ -88,9 +88,14 @@ Feature: check when settlement data precision is different/equal to the settleme
       | party3 | USD   | ETH/DEC19 | 132000  | 99866000  |
       | lp1    | USD   | ETH/DEC19 | 6600000 | 492200000 |
 
+    When time is updated to "2021-02-11T16:35:24Z"
```
EVODelavega, line 91: Can we not use blocks for this (`When the network moves ahead "N" blocks`). It's a lot easier to read through a scenario using relative time/blocks compared to some arbitrary timestamp and date-time strings. It's also more representative for what actually happens in reality (moving the network by N blocks updates the time 1 second at a time, calling the `OnTick` callbacks each time).

core/integration/features/settlement/0053-PERP-024_Closeout.feature
EVODelavega, line 1: Why was this test removed? Going by its description (funding payment triggering a closeout), I think it's something worth covering with a test

core/integration/features/settlement/0019-MCAL_018.feature
witgaw: Not really, the test moves time by 6 months, so doing it via N blocks would be inefficient.

core/integration/features/settlement/0053-PERP-024_Closeout.feature
witgaw: `core/integration/features/settlement/0053-PERP-024_LossSocialisation.feature` does that already. This test wasn't adding anything useful and the behaviour described there wasn't even in line with the AC.

core/integration/features/settlement/0019-MCAL_018.feature
EVODelavega: you can set the block duration to 86400 (24h) and move by 180 blocks, or 864000 and move 18 blocks to move time by 6 months

witgaw: Great, there are intermediate steps that move 5 blocks ahead. 

I don't understand why we have "time is updated to" step available at all then if it's such a problem.

PR #11361, opened 2024-06-07
feat: nullchain now calls prepare proposal to acknowledge tx delays

### LINKED ISSUE #11360: Investigate nullchain enhancment for transaction reordering testing
To enable testing of https://github.com/vegaprotocol/vega/pull/11338 we'd like to ideally have functionality to do so in nullchain / vega market sim. 

@wwestgarth says it may be not too hard.

wwestgarth:
closes #11360

### DIFF
--- a/core/blockchain/nullchain/nullchain.go
+++ b/core/blockchain/nullchain/nullchain.go
@@ -55,6 +55,7 @@ type TimeService interface {
 
 type ApplicationService interface {
 	InitChain(context.Context, *abci.RequestInitChain) (*abci.ResponseInitChain, error)
+	PrepareProposal(_ context.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error)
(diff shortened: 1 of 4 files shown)


core/blockchain/nullchain/nullchain.go
```diff
@@ -187,22 +188,35 @@ func (n *NullBlockchain) processBlock() {
 		n.log.Debugf("processing block %d with %d transactions", n.blockHeight, len(n.pending))
 	}
 
+	// prepare it first
+	ctx := context.Background()
```
EVODelavega, line 192: I'd prefer we use the main context here, but not sure if it matters all that much

EVODelavega (approved):

core/blockchain/nullchain/nullchain.go
wwestgarth: Core doesn't use the context in the `PrepareProposal` atm. In-real-life it would come from tendermint and I don't know what stuff it throw in it so that we can emulate it, and also the nullchain doesn't have a main context that it uses throughout so I think I'll leave it as is for now.

PR #11373, opened 2024-06-11
Feat 11028

### LINKED ISSUE #11028: Create API to estimate order book depth based on defined vAMM shape
# API Overview

When users are creating or amending their AMM shape we need to be able to give them information on what that will actually mean for them as a user.  What shape they're quoting, what it means in terms of range, loss at range, liquidation price etc, so they can make informed decisions.

Since this API will need thorough testing and is complex functionality it has been explicitly included in the spec.

## API request details

- [ ] Provide an API that delivers the functioinality described in the [0014-NP-VAMM-bounds-estimations](https://github.com/vegaprotocol/specs/pull/2019/files) spec

## Questions 

## API test scenarios

GIVEN (setup/context) 
WHEN (action) 

karlem:
Closes #11028


jeremyletang (dismissed):

EVODelavega (dismissed):

candida-d (dismissed):

datanode/api/trading_data_v2.go
```diff
@@ -5559,3 +5559,89 @@ func (t *TradingDataServiceV2) ListAMMs(ctx context.Context, req *v2.ListAMMsReq
...
+	if overflow || basePrice.IsNegative() {
+		return nil, formatE(ErrInvalidBasePrice)
+	}
+
+	upperPrice, overflow := num.UintFromString(req.UpperPrice, 10)
+	if overflow || upperPrice.IsNegative() {
+		return nil, formatE(ErrInvalidUpperPrice)
+	}
+
+	lowerPrice, overflow := num.UintFromString(req.LowerPrice, 10)
```
wwestgarth, line 5580: Don't know whose taking over this PR but we should validate that `lower price < base price < upper price` otherwise the maths is going to go wild.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -5009,3 +5016,38 @@ message AMMEdge {
...
   string cursor = 2;
 }
+
+message EstimateAMMBoundsRequest {
+  // Base price of the AMM pool, the price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places.
+  string base_price = 1;
+  // Upper price of the AMM pool, the price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places.
+  string upper_price = 2;
```
wwestgarth, line 5026: This and `lower_price` should be optional since you can have a AMM that is only one sided i.e can only go long, and can only go short.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -5009,3 +5016,38 @@ message AMMEdge {
...
+  // Lower price of the AMM pool, the price is an integer, for example `123456` is a correctly
+  // formatted price of `1.23456` assuming market configured to 5 decimal places.
+  string lower_price = 3;
+  // Leverage at the upper price of the AMM pool.
+  string leverage_at_upper_price = 4;
```
wwestgarth, line 5031: This and `leverage_at_lower_price` should also be optional, and if they aren't provided we just ust the market risk factors.

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -5009,3 +5016,38 @@ message AMMEdge {
...
+  // Leverage at the lower price of the AMM pool.
+  string leverage_at_lower_price = 5;
+  // Amount of the asset that the party is willing to commit to the AMM pool.
+  string commitment_amount = 6;
+  // Market ID to filter for.
```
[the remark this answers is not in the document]
EVODelavega: done

protos/sources/data-node/api/v2/trading_data.proto
EVODelavega: done

protos/sources/data-node/api/v2/trading_data.proto
EVODelavega: done

datanode/api/trading_data_v2.go
EVODelavega: checks added

EVODelavega (approved):

wwestgarth (approved):

PR #11396, opened 2024-06-19
feat: improve datanode snapshot creation

jeremyletang:
As of now, the snapshot are created in a sequential and blocking way in the datanode. This means that while a snapshot is being taken, no block can be processed.

The following approach is made:
- the database is locked with a transaction
- queries are generated
- one by one the query are:
 - executed
 - the result piped into the file system
- finally the lock is released, and later the files are added to ipfs.

The bottle neck here is that the results are being save on the fs as they arrive, which is unecessary and amount for 95% of the time spent snapshoting (and so blocking anything else).

To prevent this, we keep those results from the database in buffers, and only save them to file via a worker go routine.

### DIFF
--- a/datanode/networkhistory/service.go
+++ b/datanode/networkhistory/service.go
@@ -70,8 +70,26 @@ func New(ctx context.Context, log *logging.Logger, chainID string, cfg Config, c
 		datanodeGrpcAPIPort: datanodeGrpcAPIPort,
 	}
 
+	go func() {
+		ticker := time.NewTicker(5 * time.Second)
+		for {
+			select {
+			case <-ctx.Done():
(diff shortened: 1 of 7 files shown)


datanode/networkhistory/service.go
```diff
@@ -68,10 +71,35 @@ func New(ctx context.Context, log *logging.Logger, chainID string, cfg Config, c
...
+			}
+
+			ticker := time.NewTicker(5 * time.Second)
+			for {
+				select {
+				case <-ctx.Done():
+					f()
+					return
+				case <-ticker.C:
+					f()
```
EVODelavega, line 97: why not call `PublishSegments` here, too (if there are any new segments to publish)?

PR #11407, opened 2024-06-24
feat: add derived key support to paid liquidity API

### LINKED ISSUE #11026: Create API to show AMM rewards & fees being paid to parent keys
# API Overview

Since it has been agreed by protocol design / core that rewards for AMMs will be paid to the parent key and not direct to the subkey, we will need an API that tells us what the parent keys have been paid.

UPDATE: We believe that by default the fees and rewards will all turn up on the normal APIs anyway and be sent to the parent key anyway.  Therefore the question here is how to make it clear on each of those APIs that the payment is a result of the derived party key. Need to define the full list of APIs we think this impacts. 

## API request details

- [ ] Ensure that the rewards API has derived party key in it 
- [ ] Ensure that the fee stats API has derived party key in it 

- [ ] Create a new summary API that tells us per market and per parent key
  - [ ] Maker fees received by parent keys as a result of subkey AMM
  - [ ] LP fees received by parent keys as a result of subkey AMM
  - [ ] Rewards paid to parent keys as a result of dispatch metric = maker fees received
  - [ ] Rewards paid to parent keys as a result of dispatch metric = LP fees received

  - [ ] Parent key
  - [ ] Subkey
  - [ ] Market ID
  - [ ] Fee / Reward type
  - [ ] Epoch to
  - [ ] Epoch from

```

```

## Questions 

## API test scenarios

GIVEN (setup/context) 
WHEN (action) 

EVODelavega:
Closes #11026


mattrussell36 (dismissed):

datanode/api/trading_data_v2.go
```diff
@@ -2534,7 +2534,15 @@ func (t *TradingDataServiceV2) ListPaidLiquidityFees(ctx context.Context, req *v
 		assetID = ptr.From(entities.AssetID(*req.AssetId))
 	}
 
-	stats, pageInfo, err := t.paidLiquidityFeesStatsService.List(ctx, marketID, assetID, req.EpochSeq, req.PartyIds, pagination)
+	partyIDs := req.PartyIds
+	if req.IncludeDerivedParties != nil && *req.IncludeDerivedParties {
+		subKeys, err := t.ammPoolService.GetSubKeysForParties(ctx, partyIDs, marketID)
```
wwestgarth, line 2539: Theres already a:
```
func (t *TradingDataServiceV2) getDerivedParties()
```
method that all the other API's with a similar `includeDerivedParties` flag use. It gets the amm-keys in a slightly different way, I don't have an opinion on which is best, but we should use the same function everywhere so we don't end up with inconsistencies between API.

EVODelavega: Yeah, I saw that method, decided not to use it, because in the way I'd call it here, it'd get all markets, and generate all derived keys for the party/parties (this endpoint is most likely going to be called for a single party anyway). I figured hitting the DB with a single query that may or may not return any derived keys, and then fetching the data for a set of keys that all exist made more sense to me than to use a set of keys that may not exist. That method also returns the data as a map, so FWIW, it just _looked_ a bit messy to iterate over a map just to append to the slice.

wwestgarth: Makes sense. Can we switch the other API to use your function then?

EVODelavega: and that is done now

candida-d (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -3115,6 +3115,8 @@ message ListPaidLiquidityFeesRequest {
   repeated string party_ids = 4;
   // Pagination controls.
   optional Pagination pagination = 5;
+  // Whether to return all derived parties from AMMs for the given party.
```
candida-d, line 3118:
```suggestion
  // Whether to return all derived parties from AMMs for the given party. If used, party ID is required.
```

Gave this some thought and I think since we have the extra note in GraphQL, it could be useful here too.

EVODelavega: updated

candida-d (approved):

mattrussell36 (approved):

jeremyletang (approved):

PR #11411, opened 2024-06-26
feat: add heartbeat to eth-forwarder so last-seen block heights stay …

### LINKED ISSUE #11334: [Feature]: Add a "heartbeat" for bridge contract chain events
# Feature Overview

**In order for** validator nodes to catch up Arbitrum events after a protocol-upgrade
**We will** add a "heartbeat" to the forwarder
**So that** the "last seen" block for inactive bridge contracts (e.g the multisig) remains recent

When a node is started from a snapshot it will use the block height of the last-seen event for a bridge contract as the block to start reading logs from. This is so that no events are missed during any down-time. For bridge contracts where there is not much activity the last-seen event could be a very long time ago, for example on the Ethereum bridge the multisig control contract's last signer event was 190 days ago.

What this means is that a validator node starting from a snapshot today would have to read the logs for 190 days worth of ethereum blocks, and this can take some time. 

On Arbitrum the problem is much worse since the block time is very low and 5 blocks are created per second. Also the maximum block interval that some Arbitrum Archival nodes allow to be queried at once is very small at `500` blocks. So to be able to read logs from the "last-seen" even could take a very long time.

What we need is to periodically send in a dummy heartbeat transaction that says to the network "there were no events, but we've checked logs up to block N", so that on snapshot restore the validators can simply check for new events from block N instead.

Note that this will will be similar to the heartbeat added to the eth-call engine here https://github.com/vegaprotocol/vega/pull/10872

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
closes #11334 

The Ethereum event forwarder will now send out a "heartbeat" chain event if no event-logs are seen on a contract for over an hour (configurable). This is so that the "last-seen" block saved into the snapshot remains relatively up to date so that after a protocol upgrade fewer block-spans need to be rechecked.

Other changes worth noting:
- All of the per-contract last seen block heights are saved in "tracker" engine in the `evtforward` package, instead of being dotted around each individual engine that processes a chain event.
- Its this engine which verifies the heartbeats
- I've added a `block-time` field to the `EthereumConfig` network parameter, like we have in the `EVMConfig` parameter, so core is aware of the expected block duration on Ethereum.


protos/sources/vega/chain_events.proto
```diff
@@ -4,6 +4,18 @@ package vega;
 
 option go_package = "code.vegaprotocol.io/vega/protos/vega";
 
+// Heartbeat represents that the ERC20 bridge is being watched even though no events happened.
+message ERC20Heartbeat {
+  // Contract address that is being listened too
```
candida-d, line 9:
```suggestion
  // Contract address that is being listened to
```

candida-d (dismissed):

candida-d (approved):

jeremyletang (approved):

PR #11427, opened 2024-07-03
fix v0.76.8 experiment

jeremyletang:
### **_DO NOT MERGE_**

# feat: improve datanode snapshot creation

As of now, the snapshot are created in a sequential and blocking way in the datanode. This means that while a snapshot is being taken, no block can be processed.

The following approach is made:
- the database is locked with a transaction
- queries are generated
- one by one the query are:
 - executed
 - the result piped into the file system
- finally the lock is released, and later the files are added to ipfs.

The bottle neck here is that the results are being save on the fs as they arrive, which is unecessary and amount for 95% of the time spent snapshoting (and so blocking anything else).

To prevent this, we keep those results from the database in buffers, and only save them to file via a worker go routine.

Cache buffer size in datanode snapshot.


jeremyletang:
```
{
   "system-tests": "release/v0.76.3"
}

```

PR #11433, opened 2024-07-10
feat: add AMM volume to market-depth API

### LINKED ISSUE #11020: Enhance Market Depth API to include vAMM volume
# API Overview

We will need the vAMM volume on the market depth API so that we can show this in the order book for users. 

Our understanding  / assumption is that for performance reasons we cannot do this across the whole order book, since AMM volume goes to infinity price in theory.  So the assumption is that you will provide it in some range of the mid, and to some level of granularity, and we will need to work out what is achievable through performance testing.

Outside this range we request some sort of estimate of the remainder of the shape, again probably at configurable resolution.  To keep this info in one place I've assumed this will be on the same API and included as part of this ticket, but am open to other implementation options like a seperate API for that.

## API request details

- [ ] Include real known AMM volume at each price level on the `market depth` API
- [ ] Include the estimated AMM volume at each price level 

NB. The resolution of the price levels provided by AMM, the range between which the AMM volume is real, and outside which it is estimated should all be defined by network parameters and optimised during performance testing.  

  - [ ] Show only CLOB volume price levels
  - [ ] Show only AMM volume price levels
  - [ ] Show levels with both

```
{
    "price": "17250",
    "numberOfOrders": "1",
    "volume": "100",
    "volumeAAM": "0",
    "volumeAAMEstimated": "0"
},
{
    "price": "17249",
    "numberOfOrders": "1",
    
    // normal amount on book
    "volume": "100",      
    
    // amount on book from aam                
    "volumeAAM": "10",
    
    // I think this would always zero here?
    "volumeAAMEstimated": "0"
},
{
    "price": "17248",
    "numberOfOrders": "1",
    
    // no normal orders
    "volume": "0",            
    
    // but some aam              
    "volumeAAM": "100",
    "volumeAAMEstimated": "0"
},
{
    "price": "17247",
    "numberOfOrders": "1",
    
    // no normal orders
    "volume": "0",  
    
    // no aam 
    "volumeAAM": "0", 
    
    // but 500 estimated
    "volumeAAMEstimated": "500"
}
```

## Questions 
~TBC on the performance questions and whether we set these in the UI or if they are dictated by the network.  Or maybe the max is dictated by the network but there is configurability beneath that in the UI?  Dont know.~

~This configuarability would need to be avalable on both the actual order book volume that trades can be matched with AND the estimated volume out to the extremes of the range and be individually configurable so it can be optimised.~

All config around number of price levels / granularity will be handled by network params.

## API test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
closes #11020


jeremyletang (dismissed):

datanode/gateway/graphql/schema.graphql
```diff
@@ -2890,6 +2890,12 @@ type PriceLevel {
   "The total remaining size of all orders at this level (uint64)"
   volume: String!
 
+  "The total volume of all AMM's at this level (uint64)"
```
candida-d, line 2893:
```suggestion
  "The total volume of all AMMs at this level (uint64)"
```

datanode/gateway/graphql/schema.graphql
```diff
@@ -2890,6 +2890,12 @@ type PriceLevel {
...
+  ammVolume: String!
+
+  "The total estimated volume of all AMM's at this level (uint64)"
```
candida-d, line 2896:
```suggestion
  "The total estimated volume of all AMMs at this level (uint64)"
```

protos/sources/vega/vega.proto
```diff
@@ -666,6 +666,10 @@ message PriceLevel {
   uint64 number_of_orders = 2;
   // Volume at the price level.
   uint64 volume = 3;
+  // Volume of AMM's at the price level.
```
candida-d, line 669:
```suggestion
  // Volume of AMMs at the price level.
```

protos/sources/vega/events/v1/events.proto
```diff
@@ -96,6 +98,13 @@ message AMM {
     // Leverage at lower bounds.
     optional string leverage_at_lower_bound = 5;
   }
+
+  message Curve {
+    // Unit liquidity on the given AMM curve.
```
candida-d, line 103: I can't figure out what this string is telling me - is it a unit of liquidity, the entirety of the liquidity, or maybe a third option...

mattrussell36 (dismissed):

candida-d (approved):

jeremyletang (approved):

PR #11471, opened 2024-07-17
chore: update uint256 to 1.3.0 and decimal to 1.4.0

jeremyletang:
I've just noticed that both the uin256 and decimal package have had recent (ish) updates, and both recent release comes with performances improvement. The codebase relying heavily on both of these, updating them can only be beneficial.


EVODelavega (approved):

ze97286:
I've already had it in the upgrade to cometbft 1.0

PR #11476, opened 2024-07-18
fix: fail checkTx downstream for delayed transactions so they don't g…

### LINKED ISSUE #11474: [Bug]: Aggressive transactions delayed for `>1` blocks causes network to stop
### Problem encountered

When a market is proposed with transaction reordering enabled and `market.aggressiveOrderBlockDelay == 2` submiting an order causes the network to stop producing blocks and pumps out the following logs:
```
ERROR	core.protocol.pow	pow/engine.go:528	replay attack: txHash already used	{"tx-hash": "d41dcce356dbace4c6949d8432755bb78276d5db1d355677bb62500a53e45663", "tid": "bb18c7f6b323d2251eebc1f854798af225c58af1387139a7c203eb407c31011c", "party": "513e0a30e52ac13fb22eb66d32822db4d70cb17830725144d08b4a5be6328143", "command": "Submit Order"}
ERROR	core.protocol.pow	pow/engine.go:390	pow error	{"error": "transaction hash already used"}
.
.
.
ERROR	tendermint	consensus/state.go:1391	prevote step: state machine rejected a proposed block; this should not happen:the proposer may be misbehaving; prevoting nil	{"err": null}
```

This seems to be because an order that is prepared in block `n` and delayed for block `n+2` gets reconsidered when preparing block `n+1` and is included for another delay in block `n+3`.

system test branch: `txn-ordering`
test name:  `test_transaction_reordering`
vega-branch with lots of logs added: `tx-ordering-logs`

### Observed behaviour

Network stops producing blocks

### Expected behaviour

Network keeps producing blocks.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

ze97286:
…et included in more than one block

Closes #11474

### DIFF
--- a/core/txcache/cache.go
+++ b/core/txcache/cache.go
@@ -17,8 +17,10 @@ package txcache
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"sort"
+	"sync"
 
 	"code.vegaprotocol.io/vega/core/nodewallets"
 	"code.vegaprotocol.io/vega/core/txn"
@@ -27,6 +29,8 @@ import (
 	"code.vegaprotocol.io/vega/libs/proto"
 	commandspb "code.vegaprotocol.io/vega/protos/vega/commands/v1"
 	snapshotpb "code.vegaprotocol.io/vega/protos/vega/snapshot/v1"
+
+	"github.com/cometbft/cometbft/crypto/tmhash"
 )
 
 func NewTxCache(commander *nodewallets.Commander) *TxCache {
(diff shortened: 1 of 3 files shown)


core/txcache/cache.go
```diff
@@ -44,6 +48,34 @@ type TxCache struct {
 	numBlocksToDelay uint64
 	// no need to include is snapshot - is updated when markets are created/updated/loaded from snapshot
 	marketToDelayRequired map[string]bool
+	// map of transactions that have been picked up for delay
+	cachedTxs map[string]struct{}
```
wwestgarth, line 52: Need to init this new map in `NewTxCache()` otherwise it'll panic.

wwestgarth (approved):

PR #11478, opened 2024-07-18
feat: add test for LP Fee Fraction

Jiajia-Cui:


Jiajia-Cui:
close it as the LP system will be updated in the next release

PR #11485, opened 2024-07-22
fix: handle point ranges in AMM matching

### LINKED ISSUE #11484: [Bug]: Panic in overnight run -- AMM uncrossing does not produce a trade
### Problem encountered

A panic occurred in the full system test run over the weekend:
```
[2024-07-20T03:26:42.905Z] goroutine 101 [running]:
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/cmd/vega/commands/node.(*Command).Run.func1.1()
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/cmd/vega/commands/node/node.go:166 +0x54
[2024-07-20T03:26:42.905Z] panic({0x55af8c0?, 0x89a4c40?})
[2024-07-20T03:26:42.905Z] 	/usr/local/go/src/runtime/panic.go:914 +0x21f
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/libs/num.(*Uint).Mul(...)
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/libs/num/uint.go:347
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/core/liquidity/v2.(*Engine).doesLPMeetsCommitment(0xc000ce22c0, {0xc008e61d00, 0x40}, 0x0, 0xc008169fc0, {0xc0063fdce0?, 0x0?})
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/core/liquidity/v2/sla.go:187 +0x4ab
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/core/liquidity/v2.(*Engine).EndBlock(0xc000ce22c0, 0xc008169ae0?, 0xc008169560?, {0xc0063fdce0?, 0x8169ae0?})
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/core/liquidity/v2/sla.go:65 +0xe8
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/core/execution/common.(*MarketLiquidity).EndBlock(...)
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/core/execution/common/liquidity_provision.go:396
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/core/execution/future.(*Market).BlockEnd(0xc005f30700, {0x6677ec8, 0xc006177e90})
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/core/execution/future/market.go:1198 +0x1597
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/core/execution.(*Engine).BlockEnd(0xc00109a380?, {0x6677ec8, 0xc006177e90})
[2024-07-20T03:26:42.905Z] 	/jenkins/workspace/common/system-tests-wrapper/vega/core/execution/engine.go:1388 +0x42
[2024-07-20T03:26:42.905Z] code.vegaprotocol.io/vega/core/processor.(*App).OnEndBlock(0xc000188e00, 0x7fbf3c17def0?)
```

Jenkins run:
https://jenkins.vega.rocks/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/63743/pipeline/614

Looks to be AMM related where indicative-price-and-volume says we can uncross, but when we do uncross no trades occur.

This causes the mark-price to be nil because there has been no last-trade.

### Observed behaviour

Panic

### Expected behaviour

No panic

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

https://jenkins.vega.rocks/blue/organizations/jenkins/common%2Fsystem-tests-wrapper/detail/system-tests-wrapper/63743/pipeline/614

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #11484

### DIFF
--- a/core/execution/amm/engine.go
+++ b/core/execution/amm/engine.go
@@ -130,6 +130,7 @@ type Engine struct {
 	// gets us from the price in the submission -> price in full asset dp
 	priceFactor    num.Decimal
 	positionFactor num.Decimal
+	oneTick        *num.Uint
 
 	// map of party -> pool
 	pools    map[string]*Pool
@@ -157,6 +158,7 @@ func New(
 	marketActivityTracker *common.MarketActivityTracker,
 	parties common.Parties,
 ) *Engine {
+	oneTick, _ := num.UintFromDecimal(priceFactor)
 	return &Engine{
 		log:                   log,
 		broker:                broker,
@@ -173,6 +175,7 @@ func New(
 		priceFactor:           priceFactor,
 		positionFactor:        positionFactor,
 		parties:               parties,
+		oneTick:               oneTick,
 	}
 }
 
@@ -491,6 +494,21 @@ func (e *Engine) partition(agg *types.Order, inner, outer *num.Uint) ([]*Pool, [
 	// switch so that inner < outer to make it easier to reason with
 	if agg.Side == types.SideSell {
 		inner, outer = outer, inner
+	}
+
+	// if inner and outer are equal then we are wanting to trade with AMMs *only at* this given price
+	// this can happen quite easily during auction uncrossing where two AMMs have bases offset by 2
+	// and the crossed region is simply a point and not an interval. To be able to query the tradable
+	// volume of an AMM at a point, we need to first convert it to an interval by stepping one tick away first.
+	// This is because to get the BUY volume an AMM has at price P, we need to calculate the difference
+	// in its position between prices P -> P + 1. For SELL volume its the other way around and we
+	// need the difference in position from P - 1 -> P.
+	if inner.EQ(outer) {
+		if agg.Side == types.SideSell {
+			outer = num.UintZero().Add(outer, e.oneTick)
+		} else {
+			inner = num.UintZero().Sub(inner, e.oneTick)
+		}
 	}
 
 	if inner != nil {
--- a/core/execution/amm/pool.go
+++ b/core/execution/amm/pool.go
@@ -128,7 +128,7 @@ func NewPool(
 	positionFactor num.Decimal,
 	maxCalculationLevels *num.Uint,
 ) (*Pool, error) {
-	oneTick, _ := num.UintFromDecimal(num.DecimalOne().Mul(priceFactor))
+	oneTick, _ := num.UintFromDecimal(priceFactor)
 	pool := &Pool{
 		log:                  log,
 		ID:                   id,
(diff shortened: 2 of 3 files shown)


core/execution/amm/engine.go
```diff
@@ -157,6 +158,7 @@ func New(
 	marketActivityTracker *common.MarketActivityTracker,
 	parties common.Parties,
 ) *Engine {
+	oneTick, _ := num.UintFromDecimal(num.DecimalOne().Mul(priceFactor))
```
ze97286, line 161: I don't understand a couple of things about this:
1. this will always be 0 if the asset has less decimals than the market, is that ok?
2. what's the point of multiplying by 1?
3. why should it be price factor rather than just 1 if it's prices in asset than adjusting by 1 is what you want isn't it?

wwestgarth: 1) probably not, but none of this has been tested with negative DP's yet since they appeared half-way through the AMM feature branch's life and I just had to resolve the conflicts. Its been on my list to revist for a while, I'll make a ticket for it now.
2) there is no point, I'll fix it up
3) the matching algorithm for AMMs as defined in the spec works to the resolution of order-price-levels, there is no need for it to be more granual because any incoming orders will only be at these price-factor price levels and so knowing that an AMM's fair-price is "somewhere between two price levels" is enough. Whenever we need to view an AMM as orders, or if we need to reason about what a "unit-step" is along the AMM's curve we only need to consider the order-price levels to ensure everything remains uncrossed.

For example here in this bug, when we need to uncross at auction end, we "expand" the AMM in orders to this one-tick granuality because thats all we need to ensure we remain uncrossed afterwards. Then when we submit those orders to uncross, we again only need to reason in this one-tick granuality. 

We could instead view "unit-steps" along the AMM's curve in asset prices but this would not only cause a performance hit because there would be so many. We would also need to enforce that the commitment of an AMM is sufficient to create at least 1 volume between these "unit-steps" on the AMM curve so that the maths doesn't break down with `< 1` volume price changes. If we use asset-prices for our "unit-steps" then the minimum commitment for an AMM would have to me much much higher so that this remains true with smaller steps.

ze97286 (approved):

PR #11496, opened 2024-07-24
test: add coverage for protective auction ACs

EVODelavega:
Also included some more unit testing to processor package, changed the `log.Fatal` call in the protocol upgrade flow to a `log.Panic` to be able to recover in a unit test, to check that code-path.

### DIFF
--- a/core/execution/future/market.go
+++ b/core/execution/future/market.go
@@ -1502,7 +1502,7 @@ func (m *Market) EnterLongBlockAuction(ctx context.Context, duration int64) {
 	if !m.canTrade() {
 		return
 	}
-	// markets that are suspended via governance, or are in monitoring auction should be unaffected by long block auctions.
+	// markets that are suspended via governance should be unaffected by long block auctions.
 	if m.mkt.TradingMode == types.MarketTradingModeSuspendedViaGovernance {
 		return
 	}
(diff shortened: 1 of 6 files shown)


core/execution/future/market.go
```diff
@@ -1502,6 +1502,10 @@ func (m *Market) EnterLongBlockAuction(ctx context.Context, duration int64) {
 	if !m.canTrade() {
 		return
 	}
+	// markets that are suspended via governance, or are in monitoring auction should be unaffected by long block auctions.
```
ze97286, line 1505: you want to drop the second part of the comment

core/execution/spot/market.go
```diff
@@ -492,6 +492,11 @@ func (m *Market) EnterLongBlockAuction(ctx context.Context, duration int64) {
 		return
 	}
 
+	// markets in monitoring or governance auction are unaffected by long block auctions.
```
ze97286, line 495: and the first part here :)

core/execution/future/market.go
EVODelavega: true

core/execution/spot/market.go
EVODelavega: done

core/execution/spot/market.go
```diff
@@ -492,6 +492,11 @@ func (m *Market) EnterLongBlockAuction(ctx context.Context, duration int64) {
 		return
 	}
 
+	// markets in governance auction are unaffected by long block auctions.
+	if m.mkt.TradingMode == types.MarketTradingModeSuspendedViaGovernance {
+		return
+	}
+
 	m.mkt.State = types.MarketStateSuspended
 	m.mkt.TradingMode = types.MarketTradingModelLongBlockAuction
```
ze97286, line 501: I think this is still wrong you know? or at least out of place. If we're not in an auction then it's correct. But if we are in an auction I'm not sure we should do either of these.

jeremyletang (approved):

PR #11498, opened 2024-07-24
fix: improve performance with the ackedEvent store

### LINKED ISSUE #11497: Performance issue with storing the acked events
In the event forwarder, the ackedEvents data structure keeps track of the event that the chain haver acknowledge for a given duration. It uses a treeset, containing buckets at given timestamps.

We notice on staging that this is code becomes very slow (chain pretty much stoopped) when lots of events are processed (basically iterating over the tree, then the slice in the buckets).

See pprof graph:
Screenshot 2024-07-24 at 11 52 10

jeremyletang:
Two things have been done here:
- first the buckets are for one hour instead of one per timestamps.
- hashes are stored in a map not a slice anymore to avoid long iterations

close #11497

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - [1343](https://github.com/vegaprotocol/core-test-coverage/issues/1343) - Added coverage for `0093-TRTO-011`.
 - [1344](https://github.com/vegaprotocol/core-test-coverage/issues/1344) - Added coverage for `0093-TRTO-012`.
 - [1382](https://github.com/vegaprotocol/core-test-coverage/issues/1382) - Added coverage for `0093-TRTO-013`.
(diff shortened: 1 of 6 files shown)


wwestgarth:
In one of the previous panics about eth-call events being annoying, I lazily made a copy of `acked_events.go` in `core/datasource/external/ethverifier`...

Its probably worth copy+pasting this fix into that file too (or combining somehow if you can be bothered).

jeremyletang:
> In one of the previous panics about eth-call events being annoying, I lazily made a copy of `acked_events.go` in `core/datasource/external/ethverifier`...
> 
> Its probably worth copy+pasting this fix into that file too (or combining somehow if you can be bothered).

Addressed and rebased.

wwestgarth (approved):

PR #11534, opened 2024-08-07
fix: estimate AMM bounds now reports commitment issues and respects m…

### LINKED ISSUE #11426: [Feature]: Add "minimum commitment" checks to `EstimateAMMBounds`
# Feature Overview

When creating an AMM there are two reasons why the submission can fail due to the choice of commitment size:
1) the commitment is less that the network parameter `market.amm.minCommitmentQuantum`
2) commitment is too low with respect to the width of the bounds resulting in 0 volume price levels

Therefore it would make sense to add to the `EstimateAMMBounds` API a field in the response that will indicate if one of these two conditions has been breached. This will allow a user who is configuring their AMM on a fancy website with sliders can see when they've set an invalid AMM definition.

This was requested by @candida-d 

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

### LINKED ISSUE #11526: [Bug]: `EstimateAMMBounds` API doesn't not match up with the state of actual AMM's on stagnet.
### Problem encountered

Taking an AMM currently active on stagnet and passing its definition through the `EstimateAMMBounds` API shows that the AMM's position is outside of the max/min position reported by the API.

My hunch is that the API is wrong, and it is not taking into account the market's price/position decimals.

### Observed behaviour

The `EstimateAMMBounds` API does not reflect real life. 

### Expected behaviour

The `EstimateAMMBounds` API returns correct results.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

wwestgarth:
closes #11526 
closes #11426


jeremyletang (dismissed):

protos/sources/data-node/api/v2/trading_data.proto
```diff
@@ -5085,6 +5098,8 @@ message EstimateAMMBoundsResponse {
   string liquidation_price_at_upper = 5;
   // Estimated price below the lower bound at which the commitment will be lost.
   string liquidation_price_at_lower = 6;
+  // If populated the bounds could be calculated but the AMM is invalid for other reasons.
+  optional AMMError amm_error = 7;
```
wwestgarth, line 5102: I've forgotten to add this to graphql.

jeremyletang: Oh graphql...

jeremyletang (dismissed):

candida-d (approved):

jeremyletang (approved):

PR #11555, opened 2024-08-12
fix: specify end time when fetching new candles

### LINKED ISSUE #11544: [Bug]: Candles stream does not handle empty/zero-volume candles very well
### Problem encountered

On discord [a user pointed out](https://discord.com/channels/720571334798737489/720571334798737491/1270821626493272085) that the candles stream is a bit odd on markets where no trades occur, and this seems to be due to the handling of empty candles.

On testnet the market `89cc0f6d1616a14c79e7e9a310ef47bdbee0a6874654581ee1f4999c4fcdcd96` is inactive where the last trade occurred on the 20th Jun. If you stream any candles for this market you receive nothing apart from the old candle from months ago where you would expect updates with empty candles:
```
curl https://api.n00.testnet.vega.rocks/api/v2/stream/candle/data?candleId=trades_candle_15_minutes_89cc0f6d1616a14c79e7e9a310ef47bdbee0a6874654581ee1f4999c4fcdcd96
```

The reason for this seems to be because when you query candles with no end-timestamp you do not get trailing empty candles. This can bee seen here:
```
https://api.n00.testnet.vega.rocks/api/v2/candle?candleId=trades_candle_5_minutes_89cc0f6d1616a14c79e7e9a310ef47bdbee0a6874654581ee1f4999c4fcdcd96
```

but if you do supply an end-timestamp you do get trailing empty candles:
```
https://api.n00.testnet.vega.rocks/api/v2/candle?candleId=trades_candle_5_minutes_89cc0f6d1616a14c79e7e9a310ef47bdbee0a6874654581ee1f4999c4fcdcd96&toTimestamp=1723112638634349000
```

The way the candle stream service asks for candles updates is without an end-timestamp:
```
candles, _, err = s.candleSource.GetCandleDataForTimeSpan(ctx, s.candleID, &start, nil, entities.CursorPagination{})
```

which in times with no trades will result in no empty candles coming through, and the odd effect where once a trade does occur we will suddenly get *all* of the empty candles through to the stream at the same time and then the non-empty candle. If a trade never comes through then candle stream remains stuck on the last candle with volume.

### Observed behaviour

The stream-candle endpoint does not handle empty-candles very well.

### Expected behaviour

The stream-candles endpoint works as expected when empty candles exist.

### Steps to reproduce

```bash
1. Submit `xyz` transaction
2. With this payload
3. Query `abc` API
4. Observe the error
```

### Software version

develop

### Failing test

_No response_

### Jenkins run

_No response_

### Configuration used

_No response_

### Relevant log output

_No response_

EVODelavega:
Closes #11544

### DIFF
--- a/datanode/candlesv2/candle_updates.go
+++ b/datanode/candlesv2/candle_updates.go
@@ -74,6 +74,7 @@ func (s *CandleUpdates) run(ctx context.Context) {
 	defer closeAllSubscriptions(subscriptions)
 
 	ticker := time.NewTicker(s.config.CandleUpdatesStreamInterval.Duration)
+	defer ticker.Stop()
 	var lastCandle *entities.Candle
 
 	errorGettingCandleUpdates := false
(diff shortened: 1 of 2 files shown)


datanode/candlesv2/candle_updates.go
```diff
@@ -198,7 +190,7 @@ func (s *CandleUpdates) getCandleUpdates(ctx context.Context, lastCandle *entiti
 	if lastCandle != nil {
 		start := lastCandle.PeriodStart
 		var candles []entities.Candle
-		candles, _, err = s.candleSource.GetCandleDataForTimeSpan(ctx, s.candleID, &start, nil, entities.CursorPagination{})
+		candles, _, err = s.candleSource.GetCandleDataForTimeSpan(ctx, s.candleID, &start, &now, entities.CursorPagination{})
 
 		if err != nil {
 			return nil, fmt.Errorf("getting candle updates:%w", err)
```
wwestgarth, line 196: Theres a little bit below here:
```
for _, candle := range candles {
	if candle.LastUpdateInPeriod.After(lastCandle.LastUpdateInPeriod) {
		updates = append(updates, candle)
	}
}
```

which I think will stop the empty candles being added because for an empty candle, for whatever reason `LastUpdateInPeriod` ends up being equal to the value of `LastUpdateInPeriod` for the last real candle.

EVODelavega: You're right, I'll see if I can replace the `locf` in the subquery with an `interpolate` call. I think ought to set the last update in period to the correct value.

datanode/candlesv2/candle_updates.go
```diff
@@ -198,14 +190,14 @@ func (s *CandleUpdates) getCandleUpdates(ctx context.Context, lastCandle *entiti
...
 		}
 
 		for _, candle := range candles {
-			if candle.LastUpdateInPeriod.After(lastCandle.LastUpdateInPeriod) {
+			if candle.PeriodStart.After(lastCandle.PeriodStart) {
```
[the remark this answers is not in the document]
EVODelavega: fixed

wwestgarth (approved):

PR #11567, opened 2024-08-13
feat: change behaviour of stop orders

### LINKED ISSUE #8777: remove reduce only stop loss
https://github.com/vegaprotocol/specs/pull/1825

The above spec change removes the requirement for a stop loss to be reduce-only. In spot markets this is not an entirely obvious restriction as there may be no particularly obvious side of the trade to be the risk-reducing one. In the interests of keeping things simple we will thus remove this restriction from futures too for now to keep them in line.

This issue is to implement this on the core side

EVODelavega:
Closes #8777

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🚨 Breaking changes
 
-- [](https://github.com/vegaprotocol/vega/issues/xxx)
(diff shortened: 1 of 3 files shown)


EVODelavega:
Spec update PR: https://github.com/vegaprotocol/specs/pull/2339

jeremyletang (approved):

PR #11581, opened 2024-08-14
fix: send empty factors for party that does no quality

wwestgarth:
When a party wasn't eligible for a volume discount we were sending `nil` in the event instead of `types.EmptyFactors`. This was causing the data-node problems when it was trying to read the JSON blob for it and unmarshalling back into a `vega.DiscountFactors{}` where it didn't expect it to be `nil`.

### DIFF
--- a/core/volumediscount/engine.go
+++ b/core/volumediscount/engine.go
@@ -210,9 +210,9 @@ func (e *Engine) computeFactorsByParty(ctx context.Context, epoch uint64) {
 		// if the party hasn't qualified, then still send the stats but with a zero factor
 		if !qualifiedForTier {
 			evt.Stats = append(evt.Stats, &eventspb.PartyVolumeDiscountStats{
-				PartyId:        party.String(),
-				DiscountFactor: "0",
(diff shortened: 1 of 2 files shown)


jeremyletang (dismissed):

core/volumediscount/engine.go
```diff
@@ -210,9 +210,10 @@ func (e *Engine) computeFactorsByParty(ctx context.Context, epoch uint64) {
 		// if the party hasn't qualified, then still send the stats but with a zero factor
 		if !qualifiedForTier {
 			evt.Stats = append(evt.Stats, &eventspb.PartyVolumeDiscountStats{
-				PartyId:        party.String(),
-				DiscountFactor: "0",
-				RunningVolume:  notionalVolume.Round(0).String(),
+				PartyId:         party.String(),
+				DiscountFactor:  "0",
```
ze97286, line 214: do we need to keep this?

ze97286 (dismissed):

jeremyletang (approved):

PR #11648, opened 2024-09-04
11644 live only amms

### LINKED ISSUE #11644: [Feature]: Add `liveOnly` flag to AMM API
# Feature Overview

**In order to** be able to easily view only live-amms from the AMM API
**We will** add a "live-only" flag similar to on the order's API
**So that** any AMM without state `ACTIVE` or `REDUCE_ONLY` will be filtered from the response.

## Specs

## Tasks
- [ ]
- [ ]

## Feature test scenarios

GIVEN (setup/context) 
WHEN (action) 

wwestgarth:
closes #11644 

Adds a `liveOnly` flag to the AMM API that mirrors the `liveOnly` flag in the orders API. When set only AMM's with state `STATUS_ACTIVE` or `STATUS_REDUCE_ONLY` will be returned.

I've also removed a lot of the rejection events, because they are essentially duplicated of the transaction-result event and there is no point them existing and clogging up the AMM table.


jeremyletang (dismissed):

candida-d:
Shouldn’t this also have a GQL component?

jeremyletang (approved):

